PR #2, opened 2010-11-02
Fix lru leak

jlouis:
Hi,

This patch set fixes a (memory) leak in the LRU cache which I saw when running some Erlang Map/Reduce queries through the riakc_pb_socket interface. The commit message describes the problem and the one-line fix that plugs the leak in question.

Test case also provided.


kevsmith:
Thanks for catching the leak. I've merged your fix into master.

PR #6, opened 2010-11-22
Revised pull request

kevsmith:
New version of Friday's pull request which now includes a fixed erlang_js dependency. Tried amending the prior commit but ran into merge issues upstream so I figured it would be cleaner to generate a new pull request.


beerriot:
The latest version of this branch seems incompatible with Wriaki.  Wriaki uses map/reduce queries to produce its history pages.  On the tip of current riak_kv, they work fine, but on the tip of this branch they fail with a timeout.

I should be able to boil this down to a few simple curls, but if you already have Wriaki built, reproduction is easy.

Steps to reproduce:
1. Start riak.
2. Start wriaki.
3. Load wriaki in a browser (http://localhost:8000/)
3a. Signup (click 'signup' in the top right, then enter 'bryan' in all the signup fields, then click submit)
3b. Edit the welcome page (click 'edit', add 'Foobar!' to the text, comment on the change as 'fb', click submit)
3c. View the history (click 'history')

Expected: A list of one revision on the history page will be displayed.

Observed: A timeout error is displayed.  No errors display in the Riak console.  Timeout error in the wriaki console contains only client-side stacktrace.

Update: test case without Wriaki installed:

First load the two objects needed:

```
$ curl -X PUT -H "content-type:application/json" -H "Link: </riak/user/bryan>; riaktag=\"editor\"" http://localhost:8098/riak/archive/20b3aa.V2VsY29tZQ== --data @-
{"version":"20b3aaa","timestamp":1290453396,"text":"= This page describes Welcome =\n\nFoobar!","message":"fb"}

$ curl -X PUT -H "content-type:application/json" -H "Link: </riak/archive/20b3aa.V2VsY29tZQ%3D%3D>; riaktag=\"1290453396\"" http://localhost:8098/riak/history/V2VsY29tZQ== --data @-
""
```

Then run the map/reduce query:

```
$ curl -X POST -H "content-type: application/json" http://localhost:8098/mapred --data @-
{"inputs":[["history","V2VsY29tZQ=="]],"query":[{"link":{"bucket":"archive","tag":"_","keep":false}},{"map":{"language":"javascript","arg":"","keep":true,"source":"function(v) { var json = JSON.parse(v.values[0].data); var summary = { version: json.version, timestamp: json.timestamp, message: json.message}; var links = v.values[0].metadata.Links; for(var i in links) { if (links[i][2] == \"editor\") { summary.editor = links[i][1]; } } return [summary]; }" }}]}
```

The output should look like this (but hangs instead):
    [{"version":"20b3aaa","timestamp":1290453396,"message":"fb","editor":"bryan"}]

beerriot:
riak_cilent:list_buckets also seems to be broken on this branch.

Steps to reproduce:
1. Start riak.
2. {ok, C} = riak:local_client().
3. C:list_buckets().

Expected: A tuple of the form {ok, [binary()]}, where each binary in the list is a bucket name.

Observed: Error with stacktrace.  Here's a gist of an example run on a fresh, empty, single-node Riak cluster:
https://gist.github.com/30640402f9bfc38256be

kevsmith:
The hanging MapReduce query is caused by missing config entries in app.config. When I was originally merging this work in there was a set of changes to vars.config, vars/dev*.config, and files/app.config to support the new code. Running the code example with the fixed up app.config returns the correct result.

beerriot:
I have just confirmed that these patches fix the bugs I found in the comments above (listing buckets and that hung map/reduce query).

PR #7, opened 2010-11-25
eper tag

ghost:
Tell rebar eper-0.60 is a tag so that get-deps works reliably
and correctly. The embedded rebar binary needs to be updated
to at least basho/rebar@064195d.


dizzyd:
We'll revisit this after the 0.14 release, most likely.

dizzyd:
This was fixed coincidentally when I was upgrading rebar a week or so ago.

PR #10, opened 2010-11-29
feature/feature1 [JIRA: RIAK-2395]

dizzyd:
Please review these patches for bz508.


jonmeredith:
Passes my testing.  Verified test case in original bug as reported by Kevin works.

dizzyd:
Merged to mainline.

PR #12, opened 2010-12-01
Bz882 cluster info

slfritchie:
Reviewer: anyone interested

I've rearranged some of the functions in riak_kv_console.erl to do
double-duty in the name of reusing useful code: spit stuff out "verbose"ly
when riak_kv_console:Foo([]) is called (old-style) and avoid io:format()
calls when riak_kv_console:Foo(quiet) is called (new-style).


beerriot:
+1 looks good & works for me (and riak_kv_console-based riak-admin commands still work as well)

PR #14, opened 2010-12-08
Dss status refactor

dizzyd:
Please review and test these changes to the console/cluster_info code. Without this change, we are unable to run basho_expect to completion.


jonmeredith:
Works for me - created a 4 node cluster and checked transfers (pending transfers, plus vnodes not running)/ringready work and checked the cluster_info stuff all works.

PR #17, opened 2010-12-16
Bz937 filter keys

beerriot:
Just needed an extra function clause to match the format of the key filter produced by riak_client:filter_keys.

Fixes https://issues.basho.com/show_bug.cgi?id=937


beerriot:
Doh. I forgot this branch was made from 0.14 instead of master.  I'll send another pull request from the correct branch shortly.

PR #18, opened 2010-12-16
Fix for Issue 937 regarding riak_client:filter_keys

beerriot:
riak_client:filter_keys produces a different "filter" format than other functions.  This patch fixes the issue in the same manner that bucket-listing was fixed: by adding a function clause to the filter builder to handle the alternate format.


beerriot:
Merged into riak_kv-0.14 branch

PR #19, opened 2010-12-16
Proposed fix for spurious JS timeouts (bz #938)

kevsmith:
This patch addresses spurious timeouts generating during periods of heavy Javascript VM usage.


dizzyd:
Merged into riak_kv-0.14.

PR #23, opened 2010-12-20
Bz933 missing key map

beerriot:
These patches prevent map/reduce jobs from failing (with 'exhausted_preflist' errors) when non-existent bucket/keys are sent to map phases.  They restore the 0.13.0 behavior of evaluating Erlang map functions with the {error, notfound} tuple, and of substituting a "not_found" result in lieu of evaluating Javascript map functions.


beerriot:
The final commit in this request (4f01f16, "configurable notfound handling...") is not required to fix the bug, but does enable testing this behavior more easily.  See https://github.com/beerriot/mapred_verify/commit/5f7f88d9f8acfb5c5b8faf2c744ef74663a0b060 for example usage of the options implemented.

argv0:
+1, looks good to me.

beerriot:
Merged into riak_kv-0.14 branch.

PR #25, opened 2011-01-05
Bug 964 - riak_kv_mapred_term doesn't understand key-filters (Protobuffs) 

seancribbs:
Fixes bug where JSON format understands key filters but Erlang-terms format does not.


beerriot:
+1 merge

PR #26, opened 2011-01-10
no prune on read repair - for bz://968

jonmeredith:
Stop pruning vclocks on read repair.

 This fixes a case of endless read-repair.  If the get is on an object with a vector clock that will be truncated on the next put (e.g. after being allowed to grow temporarily large with the young vclock logic) and one of the vnodes holds an ancestor value then the read repair to the out of date node will be truncated and in conflict with the original value again.  Each get of the object will trigger a read repair.

It is safer to allow the vclocks to grow a little more in the read repair case than to truncate.

Fix for bz://968 | behavior seen in zd://387

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -74,7 +74,7 @@
                   robj :: term(),
                   reqid :: non_neg_integer(),
                   bprops :: maybe_improper_list(),
-                  prunetime :: non_neg_integer()}).
+                  prunetime :: undefined | non_neg_integer()}).
 
 %% TODO: add -specs to all public API funcs, this module seems fragile?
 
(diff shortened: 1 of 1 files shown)


argv0:
+1 on merge - looks good and seems to work based on my simple tests.

slfritchie:
+1.  This patch avoids huge memory explosions and changes them to only big.  In combination with etc/app.config changes of:

   {big_vclock, 1000},
   {young_vclock,-1},

... the not-recommended workload suffers no unexpected memory growth.  (Memory growth from FSMs slowing piling up until they reach their 30/60? sec timeout limit is expected.)

slfritchie:
Oh, this has indeed been merged, by
changeset 47e9d9898573ea91f62999bbc54baad08fc6a9d2 and
is also present in the riak_kv-0.14 release branch.

PR #27, opened 2011-01-11
Patch A for BZ 969

beerriot:
This patch solves the cache limit overrun bug described in https://issues.basho.com/show_bug.cgi?id=969.

The method for solving the bug that this patch uses is to convert riak_kv_lru into a gen_server.  This serializes all modifications to the cache.  Unfortunately, it also serializes all cache lookups, but that could probably be changed in a second commit (by using protected tables, and exposing their names).

Another option was to serialize all modifications through riak_kv_mapred_cache.  This patch adds the riak_kv_lru gen_server instead, in order to limit the changes to one file.  Serializing through riak_kv_mapred_cache would have required editing riak_kv_mapred_cache, riak_kv_mapper, and riak_kv_lru.

### DIFF
--- a/src/riak_kv_lru.erl
+++ b/src/riak_kv_lru.erl
@@ -33,7 +33,13 @@
          max_size/1,
          clear/1,
          clear_bkey/2,
-         destroy/1]).
+         destroy/1,
+         table_sizes/1]).
+-export([init/1,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         terminate/2]).
 
 -record(kv_lru, {max_size,
                  bucket_idx,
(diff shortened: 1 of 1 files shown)


jonmeredith:
I was just thinking in terms of parallelism... whatever wanted the value will be able to get on with its activities rather than blocking.

jonmeredith:
+1 - Build the server and played with it a bit - basic m/r functionality still works for me.

beerriot:
Rebased to riak_kv-0.14 branch for possible point release, then merged there and to master as

https://github.com/basho/riak_kv/commit/48648fc18a9703c5920dd980621b52324aa611a4

PR #28, opened 2011-01-17
Bz325 mapred builtins min max

seancribbs:
Fixes Riak.reduceMin and Riak.reduceMax. Also reformats some uneven indentation.

Javascript's sort() function sorts lexically and thus will not sort numbers correctly without a comparator function. Reduce/fold is the proper way to find a minimum or maximum.


slfritchie:
+1, will merge & push shortly.

PR #30, opened 2011-01-26
Bz982 - js_reload not working

beerriot:
This patch fixes problems with js_reload, caused by the split of JS VMs into multiple pools in 0.14 instead of one big pool in 0.13.  All VMs in all pools are reset if none is specified (as with default 'riak-admin js_reload' behavior).


dizzyd:
+1. Thanks for the nice explanations. :)

PR #31, opened 2011-01-27
Key count reduce function

beerriot:
The intention of this patch is to provide a suitable function for
counting the keys in a bucket via keylist+map/reduce.  Using this
function saves bandwidth over counting on the client side (with
?keys=stream, for example) because keys are never sent to the client.
Using this function saves I/O over a query including a map phase because
the values stored under the keys are never read from disk.

It still uses the regular list-keys process under the hood, but it at least abstracts the framework necessary for handling the streamed keylists.

See the commit message or edoc for example usage.

### DIFF
--- a/src/riak_kv_mapreduce.erl
+++ b/src/riak_kv_mapreduce.erl
@@ -32,7 +32,8 @@
          reduce_sort/1,
          reduce_string_to_integer/1,
          reduce_sum/1,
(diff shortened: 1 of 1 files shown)


beerriot:
Ew. The test-case had horrible typos, meaning I completely flubbed running it.  Closing this and torching the branch.  Will rebranch, recommit and reopen when finished.

PR #32, opened 2011-01-27
Key count reduce (Corrected)

beerriot:
The intention of this patch is to provide a suitable function for
counting the keys in a bucket via keylist+map/reduce. Using this
function saves bandwidth over counting on the client side (with
?keys=stream, for example) because keys are never sent to the client. Using this function saves I/O over a query including a map phase because
the values stored under the keys are never read from disk.

It still uses the regular list-keys process under the hood, but it at least abstracts the framework necessary for handling the streamed keylists.

See the commit message or edoc for example usage.

### DIFF
--- a/src/riak_kv_mapreduce.erl
+++ b/src/riak_kv_mapreduce.erl
@@ -83,13 +83,13 @@ map_object_value(Acc) ->
 %%      If the RiakObject is the tuple {error, notfound}, the
 %%      behavior of this function is defined by the Action argument.
 %%      Values for Action are:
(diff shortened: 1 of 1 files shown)


dreverri:
Please update the following KB entry when this pull request is accepted:
https://help.basho.com/entries/454778

slfritchie:
I don't see a BZ ticket or an AZ ticket for this, bummer.
But some fiddling with the Erlang shell and with curl suggest
that they work, at least in small cases.  Merge, takk.

PR #33, opened 2011-01-28
Bz971 stats latency

slfritchie:
Pull request 2 of 3 for bz971.  Only minor changes here, iirc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,31 +5,24 @@
 all: deps compile
 
 compile:
-	@echo Warning: deps/riak_core is a symlink
 	./rebar compile
 
 deps:
-	@echo Warning: deps/riak_core is a symlink
 	./rebar get-deps
 
 clean:
-	@echo Warning: deps/riak_core is a symlink
 	./rebar clean
 
 distclean: clean 
-	@echo Warning: deps/riak_core is a symlink
 	./rebar delete-deps
 
 eunit:
-	@echo Warning: deps/riak_core is a symlink
 	./rebar skip_deps=true eunit
 
 docs:
-	@echo Warning: deps/riak_core is a symlink
 	./rebar skip_deps=true doc
 
 dialyzer: compile
-	@echo Warning: deps/riak_core is a symlink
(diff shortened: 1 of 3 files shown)


jonmeredith:
I still feel a bit queasy about replacing spiraltime with an on-disk structure for stats.  After talking to Scott he said he's load tested it and it had no measurable effect on the cluster performance.  Given that we don't understand why the in-memory approach is having problems and customers are experiencing it, fixing it this way is acceptable until we can revisit stats collection. 

The code as is will generate a lot of files on disk with no cleanup.  It would be better if riak_kv_stat passed a directory into slide to use for storing the files in and cleaned that up on startup.   We should introduce an application environment variable and update the release templates so we can control the location. 

I don't think we want the OS pid used for the directory as that will make it hard to do cleanup across restarts, especially if there are multiple instances of riak running on a server for testing.

slfritchie:
Jon: I've addressed all of the review items that you've outlined.  I need to add one thing, which is a general dir cleanup of the slide dir when riak_kv_stat starts.

Re OS pid: we need a unique something to keep multiple Riak instances on the same box separated.  Having riak_kv_stat remove the dir on its init() will be sufficient, even in case of OTP app restart.

jonmeredith:
On the tmp dir name stuff - agreed it should have a PID on it if there's nothing in the app.config, but if the user has set one you can use it with impunity.

slfritchie:
Trap exit fix added, learn something new every day.
The app.config thingie was added to https://github.com/basho/riak_core/pull/11

slfritchie:
Notes on before/after performance measurements to address queasiness :-)
https://issues.basho.com/show_bug.cgi?id=971 comments 13 and 14.

I don't think that I've overlooked any other review comments, waiting for final
stamp of approval from Jon.

jonmeredith:
+1 merge it!!!

slfritchie:
Odd that this didn't auto-close, but it didn't, and it should be closed, since
all mergeification has been done.

PR #34, opened 2011-02-01
Fix for badmatch error in riak_kv_map_master on restart (bz://989, az #112)

seancribbs:
This is a temporary fix for too-strict pattern matching in fetching entries from the bitcask backing riak_kv_map_master.  Longer term, viability and utility of the MR queue should be considered.


slfritchie:
These are simple & clear fixes for the reported problem, many thanks.
I'm going to skip adding a unit test to check for regressions: MR queue
refactorings are likely to disrupt whatever unit test setup I can cobble
together, I think.

PR #36, opened 2011-02-09
Bz909 possible atom leak

kellymclaughlin:
This fixes the potential atom leak described in BZ909 and also adds some improved checking and error feedback for invalid RW, W, and DW values.

### DIFF
--- a/src/riak_kv_delete.erl
+++ b/src/riak_kv_delete.erl
@@ -81,7 +81,7 @@ reap(Bucket, Key, Timeout) ->
 
 delete_test_() ->
     %% Start erlang node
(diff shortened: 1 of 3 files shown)


slfritchie:
Kelly: would you mind adding some eunit tests for the put & delete FSMs?

It's interesting that the get FSM doesn't require mods: atom() > integer()
is always true.

slfritchie:
I saw 10 clean "make eunit" runs in a row, so I'd say it's good to merge.

slfritchie:
Hrm, I don't know why GitHub's auto-magic didn't close this
pull request, but it looks like Kelly merged his branch to "master",
so I'll close this.

PR #38, opened 2011-02-12
Specify Erlang map reduce functions as strings

dreverri:
This commit allows users to specify Erlang map reduce functions as strings:
REST:
curl http://127.0.0.1:8098/mapred -XPOST -H 'content-type: application/json' -d '{"inputs":[["b","k1"],["b","k2"]], "query":[{"map":{"language":"erlang","source":"fun(_,_,_)-> [ok] end."}}]}'
["ok","ok"]

PBC:
(riak@127.0.0.1)3> {ok,Pid}=riakc_pb_socket:start_link ("127.0.0.1", 8087).
{ok,<0.12379.1>}
(riak@127.0.0.1)4> riakc_pb_socket:mapred(Pid, [{<<"b">>, <<"k1">>},{<<"b">>, <<"k2">>}], 
(riak@127.0.0.1)4>                        [{map, {strfun, "fun(_,_,_) -> [ok] end."}, none, true}]). 
{ok,[{0,[ok,ok]}]}

### DIFF
--- a/ebin/riak_kv.app
+++ b/ebin/riak_kv.app
@@ -83,6 +83,15 @@
          {stats_urlpath, "stats"},
 
          %% Secondary code paths
-         {add_paths, []}
+         {add_paths, []},
+
+         %% Allow Erlang MapReduce functions to be specified as
+         %% strings.
+         %%
+         %% !!!WARNING!!!
+         %% This will allow arbitrary Erlang code to be submitted
+         %% through the REST and Protocol Buffers interfaces. This
+         %% should only be used for development purposes.
+         {allow_strfun, false}
         ]}
  ]}.
--- a/src/riak_kv_mapred_json.erl
+++ b/src/riak_kv_mapred_json.erl
@@ -283,21 +283,14 @@ parse_step(<<"javascript">>, StepDef) ->
             {ok, {jsanon, Source}}
     end;
 parse_step(<<"erlang">>, StepDef) ->
-    case bin_to_atom(proplists:get_value(<<"module">>, StepDef)) of
-        {ok, Module} ->
-            case bin_to_atom(proplists:get_value(<<"function">>, StepDef)) of
-                {ok, Function} ->
-                    {ok, {modfun, Module, Function}};
-                error ->
-                    {error, ["Could not convert \"function\" field value"
-                             " to an atom in:"
-                             "   ",mochijson2:encode({struct, StepDef}),
-                             "\n"]}
-            end;
-        error ->
-            {error, ["Could not convert \"module\" field value"
-                     " to an atom in:"
-                     "   ",mochijson2:encode({struct, StepDef}),"\n"]}
+    case extract_fun_type(StepDef) of
+        {source, Source} ->
+            {ok, {strfun, Source}};
+        {bucket_key, {Bucket, Key}} ->
+            {ok, {strfun, {Bucket, Key}}};
+        {modfun, M, F} ->
+            {ok, {modfun, M, F}};
+        Else -> Else
     end;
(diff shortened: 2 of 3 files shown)


seth:
This looks good to me.  Thanks for making the time to work this up.  I think it will be quite useful.

slfritchie:
Hi, all.  What's to keep folks from adding 'os:cmd("rm -rf ./\* ./.??*")' to their map/reduce phases?

seth:
There is no protection (you knew that, of course).  I can appreciate the desire to have the REST API be "safe" and perhaps that would be an argument to make this functionality easy to turn off via config.  I don't think it is a reason not to have the feature at all.

The safety feature seems most valuable in a situation where one is accepting untrusted map/reduce REST API requests.  That seems like a fairly specialized use case, akin to giving users access to a db client connection and allowing arbitrary SQL.  Sounds fun :-)

On the plus side, allowing erlang funs as strings allows those working in Erlang already to have a fast way issue M/R queries without switching languages.

kellymclaughlin:
I can see how useful this is, but it's also potentially very dangerous as Scott points out. IMO, if you're going to allow this there should be a config option and that option should be disabled by default to protect unsuspecting users.

beerriot:
For the sake of discussion:

It's important to note that the string-function must be a _complete_ Erlang expression — including the dot at the end!  Forgetting the dot generates a parse error.

Speaking of errors, here are what a couple look like:

Forgetting the dot at the end, over HTTP:

```
<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request:<br><pre>{error,{error,{badmatch,{error,{1,erl_parse,["syntax error before: ",[]]}}},
              [{riak_kv_mapred_query,define_anon_erl,1},
               {riak_kv_mapred_query,check_query_syntax,2},
               {riak_kv_mapred_query,start,6},
               {riak_client,mapred_stream,5},
               {riak_client,mapred,5},
               {riak_kv_wm_mapred,process_post,2},
               {webmachine_resource,resource_call,3},
                   {webmachine_resource,do,3}]}}</pre><P><HR><ADDRESS>mochiweb+webmachine web server</ADDRESS></body></html>
```

Using an unbound variable:

```
<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request:<br><pre>{error,{error,{unbound_var,'O'},[{erl_eval,expr,2}]}}</pre><P><HR><ADDRESS>mochiweb+webmachine web server</ADDRESS></body></html>
```

So, very Erlangy, which you'd expect writing Erlang code, though easily confused as to whether it's a problem with the written function or something else, imo.

Protocol buffers (riakc_pb_socket client) is a little more harsh:

```
{error,disconnected}

=ERROR REPORT==== 24-Feb-2011::16:43:13 ===
** Generic server <0.33.0> terminating 
** Last message in was {tcp_closed,#Port<0.3304>}
** When Server state == {state,"127.0.0.1",8087,false,false,undefined,
                               undefined,
                               {[],[]},
                               1,[],infinity,100}
** Reason for termination == 
** disconnected
```

The tracebacks seen in the HTTP error messages are printed in the SASL log.

Using the native Erlang client ('riak_client') prints the same error messages on the Riak console.

Anyone else have strong opinions on whether this level of error reporting is acceptable, or whether things should be a little more helpful?

kellymclaughlin:
The changes look good and everything tests out ok. Before I give the go ahead for merge, I think we should add a test case to riak_kv_mapred_json to show that a job with an erlang strfun parses ok.

kellymclaughlin:
+1 to merge.

PR #39, opened 2011-02-15
R unsatisfied but object found

krestenkrab:
Here's a proposed fix for the problem we have.

Let riak_client:get optionally return `{r_value_unsatisfied, R, P}`, where P is the number of ok replies, and R is the requested R value.

This changes the riak_kv_get_fsm logic in the cases where it receives 
`{error,  _}` to 
- not reply {error, notfound} unless the error is the Nth reply
- reply {r_value_unsatisfied, R, P} iff 0 < P < R and all errors
  from vnodes were notfound's, i.e. it is likely that a read-repair
  is imminent which will fix the issue.

So, effectively this lets us differentiate between two cases that currently reply `{error, notfound}`:
- the object is really not there (either it was deleted or never there)
- the object is available in _less than r vnodes_, but it is in there somewhere.

When the webmachine front receives a {r_value_unsatisfied, R, P}, it will issue an {halt, 503}.

So, this slows down the failure cases somewhat: it is slower to see the difference between a 404 "really-nothing-there" and a 503 "we have something, but the ?r=R request parameter was not satisfied.

For our usage cases, we really need to be able to make this distinction.

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -216,6 +216,7 @@ mapred_dynamic_inputs_stream(FSMPid, InputDef, Timeout) ->
 %%       {error, notfound} |
 %%       {error, timeout} |
 %%       {error, {n_val_violation, N::integer()}} |
+%%       {error, {r_val_unsatisfied, R::integer(), Replies::integer()}} |
 %%       {error, Err :: term()}
 %% @doc Fetch the object at Bucket/Key.  Return a value as soon as the default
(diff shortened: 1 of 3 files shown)


krestenkrab:
This is now on master (from az167-add-pr-to-get-fsm), so there is no need for this any more.

Vagabond:
The other half (the client side) is being worked on in az171-add-WP-RP-and-insufficent-response-to-clients for both riak_kv and riak-erlang-client.

PR #40, opened 2011-02-15
Bz605 put fsm timeout

krestenkrab:
Straight forward fix for https://issues.basho.com/show_bug.cgi?id=605

I have no means to trigger the error condition, but someone on #riak just experienced this very problem today.


beerriot:
Hi, Kresten.  We try to keep our source files tab-free.  Could you please resubmit this patch with spaces instead of tabs for indentation?

PR #41, opened 2011-02-24
Purify KV FSMs

jonmeredith:
I've spent some time refactoring the get FSM today in preparation for changes and thought I'd share progress to see if we think this is a viable approach.  I've created branches in riak_core and riak_kv called jdm-purify-fsms and have pushed my latest and greatest including some code I forgot to commit end of Wednesday.

Here's the broad strokes of what I've done:
- split the initialized state into prepare and execute.  Any side effects are isolated to the prepare function.
- added an alternate start function that takes a proplist to fudge the state and drops into the execute state.
- refactored the vnode response handling into a single function
- refactored the reply-to-client decision into a function
- replaced the preference list code with the riak_core_vnode_master based commands and added an alternate variant of the preflist code that provides the vnode type of primary/fallback (needed for deletes, but will be vital for adding RP-responses-from-primaries condition).
- taught the vnode_master commands to recognize Pids in the preference list and send messages there directly instead of going through a vnode master.
- taught riak_kv_vnode:put to accept bucket properties in options so it didn't try and get the ring while in 'pure' mode
- made read repair supply bucket properties.
- cleaned up the strange spawning/individual sync deletes - I don't believe they provide any additional protection against races.  We need to add conditional matching on the vclock soon anyway.

The refactored code still passes all the original quickcheck tests and I've added a couple of examples of how it can be used in 'pure' mode without any other parts of the system spun up.

./rebar eunit app=riak_kv suite=riak_kv_get_fsm if you'd like to see it run.

riak_kv_put_fsm is causing problems with running the whole test suite, I'll clean that up when I refactor the put FSM.

Also requires pull request https://github.com/basho/riak_core/pull/19 for riak_core.

Looking for comments on
- if this seems like a good approach
- reasons why any of the changes are bad or breaks something
- if the delete change is acceptable
- if the readrepair change is acceptable, and if we think it is ok in production or if it needs to be conditional for testing.  It will mean more data sent per read repair.

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -34,7 +34,6 @@
                 fail_threshold :: pos_integer(),
                 allowmult :: boolean(),
                 preflist2 :: riak_core_apl:preflist2(),
-                waiting_for=[] :: [{pos_integer(), atom(), atom()}],
                 req_id :: pos_integer(),
                 starttime :: pos_integer(),
                 replied_r = [] :: list(),
@@ -43,7 +42,6 @@
                 num_r = 0,
                 num_notfound = 0,
                 num_fail = 0,
-                repair_sent = [] :: list(),
                 final_obj :: undefined | {ok, riak_object:riak_object()} |
                              tombstone | {error, notfound},
                 timeout :: pos_integer(),
@@ -60,7 +58,7 @@ start(ReqId,Bucket,Key,R,Timeout,From) ->
 init([ReqId,Bucket,Key,R,Timeout,Client]) ->
     StateData = #state{client=Client,r=R, timeout=Timeout,
                 req_id=ReqId, bkey={Bucket,Key}},
-    {ok,prepare,StateData,0}.
+    {ok,prepare,StateData,0};
 
 prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key}}) ->
     StartNow = now(),
@@ -95,7 +93,7 @@ execute(timeout, StateData0=#state{timeout=Timeout, n=N, r=R0, req_id=ReqId,
             riak_kv_vnode:get(Preflist, BKey, ReqId),
             StateData = StateData0#state{n=N,r=R,fail_threshold=FailThreshold,
                                          allowmult=AllowMult,
-                                         waiting_for=Preflist2,tref=TRef},
+                                         tref=TRef},
             {next_state,waiting_vnode_r,StateData}
     end.
 
@@ -136,7 +134,7 @@ finalize(StateData) ->
 
 really_finalize(StateData=#state{allowmult = AllowMult,
                                  final_obj = FinalObj,
-                                 waiting_for=Sent,
+                                 preflist2=Sent,
                                  replied_r=RepliedR,
                                  bkey=BKey,
                                  req_id=ReqId,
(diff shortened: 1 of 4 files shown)


slfritchie:
My comments:
- My branch went so far as to make the FSM completely pure (see the
  "slf-pure-client-fsm-testing" branch):
  - All events and messages into the FSM are simulated by a
    gen_fsm-compatible reusable wrapper.
  - All messages out of the FSM are intercepted prior to calling
    erlang:'!' by pure code and added to a per-FSM list.
  - Calls to impure functions are either executed by erlang:apply/3 (for
    "real" use) or end up being calls to proplists:get_value/2,3 (for
    testing use).
- Jon's solution isn't 100% pure.  The tradeoff is fewer lines of code
  disturbed.  Further, Jon avoids using erlang:apply/3 which should
  confuse Dialyzer much less than my scheme requires.  (Well, my
  scheme doesn't really require erlang:apply/3, but the alternative to
  using apply() is very verbose and tedious.)  Less disturbance is good.
- Jon's scheme needs fewer lines of code for test setup & assertion
  checking appears to be less.  Also good.
- I'm a bit worried about the creation of extra procs and reliance on
  some impure code.  We've already seen hard-to-debug unit test
  failures when the order of previously-run unit tests changes: e.g.,
  registered procs not killed, application environment variables
  tweaked to weird/nonsense values.  This approach will require a more
  disciplined writing style ... especially if applied to something
  like replication, which could add at least one other complication of
  impossible-to-share resources like TCP ports.
- I can envision some tests wanting to check message ordering between
  all processes in a test.  It should be feasible to ask the tracer
  server for all events and then check ordering, right?  Will we have
  to be defensive about late-arriving messages screwing up tests that
  do that?
- The one problem that I see with the existing test code in the get
  FSM is that there's no way to control the order in which messages
  from the vnode procs arrive at the FSM process.  Some tests are
  going to want to strictly order those messages.
  - A solution to the ordering is to put a proxy in the middle, I
    suppose?  The FSM under test specifies the proxy's pid for the vnode
    procs to send their replies, and the harness tells the proxy which
    order to send the vnode replies back to the FSM under test.
  - FSM replies are tagged with the partition number, so the proxy could
    unambiguously identify the source of each reply.
  - In general, though, such a proxy wouldn't be able to identify an
    arbitrary message and thus couldn't deterministically reorder
    messages to a test author's satisfaction.  The question is, how
    often is that going to happen ... and how painful would it be to
    make further changes to message structures to make the ambiguous
    message unambiguous?

jonmeredith:
Thanks for the thoughtful comments. It's true it's not fully pure - and meck didn't end up working as well as I had hoped which is why the riak_core_tracer code is in there.

I've encountered the message ordering problem with the put FSM test.  I ended up using a proxy process (just like the get FSM test does).  When I free up from current tasks I'll complete the test and push it up for review.

I understand your concerns about spinning up extra procs - although I think we can use the fact that the FSM spins up in a process to defend against the extra messages problem.  As long as we don't send messages to registered processes we'll be fine - by creating processes each time we guarantee they won't interfere with one another.

I'd really like to use this as the basis for the RP/WP changes I have planned.  Do you think it is acceptable enough to merge and use as a basis for the changes or do we need to address specific issues first?

Thanks again, Jon.

slfritchie:
I think it's good enough (unless the put test looks really ugly), but this change of philosophy also big enough that at least another pair of eyes would be helpful.

beerriot:
Adding to the comment pile:
- I really like the breaking out of add_vnode_result and enough_results.  It definitely makes things clearer.
- I'm a little worried about defaults-drift from the bucket_props function in the test section of riak_kv_get_fsm, but at the same time, it's also nice to see a static capture of what is reasonably expected to be there usually.

+1 for merge

PR #42, opened 2011-02-24
Az127 flush ring

slfritchie:
Jon, I added one changeset to remove the debugging goop
that you'd temporarily added & a bit of comment line wrapping.
Otherwise,


jonmeredith:
Abandon this pull request - riak_core_ring_manager is not running when reip is executed.  Will submit a new one with a comment to that effect.

PR #45, opened 2011-03-02
Properly convert link headers returned from precommit hooks

dreverri:
Fixes bug 420:
https://issues.basho.com/show_bug.cgi?id=420


kellymclaughlin:
+1 - merge it

PR #47, opened 2011-03-08
Td p16 improve supervision and linking

kellymclaughlin:
The biggest changes here are to add supervisors to create and watch the processes for get, put, delete, key listings. The remaining changes are to change some spawn calls to spawn_link.

### DIFF
--- a/src/riak_kv_gb_trees_backend.erl
+++ b/src/riak_kv_gb_trees_backend.erl
@@ -37,7 +37,7 @@
 % @spec start(Partition :: integer(), Config :: integer()) ->
 %                        {ok, state()} | {{error, Reason :: term()}, state()}
 start(_Partition, _Config) ->
-    Pid = spawn(fun() -> 
+    Pid = spawn_link(fun() -> 
         {A1,A2,A3} = now(),
         random:seed(A1, A2, A3),
         tree_loop(gb_trees:empty()) 
--- a/src/riak_kv_gb_trees_backend.erl
+++ b/src/riak_kv_gb_trees_backend.erl
@@ -47,8 +47,15 @@ start(_Partition, _Config) ->
 % @spec stop(state()) -> ok | {error, Reason :: term()}
 stop(#state { pid=Pid }) -> 
     Ref = make_ref(),
-    Pid ! {stop, self(), Ref},
-    receive {stop_response, Result, Ref} -> Result end.
+    %% Make sure the backend process is alive or 
+    %% the call to receive will wait indefinitely.
+    case is_process_alive(Pid) of
+        true ->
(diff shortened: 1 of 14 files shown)


kellymclaughlin:
I replaced the start functions in riak_kv_put_fsm as well for the same reason.

src/riak_kv_put_fsm.erl
```diff
@@ -465,12 +474,14 @@ start_test_() ->
 
 successful_start() ->
     W = DW = 1,
+    process_flag(trap_exit, true),
```
jonmeredith, line 479: Could we wrap the whole start_test block in a {spawn, ...} - it makes me nervous changing the process flag on the eunit process that could go on and run other tests.

src/riak_kv_gb_trees_backend.erl
```diff
@@ -37,7 +37,7 @@
 % @spec start(Partition :: integer(), Config :: integer()) ->
 %                        {ok, state()} | {{error, Reason :: term()}, state()}
 start(_Partition, _Config) ->
-    Pid = spawn(fun() ->
```
jonmeredith, line 40: I tried killing the gb_trees process but unfortunately there's no error report as riak_core_vnode sets process_flag to true and then ignores any handle_info messages.
Rather than fix here and slow things up I'll add a story to fix it.

jonmeredith:
Get/put/delete all test fine for me - same baseline as master/HEAD.  Just list_keys left to stress.

%% Test for tech debt p16 - deliberately keep data small, testing overhead of supervision
{mode, {rate, 10}}.
{duration, 480}.
{concurrent, 100}.
{driver, basho_bench_driver_riakc_pb}.
{code_paths, ["deps/stats",
              "deps/riakc",
              "deps/protobuffs"]}.
{key_generator, {int_to_bin, {uniform_int, 1000}}}.
{value_generator, {fixed_bin, 4}}.
{riakc_pb_ips, ["192.168.1.132"]}.
{riakc_pb_replies, 1}.
{operations, [{get, 45}, {update, 45}, {delete, 10}]}.

jonmeredith:
list_keys seems happy enough after re-testing.  Looks like I had a b0rked bitcask.

jonmeredith:
Rechecked after transient -> temporary change.

+1 on merge.

kellymclaughlin:
I made another change to add a handle_exit callback to the riak_core_vnode behavior and added an implementation of it in riak_kv_vnode. If you don't mind have a look a see what you think of this change. I think it's more along the lines of what you were envisioning.

jonmeredith:
+1 looks good for merge.

PR #49, opened 2011-03-16
az://168 put fsm eqc

jonmeredith:
Here is my pre-refactoring test for the put FSM.

Reviewers: Scott plz :)


slfritchie:
First round of comments ... I'll dive into put_fsm_eqc.erl next.

I'm very, _very_ happy to see some test cases added for the put FSM.
"Cover" now says that line coverage increased from 0% -> 93%.

test/fsm_eqc_util.erl:

not_empty/1 is necessary?  eqc_gen:non_empty/1 doesn't mention
anything about 0-byte binaries.  However, the output of running
eqc_gen:sample(eqc_gen:non_empty(eqc_gen:binary())) a bunch of times
shows no instances of <<>>, but <<>> will appear if the non_empty()
call is removed.

longer_list/2: the inner resize() call looks redundant, is it doing
something useful?

A short comment on the reason for shrinking control in partval() might
be nice, I'm a bit curious.

riak_objects/0: possible TODO: use different tombstones for the 5
objects?

wait_for_req_id/1: if timeouts should never happen, then false
timeouts should happen less frequently if the 400msec timeout were
raised by 5x or 10x without interfering with test runtime.  (Thinking
about running tests on EC2 small instances, for example.)

Ditto about increasing timeout in timer:sleep(5) call in
get_fsm_qc.erl ... though that one is trickier, there isn't a way to
short-circuit timer:sleep().  Hm.  Well, that sleeping dog can
remain until some slow test CPU really causes a problem.

slfritchie:
I pushed some typo fixes & line wrapping, if you want 'em, take 'em.

There is good evil thinking in put_fsm_eqc:make_vput_replies().  Its
weirdness is that it generates more replies than are needed, but it's
only test code.  And I suppose there's extra value in having extra
replies available for the fake vnode to send in case the put FSM does
something really strange like look for even more replies?

+1 to merge

PR #50, opened 2011-03-17
bz://977 merge diff ts

jonmeredith:
Fixes case where vnode put is doing a syntactic merge but vnodes have a different timestamp.  This makes sure the same vclock is generated.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -320,7 +320,7 @@ prepare_put(#state{mod=Mod,modstate=ModState}, #putargs{bkey=BKey,
                                                         reqid=ReqID,
                                                         bprops=BProps,
                                                         prunetime=PruneTime}) ->
-    case syntactic_put_merge(Mod, ModState, BKey, RObj, ReqID) of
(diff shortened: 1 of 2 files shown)


slfritchie:
Sorry I didn't get this reviewed at the end of last week: I wanted to ponder the
meaning of prunetime a bit more.  prunetime does indeed seem to come from
the original put req's start_time member, which seems good.  If nothing bad can
happen when a read repair causes a syntactic  merge (prunetime will be
'undefined' and not the put's start_time), then +1 to merge. Otherwise, what
should the read repair + merge do?

jonmeredith:
The prune time gets passed in for read repairs (from the get timestamp), so that isn't a problem.

readrepair(Preflist, BKey, Obj, ReqId, StartTime, Options) ->
    put(Preflist, BKey, Obj, ReqId, StartTime, [rr | Options], ignore).

For handoffs it calls syntactic_put_merge/5 which adds a timestamp and calls syntactic_put_merge/6 with a timestamp.

I had a go at beating it up and I couldn't break it.  Merging.

PR #51, opened 2011-03-18
Cleanup put FSM

jonmeredith:
Similar refactor to the get FSM.

Fixes multiple issues found while writing put_fsm_eqc:
- Checks if DW met on last required W and returns immediately.  Before waited for next DW to respond (and timed out if it never came).
- Precommit hooks now chain correctly - garbage/failure is detected and the put aborted without calling the next hook.
- Postcommit hooks run on the put FSM process rather than spawning an unnecessary process and copying the object into it.
- Badly formatted hooks return an error rather than crashing the put FSM.

This is a big refactor of a critical piece.  I need at least two reviewers please.


jonmeredith:
I'll rebase this on top of the put fsm test commit. Too big.

PR #52, opened 2011-03-18
Bz1007 inconsistent map phase behavior

kellymclaughlin:
These changes along with the accompanying changes to erlang_js resolve the inconsistent handling of 'notfound' errors between erlang and javascript map functions. These changes adopt option number 2 that was outlined in the [bugzilla ticket](https://issues.basho.com/show_bug.cgi?id=1007): all 'notfound' errors are passed to the map functions and may be handled there. In the case the errors are not handled by the map functions, the error output returned from using erlang and javascript functions in HTTP MapReduce queries are consistent now. 

There is still more work to be done to cleanup and improve MapReduce error handling, but I found it to be out of scope of this ticket.


kellymclaughlin:
After considering this for a while I have decided that there are still too many inconsistencies between the handling of notfound errors between erlang and javascript so I'm closing this pull request for now and will re-open when I resolve the inconsistencies more to my liking.

PR #54, opened 2011-03-18
Cleanup put FSM

jonmeredith:
Fixes multiple issues found while writing put_fsm_eqc:
- Checks if DW met on last required W and returns immediately. Before waited for next DW to respond (and timed out if it never came).
- Precommit hooks now chain correctly - garbage/failure is detected and the put aborted without calling the next hook.
- Postcommit hooks run on the put FSM process rather than spawning an unnecessary process and copying the object into it.
- Badly formatted hooks return an error rather than crashing the put FSM.

This is a big refactor of a critical piece. I need at least two reviewers please.


src/riak_kv_put_fsm.erl
```diff
@@ -79,46 +86,12 @@ start_link(ReqId,RObj,W,DW,Timeout,From,Options) ->
...
-                                resobjs=[], allowmult=AllowMult},
-            StateData = handle_options(Options, StateData0),
-            {ok,initialize,StateData,0}
-    end.
+    StartNow = now(),
+    StateData = #state{robj=RObj0, 
+                       client=Client, w=W0, dw=DW0,
+                       req_id=ReqId, timeout=Timeout, options = Options0,
+                       startnow=StartNow},
+    {ok,prepare,StateData,0}.
```
argv0, line 94: do these state-transitions via gen_fsm timeouts cause any performance impact?

src/riak_kv_put_fsm.erl
```diff
@@ -157,153 +133,152 @@ handle_options([{returnbody, false}|T], State) ->
...
+                        ok ->
+                            {next_state, postcommit, StateData1, 0};
+                        {ok, _} ->
+                            {next_state, postcommit, StateData1, 0};
+                        _ ->
+                            {stop, normal, StateData1}
+                    end;
+                {false, StateData} ->
+                    {next_state, waiting_vnode, StateData}
+            end
```
argv0, line 241: perhaps generalize this code into a function ('maybe_reply'), since it looks like it's repeated in waiting_vnode.

argv0:
+1 on merge:  reviewed and did some testing, see comments above.

src/riak_kv_put_fsm.erl
argv0, line 94: answering my own question - they're just a receive... after Timeout loop, so it shouldn't be a perf impact.

jonmeredith:
Thanks for the review - I agree about factoring out the common code.  I'm going to be making some more changes to the put FSM to add the WP parameter so I'll include it in that change set instead so we don't have to go through another round of double-checking.

slfritchie:
+1 to merge. Eyeballing the code looks good, with the possible exception
of the # of return values from pre-/post-commit hooks.  But that rats nest is
small enough to punt to another ticket.

Nothing puked with basho_expect, using my usual multi-OS cluster of:
  HOSTS="ubu9-32-0 ubu9-64-0 osol-0 sol10-0 centos5-64-0"

jonmeredith:
Rebased and merged (don't worry, I diffed first) so github didn't pick it up.  Closed manually.

PR #55, opened 2011-03-25
Timeout listing keys if "tokenize" filter used and key has fewer than _n_ tokens

russelldb:
Choice was either
1. return the key untransformed
2. do not return the key

I went with 2

### DIFF
--- a/src/riak_kv_mapred_filters.erl
+++ b/src/riak_kv_mapred_filters.erl
@@ -79,7 +79,14 @@ to_lower(_) ->
 tokenize(Args) ->
     Seps = riak_kv_mapred_filters:to_string(lists:nth(1, Args)),
     TokenNum = lists:nth(2, Args),
-    fun(V) -> list_to_binary(lists:nth(TokenNum, string:tokens(to_string(V), Seps))) end.
+    fun(V) ->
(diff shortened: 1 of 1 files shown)


src/riak_kv_mapred_filters.erl
```diff
@@ -79,7 +79,14 @@ to_lower(_) ->
 tokenize(Args) ->
     Seps = riak_kv_mapred_filters:to_string(lists:nth(1, Args)),
     TokenNum = lists:nth(2, Args),
-    fun(V) -> list_to_binary(lists:nth(TokenNum, string:tokens(to_string(V), Seps))) end.
+    fun(V) ->
+	    Tokens = string:tokens(to_string(V), Seps),
+	    case length(Tokens) < TokenNum of
+		true -> [];
```
seancribbs: Minor thing, indentation is kind of weird here.

russelldb: Hmm, so it is. I just hit C-c C-q in Aquamacs...and for shame there be tabs. :-$
Have reformatted and resubmitted.

russelldb:
This looks fixed, running against 1.0+ it is no longer an issue

PR #56, opened 2011-03-28
Catch all errors and prevent them propogating back up to the user

Vagabond:
In cases where we can, we print out a friendly error message, in other
cases we indicate to the user that the full error is in the log file.


jonmeredith:
Looks good code wise.  We're getting a mix of styles for console messages.  It would be even better if the new messages followed the same style.

PR #57, opened 2011-03-30
fix call to riak_object:syntactic_merge to properly pass timestamp

argv0:


jonmeredith:
+1 merge - thanks Andy, great catch.

PR #58, opened 2011-03-30
Remove unecessary call to riak_kv_delete_sup:start_link

Vagabond:
The delete supervisor is actually started by do_dep_apps and this extra
call to start it just makes the setup fail.


Vagabond:
There's more issues involved here, closing until I've got them all fixed.

PR #59, opened 2011-03-31
Az211 run tests from toplevel

Vagabond:
Tested on OSX, Solaris and Linux, all tests pass in all applications from the toplevel now.

### DIFF
--- a/src/riak_kv_multi_backend.erl
+++ b/src/riak_kv_multi_backend.erl
@@ -258,7 +258,7 @@ eqc_test() ->
     % Start the ring manager...
     crypto:start(),
     {ok, P1} = riak_core_ring_events:start_link(),
-    {ok, P1} = riak_core_ring_manager:start_link(test),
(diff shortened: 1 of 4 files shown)


src/riak_kv_multi_backend.erl
```diff
@@ -197,28 +197,68 @@ assert(false, Error) -> throw({?MODULE, Error}).
...
 -ifdef(EQC).
 %% @private
+
 eqc_test() ->
     % Start the ring manager...
     crypto:start(),
-    riak_core_ring_events:start_link(),
-    riak_core_ring_manager:start_link(test),
+    {ok, P1} = riak_core_ring_events:start_link(),
+    {ok, P1} = riak_core_ring_manager:start_link(test),
```
kellymclaughlin: Looks like the tuple here should be {ok, P2}.

kellymclaughlin:
Good to merge.

PR #60, opened 2011-03-31
Other half of proposed fix for BZ1015

beerriot:
This is the other half of the fix for https://issues.basho.com/1015 (the first half is https://github.com/basho/luke/pull/4).

Prior to this pull, riak_kv_mapper was monitoring the luke_phase pid.  This meant that when the phase exited, the mapper received a message in its inbox.  Unfortunately, it may be some time (possibly forever, in a hung-process situation) before the mapper received that 'DOWN' message.  This patch removes the monitor, and instead links to the luke_phase, such that it is torn down immediately when the phase exits.

The patch also links the mapper to its JavaScript VM, if it uses one.  This is necessary to ensure that the VM is returned to the pool.  Using this method means that the JS VM is killed when the mapper process dies abnormally, so the JS supervisor spins up a new JS VM to replace it.  If this link is not in place, the JS VM will never be returned to the pool if the mapper dies abnormally.

WARNING: In the current version of erlang_js, linking and exiting in this manner appears to cause the entire Erlang VM to segfault (at least on OTP R14B01, 64-bit, Mac OS X 10.6.7).  This problem should be fixed before this pull request is honored.

### DIFF
--- a/src/riak_kv_mapper.erl
+++ b/src/riak_kv_mapper.erl
@@ -62,13 +62,19 @@ start_link(VNode, Id, QTerm, MapInputs, PhasePid) ->
 init([VNode, Id, QTerm0, MapInputs, PhasePid]) ->
     QTermFun = xform_link_walk(QTerm0),
     {_, _, ReqId} = erlang:now(),
-    process_flag(trap_exit, true),
     gen_fsm:send_event(PhasePid, {register_mapper, Id, self()}),
-    erlang:monitor(process, PhasePid),
+    erlang:link(PhasePid),
     riak_kv_stat:update(mapper_start),
     {ok, CacheRef} = riak_kv_mapred_cache:cache_ref(),
     CacheKeyBase = generate_cache_key_base(QTermFun(undefined)),
     {ok, VM} = reserve_jsvm(QTermFun(undefined)),
+    %% we need some way to reclaim the JS VM if it is busy doing something
+    %% when the rest of the MapReduce phase exists (e.g. on timeout)
+    %% easiest method is simply to link, such that the VM is also killed,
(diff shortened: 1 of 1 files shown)


Vagabond:
This also makes the VM segfault on CentOS 64 bit with R14B02.

I think the code looks OK and the VM segfault is caused by the port driver freeing memory out from under the JS VM or something. The backtrace is consistant every time.

Ok to merge pending fixing the segfault.

Vagabond:
You should rebase this branch on top of master, as you missed out of the commit fixing the call to start_js on line 170 of riak_kv_js_manager.

beerriot:
This branch was based on 0.14.1 because there was a chance we'd want to get this fix into a point release.  If that's still the case, we'll need to cherry-pick the start_js fix.  Pinging @dizzyd: is this point release material or no?

beerriot:
Merged to 0.14 with additional:
- start_js fix: https://github.com/basho/riak_kv/commit/f4d7f1d496058e710854e3e2cf9c1c43280e0304
- "other half" dependency: https://github.com/basho/riak_kv/commit/137f2d78cebfa8dcdf4562735df37bde87eaf45c
- erlang_js segfault fix: https://github.com/basho/riak_kv/commit/f18d4de831590b229123539fa221df622be0eeee

Merged to master as:
- rebased patch: https://github.com/basho/riak_kv/commit/c1c65796576c529574f9dcf2e8a8e205f1c46d57
- "other half" dependency: https://github.com/basho/riak_kv/commit/7cc870b61e1467b4d921987dc302da03fd2e3b9f
- erlang_js segfault fix: https://github.com/basho/riak_kv/commit/9c21abe496b03d8a3c739fb0f8f4c68be5258b11

PR #61, opened 2011-03-31
bz://599 add information from erlang:memory() to stats output

argv0:


jonmeredith:
+1 merge. Tests fine.

PR #63, opened 2011-04-05
Bz1059 syntactic merge sans increment

krestenkrab:
See https://issues.basho.com/show_bug.cgi?id=1059 for details.


krestenkrab:
I rolled back part of this patch to undo the apply_updates call in riak_client:put, so that precommit hooks are run without updates applied.

PR #66, opened 2011-04-06
Generate keys for objects put without a key and return them to the client

Vagabond:
Generate keys for objects put without a key and return them to the client

### DIFF
--- a/src/riak_kv_pb_socket.erl
+++ b/src/riak_kv_pb_socket.erl
@@ -184,7 +184,17 @@ process_message(#rpbputreq{bucket=B, key=K, vclock=PbVC, content=RpbContent,
                            w=W0, dw=DW0, return_body=ReturnBody}, 
                 #state{client=C} = State) ->
 
-    O0 = riak_object:new(B, K, <<>>),  
+    case K of
+        undefined ->
+            % Generate a key, the user didn't supply one
+            Key = list_to_binary(riak_core_util:unique_id_62()),
+            ReturnKey = Key;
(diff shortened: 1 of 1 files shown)


src/riak_kv_pb_socket.erl
```diff
@@ -192,12 +202,17 @@ process_message(#rpbputreq{bucket=B, key=K, vclock=PbVC, content=RpbContent,
     DW = normalize_rw_value(DW0),
     Options = case ReturnBody of 1 -> [returnbody]; _ -> [] end,
     case C:put(O, default_w(W), default_dw(DW), default_timeout(), Options) of
+        ok when is_binary(ReturnKey) ->
```
jonmeredith, line 205: Why the need for a special case?  protobuffers library does not include any fields marked as undefined, so you should be safe without this clause. (Did not test).

jonmeredith:
Tested with the erlang client and the python client - the python client has errors until you merge this branch into master.  I'm assuming it's an issue with the new protobuffs code that was resovled somewhere else.

Vagabond:
Yes, that is the issue.

src/riak_kv_pb_socket.erl
Vagabond, line 205: Because I want to distinguish the 'return the generated key' vs. the 'return_body' case. When you are generating the key serverside and don't have return_body the API returns {ok, Key} instead of just 'ok'.

jonmeredith:
+1 merge.

PR #67, opened 2011-04-06
Fix for BZ1006 - Objects not found in map-reduce are no longer retried on other vnodes in preflist 

kellymclaughlin:
Change riak_kv_map_phase so that the other preflist entries are checked before giving up when an object is not found on the vnode initially selected for a map phase.


kellymclaughlin:
I pushed another change to handle the not found errors generated by erlang map functions also. The erlang map functions actually cache the notfound result which is probably not the best behavior, but this change at least sets up preference list retries once that behavior is corrected.

PR #69, opened 2011-04-07
remove unused module riak_kv_map_localphase.erl

argv0:
grep reveals no use, mapred_verify passes. 

can I get an independent run of mapred_verify on this branch before merge?


slfritchie:
No errors with basho_expect's ./RUN-systests.sh suite, merge plz.

PR #70, opened 2011-04-07
Unique list of siblings while merging.

jonmeredith:
Prior to the fix for bz://977 when performing syntactic put merge
each of the vndoes used a local timestamp for the vclock increment.
If the vnodes were not updating inside the same second the siblings
would be duplicated.   This will repair any objects with duplicated
siblings.

Thanks to Kresten for suggestions.


argv0:
+1 on merge- tests passed.   Spent a while puzzling over any negative side-effects, can't think of a case where this would hurt.

PR #71, opened 2011-04-08
Cleanup and performance

eriksoe:
Minor code simplifications; non-aggressive performance-improving changes.


Vagabond:
I reviewed it and it looked OK, but I didn't really see any performance differences. Maybe if there's a lot of siblings it'd help.

Merged.

PR #72, opened 2011-04-08
Fix noproc error for riak_core_ring_manager by just setting ring global.

jonmeredith:
Ran into problems with this crashing while testing the put FSM.  It still crashed when run as the only test.

Sidestep the problem by just setting the ring in mochiglobal without having to start up the ring manager at all.


argv0:
+1 on merge.  I've had to do this to several other unit tests as well.

PR #73, opened 2011-04-12
Bz1062 keylister error handling

kellymclaughlin:
These changes add error handling when starting key lister processes and changes riak_kv_keylister_master to use the client timeout value when starting these processes instead of relying on the default timeout value.

The goal of these changes is primarily to avoid 2 situations:
- A timeout occurring on a single node when starting the key lister causing the entire key listing process to fail and leaving the client process unaware and lingering until the client-supplied timeout (defaults to 8 minutes in riak_client) has expired. 
- All nodes timeout while starting the key listers and the keys fsm waits around for key listings to be reported back instead of promptly detecting the failure scenario and returning an error to the client.

In the former situation, the key listing should still be able to succeed in most cases and in the latter situation, the client should be notified as quickly as possible so that resources can be freed and appropriate steps taken.

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -527,7 +527,8 @@ wait_for_listkeys(ReqId, Timeout) ->
 wait_for_listkeys(ReqId,Timeout,Acc) ->
     receive
         {ReqId, done} -> {ok, lists:flatten(Acc)};
-        {ReqId,{keys,Res}} -> wait_for_listkeys(ReqId,Timeout,[Res|Acc])
+        {ReqId,{keys,Res}} -> wait_for_listkeys(ReqId,Timeout,[Res|Acc]);
+        {ReqId, Error} -> {error, Error}
     after Timeout ->
             {error, timeout, Acc}
     end.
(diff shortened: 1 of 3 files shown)


beerriot:
I saw two different error reports printed out while playing with these patches.  Gist: https://gist.github.com/034d83ac037d459643eb

Essentially, once, a timeout was caught by the initialize state, but another time an EXIT message made it to handle_info.  However, the EXIT message had a Reason that was not 'normal' (it was {error, flow_timeout} instead), so a 'badmsg' error was printed instead of the message about the nodes all being down.

In both cases, however, the result returned to the client was {error, timeout}, as expected, so perhaps this state is okay?

beerriot:
+1, works for me otherwise

kellymclaughlin:
I went ahead and pushed another small change so that when handle_info receives an 'EXIT' message with a reason other than normal it is used as the reason in the stop tuple to provide better feedback than the generic badmsg catchall.

beerriot:
I like it

PR #74, opened 2011-04-12
az://170 Add PW to put FSM

jonmeredith:
Refactor of the put FSM 
- Adopted new API for put FSM and updated riak_client.  The preferred interface takes Client:put(RObj, Options).
- Support new {pw, NumPrimaries} option to make sure that at least NumPrimaries of the entries in the preflist are the vnode owners - not fallbacks.  This can be used to get stronger consistency if desired.  For example with N=3, PR=2, R=2, PW=2, W=2, DW=2 you will always read your writes, but only one node can fail. Any more failure than that reads and writes will be denied.
- Added details option to return per-request details - currently only timing is supported. This will be extended with things like 'bestnode' and perhaps request parameters/vnode result counts in the future.
- Timing details breaks down request times by stage in the put FSM so you can see how long preparation (looking up the ring/bucket props/pref list), precommit, execution, waiting for responses.  If you want to add a debug tracer you can also see how long the postcommit takes.
- Added two new states - precommit is broken out for timing purposes and a finish state has been added as a point where all requests end for updating stats.  The behavior is the same as the current put FSM in that we don't track stats on failure, however we will be able to when we can spend some time improving the stats module.


src/riak_kv_put_fsm.erl
```diff
@@ -33,13 +33,33 @@
...
+
+-type detail_info() :: timing.
+-type detail() :: true |
+                  false |
+                  [detail_info()].
+
+-type option() :: {pw, non_neg_integer()} | %% Min number of primary (owner) vnodes participating
+                  {w, pos_integer()} |      %% Minimum number of vnodes receiving write
+                  {dw, pos_integer()} |     %% Minimum number of vnodes completing write
+                  {timeout, pos_integer() | infinity} |
```
argv0: possibly use the built-in timeout() type (non_neg_integer() | infinity)?

src/riak_kv_put_fsm.erl
```diff
@@ -33,13 +33,33 @@
...
+
+
+-type detail_info() :: timing.
+-type detail() :: true |
+                  false |
+                  [detail_info()].
+
+-type option() :: {pw, non_neg_integer()} | %% Min number of primary (owner) vnodes participating
+                  {w, pos_integer()} |      %% Minimum number of vnodes receiving write
+                  {dw, pos_integer()} |     %% Minimum number of vnodes completing write
```
argv0: shouldn't w,dw be "one | quorum | all | default | pos_integer()"  and pw be 'one | quorum | all | default | non_neg_integer()' ?

i have these types defined in my spec cleanup branch, so feel free to ignore this for this PR - it's on my radar.

src/riak_kv_put_fsm.erl
```diff
@@ -157,11 +179,19 @@ prepare(timeout, StateData0 = #state{robj = RObj0}) ->
...
     
 %% @private
-validate(timeout, StateData0 = #state{n=N, w=W0, dw=DW0, bucket_props = BucketProps, 
-                                      options = Options0, preflist2 = Preflist2}) ->
+validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
+                                      options = Options0,
+                                      n=N, bucket_props = BucketProps,
+                                      preflist2 = Preflist2}) ->
+    Timeout = get_option(timeout, Options0, ?DEFAULT_TIMEOUT),
+    PW0 = get_option(pw, Options0, ?DEFAULT_PW),
```
argv0: there's already symbolic defaults in the bucket props - see riak_kv_app.erl:51 or so - maybe we could get rid of some of the DEFAULT_X macros by using bucket defaults?

src/riak_kv_put_fsm.erl
```diff
@@ -173,100 +203,93 @@ validate(timeout, StateData0 = #state{n=N, w=W0, dw=DW0, bucket_props = BucketPr
...
-            case Reply of
-                ok ->
-                    {next_state, postcommit, StateData2, 0};
-                {ok, _} ->
-                    {next_state, postcommit, StateData2, 0};
-                _ ->
-                    {stop, normal, StateData2}
-            end;
+            process_reply(Reply, StateData2);
+
```
argv0: nit:  zap whitespace here?

src/riak_kv_put_fsm.erl
```diff
@@ -33,13 +33,33 @@
...
+-type detail_info() :: timing.
+-type detail() :: true |
+                  false |
+                  [detail_info()].
+
+-type option() :: {pw, non_neg_integer()} | %% Min number of primary (owner) vnodes participating
+                  {w, pos_integer()} |      %% Minimum number of vnodes receiving write
+                  {dw, pos_integer()} |     %% Minimum number of vnodes completing write
+                  {timeout, pos_integer() | infinity} |
+                  {detail, detail()}.       %% Request additional details about request
```
argv0: this should be {details, detail()} (details plural).

src/riak_kv_put_fsm.erl
jonmeredith: yes it should - something like symbolic_rw() | non_neg_integer().
Would be good to define that in one place as it will be shared across multiple FSMs.
I'll leave it for now for your cleanup branch.

argv0:
enthusiastic +1.  code is much more readable now.

tested/verified existing w/dw functionality works, tested pw functionality by taking nodes down / modifying vnode code to return failures/timeouts.

verified pre/post commit hooks still work.

justinsheehy:
+1

Solid.

argv0:
@justinsheehy :  I was wondering the same thing - IIRC at one time you were able to do a "mongodb-style" write with W=0.  Do we want to preserve that behavior?   We don't have to make a decision before merging this, but maybe it's worth thinking about.   A read with R=0 to force a read-repair could be useful too, perhaps.

jonmeredith:
W=0, DW=0 still work and are tested by eqc, so we should change those to non_neg_integer.

PR #76, opened 2011-04-14
Az171 add wp rp and insufficent response to clients

Vagabond:
This adds server side support for the new options in the Put and Get FSMs; pr, pw, basic_quorum and notfound_ok.

It also does some incidental code cleanups.

### DIFF
--- a/src/riak_kv_pb_socket.erl
+++ b/src/riak_kv_pb_socket.erl
@@ -169,9 +169,9 @@ process_message(#rpbgetreq{bucket=B, key=K, r=R0, pr=PR0, notfound_ok=NFOk,
                            basic_quorum=BQ}, #state{client=C} = State) ->
     R = normalize_rw_value(R0),
     PR = normalize_rw_value(PR0),
-    case C:get(B, K, [{r, default_if_undef(R)}, {pr, default_if_undef(PR)},
-                      {notfound_ok, default_if_undef(NFOk)},
-                      {basic_quorum, default_if_undef(BQ)}]) of
+    case C:get(B, K, make_option(r, R) ++ make_option(pr, PR) ++
+                     make_option(notfound_ok, NFOk) ++
+                     make_option(basic_quorum, BQ)) of
         {ok, O} ->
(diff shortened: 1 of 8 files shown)


src/riak_kv_pb_socket.erl
```diff
@@ -165,10 +165,13 @@ process_message(rpbgetserverinforeq, State) ->
...
-process_message(#rpbgetreq{bucket=B, key=K, r=R0}, 
-                #state{client=C} = State) ->
+process_message(#rpbgetreq{bucket=B, key=K, r=R0, pr=PR0, notfound_ok=NFOk,
+                           basic_quorum=BQ}, #state{client=C} = State) ->
     R = normalize_rw_value(R0),
-    case C:get(B, K, default_r(R)) of
+    PR = normalize_rw_value(PR0),
+    case C:get(B, K, [{r, default_if_undef(R)}, {pr, default_if_undef(PR)},
+                      {notfound_ok, default_if_undef(NFOk)},
+                      {basic_quorum, default_if_undef(BQ)}]) of
```
jonmeredith: The FSMs will supply their own default, so you could do something like
make_option(_Opt, undefined) -> [];
make_option(_Opt, default) -> [];
make_option(Opt,Val) -> [{Opt,Val}]

and make options as make_option(r,R)++make_option(pr,PR)++make_option(notfound_ok,NFOk)++make_option(basic_quorum, BQ)++Options

Not sure which is more efficient - means less terms to copy around, but it probably makes very little difference :)

Vagabond: I don't have a strong preference although I do find all those ++s ugly.

maybe [make_option(r, R), make_option(pr, PR), ...]?

jonmeredith:
Code looks good and tests well.  If you agree with passing a minimal list of parameters in then push the changes and I'll re-test.
Will be good to merge after that unless you have objections.

jonmeredith:
Thx for fixing.
+1 merge.

Vagabond:
Merged

PR #79, opened 2011-04-18
Broke out guts of new put/2 function into do_put to make work with R13B04

jonmeredith:
Broke out guts of new put/2 function into do_put to make work with R13B04.

Please test this works with R13B04 and R14B01.


argv0:
https://github.com/basho/riak_kv/pull/80 may be an alternate approach.

jonmeredith:
Withdrawn, I prefer Andy's fix.

PR #80, opened 2011-04-19
bz://1073 use THIS:Fun for qualified calls

argv0:


russelldb:
Since I committed the dodgy fix this might be a meaningless, but I'm +1 on this (having tried it on r13b04)

jonmeredith:
+1 merge, testing on R13B04 and R14B01.  You learn something new every day.

PR #81, opened 2011-04-19
Add "backlog" tcp option to pb socket opts

russelldb:
When (not that) many connections are opened simultaneously (as few as 30) by the PB client, often a few (or more) throw "connection reset by peer" exceptions.

Expose the backlog tcp option as an application env parameter to mitigate problem.

See bz://1075 for more details

Defaults to the (default) value of 5, can be set in app.config for riak_kv app as

```
{riak_kv, [
     {pb_backlog, 128}
    ]
}
```

is this the right approach?

what do I do to get this into:

a) the app.config?
b) the wiki?

### DIFF
--- a/src/riak_kv_pb_listener.erl
+++ b/src/riak_kv_pb_listener.erl
@@ -38,7 +38,9 @@ start_link() ->
 init([PortNum]) -> 
     {ok, #state{portnum=PortNum}}.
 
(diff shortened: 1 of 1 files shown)


Vagabond:
+1

russelldb:
Please cast your eye over:

https://github.com/basho/riak_ee/pull/11
https://github.com/basho/riak/pull/29
https://github.com/basho/riak_search/pull/49

if you have a minute

PR #82, opened 2011-04-20
Rz fix delete

rzezeski:
While trying to get all the Search tests to pass I found a bug in the delete functionality of KV that only manifested itself when the Search KV precommit hook is installed.  Upon further investigation I found that the `riak_kv_delete_sup` was spawning `riak_kv_delete` in such a way that the actual work of delete was being done **ON** the supervisor's container process.  This is bad because it will block any other delete request that comes in on that node and is why the Search KV hook was failing during delete (because there is an inner delete inside of it).  This also violates the interface for the supervisor which expects initialization to do as little work as possible and to return `{ok, Pid}`.

Please see commit comments for more info.

### DIFF
--- a/src/riak_kv_delete.erl
+++ b/src/riak_kv_delete.erl
@@ -28,7 +28,11 @@
 -include_lib("eunit/include/eunit.hrl").
 %-endif.
 
--export([delete/6]).
+-export([start_link/6, delete/6]).
+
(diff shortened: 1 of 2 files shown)


rzezeski:
One more thing since it's not obvious...the search tests I'm talking about are executed via the `run_all.sh` script under `tests`.

kellymclaughlin:
Looks good. +1 to merge.

PR #85, opened 2011-04-22
Fix vtag test

rzezeski:
fixes: az325

In order for this test to work on R13B04 the crypto app needs to be
started first.


jonmeredith:
+1 merge.

PR #86, opened 2011-04-22
Add Reply parameter to handle_not_found_reply so that the notfound errors

kellymclaughlin:
Add Reply parameter to handle_not_found_reply so that the notfound errors
are present if requested by the client.

Fixes: az327

### DIFF
--- a/src/riak_kv_map_phase.erl
+++ b/src/riak_kv_map_phase.erl
@@ -70,9 +70,9 @@ handle_event({mapexec_reply, VNode, BKey, Reply, Executor}, #state{fsms=FSMs, ma
         true ->
             case Reply of
                 [{not_found, _, _}] ->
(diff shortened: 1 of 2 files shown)


slfritchie:
Kelly, is it feasible to cook up an EUnit test without burning a lot of time?
I'm not familiar with the m/r code, but if there's a public API option to get
extra info, it'd be cool to know that that option is being honored.

kellymclaughlin:
I added a couple of test cases to show that you can choose to include or filter out the notfound errors when using map_object_value. This is only for the erlang side of map functions. The JS side does it a little differently. There is also some test cases that exercise this code in riak erlang client's live node tests. Look around line 1798 of riakc_pb_socket.erl.

slfritchie:
Thanks, looks good to merge.

PR #88, opened 2011-05-05
move json encoding out of riak_object and into riak_object_util

argv0:
This is a first step at trying to a) clean up riak_object, and b) move towards a more generic versioned object implementation

IMO, these are the major problems with riak object
- conflation of values vs. full objects :  values should be {vclock, content} pairs, objects should be {key, [value]}
- conflation values and operations on values:  updatemetadata and updatevalue are awkward and shouldn't be stored with the object
- overspecification of key structure:  shouldn't assume key structure like riak_kv's {B,K} tuples.
- overspecification of value structure:  riak_kv-specific {metadata, content}

I'm not sure how to represent updates yet, but I don't think it should be inside the object/value representations.

I imagine a generic versioned object to look something like this:

(in riak_core):

-type versioned_value(ValueType) :: {vclock(), ValueType}.
-type versioned_object(T) :: {key(), [versioned_value(T)]}.

(in riak_kv):
-type riak_kv_value() :: {dict(), term()}.
-type riak_object() :: riak_core::versioned_object(riak_kv_value()).

the 'versioned_object' module would be parameterized (but not a parameterized module - no dialyzer support) with a module implementing operations/callbacks on T like equal, compare, maybe some functions to help with merges.

the versioned_object module would also implement all operations over multiple values (reconcile,merge,etc).


argv0:
Almost forgot the actual motivation for this pull request in the above rant.  riak_object also includes a bunch of JSON serialization logic that is the lowest hanging fruit in this problem.   This pull request moves that logic into a separate module.

seancribbs:
@argv0 You should update this PR and resubmit.

engelsanchez:
I'm revisiting this in #601. Let's see if we can close these 2 year old PRs for good :)

rzezeski:
Closing this since 601 replaces it.

PR #89, opened 2011-05-06
Break out core get FSM logic

jonmeredith:
This breaks out the core logic of the get FSM so it can be reused for modelling and other places.

It also fixes an edge case with deletion - now deletion only takes place if all vnodes respond with the same object and there are no read repairs.  Otherwise it is possible for vnodes to be an ancestor of the deleted object without the tombstone in the metadata.  If the delete is not processed on the ancestor vnodes, but succeeds on the tombstone vnodes the object would be resurrected (could happen on an out-of-disk-space bitcask node).

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -32,12 +32,20 @@
          handle_info/3, terminate/3, code_change/4]).
 -export([prepare/2,validate/2,execute/2,waiting_vnode_r/2,waiting_read_repair/2]).
 
+-type detail() :: timing |
+                  vnodes.
+-type details() :: [detail()].
+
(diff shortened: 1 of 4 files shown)


src/riak_kv_get_fsm.erl
```diff
@@ -43,24 +43,15 @@
```
argv0, line 51: ^^^ need to update the options() spec to include details?

jonmeredith:
Just to be thorough :)

OK, passed 10000000 tests
true

argv0:
experience quickcheck.

argv0:
+1 to merge

jonmeredith:
Also received +1 by email from Justin Sheehy.

PR #91, opened 2011-05-06
bz://1067:  remove unused try_cast and fallback functions from riak_kv_ut

argv0:
bz://1067:  remove unused try_cast and fallback functions from riak_kv_util


rzezeski:
+1 to merge

PR #92, opened 2011-05-10
Put core

jonmeredith:
Similar to the get core pull request, breaks the put request logic out into a purely functional core.

### DIFF
--- a/src/riak_kv_put_core.erl
+++ b/src/riak_kv_put_core.erl
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% riak_kv_get_core: Riak get logic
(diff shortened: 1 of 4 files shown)


src/riak_kv_put_core.erl
```diff
@@ -0,0 +1,145 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_get_core: Riak get logic
```
seancribbs: (typo) riak_kv_put_core: Riak put logic

beerriot:
just reading, not running, the translation makes sense to me

argv0:
Reviewed, dialyzed, tested.  +1 on merge.

PR #94, opened 2011-05-11
Ensure timestamp passed to riak_object:syntactic_merge on read repair.

jonmeredith:
Fixes: bz://1094

If the vnode receives a read repair where the RR object is in conflict with the vnode object, syntactic_put_merge would introduce a new vclock entry with an undefined timestamp. This is because read repairs should not trigger vclock pruning and we use the prunetime for deciding whether or not to prune the vclock and for passing into syntactic_put_merge to use for vclock increments.

Fixed by adding an explicit start time field in #putargs and use that in the syntactic_put_merge.

I have been unable to reproduce this outside of the delete_bug code attached to bz://260 as it is very hard to get the vnode/fsm into that state despite lots of fiddling with net kernel and temporarily hacking riak_kv_vnode.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -76,7 +76,8 @@
                   robj :: term(),
                   reqid :: non_neg_integer(),
                   bprops :: maybe_improper_list(),
-                  prunetime :: undefined | non_neg_integer()}).
+                  starttime :: non_neg_integer(),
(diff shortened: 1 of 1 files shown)


slfritchie:
Note: bug recipe in https://issues.basho.com/show_bug.cgi?id=1091

argv0:
Verified old buggy behavior occurs on current with Scott's recipe, and that this patch fixes it. 

+1 on merge.

PR #96, opened 2011-05-12
Expose tombstone vclocks as in Kresten's 'expose-tombstone' branch

Vagabond:
This patch provides the same functionality as Kresten's did but its
optional (although always enabled for the REST API). The local client
and PB clients must use the 'deletedvclock' option to enable it.


jonmeredith:
Code looks good.  The get FSM eqc test needs to be extended with the new option.
Will be good to merge once that's there.

jonmeredith:
+1 merge.

PR #97, opened 2011-05-12
Reuse clientID when doing a delete to prevent clientID explosion

Vagabond:
Currently we allocate a new client ID when we write the tombstone. This is not necessary. However, we still do use a new clientID for the async get.


jonmeredith:
+1 merge, testing with https://github.com/basho/riak_kv/pull/96 already applied.

PR #98, opened 2011-05-16
Az289 listkeys refactoring

kellymclaughlin:
This pull request represents a substantial refactoring of the code that handles key listing. The goal was to simplify the code and also make it more robust such that if the initial attempt at key listing fails due to a node being down there is at least an attempt to retry the operation using different nodes. This is the first step towards generalizing the type of behavior required by operations like key listing so that in the future operations that require similar _coverage_ behavior will not have to be hacked onto the key listing code. Some of the major changes are as follows:
- `riak_kv_keylister_master` has been removed and `riak_kv_keylister` processes are started directly via `riak_kv_keylister_sup`.
- `riak_kv_keys_fsm` has been refactored to be stylistically similar to the get and put fsms. They now also include `test_link` function that I copied from the get fsm. Hopefully they will be used to facilitate qc tests for key listing in the near future.
- The use of bloom filters in `riak_kv_keylister` have been removed.
- The bloom filter in `riak_kv_keys_fsm` that was used to determine key uniqueness has been removed. Instead we now rely on the characteristics of the ring and filtering based on VNode positions in preference lists to ensure that the list of keys returned contains only unique entries.
- The term coverage is used to represent the type of behavior needed by key listing. This is an effort to move towards generalization. 
- The term position is used in the code to indicate positions in a preference list. _Position_ was used to avoid confusion with the term _index_ used in reference to VNodes.
- If the initial coverage plan fails due a node being down or the attempt to start a `riak_kv_keylister` process failing, a attempt is made to calculate and execute one or more backup plans so that the requested operation might still succeed within the timeout. The number of retries that are attempted and the characteristics (i.e. number of nodes and vnodes involved) of those backup plans is dependent upon the number of partition, the bucket n_val, the number of total physical nodes, and the number of available physical nodes. 

I have attempted to provide inline documentation to explain the impetus for different steps in the coverage plan calculation, but there may be things that did not make it out of my head and into the comments. Please make requests for additional comments if something is unclear or it otherwise seems appropriate. 

There are still things that can be optimized and improved, but since this is a fairly substantial change I wanted to go ahead and get more sets of eyes on it. I would like to get **at least 2** different reviewers on this please.


slfritchie:
Kelly/other reviewers, what changes and/or additions to basho_expect would be
worthwhile?  From what I can see:
- no basho_expect test is currently explicitly testing listkeys
- several basho_expect tests indirectly test listkeys, i.e. map/reduce tests by
  PHP/Python/JS client unit tests that are run by b_e)

The client library unit tests are probably sufficient only for finding gross regressions.
Those client lib tests have no awareness of how many Riak nodes are in the cluster
that they're talking to.  I suggest that it's unreasonable to modify those tests for
cluster awareness -- if anyone disagrees, feel free to hack the code.

To add explicit listkeys tests in a test that is indeed aware of multiple nodes, basho_exect
has a few such tests: from the `RUN-systests.sh` script:

```
for test in ./Run-systest-cluster-membership \
            ./Run-regression-753 \
            ./Run-systest-backup-and-restore \
            ./Run-systest-riak-repl-2x2 
```

The cluster membership systest is likely a good place to start ... or at least
a good place to start thinking about how a test with individual Riak nodes
fail and recover while listkeys operations are ongoing.

```
egrep '\.(step|prereq)' basho_expect/*cluster*py | sed 's/.*\.py:  *//'
```

NOTE: there are two spaces in the `sed` regexp.

kellymclaughlin:
I totally agree about not modifying the existing unit tests for cluster awareness. I will look at the tests you point out as a starting place for a listkeys test and then probably have to come asking for help.

slfritchie:
A couple of other notes:
- There aren't any EUnit tests for listkeys code
- Dialyzer has a couple of comments:

```
     riak_kv_keys_fsm.erl:94: Invalid type specification for function riak_kv_keys_fsm:start_link/5. The success typing is (non_neg_integer(),binary() | {binary(),fun()} | {'filter',binary(),[{atom(),atom(),byte()}]},'infinity' | non_neg_integer(),atom(),pid()) -> 'ignore' | {'error',_} | {'ok',pid()}
     riak_kv_keys_fsm.erl:573: The pattern <List, 0> can never match the type <[[{_,_}]],1>
```

slfritchie:
list_buckets filtering appears to be sub-optimal.  On a freshly installed, empty data
cluster, if I do the following, I see far more than a single bucket:

```
{ok, C} = riak:local_client().
[ok = C:put(riak_object:new(<<"b">>, list_to_binary([$k, integer_to_list(X)]), <<"val!">>)) || X <- lists:seq(1,10*1000)].
length(element(2, C:list_buckets())).
```

slfritchie:
When I use this patch to riak_kv_vnode.erl, I see a timeout error when I try using
list_buckets or list_keys.  Is the new code supposed to handle this kind of crash/interruption?

```
> {ok, C} = riak:local_client().
> [ok = C:put(riak_object:new(<<"b">>, list_to_binary([$k, integer_to_list(X)]), <<"val!">>)) || X <- lists:seq(1,10*1000)].
> timer:tc(C, list_keys, [<<"b">>, 70*1000]).
{8772235,{error,timeout}}
> timer:tc(C, list_keys, [<<"b">>, 3*70*1000]).
{26275896,{error,timeout}}

--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -439,8 +439,10 @@ do_list_bucket(ReqID,Bucket,Mod,ModState,Idx,State) ->
 %% @private
 do_list_keys(Caller,ReqId,Bucket,Idx,Mod,ModState) ->
     F = fun({_, _} = BKey, _Val, Acc) ->
+true = (BKey /= {<<"b">>, <<"k1">>}),
                 process_keys(Caller, ReqId, Idx, Bucket, BKey, Acc);
            (Key, _Val, Acc) when is_binary(Key) ->
+true = (Key /= <<"k1">>),
                 %% Backend's fold gives us keys only, so add bucket.
                 process_keys(Caller, ReqId, Idx, Bucket, {Bucket, Key}, Acc)
         end,
```

kellymclaughlin:
- I will work on some eunit tests.
- The bucket listing is broken right now. I just have to decide the least worst way to fix it at this juncture.
- I would not expect the code to currently handle the failure you introduced with the patch to riak_kv_vnode. Right now it will handle node failures until all of the key listings have been initiated. I wanted to get a sanity check on my approach before I took the error handling any further. The best way I found to test the failures it should currently handle is to add a sleep call right before [this](https://github.com/basho/riak_kv/blob/51b3687ca5f38ff153ea38d66a61aeb92b813119/src/riak_kv_keys_fsm.erl#L502) line and halt a node involved in the operation from the console before the call to start_keylister.

beerriot:
I've just tried to list an N=3 bucket on a four node cluster, with one node down and got `insufficient_nodes_available`.  I'm sure that's "correct", but it's a bit surprising.  It would be nice to be able to get a list of what other nodes think are available.

To reproduce:
1. setup a 4-node cluster
2. write some keys

```
{ok, C} = riak:local_client().
[ C:put(riak_object:new(<<"lkt">>, list_to_binary(integer_to_list(N)), N), 3) || N <- lists:seq(1, 1000) ].
```
1. Kill a node

```
bin/riak stop
```
1. List keys

```
C:list_keys(<<"lkt">>).
```

Expected: a list of keys, possibly missing some entries.
Actual: the error `insufficient_nodes_available`

src/riak_kv_keys_fsm.erl
```diff
@@ -20,265 +20,569 @@
...
+%% ====================================================================
+%% Internal functions
+%% ====================================================================
 
-finish(StateData=#state{req_id=ReqId,client=Client,client_type=ClientType, listers=[]}) ->
+%% @private
+finish({error, Error}, StateData=#state{from={raw, ReqId, ClientPid}, client_type=ClientType}) ->
     case ClientType of
-        mapred -> 
+        mapred ->
```
jonmeredith, line 244: Update comment - could be here on a timeout too.

src/riak_kv_keys_fsm.erl
```diff
@@ -20,265 +20,568 @@
...
 
 %% @private
-handle_info({'EXIT', Pid, Reason}, _StateName, #state{client=Pid}=StateData) ->
-    {stop,Reason,StateData};
-handle_info({_ReqId, {ok, _Pid}}, StateName, StateData=#state{timeout=Timeout}) ->
-    %% Received a message from a key lister node that 
-    %% did not start up within the timeout. Just ignore
-    %% the message and move on.
-    {next_state, StateName, StateData, Timeout};
-handle_info(_Info, _StateName, StateData) ->
```
jonmeredith, line 259: I think these functions can do the rotation without as much recursion/sorting.  You'd have to check them though - not even sure if they'll compile.

left_rotate(List, Rotations) ->
    {T, H} = lists:split(Rotations, List),
    H ++ T.

right_rotate(List, Rotations) ->
    {T, H} = lists:split(length(List) - Rotations, List)
    H ++ T.

src/riak_kv_keys_fsm.erl
```diff
@@ -20,265 +20,448 @@
...
-                luke_flow:add_inputs(Client, [{Bucket,K} || K <- Acc])
-            catch _:_ ->
-                    exit(self(), normal)
-            end;
-        plain -> Client ! {ReqId, {keys, Acc}}
-    end,
-    ok;
-process_keys([K|Rest],Bucket,ClientType,Bloom,ReqId,Client,Acc) ->
-    case ebloom:contains(Bloom,K) of
+compare_final_vnode({CA,VA}, {CB, VB}) ->
```
kellymclaughlin: The purpose of this function is to optimize the physical node selection in the case where the number of partitions is not evenly divisible by the bucket n_val and either the bucket n_val or the number of nodes is evenly divisible by the other. Using only the compare_next_vnode function for all vnode selection results in an additional physical node being used in the coverage plan in these cases. I felt this was a worthwhile optimization and I have not seen any adverse effects for other cases.

src/riak_kv_keys_fsm.erl
```diff
@@ -20,265 +20,438 @@
...
+                                               end,
+                                               {CoverageVNode, Acc1};
+                                           false ->
+                                               {CoverageVNode, Acc}
+                                       end
+                               end,
+            {CoverageVNodes, FilterVNodes} = lists:mapfoldl(CoverageVNodeFun, [], CoveragePlan),
+            NodeIndexes = group_indexes_by_node(CoverageVNodes, []),
+            {NodeIndexes, FilterVNodes, length(CoverageVNodes)};
+       {insufficient_vnodes_available, _}  ->
```
jonmeredith: The case should be {insufficient_vnodes_available, _KeySpace, _Coverage} - find_coverage returns a 3-tuple.

jonmeredith:
Kelly/I discussed this stuff over the phone, but here are some notes for the pull-request-record.

There are four handle_info variants for the non-bucket listing case.  They each do variants of filtering and create functions per lk message handled.  Instead we could create a function per-Idx that does all necessary filtering and is compiled just once.

```
Filter = lists:keyfind(Idx, 1, IdxFilters),
case [BKey || BKey <- BKeys, Filter(BKey)] of
   [] ->
     ok;
   FilteredKeys ->
     gen_fsm:send_event(Caller, {ReqId, {kl, Idx, node()}, FilteredKeys})
end
```

Of course you have to create the filter function, but it only has to happen once.   Ideally a lot of work that is done in the keylister now should be moved into a filter function on list keys inside the k/v vnode.  The filter function will be useful when we perform that refactor so it's a good evolutionary change.

The keyspace filtering function could be optimized to work out the first vnode index responsible for the bkey just using the number of partitions rather than computing the preflist.  This would be a good addition to riak_core_ring and is probably generally useful.

riak_kv_keys_fsm sets the process_flag and handles an 'EXIT' message that never comes from the client.  It doesn't need to do that any more and should create a monitor against all the keylisters it creates so it can see if they die rather than waiting for the timeout.  If the spinning up keylisters / starting the coverage plan were split into two states in the FSM then you could check for 'DOWN' messages and replan before you get to the execute phase - and fail early if it happens during listing (resubmitting to an alternate vnode for the failed keyspace would probably be bad as you may get duplicate bkeys in the result).

`list_keys/2` and `waiting({listkeys, Vnode}` are not called and can be removed if they aren't used during rolling upgrades.

Check what happens if discovering a downed node changes the coverage plan and a vnode is no longer used for coverage - does it shut down correctly, will it have start sent?  Perhaps the vnodes / filtering should be sent with the start message rather than updated separately.

kellymclaughlin:
I have made changes to address all of the things in Jon's previous comment except adding the new function to riak_core_ring. That is deferred until this pull request is merged and then it will be handled as a separate task.

src/riak_kv_keys_fsm.erl
```diff
@@ -20,265 +20,468 @@
...
+                                       VNodeIndex = (Position rem PartitionCount) * RingIndexInc,
+                                       Node = riak_core_ring:index_owner(Ring, VNodeIndex),
+                                       CoverageVNode = {VNodeIndex, Node},
+                                       case length(KeySpaces) < NVal of
+                                           true ->
+                                               %% Get the VNode index of each keyspace to
+                                               %% use to filter results from this VNode.
+                                               KeySpaceIndexes = [(((KeySpaceIndex+1) rem PartitionCount) * RingIndexInc) 
+                                                                  || KeySpaceIndex <- KeySpaces],
+                                               case proplists:get_value(Node, Acc) of
```
jonmeredith: orddict:append(VNodeIndex, KeySpaceIndices, Acc) would save a few lines here.

src/riak_kv_keylister.erl
```diff
@@ -71,55 +105,170 @@ handle_event(_Event, StateName, State) ->
...
+    %% use that to determine the preference list to
+    %% use in filtering the keys from this VNode.
+    ChashKey = riak_core_util:chash_key({Bucket, Key}),
+    hd(riak_core_ring:preflist(ChashKey, Ring)).
+
+%% @private
+build_vnode_filter_check(FilterVNodes, PrefListFun) ->
+    %% Create a function to check which VNodes should be filtered.
+    fun({Index, _}, Filters) ->
+            case proplists:get_value(Index, FilterVNodes) of
```
jonmeredith: If you pass the vnode index into the builder function you can do the proplists:get_value _once_ at filter generation time so you don't have to do it per-key.

jonmeredith:
I think there is something wrong with the keyspace mapping.  I have a 6 node / 16 partitions cluster set up with ownership set as

<pre>
[{0,'dev1@127.0.0.1'},
 {91343852333181432387730302044767688728495783936,  'dev2@127.0.0.1'},
 {182687704666362864775460604089535377456991567872,  'dev3@127.0.0.1'},
 {274031556999544297163190906134303066185487351808,  'dev4@127.0.0.1'},
 {365375409332725729550921208179070754913983135744,  'dev6@127.0.0.1'},
 {456719261665907161938651510223838443642478919680,  'dev1@127.0.0.1'},
 {548063113999088594326381812268606132370974703616,  'dev2@127.0.0.1'},
 {639406966332270026714112114313373821099470487552,  'dev3@127.0.0.1'},
 {730750818665451459101842416358141509827966271488,  'dev4@127.0.0.1'},
 {822094670998632891489572718402909198556462055424,  'dev6@127.0.0.1'},
 {913438523331814323877303020447676887284957839360,  'dev1@127.0.0.1'},
 {1004782375664995756265033322492444576013453623296,  'dev2@127.0.0.1'},
 {1096126227998177188652763624537212264741949407232,  'dev3@127.0.0.1'},
 {1187470080331358621040493926581979953470445191168,  'dev4@127.0.0.1'},
 {1278813932664540053428224228626747642198940975104,  'dev5@127.0.0.1'},
 {1370157784997721485815954530671515330927436759040,  'dev1@127.0.0.1'}]
</pre>


I've written 32 keys to is using <pre>systest:write(32)</pre> and checked the keys are there with <pre>systest:listkeys(1, 32, <<"systest">>).</pre>.  With all nodes up I get correct results, but with dev4/5/6 down I get missing keys

<pre>(dev1@127.0.0.1)90> systest:listkeys(1, 32, <<"systest">>).
[<<0,0,0,3>>,
 <<0,0,0,6>>,
 <<0,0,0,7>>,
 <<0,0,0,9>>,
 <<0,0,0,10>>,
 <<0,0,0,12>>,
 <<0,0,0,13>>,
 <<0,0,0,14>>,
 <<0,0,0,17>>,
 <<0,0,0,19>>,
 <<0,0,0,20>>,
 <<0,0,0,23>>,
 <<0,0,0,24>>,
 <<0,0,0,25>>,
 <<0,0,0,26>>,
 <<0,0,0,30>>,
 <<0,0,0,31>>,
 <<0,0,0,32>>]
</pre>


Using the tracer to find the plan it is trying to use I think it is valid

<pre>

(<0.17777.1>) returned from riak_kv_keys_fsm:find_coverage/3 -> {ok,
                                                                 [{2,[0,1,15]},
                                                                  {5,[2,3,4]},
                                                                  {7,[5,6]},
                                                                  {10,[7,8,9]},
                                                                  {12,[10,11]},
                                                                  {15, [12,13,14]}]}
</pre>


To map those keyspaces to the chash ranges the vnode is responsible for you should use

<pre>
Q = 16.
RingWedge = RingTop / Q.
FirstChash = <<(KeySpace * RingWedge):160>>.
LastChash = <<((KeySpace+1)*RingWedge) - 1):160>>.
</pre>


Something is wrong somewhere...

kellymclaughlin:
I am unable to replicate the keyspace problem. I used a devrel cluster with your ring files and manually assigned the nodes to the same pattern in my vm cluster. In both cases I brought down nodes 4,5, and 6 and I have only gotten the correct results in all attempts on both clusters.

Here's some example output I got resulting from the same coverage plan on the devrel cluster:

<pre><code>
(dev1@127.0.0.1)40> {ok,Res}=C:list_keys(<<"bucket">>).
CoveragePlan: [{2,[0,1,15]},
               {5,[2,3,4]},
               {7,[5,6]},
               {10,[7,8,9]},
               {12,"\n\v"},
               {15,[12,13,14]}]
{ok,[<<"22">>,<<"20">>,<<"16">>,<<"23">>,<<"15">>,<<"14">>,
     <<"8">>,<<"11">>,<<"25">>,<<"9">>,<<"5">>,<<"12">>,<<"30">>,
     <<"6">>,<<"28">>,<<"1">>,<<"21">>,<<"7">>,<<"13">>,<<"31">>,
     <<"32">>,<<"29">>,<<"3">>,<<"19">>,<<"2">>,<<"4">>,<<...>>|...]}
(dev1@127.0.0.1)41> length(Res).                       
32
</code></pre>


I loaded data using this statement:

`[C:put(riak_object:new(<<"bucket">>, list_to_binary(integer_to_list(X)), <<"val">>)) || X <- lists:seq(1,32)].`

jonmeredith:
My fault - set backend to ETS like an idiot.  Transient data is transient.

jonmeredith:
I've been playing around with the PBC interface today.  If there are insufficient nodes available it sends the error message  but there is no handler for it in riak_kv_pb_socket:handle_info.

<pre>
(<0.888.0>) returned from riak_kv_keys_fsm:finish/2 -> {stop,normal,
                                                        {state,<<"systest">>,
                                                         plain,
                                                         {raw,19007209,
                                                          <0.474.0>},
                                                         {<<"systest">>,[]},
                                                         undefined,0,60000}}
</pre>

jonmeredith:
riak_client:stream_list_keys crashes when run as

<pre>
systest:write(1)
{ok,C}=riak:local_client().
C:stream_list_keys(<<"systest">>).
3:56
flush().
</pre>

Looks like the filter function is called with the wrong arity.

jonmeredith:
Crashing case for June 07, 2011 is resolved for me.

jonmeredith:
I've been doing some load testing with basho bench.  The results look good and stable with listkeys running on it's own degrading nicely (i.e. not intermixed get/puts).  I did a test where I killed one of the nodes during a basho_bench run and some of the sockets dropped.  

<pre>
=ERROR REPORT==== 14-Jun-2011::03:27:29 ===
** State machine <0.24044.0> terminating 
** Last message in was {'EXIT',<10971.3657.0>,shutdown}
** When State == waiting_kl
**      Data  == {state,<<"test">>,plain,
                        {raw,90111964,<0.20346.0>},
                        <<"test">>,22,12,60000}
** Reason for termination = 
** {node_failure,shutdown}
</pre>


Which led me to

<pre>
handle_info({'EXIT', _Pid, Reason}, _StateName, StateData) ->
    {stop, {node_failure, Reason}, StateData};
</pre>


I think we should be calling finish/2 here rather than stop.  That way the client finds out early that they should resubmit the list keys.

jonmeredith:
Above change resolves the issue above.
Tested key filters/map reduce with a few nodes down.
Checked performance against current version with list keys and listing keys while under update load.  Performed faster than existing code without crashing.

+1 merge.

PR #100, opened 2011-05-17
bz://1095: result non-determinstic with allow_mult=false and siblings have same timestamp

jtuple:


jonmeredith:
Code looks good.

Could you add a couple of extra tests than demonstrate the correct behavior with different last modified stamps and prove the correct ordering with and without deletion involved.

jonmeredith:
New tests check out - merge at will.

jonmeredith:
Merged onto riak_kv-0.14 branch, commits

910bdb3955c73cba4aca191d9f870d29f4850807
9b2750a5770d54532b4d2955820f6c0b86d6b775
d5b08e1a906c0a0f1d92c5db1b2022a4ce1b6370

PR #101, opened 2011-05-23
bz1104 concurrent mapred errors

kellymclaughlin:
Fix problem of erroneous data being returned while running concurrent
MapReduce jobs.

Fixes: az432 bz1104

The riak_kv_js_vm module used a finite set of 25 integers appended to
to a variable name as a means to identify anonymous javascript functions
in order to reuse functions and make MapReduce processing more efficient.
The limit was set to 25 presumably to prevent too much resource consumption
by the JS VM.

The bug described by bz1104 is due reuse of this small set of ids and
results in unexpected code being executed and MapReduce results being
contaminated. The solution is to move the assembly of anonymous
javascript functions from erlang_js into the riak_kv_js_vm module.
This is does not decrease the efficiency of MapReduce execution because
each javascript function is compiled before each execution and
destroyed after execution. The only remnant of a previously executed
anonymous function was a local variable that could be reused. By moving
the function that assembles the anonymous functions to riak_kv_js_vm and
eliminating lookups for any previously defined functions the code is
simplified, there is no concern about id reuse, and there is no chance
of memory accumulation on the JS VM due to large numbers of anonymous
javascript functions being used.

Dan created a ruby script that easily demonstrates this issue and can be used to verify the fix. It is attached to the bugzilla ticket [here](https://issues.basho.com/show_bug.cgi?id=1104).

### DIFF
--- a/src/riak_kv_js_vm.erl
+++ b/src/riak_kv_js_vm.erl
@@ -232,7 +232,7 @@ code_change(_OldVsn, State, _Extra) ->
 
 %% Internal functions
 define_invoke_anon_js(JS, Args, #state{ctx=Ctx}=State) ->
-    JSFun =  define_anon_js(JS, Args, State),
+    JSFun =  define_anon_js(JS, Args),
     case invoke_js(Ctx, JSFun) of
         {ok, R} ->
             {{ok, R}, State};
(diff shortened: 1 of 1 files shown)


slfritchie:
Seems fine, with one small exception (see below): the Python system test passes
on my local machine.  I haven't subjected this to a full basho_expect regime, though.

It looks like the anon_funs member of the #state record is no longer being used
anywhere.  If so, it should be removed from the record definition.

PR #102, opened 2011-05-25
Add disable Nagle option for PB sockets

b:
Fix for 1108


russelldb:
Merged from a different branch. Many thanks.

PR #104, opened 2011-05-26
Erase the reference to a map executor in the case of a notfound

kellymclaughlin:
instead of decrementing the input counter.

Fixes: az445 bz1112

When a notfound error occurs during a map phase the phase creates a new map plan and retries using a difference preference list entry. A lingering reference to the original map phase executor used to track input counts was the source of the error. The solution is to remove the reference to the original map phase executor in the case of a notfound error instead of simply decrementing the associated counter.

A simple case to reproduce the original error and verify the fix is outlined [here](https://issues.basho.com/show_bug.cgi?id=1112).


beerriot:
Cherry-picking this commit to the riak_kv-0.14 branch, a new test in mapred_verify (https://github.com/basho/mapred_verify/commit/056454e29f82ad9f181aa6906ff8bf13b9e3826c) fails without this change, but succeeds with it.

+1 for merging to 0.14 branch

PR #106, opened 2011-05-26
Map phase retries on not_found causes "500 Internal Server Error" response (function_clause) when nodes are down

kellymclaughlin:
Catch exit with reason exhausted_preflist on the call to schedule_input after a nodedown message is received.

Fixes: az449 bz1113 bz1114

The schedule_input function in riak_kv_map_phase calls itself
in the case that it catches a nodedown exit while trying to
start mapper processes. If all the primary preference list
entries have been checked or their nodes are down this inner
call to schedule_input will result in an exit with reason
exhausted_preflist. Instead of just having the map phase
process exit, this change catches the exit and returns an
error to the caller of schedule_input so that notfound results
can be returned in the case of bz1113 or a more descriptive
error message can be return in the case of bz1114.

Steps to reproduce can be found [here](https://issues.basho.com/show_bug.cgi?id=1113) and [here](https://issues.basho.com/show_bug.cgi?id=1114). I did not add any testing to mapred_verify for this issue because the problems only show up on cluster setups where a portion of the nodes are down, but maybe this is a good test that basho_expect could handle.

### DIFF
--- a/src/riak_kv_map_phase.erl
+++ b/src/riak_kv_map_phase.erl
@@ -43,9 +43,13 @@ handle_input(Inputs0, #state{fsms=FSMs0, qterm=QTerm, mapper_data=MapperData}=St
     case length(Inputs1) > 0 of
         true ->
             ClaimLists = riak_kv_mapred_planner:plan_map(Inputs1),
-            {NewFSMs, _ClaimLists1, FsmKeys} = schedule_input(Inputs1, ClaimLists, QTerm, FSMs0, State),
-            MapperData1 = MapperData ++ FsmKeys,
-            {no_output, State#state{fsms=NewFSMs, mapper_data=MapperData1}};
+            case schedule_input(Inputs1, ClaimLists, QTerm, FSMs0, State) of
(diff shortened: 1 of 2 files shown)


beerriot:
Would you mind running mapred_verify, anyway?  I'm getting an error in the first "Erlang/Erlang, discrete entries" test, very similar to the error described in bz1114 after setting up a 3-node dev cluster (make stagedevrel) and stopping dev1 and dev3.  Could you check whether you see the same, or if I've misconfigured something?

The tests in the BZ tickets do pass, though.

kellymclaughlin:
The failure of the discrete entries test was actually a slightly different, but similar problem. I made another commit to address that issue, but in the case you describe failure is expected due to some of the inputs having preflists with entries from only 2 nodes including some whose entries are only on dev1 and dev3. The most recent commit will make the error message more indicative of what actually happened though. The behavior could be changed so that partial results could be returned in a case like this, but I am not sure if that's a desired behavior or not. I can see pros and cons to it.

beerriot:
Looks good to me, +1 merge

PR #107, opened 2011-05-27
Adds proper URL decoding to Riak's REST API

jtuple:
Addresses bz://617 and bz://1003.

New sane behavior where Riak URL decodes buckets/keys sent as URLs and Headers over REST API. Behavior is configurable through 'http_url_decoding' app.config parameter where 'on' defaults to new behavior, and missing/anything else defaults to compatibility mode where links are decoded but buckets/keys are not.

This pull request also includes migration support that can help users migrate an existing cluster with encoded bucket/keys to one without. This is handled in riak_kv_encoding_migrate.erl.

### DIFF
--- a/src/riak_kv_encoding_migrate.erl
+++ b/src/riak_kv_encoding_migrate.erl
@@ -1,0 +1,228 @@
+%% Support code to help migrate clusters with URL encoded buckets/keys.
+%%
+%% Usage:
+%% 1. Attach to a running riak console.
+%%
+%% 2. Check if cluster needs migration:
+%%      1> riak_kv_encoding_migrate:check_cluster().
+%%
+%% 3. If migration necessary and safe, migrate objects:
+%%      1> riak_kv_encoding_migrate:migrate_objects().
+%%
+%% 4. If all went well, delete migrated objects:
+%%      1> riak_kv_encoding_migrate:delete_migrated_objects().
+%%
+%% This module also provides the post-commit function 'postcommit_rewrite'
+%% that can be installed on a bucket to perform live copying of inserted
+%% objects. This could be used to perfrom live migration of new objects
+%% while the above migration process is running in background copying
+%% existing objects.
+
+-module(riak_kv_encoding_migrate).
+-export([check_cluster/0, migrate_objects/0, delete_migrated_objects/0]).
+-export([map_check/3, reduce_check/2, map_rewrite_unsafe/3,
+         postcommit_rewrite/1, map_delete_unsafe/3]).
+-export([test_migration/0]).
+
+%% Check if the cluster contains unsafe values that need to be migrated
+check_cluster() ->
+    {ok, RC} = riak:local_client(),
+    riak_kv_mapred_cache:clear(),
+    {ok, Buckets} = RC:list_buckets(),
+    case Buckets of
+        [] ->
+            io:format("Cluster is empty. No migration needed.~n", []),
+            {ok, empty};
+        _ ->
+            MR = [RC:mapred(Bucket, [map_check(), reduce_check()])
+                  || Bucket <- Buckets],
+            {ok, [Res]} = lists:foldl(fun erlang:max/2, 0, MR),
+            case Res of
+                0 ->
+                    io:format("Cluster does not contain URL unsafe values. "
+                              "No migration needed.~n", []),
+                    {ok, not_needed};
+                1 ->
+                    io:format("Cluster contains URL unsafe values. "
+                              "Migration needed.~n", []),
+                    {ok, needed};
+                2 ->
+                    io:format("Cluster contains URL unsafe values. However,~n"
+                              "one or more bucket/key exists in both URL~n"
+                              "encoded and non-encoded form. Custom migration "
+                              "necessary.~n", []),
+                    {ok, custom}
+            end
+    end.
+
+map_check() ->
+    {map, {modfun, ?MODULE, map_check}, none, false}.
+map_check(RO, _, _) ->
+    case check_object(RO) of
+        safe ->
+            [0];
+        {_, _, B2, K2} ->
+            {ok, RC} = riak:local_client(),
+            case RC:get(B2, K2) of
+                {error, notfound} ->
+                    [1];
+                _ ->
+                    [2]
+            end
+    end.
+
+reduce_check() ->    
+    {reduce, {modfun, ?MODULE, reduce_check}, none, true}.
+reduce_check(L, _) ->
+    [lists:foldl(fun erlang:max/2, 0, L)].
+
+%% Perform first phase of migration: copying encoded values to
+%% unencoded equivalents.
+migrate_objects() ->
+    {ok, RC} = riak:local_client(),
+    riak_kv_mapred_cache:clear(),
+    {ok, Buckets} = RC:list_buckets(),
+    [RC:mapred(Bucket, [map_rewrite_unsafe()]) || Bucket <- Buckets],
+    io:format("All objects with URL encoded buckets/keys have been copied to "
+              "unencoded equivalents.~n", []).
+
+map_rewrite_unsafe() ->
+    {map, {modfun, ?MODULE, map_rewrite_unsafe}, none, true}.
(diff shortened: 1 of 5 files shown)


src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,228 @@
```
beerriot, line 0: This file needs a license header.

test/rest_url_encoding_test.erl
```diff
@@ -0,0 +1,241 @@
```
beerriot, line 0: This file should probably have a license header as well.  Some in test/ do, others don't.

test/rest_url_encoding_test.erl
```diff
@@ -0,0 +1,241 @@
...
+    httpc:request(put, {URL ++ "/riak/rest_links/compat",
+                        [{"Link", "</riak/rest1/%40compat>; riaktag=\"tag\""}],
+                        "text/plain",
+                        "Test"},
+                  [], []),
+
+    %% Retrieve link header and check that it is singly encoded, corresponding
+    %% to a decoded internal link.
+    {ok, {_, Headers, _}} = 
+        httpc:request(get, {"http://localhost:8091/riak/rest_links/compat",
```
beerriot: change `"http://localhost:8091` to `URL ++ "`, and in other places below (or test fails for non-dev1:8091 nodes)

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,228 @@
...
+check_cluster() ->
+    {ok, RC} = riak:local_client(),
+    riak_kv_mapred_cache:clear(),
+    {ok, Buckets} = RC:list_buckets(),
+    case Buckets of
+        [] ->
+            io:format("Cluster is empty. No migration needed.~n", []),
+            {ok, empty};
+        _ ->
+            MR = [RC:mapred(Bucket, [map_check(), reduce_check()])
```
beerriot: This MapReduce query is going to read every object stored in the cluster.  Since check_object/1 only needs the bucket and key to make its decision, this request could be converted to reduce-only (no map phase), and not have to read any objects from disk.  The reduce phase will receive a list of bucket/key pairs, and can operate directly on them.  Most of an example of doing this can be found here: https://github.com/basho/riak_kv/commit/b5e969008343e466f245a2faadf1f364b7cba915

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,228 @@
...
+    end.
+
+reduce_check() ->    
+    {reduce, {modfun, ?MODULE, reduce_check}, none, true}.
+reduce_check(L, _) ->
+    [lists:foldl(fun erlang:max/2, 0, L)].
+
+%% Perform first phase of migration: copying encoded values to
+%% unencoded equivalents.
+migrate_objects() ->
```
beerriot: IMO, it would be useful to have this log the list of objects it converts, whether it do so in a file, or by returning the list of converted keys, or by just printing them on the console.  It would help with troubleshooting later know have a record of which things converted, and how.

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,228 @@
...
+    case check_object(RO) of
+        safe ->
+            ok;
+        {_, _, B2, K2} ->
+            copy_object(RO, B2, K2),
+            ok
+    end.
+
+%% Perform second phase of migration: delete objects with encoded buckets/keys.
+delete_migrated_objects() ->
```
beerriot: Same as migrate_objects: logging would be a good idea here.

jtuple:
Forgot to mention this earlier: my last set of commits should address all review items.

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,282 @@
...
+                      "necessary.~n", []),
+            {custom, []};
+        {_, true} ->
+            io:format("Cluster contains doubly encoded values. Custom "
+                      "migration necessary.~n", []),
+            {custom, []}
+    end.
+
+%% Returns a list of URL encoded objects needing migration.
+get_encoded_objects(RC) ->
```
beerriot: Let's export a zero-arity form of this function as well, for people to use  if they hit the `{custom, []}` cases above.  I might suggest naming it `get_encoded_keys` as well, just to not confuse people when they don't get `riak_object` records back.

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,282 @@
...
+
+%% Determine if it is safe to perform migration (no bucket/key conflicts).
+check_safe(RC, EObjs) ->
+    EObjs2 = [decode_name(Name) || Name <- EObjs],
+    MR = RC:mapred(EObjs2, [map_safe(),
+                            riak_kv_mapreduce:reduce_set_union(true)]),
+    case MR of
+        {ok, [unsafe]} ->
+            unsafe;
+        _ ->
```
beerriot: This will return `safe` if MR is an error tuple.

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,282 @@
...
+            safe
+    end.
+
+map_safe() ->
+    {map, {modfun, ?MODULE, map_safe}, none, false}.
+map_safe({error, notfound}, _, _) ->
+    [];
+map_safe(_, _, _) ->
+    [unsafe].
```
beerriot: It might be more user-friendly to return `[{riak_object:bucket(O), riak_object:key(O)}]` here.  Skip the reduce phase in `check_safe/2`, and match on `[]` for the safe case, but return the list of unsafe keys for a user to comb through, without having to repeat this dig manually.

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,282 @@
...
+    case {B2, K2} of
+        {B1, K1} ->
+            Objs;
+        _ ->
+            [{B1, K1} | Objs]
+    end.
+
+%% Returns true if any value is double (or more) encoded.
+check_double_encoding(EObjs) ->
+    lists:any(fun(Name) ->
```
beerriot: It might be more user-friendly to `lists:filter` here instead, so `check_cluster/2` can return a list of the keys it doesn't know how to migrate, instead of just saying that some exist.

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,282 @@
...
+    io:format("All objects with URL encoded buckets/keys have been copied to "
+              "unencoded equivalents.~n", []).
+
+map_rewrite_encoded() ->
+    {map, {modfun, ?MODULE, map_rewrite_encoded}, none, true}.
+map_rewrite_encoded({error, not_found}, _, _) ->
+    [];
+map_rewrite_encoded(RO, _, _) ->
+    {_, _, B2, K2} = decode_object(RO),
+    copy_object(RO, B2, K2),
```
beerriot: Probably worth checking the return of `copy_object/3`.  If it's not `ok`, maybe return the bucket/key as the result, so the user can check up on failures?

src/riak_kv_encoding_migrate.erl
```diff
@@ -0,0 +1,282 @@
...
+            ok;
+        _ ->
+            copy_object(RO, B2, K2),
+            ok
+    end.
+
+%% Perform second phase of migration: delete objects with encoded buckets/keys.
+delete_migrated_objects(EObjs) ->
+    {ok, RC} = riak:local_client(),
+    [RC:delete(B, K) || {B, K} <- EObjs],
```
beerriot: Probably worth checking that all deletes returned `ok`.

jtuple:
Latest commit should address the remaining issues. There are a few cases were we can't really recover gracefully, so I'm just letting a match fail (eg. map_unsafe when the MR job fails), but otherwise I believe we handle all cases where it makes sense to provide the user with "bad objects" that can be dealt with manually.

beerriot:
+1 merge

PR #109, opened 2011-06-03
2I Development - AZ456, AZ457

rustyio:
To try this code locally, clone http://github.com/basho/riak, switch to branch "rtk-dev1", and build. This sets up a Riak source tree using the rtk-dev1 branch of riak_kv, and master branch of merge_index.
# Update Riak REST interface to extract Index headers and store in object metadata. (AZ456)

Created a malformed_index_headers/2 function to pull any headers prefixed with "x-riak-index-" from the request. Storing the headers in the object metadata under a slot called "index".

Some points for consideration:
- We _could_ have have made this "x-riak-meta-index-" in order to piggyback on existing metadata. Decided against it because:
  1. That would change the meaning of existing fields.
  2. That could causes problems if we ever want to store index fields somewhere other than object metadata.
- Do we want to use dash or underscore for field separators???
  - Using dashes fits in with existing header names.
  - Using underscores makes more sense for field names, especially later when we do some sort of query language. Field names will need some sort of delimiter, since we plan on using Hungarian notation to denote field types.
- We store the index_fields as unparsed lists:
  - This makes it easier to return return the original field to the user on read. (There is some pain in Riak Search around converting type fields back into the original values on the way out. Doesn't always work as expected.)
  - But, it means that we need to reparse the types when we need them later. This will happen at least once in the Put FSM, again in the VNode, and again if we use the metadata for filtering.
# Validate indexed fields in the Put FSM. (AZ457)

In riak_kv_put_fsm:validate/2, added a pre-commit hook that always runs and is responsible for pulling fields out of metadata and  ensuring that all of the fields parse. 

If a field fails to parse, then the hook fails just like a normal pre-commit hook, returning {fail, [Reasons]}, where Reason is one of:
- {unknown_field_type, Field}
- {field_parsing_failed, {Field, Value}}.

The pre-commit hook is stored in riak_index:validate_object/1. It calls into riak_index:validate_field/2. Types are determined based on suffix. Current types are *_id, *_int, *_float. Currently, a mapping of regular expressions to field types determines which function is used to parse the field:

```
"*_id" -> fun parse_id/1
"*_int" -> fun parse_integer/1
"*_float" -> fun parse_float/1
```

```
 # Store an object with field types...
 curl -v -X PUT \
 -d '{"bar":"baz"}' \
 -H "x-riak-index-field1_id: A" \
 -H "x-riak-INDEX-field2_int: 1" \
 -H "x-Riak-INDEX-field3_float: 3.14" \
 http://127.0.0.1:8098/riak/test/doc

 # Retrieve the object...
 curl -i http://127.0.0.1:8098/riak/test/doc?returnbody=true
```

### DIFF
--- a/src/riak_index.erl
+++ b/src/riak_index.erl
@@ -1,0 +1,337 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_index: central module for indexing.
+%%
+%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc central module for indexing.
+
+-module(riak_index).
+-export([
+         validate_object/1, 
+         validate_fields/1,
+         format_failure_reason/1
+        ]).
+
+%% -ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+%% -endif.
+
+-include("riak_kv_wm_raw.hrl").
+
+%% @type data_type_defs() = [data_type_def()].
+%% @type data_type_def() = {CompiledRegex:re:mp(), Module:module()}.
+
+%% @spec validate_object(riak_object:riak_object()) -> 
+%%         riak_object:riak_object() | {fail, term()}.
+%%
+%% @doc Validate the index fields stored in object metadata. Conforms
+%%      to the pre-commit hook interface. Return the unmodified object
+%%      if validation was successful, or {fail, [Reasons]} if validation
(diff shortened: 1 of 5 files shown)


src/riak_index.erl
```diff
@@ -0,0 +1,337 @@
...
+       {false, {field_parsing_failed, {"field_int", "4.56"}}}, 
+       F("field_int", "4.56")),
+
+    ?assertMatch(
+       {false, {field_parsing_failed, {"field_int", ".789"}}}, 
+       F("field_int", ".789")).
+
+
+validate_float_test() ->
+    %% Test validation of "*_int" fields...
```
kellymclaughlin, line 257: Should be *_float here.

src/riak_index.erl
```diff
@@ -0,0 +1,337 @@
...
+    end;
+parse_float("") ->
+    %% Explicitly checking the empty string case because otherwise our
+    %% normalization would fix it into 0.0. This stays consistent with
+    %% parse_integer/1 above.
+    {error, "Missing value."}.
+
+
+%% ====================
+%% TESTS
```
kellymclaughlin, line 189: Should add an ifdef(TEST) around here.

src/riak_index.erl
```diff
@@ -0,0 +1,337 @@
...
+       {fail, [
+               {field_parsing_failed, {"field_int", "A"}},
+               {unknown_field_type, "field_foo"}
+              ]},
+       F([
+          {"field_id", "A"},
+          {"field_int", "A"},
+          {"field_foo", "fail"},
+          {"field_float", "0.5"}
+         ])).
```
kellymclaughlin, line 337: endif to match the ifdef(TEST) here

kellymclaughlin:
Pending fix for comment typo and adding ifdef/endif to wrap the riak_index tests it's good to merge.

PR #111, opened 2011-06-06
Az419 update delete interface

Vagabond:
Allow users to specify R/W/PR/PW/DW for deletes instead of the single parameter 'RW'. Compatibility with simply passing RW is maintained, but a list is also now accepted.

### DIFF
--- a/src/riak_kv_wm_raw.erl
+++ b/src/riak_kv_wm_raw.erl
@@ -1107,13 +1107,13 @@ ensure_doc(Ctx) -> Ctx.
 %% @spec delete_resource(reqdata(), context()) -> {true, reqdata(), context()}
 %% @doc Delete the document specified.
 delete_resource(RD, Ctx=#ctx{bucket=B, key=K, client=C, rw=RW, r=R, w=W,
-        pr=PR, pw=PW}) ->
+        pr=PR, pw=PW, dw=DW}) ->
     Result = case wrq:get_req_header(?HEAD_VCLOCK, RD) of
         undefined -> 
-            C:delete(B,K,[{rw, RW}, {r, R}, {w, W}, {pr, PR}, {pw, PW}]);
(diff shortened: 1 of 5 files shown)


src/riak_kv_wm_raw.erl
```diff
@@ -1359,3 +1284,68 @@ send_precommit_error(RD, Reason) ->
...
+    case {error, Reason} of
+        {error, precommit_fail} ->
+            {{halt, 403}, send_precommit_error(RD, undefined), Ctx};
+        {error, {precommit_fail, Reason}} ->
+            {{halt, 403}, send_precommit_error(RD, Reason), Ctx};
+        {error, {n_val_violation, N}} ->
+            Msg = io_lib:format("Specified w/dw/pw values invalid for bucket"
+                " n value of ~p~n", [N]),
+            {{halt, 400}, wrq:append_to_response_body(Msg, RD), Ctx};
+        {error, {pw_val_unsatisfied, Requested, Returned}} ->
```
rzezeski: Maybe move this case adjacent to the other x_val_unsatisfied cases?

rzezeski:
I realize this isn't directly related by the dialyzer target is broken.  After fixing it I found two errors.  Maybe someone should make a ticket?

rzezeski:
It looks like you can specify a `DW` > `N` and it will not cause a `400 Bad Requst`.  Same for `RW`.  That's for the HTTP interface.

The PB interface is almost the same but `RW` > `N` will fail **iff** the object has not already been deleted.

Vagabond:
Yes, I'll fix that, although its kind of irrelevant.

rzezeski:
`DW` > `N` and `RW` > `N` are still inconsistent w/ `W`, `PW, and`PR` but for the sake of moving on I'm dropping it.

+1 to merge

PR #112, opened 2011-06-08
2I: Implement RiakKV Index Backend (AZ458)

rustyio:
Copied riak_kv_index_backend from the Riak Index prototype, and implemented some minor adjustments for changes we've made in the design of Riak Index. (Changed the `riak_index:validate_*/N` functions to `riak_index:parse_*/N` with appropriate updates to return values.)

riak_kv_index_backend works similarly to multi-backend in that it starts up other backends (riak_kv_bitcask_backend, riak_index_mi_backend) and proxies off requests. It also adds some coordination logic around put and delete requests. Namely, on a put request, we delete any old index information, pull new index information from the incoming object, and write it to the index before writing to KV. We store the old index information in a "proxy" object inside of merge_index. This allows us to be sure we are cleaning up after the old version of the object appropriately. We store this information in merge_index so that we don't accidentally run into it  while doing a KV list keys.


kellymclaughlin:
Check out [this](https://gist.github.com/1017976) diff for riak_index. It eliminates the need for the expensive use of the regular expressions to parse the fields and allows the anonymous fun in parse_fields to become the field_parser function. I have not done any direct comparisons yet, but I suspect just doing a reverse on the key and a pattern match will be more efficient.

Update: I did some a few little timing tests with this and I think it will be a nice performance improvement.

kellymclaughlin:
The riak_index_mi_backend tests fail because there is no dependency entry for merge_index in rebar.config.

kellymclaughlin:
+1 to merge pending the change to rebar.config to add merge_index.

rustyio:
Added merge_index dependency, and updated field_types to just use lists:suffix for matching field names.

PR #113, opened 2011-06-10
Az424 listkeys qc test

kellymclaughlin:
Improves the listkeys qc test to generate buckets, bucket n_vals, timeout values, and different size object lists to store in the buckets. Also varies the input between a bucket name and a key filter input tuple.

### DIFF
--- a/test/keys_fsm_eqc.erl
+++ b/test/keys_fsm_eqc.erl
@@ -59,8 +59,13 @@ setup() ->
     error_logger:tty(false),
     error_logger:logfile({open, "keys_fsm_eqc.log"}),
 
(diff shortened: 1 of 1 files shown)


jonmeredith:
Test achieves good coverage.  It's an incremental improvement on what we have, so +1 merge. 

In the next round it would be good to add failure and partial failure coverage.  I doubt my testing managed to hit the partial keylisters case in start_keylisters and when I ran the tests it didn't generate any cases when it was unable to cover the keyspace.

kellymclaughlin:
Totally agree about the next round of improvements. Thanks.

PR #114, opened 2011-06-13
2I - Implement Querying via ListKeys (AZ459)

rustyio:
Add riak_client:get_index/2 function for running index queries. This function constructs a "fake" listkeys bucket of the form {index_query, Bucket, Query} and calls `Client:list_keys/N`. List keys eventually calls `riak_kv_index_backend:fold_bucket_keys/N` on riak_kv_index_backend with the provided "fake" bucket. (NOTE: This is temporary code, we will replace it once we finish the "covering cast" work in riak_core.)

  This calls the new riak_index_mi_backend:fold_index/6 function to perform the query. For the purposes of this Kanban ticket, Query will run against a single field, and is a tuple of the form {Operation, Field, Value} where Operation is 'eq', 'lt', 'lte', etc., Field is the string field, and Value is the parsed value. Future tickets will enable more complexity, and will do some sanity checking on Field Names.

  This change also adds bucket and key indexing with an update to riak_index:parse_object/1. The bucket field is "$bucket" and the key field is "$key".

  This round of changes deviates in a few key areas from the plan we discussed:
- SKFun accepts a batch of results, rather than a single result.
- SKFun results are of the form {PrimaryKey, Properties} rather than {SecondaryKey, PrimaryKey}, and properties contains an entry for {Field, SecondaryKey}. This was done to keep in line with what merge_index already produces which has served well for Riak Search and for early Riak Index prototypes.
- We create "special" fields for bucket and key ($bucket and $key, respectively) rather than exposing operations for 'lt_pk', 'gt_pk', etc. Seemed like a more elegant way to expose the functionality, and took fewer lines of code.
- Speaking of the "$bucket" field, it seems weird to have operations like `get_index(<<"mybucket">>, [{eq, "$bucket", <<"mybucket">>}])`. Queries are already scoped to the bucket. Should we instead support a `{bucket}` operation that reads all keys for the current bucket? 
- riak_kv_index_backend takes advantage of merge_index's ability to store large data by using it to store the "proxy" object, which is the collection of postings indexed for a particular object. This is used to cleanup old postings when objects are updated or deleted. Bitcask is probably a better storage engine for this kind of data, but should we store it in the bitcask instance for the current partition, or create a new bitcask instance? Alternatively, we could avoid storing a proxy object, or store the parsed fields as additional metadata, but that would require changes to riak_kv_vnode in order to send the old copy of the object to the backend during put/delete requests.

```
 # Store an object with field types...
 curl -v -X PUT \
 -d '{"bar":"baz"}' \
 -H "x-riak-index-field1_id: A" \
 -H "x-riak-INDEX-field2_int: 1" \
 -H "x-Riak-INDEX-field3_float: 3.14" \
 http://127.0.0.1:8098/riak/mybucket/mykey

 # Retrieve the object...
 curl -i http://127.0.0.1:8098/riak/mybucket/mykey?returnbody=true

 %% Query the index...
 {ok, Client} = riak:local_client().
 Client:get_index(<<"mybucket">>, [{eq, "$key", <<"mykey">>}]).
 Client:get_index(<<"mybucket">>, [{eq, "field2_int", 1}]).
```

### DIFF
--- a/src/riak_index_mi_backend.erl
+++ b/src/riak_index_mi_backend.erl
@@ -12,6 +12,7 @@
          index/2,
          delete/2,
          lookup_sync/4,
+         fold_index/6,
          drop/1,
          callback/3
         ]).
@@ -23,6 +24,15 @@
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
+
+%% MIN_VALUE and MAX_VALUE are used during range searches. MIN_VALUE
+%% is a number because floats sort smallest in Erlang. See
+%% http://www.erlang.org/doc/reference_manual/expressions.html. MAX_VALUE
+%% is undefined because merge_index understands undefined to mean no
+%% upper bound.
+-define(MIN_VALUE, -9223372036854775808). %% -1 * (2^63)).
+-define(MAX_VALUE, undefined).
+-define(SIZE, all).
 
 %% @type state() = term().
 -record(state, {partition, pid}).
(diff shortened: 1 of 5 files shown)


kellymclaughlin:
re: The question about using bitcask to store the proxy objects, might be good to start an email or wiki discussion with the whole group to get everyone's input on that, especially the alternative option that would require changing riak_kv_vnode.

kellymclaughlin:
I got a little tripped up on having to use a binary value for the $bucket and $key fields, but strings with regular fields. Maybe we should allow either string or binary for $key and $bucket or just make it string input for all fields.

src/riak_index_mi_backend.erl
```diff
@@ -101,6 +106,55 @@ lookup_sync(State, Index, Field, Term) ->
...
+                  [{gte, Field, Term}] ->
+                      merge_index:range_sync(Pid, Index, Field, Term, undefined, ?SIZE, FilterFun);
+
+                  _ -> 
+                      {error, {unknown_query_element, Query}}
+              end,
+    
+    %% If Results is a list of results, then call SKFun to accumulate
+    %% the results.
+    Results1 = case is_list(Results) of
```
kellymclaughlin, line 141: This doesn't handle error cases if an unknown query element is used or if merge_index:rang_sync returns an error. You could do more checks here or pass the Results to SKFun without examination and handle it there. If it remains that the only tuple that can be passed to SKFun is {results, Results} then some of the clauses in SKFun can be eliminated.

rustyio, line 141: Yep, this code is intentionally simple. Didn't want to add additional error handling until we merge in the latest list_keys code, or until we nail down the shape of riak_core covering functionality.

kellymclaughlin:
All issues have been discussed and are covered. +1 to merge.

PR #118, opened 2011-06-21
Az479 empty reduce

slfritchie:
Create special callback function in riak_kv_w_reduce,
no_input_run_reduce_once(), to assist riak_pipe_fitting's decision-making
process when a fitting receives an eoi.


beerriot:
Hooray, "Erlang/JS/Erlang" and "Erlang/JS/JS" tests work on pipe now!

+1 merge, thanks!

PR #122, opened 2011-06-28
Fix syntax error in last commit

russelldb:
Replace semicolon with fullstop to terminate function def


beerriot:
+1 merge, thank you!

PR #123, opened 2011-06-28
LevelDB driver for riak.

dizzyd:
Uses Ulf Wiger's sext-encoding driver to improve efficiency when
traversing buckets, per Kresten Krab's suggestion/prototype.

### DIFF
--- a/ebin/riak_kv.app
+++ b/ebin/riak_kv.app
@@ -19,6 +19,7 @@
              riak_kv_delete,
              riak_kv_delete_sup,
              riak_kv_dets_backend,
+             riak_kv_eleveldb_backend,
              riak_kv_ets_backend,
              riak_kv_fs_backend,
              riak_kv_gb_trees_backend,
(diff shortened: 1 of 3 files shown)


jonmeredith:
eleveldb:fold_loop - should it really just return acc on an invalid_iterator?
eleveldb:get - errors don't get passed back.  how does eleveldb report disk errors?
     shouldn't it call it->status().ok() after the search?

riak_kv_eleveldb_backend:put - should not match on 'ok' and pass errors back.
riak_kv_eleveldb_backend:delete - should not match on ok and pass errors back.
riak_kv_eleveldb_backend:drop -should not match on 'ok', although caller does
riak_kv_eleveldb_backend:is_empty - any chance a disk error would hide a populated db?

riak_kv_eleveldb_backend:list - not called by riak_kv_vnode any more, should remove from
                                other backends and behaviors.

dizzyd:
Incorporated most of the suggestions -- please re-review. Note you'll need to update the eleveldb dependency to get destroy arity fixes.

I need to update eleveldb to latest leveldb code which will affect fold_loop and get -- per Jon's comments. However, I don't think there's anything serious enough pending to stop the commit...just my $0.02. :)

jonmeredith:
+1 merge

PR #124, opened 2011-06-28
Az469 function inputs

slfritchie:
Add support for `{modfun, Mod, Fun, Arg}` tuple as input to a MapReduce
job (rather than bucket name or list of BKeys).

Examples of this use are sparse.  Further compatibility tests will likely
be tricky because the current implementation is very closely tied to the
`luke` app.  (Similarly, this implementation will be closely tied to the
`riak_pipe` app because the generator function is expected to call
`riak_pipe:queue_work()` and `riak_pipe:eoi()` directly.)

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -23,18 +23,19 @@
 %% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job
 %%
 %% An uncommonly-used option for Riak KV MapReduce is the option to
-%% use third third method of specifying inputs to the beginning of the
-%% MapReduce workflow.  All three methods are:
+%% use a fourth method of specifying inputs to the beginning of the
+%% MapReduce workflow.  All four methods are:
 %%
 %% <ol>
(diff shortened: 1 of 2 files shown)


src/riak_kv_mrc_pipe.erl
```diff
@@ -20,6 +20,37 @@
 
 %% @doc Riak KV MapReduce / Riak Pipe Compatibility
 %%
+%% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job
+%%
+%% An uncommonly-used option for Riak KV MapReduce is the option to
+%% use third third method of specifying inputs to the beginning of the
```
beerriot: "third third"?

src/riak_kv_mrc_pipe.erl
```diff
@@ -20,6 +20,37 @@
...
+%% MapReduce workflow.  All three methods are:
```
beerriot: There's a fourth method: keyfilters.  `{Bucket, KeyFilter}`

beerriot:
+1 merge, thanks!

PR #125, opened 2011-06-30
AZ512 - Split up riak_kv_wm_raw module, add new HTTP API routes.

rustyio:
Split riak_kv_wm_raw into smaller modules:
- riak_kv_wm_buckets - Handle listing buckets.
- riak_kv_wm_props   - Handles setting and getting bucket properties.
- riak_kv_wm_keylist - Handles listing keys in a bucket.
- riak_kv_wm_object  - Handles object reads and writes.
- riak_kv_wm_utils - Common utilities.

Add webmachine routes for HTTP API version 2:
- `/buckets?buckets=true` - List buckets.
- `/buckets/mybucket/props` - Access bucket properties props.
- `/buckets/mybucket/keys?keys=true` - List keys.
- `/buckets/mybucket/keys/mykey` - Access an object.
- `/buckets/mybucket/keys/mykey/*` - Linkwalking.

Tested with the PHP and Python Clients:

```
# Using PHP client...
git clone git@github.com:basho/riak-php-client.git
cd riak-php-client
php unit-test.php

# Using Python client...
git clone git@github.com:basho/riak-python-client.git
cd riak-python-client
python setup.py install
python setup.py test
```

Also, created some scripts to help test the old HTTP interface vs. the new HTTP interface. The following scripts simply run a bunch of `curl` commands. By piping the output to a file, and then diffing the file against the results against a different version of code, we can see if there were any breaking changes:
- [httptest_old.sh](http://rusty.basho.com.s3.amazonaws.com/AZ512/httptest_old.sh) - Exercise the old HTTP API.
- [httptest_new.sh](http://rusty.basho.com.s3.amazonaws.com/AZ512/httptest_new.sh) - Exercise the new HTTP API.

To repeat these tests, set up the latest code on port 8098, and set up the old code (0.14.2) on port 8091. Then, run the following:

```
# Capture output from the old API on the old code.
./httptest_old.sh 8091 oldbucket > oldbucket_8091.output 

# Capture output from the old API on the new code.
./httptest_old.sh 8098 oldbucket > oldbucket_8098.output 

# Capture output from the new API on the new code.
./httptest_new.sh 8098 newbucket > newbucket_8098.output

# Compare old API on old code vs. new code.
opendiff oldbucket_8091.output oldbucket_8098.output

# Compare old API vs. new API on new code.
opendiff oldbucket_8098.output newbucket_8098.output
```

The diff's won't be _exact_, but many of the lines should be the same. Most importantly, they should contain the same object data and links.

A few review points:
- If you want to test this locally, make sure you have commit '97d37db' of Webmachine, pull request https://github.com/basho/webmachine/pull/28 (AZ460). Switch to `deps/webmachine` then run `git pull; git checkout AZ460-guard-functions`.
- `riak_kv_wm_raw.hrl` should probably change to `riak_kv_wm.hrl`, but it is a dependency of Luwak, so I left it alone.
- `riak_kv_wm_keylist.erl` calls into `riak_kv_wm_props.erl` in order to support listing props and keys at the same time. If/when we remove the old API, we can get rid of this.
- Should we have a "links" part of the URL for linkwalking? In other words, this: `/buckets/bucket/keys/key/*`, or this: `/buckets/bucket/keys/key/links/*`?


src/riak_kv_wm_buckets.erl
```diff
@@ -0,0 +1,129 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_wm_buckets - Webmachine resource for listing buckets.
+%%
+%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
```
beerriot: Probably ought to say 2011 on these now.

src/riak_kv_wm_buckets.erl
```diff
@@ -0,0 +1,129 @@
...
+%%
+%% Available operations:
+%%
+%% GET /buckets?buckets=true (NEW) 
+%% GET /Prefix?buckets=true
+%%   Get information about available buckets. Note that generating the
+%%   bucket list is expensive, so we require the "buckets=true" arg.
+
+-module(riak_kv_wm_buckets).
+-author('Bryan Fink <bryan@basho.com>').
```
beerriot: Do me a favor and remove this unfortunate vanity, please.  Let the git logs speak for themselves.

src/riak_kv_web.erl
```diff
@@ -52,10 +53,55 @@ raw_dispatch() ->
...
+
+     {[Name, bucket, key],
+      riak_kv_wm_object, Props},
+
+     {[Name, bucket, key, '*'],
+      riak_kv_wm_link_walker, Props},
+
+     %% NEW API
+     {["buckets"],
+      riak_kv_wm_buckets, Props},
```
beerriot: Using `Props` in this manner means that all Link headers point to the old API URLs.

src/riak_kv_web.erl
```diff
@@ -52,10 +53,55 @@ raw_dispatch() ->
...
+
+     {["buckets", bucket, "props"],
```
beerriot, line 81: I feel like `/buckets/BUCKET` (in addition to `…/props` and `…/keys`) should also land somewhere useful, though I maybe could be swayed, depending on what is put in the Link header.

beerriot:
`riak_kv_wm_object:malformed_link_headers/2` and `:get_link_heads/2` need to be modified to describe and accept links specified as /buckets/BUCKET/keys/KEY.

beerriot:
I get one minor failure in the JS client tests.  Item 9, "2 siblings have correct bodies" under heading 5, "Storage module: Siblings".  It looks like there's an extra line break at the head of one of the bodies.

All other JS client tests pass.  I also converted the URLs in that client + tests to the new URLs, and the tests still pass.  I can push that branch up if you're interested.

beerriot:
Confirmed that PHP and Python tests succeed and fail in the same way on this branch as they do on master (failures in search integration tests).

seancribbs:
This will make it substantially difficult to provide parallel pre/post-1.0 client APIs. -1 on the URL format.

rustyio:
Updated link formatting based on @beerriot's comments. The `Props` proplist now contains an api_version attribute, we use that to determine which version of the API to use. Removed the extra newline that was breaking the JS tests. Those all pass now.

src/riak_kv_wm_link_walker.erl
```diff
@@ -130,6 +130,14 @@
          process_post/2
         ]).
 
+-import(riak_kv_wm_utils,
```
beerriot: It seems like you converted the other modules to not import from `riak_kv_wm_utils`, but this module gained these imports.  Yet the call to `multipart_encode_body` is not imported.  Maybe best to remove this import as well for consistency?

src/riak_kv_wm_utils.erl
```diff
@@ -0,0 +1,283 @@
...
+                io_lib:format("</~s/~s/~s>; riaktag=\"~s\"",
+                              [Prefix, Bucket1, Key1, Tag1]);
+            2 ->
+                io_lib:format("</buckets/~s/keys/~s>; riaktag=\"~s\"",
+                          [Bucket1, Key1, Tag1])
+        end,
+    format_links(Rest, Prefix, APIVersion, [{?HEAD_LINK, Val}|Acc]);
+format_links([], _Prefix, _APIVersion, Acc) ->
+    Acc.
+
```
beerriot, line 183: Looks like lots of unused code commented out below. Cut it loose for the git log to claim?

beerriot:
Except for the two minor code cleanup comments, I say +1 merge.

src/riak_kv_wm_utils.erl
rustyio, line 183: Removed commented out code.

PR #126, opened 2011-07-06
Az496 coverage code changes

kellymclaughlin:
This pull request is complementary to [this](https://github.com/basho/riak_core/pull/59) pull request for `riak_core` and the two should be reviewed together. It changes key and bucket listing operations to use the generalized coverage behavior as implemented in the `riak_core` pull request. It also makes updates to `riak_kv_vnode` to support the coverage behavior. 

The `riak_kv_keylister` module and its supervisor have been eliminated and all the required filtering is now being done directly by the vnodes.

Bucket listing is now done directly instead of a hack onto key listing. Two modules have been added to support this in `riak_kv`: `riak_kv_buckets_fsm` and `riak_kv_buckets_fsm_sup`.


beerriot:
+1 merge, very cool stuff

jonmeredith:
Tested rolling upgrades and that works now. Yay.

Do we need to rename the exisiting code _legacy_blah .  If you are using an older version of the riak_client code then the names of the supervisors will be wrong.  We could fix by using the original registered name for the supervisor instead of ?MODULE.

jonmeredith:
+1 merge.

PR #127, opened 2011-07-07
az://303 vclock enhancements - *DO NOT MERGE YET*

jonmeredith:
Hey all,

Here's a pull request for node-based vclocks.  It changes the put FSM to work the way it is described in the Dynamo paper and is resistant to vclock explosion (but not sibling explosion) in the face of poorly behaved clients / multiple updates on a partitioned cluster by a client that can see both sides / syntactic put merge.

The changes
- Require a member of the preflist to coordinate a put.
- Complete the put on the local vnode first using a modified vnode put that ensures a conflicting vclock with other nodes in the cluster (which must be true as only the local node can update that vclock).
- Issue the object put successfully put on the local vnode to the remote vnodes.
- Remove syntactic put merge.
- Add an eventual consistency test that verifies this method works correctly in the face of updates during continuous partitioning of the cluster.
- Update the existing put FSM test for new behavior.

Things still to do for put FSM:
- Add a way to generate a shorter vclock nodeid (maybe initially generate from an ethernet MAC then store/retrieve from the ring, need to handle multiple instances per physical node).
- May need to use a modified vclock nodeid for fallback nodes to ensure uniqueness.

Things to do to riak_kv before we can merge
- Make any long-running vnode operation run on another thread.  Operations are serialized on the local vnode, so (for N=3) it is no longer best two latencies out of three, it's latency of local vnode put + best one of two remote vnode put latencies.

For testing it requires riak_core az303-vclock-enhancements checked out for a minor tweak to riak_core_vnode_master:command.  All k/v eunit tests should pass.

Once those changes have been made we can merge this pull request into mainline, but not before.  I'm submitting it for review now so the logic can be checked.

Cheers, Jon

P.S. DO NOT MERGE THIS YET.

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -267,13 +267,13 @@ precommit(timeout, State = #state{precommit = [Hook | Rest], robj = RObj}) ->
                                                 precommit = Rest}, 0}
     end.
 
-execute_local(timeout, StateData0=#state{robj=RObj0, req_id = ReqId,
+execute_local(timeout, StateData0=#state{robj=RObj, req_id = ReqId,
                                          timeout=Timeout, bkey=BKey,
-                                         coord_idx_node = {_Index, Node} = CoordIndexNode,
+                                         coord_idx_node = {_Index, _Node} = CoordIndexNode,
                                          vnode_options=VnodeOptions,
                                          putcore=PutCore,
                                          starttime = StartTime}) ->
-    RObj = riak_object:increment_vclock(RObj0, Node, StartTime),
+    %RObj = riak_object:increment_vclock(RObj0, Node, StartTime),
     TRef = schedule_timeout(Timeout),
     riak_kv_vnode:coord_put(CoordIndexNode, BKey, RObj, ReqId, StartTime, VnodeOptions),
     StateData = StateData0#state{robj = RObj, tref = TRef},
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -39,7 +39,10 @@
 -endif.
 -export([init/1, handle_event/3, handle_sync_event/4,
          handle_info/3, terminate/3, code_change/4]).
--export([prepare/2, validate/2, precommit/2, execute/2, waiting_vnode/2, postcommit/2, finish/2]).
+-export([prepare/2, validate/2, precommit/2,
+         execute_local/2, waiting_local_vnode/2,
+         execute_remote/2, waiting_remote_vnode/2,
+         postcommit/2, finish/2]).
 
 
 -type detail_info() :: timing.
@@ -63,6 +66,7 @@
                 n :: pos_integer(),
                 w :: non_neg_integer(),
                 dw :: non_neg_integer(),
+                coord_idx_node,
                 preflist2 :: riak_core_apl:preflist2(),
                 bkey :: {riak_object:bucket(), riak_object:key()},
                 req_id :: pos_integer(),
@@ -152,23 +156,42 @@ init({test, Args, StateProps}) ->
     {ok, validate, TestStateData, 0}.
 
 %% @private
-prepare(timeout, StateData0 = #state{robj = RObj0}) ->
+prepare(timeout, StateData0 = #state{from = From, robj = RObj,
+                                     options = Options}) ->
     {ok,Ring} = riak_core_ring_manager:get_my_ring(),
-    BucketProps = riak_core_bucket:get_bucket(riak_object:bucket(RObj0), Ring),
-    BKey = {riak_object:bucket(RObj0), riak_object:key(RObj0)},
+    BucketProps = riak_core_bucket:get_bucket(riak_object:bucket(RObj), Ring),
+    BKey = {riak_object:bucket(RObj), riak_object:key(RObj)},
     DocIdx = riak_core_util:chash_key(BKey),
     N = proplists:get_value(n_val,BucketProps),
     UpNodes = riak_core_node_watcher:nodes(riak_kv),
     Preflist2 = riak_core_apl:get_apl_ann(DocIdx, N, Ring, UpNodes),
-    StartTime = riak_core_util:moment(),
-    
-    StateData = StateData0#state{n = N,
-                                 bkey = BKey,
-                                 bucket_props = BucketProps,
-                                 preflist2 = Preflist2,
-                                 starttime = StartTime},
-    new_state_timeout(validate, StateData).
-    
+    %% Check if this node is in the preference list so it can coordinate
+    case ([IndexNode || {{_Index, Node} = IndexNode, _Type} <- Preflist2,
+                        Node == node()]) of
+        [] ->
+            %% This node is not in the preference list
+            %% forward on to the first node
+            [{{_Idx, CoordNode},_Type}|_] = Preflist2,
+            %% io:format(user, "Forwarding ~p to coordinator on ~p\n", [BKey, CoordNode]),
+            case riak_kv_put_fsm_sup:start_put_fsm(CoordNode, [From, RObj, Options]) of
+                {ok, _Pid} ->
+                    riak_kv_stat:update(coord_redir),
+                    {stop, normal, StateData0};
+                {error, Reason} ->
+                    process_reply({error, {coord_handoff_failed, Reason}}, StateData0)
+            end;
+        [CoordIdxNode|_] ->
+            %% This node is in the preference list, continue
+            StartTime = riak_core_util:moment(),
+            StateData = StateData0#state{n = N,
+                                         bkey = BKey,
+                                         bucket_props = BucketProps,
+                                         coord_idx_node = CoordIdxNode,
+                                         preflist2 = Preflist2,
+                                         starttime = StartTime},
+            new_state_timeout(validate, StateData)
+    end.
+
 %% @private
 validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
                                       options = Options0,
@@ -223,7 +246,7 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
             StateData = init_putcore(StateData3),
             case Precommit of
                 [] -> % Nothing to run, spare the timing code
-                    new_state_timeout(execute, StateData);
+                    new_state_timeout(execute_local, StateData);
                 _ ->
                     new_state_timeout(precommit, StateData)
             end
@@ -231,7 +254,7 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
 
 %% Run the precommit hooks
 precommit(timeout, State = #state{precommit = []}) ->
-    new_state_timeout(execute, State);
+    new_state_timeout(execute_local, State);
 precommit(timeout, State = #state{precommit = [Hook | Rest], robj = RObj}) ->
     Result = decode_precommit(invoke_hook(Hook, RObj)),
     case Result of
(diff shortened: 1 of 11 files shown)


src/riak_kv_put_fsm.erl
```diff
@@ -244,35 +267,81 @@ precommit(timeout, State = #state{precommit = [Hook | Rest], robj = RObj}) ->
...
+
+%% @private
+waiting_local_vnode(request_timeout, StateData) ->
+    process_reply({error,timeout}, StateData);
+waiting_local_vnode(Result, StateData = #state{putcore = PutCore}) ->
+    UpdPutCore1 = riak_kv_put_core:add_result(Result, PutCore),
+    case riak_kv_put_core:enough(UpdPutCore1) of
+        true ->
+            {Reply, UpdPutCore2} = riak_kv_put_core:response(UpdPutCore1),
+            process_reply(Reply, StateData#state{putcore = UpdPutCore2});
```
beerriot: This path short-circuits to either `postcommit` or `finish`.  Since it never goes through `execute_remote`, the replicas are never written.

Test code (on an N=3 bucket):

```
{B, K} = {<<"az303">>, <<"w=1">>},
{ok, C} = riak:local_client(),
ok = C:put(riak_object:new(B, K, ok), 1).
… wait as long as you like …
C:get(B, K, 3).  % R=2 shows the same behavior
```

Expected: `{ok, Object}`

Actual: `{error, notfound}`

jonmeredith: Nice catch, thanks!

src/riak_kv_vnode.erl
```diff
@@ -282,6 +306,52 @@ handle_exit(_Pid, _Reason, State) ->
...
+    riak_core_vnode:reply(Sender, Result),
+    riak_kv_stat:update(vnode_put).
+
+coord_put_merge(CurObj, UpdObj, CoordId, Timestamp) ->
+    %% Make sure UpdObj descends from CurObj and that CoordId is greater
+    CurVC = riak_object:vclock(CurObj),
+    UpdVC = riak_object:vclock(UpdObj),
+
+    %% Valid coord put replacing current object
+    case get_counter(CoordId, UpdVC) > get_counter(CoordId, CurVC) andalso
```
beerriot: Something that has bugged me is that puts are not truly idempotent.  That is, `C:put(O1), C:put(O1)` creates an object with two sibling values that happen to be equal[1], instead of ignoring the second put.  It seems like a vector clock comparison at the correct time should catch this.  It didn't before this patch because the vclock was incremented immediately, so the comparison found equal vclocks, not sequential ones.  I'm not sure it can again after this patch, because two writes from two clients may serialize on the same vnode.  But, I thought I'd bring up the topic in case it generated more useful discussion in the context of this patch.

[1] The objects are only logically equal, and not `=:=` because of Last-Modified metadata.

test/ec_eqc.erl
```diff
@@ -0,0 +1,844 @@
...
+            case riak_object:ancestors([CurObj, UpdObj]) of
+                [UpdObj] ->
+                    keep;
+                [CurObj] ->
+                    UpdObj;
+                _ ->
+                    riak_object:merge(CurObj, UpdObj)
+            end
+    end.
+
```
beerriot, line 872: I think the code below is duplicated from riak_kv_vnode.  Can the function just be exported from there instead?

test/put_fsm_eqc.erl
```diff
@@ -183,7 +183,7 @@ vputsecond() ->
 nodestatus() ->
     ?SHRINK(frequency([{16, primary},
                        {4, fallback},
-                       {1, down}]),
+                       {0, down}]),
```
beerriot, line 186: What's the effect of changing this `{1, down}` to `{0, down}`.  In conversation you mentioned something about the test no longer testing down nodes, and that the line should have been completely scrubbed.  True?

beerriot:
I'm a little worried about whether `node()` is unique enough for actor ID in the vclock.  The scenarios I'm thinking of are moments when the chosen coordinator may change.

The first is during handoff caused by cluster membership change.  Imagine that at T0, node N1 owns partition P.  N1 coordinates a few writes for key K, so key K has vector clock, [{N1, 5}].  At T1, we add node N2, and partition P shifts to node N2.  When handoff is complete, node N1 drops partition P.  At T2, we change cluster membership again in some way that causes partition P to shift back to node N1.  Partition P on node N1 may be asked to coordinate writes before handoff completes.  If this happens, a vclock for key K may be created such that it is an ancestor of the K's vclock stored on node N2.  When handoff completes, data stored under the ancestor vclock will be lost.  Replica partitions will also ignore the older vclock.  This would not happen for any write that was preceded by a successful read (and therefore included the existing vclock), but would happen for reads that failed as "not found", as well as for misbehaved clients.

The other case I'm thinking of is unbalanced rings, where one node owns more than one partition in a preflist.  In that state, only one of those partitions will be used as coordinator (because of the "head" match in riak_kv_put_fsm), but if that coordinator partition shifts to a different node, then the extra partition will become the coordinator.  If it has missed any updates, it may begin generating old vclocks (again, with the exception of successful reads-before-writes).  This same problem might happen if a node loses one partition in a given preflist, but gains another.

I've not written failing test code for either of these situations, so I might just be flat out wrong.  They're also a little corner-case-y, and don't apply to well-behaved read-then-write operations.  But, I still wonder if it wouldn't be better to generate a unique actor ID at vnode startup, and to use that instead of `node()`.  Even something as simple as a combination of node+partition+start_time could work.  Vector clocks would grow a little more than when using just node, but it would still be bounded by n_val \* membership_changes, if done smartly, or even n_val \* node_restarts if done naively.

src/riak_object.erl
```diff
@@ -445,8 +445,8 @@ syntactic_merge(CurrentObject, NewObject, FromClientId, Timestamp) ->
                 true ->
                     NewObject;
                 false ->
-                    increment_vclock(
-                      merge(CurrentObject, NewObject), FromClientId, Timestamp)
+                    %%increment_vclock(
```
beerriot, line 448: This change makes `riak_object:merge2_test` fail (it expects a new client id to appear in the vclock actor list after merging).

beerriot:
As far as client-visible changes go, this patch makes ClientId in riak_client (and therefore its HTTP and PB equivalents) obsolete and unused.  Everything should continue to work as expected, but it may be worth pinging the community to warn people that are doing their own vclock introspection.

slfritchie:
Jon: I'm seeing two EUnit test failures:

```
  riak_object: merge2_test...*failed*
::error:{badmatch,[node1,node2]}
  in function riak_object:merge2_test/0
```

And the second is:

```
module 'put_fsm_eqc'
  put_fsm_eqc:71: eqc_test_...ok
  put_fsm_eqc:72: eqc_test_...[0.001 s] ok
  put_fsm_eqc:73: eqc_test_...[0.009 s] ok
  put_fsm_eqc:74: eqc_test_...ok
  put_fsm_eqc:76: eqc_test_...[0.002 s] ok
  put_fsm_eqc:83: eqc_test_...FSM died:
{function_clause,
    [{gen_server,cast,
         [{riak_kv_vnode_master,<0.4073.0>},
[...]
Shrinking...(9 times)
{quorum,default,one,
    [{ancestor,
         {r_object,<<"x">>,<<"4">>,
             [{r_content,
                  {dict,1,16,16,8,80,48,
                      {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                      {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                        [[<<"X-Riak-Last-Modified">>|{1310,672109,470130}]],
                        [],[]}}},
                  <<"ancestor">>}],
             [],
             {dict,1,16,16,8,80,48,
                 {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                 {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                   [[clean|true]],
                   []}}},
             undefined}},
     {brother,
         {r_object,<<"x">>,<<"4">>,
             [{r_content,
                  {dict,1,16,16,8,80,48,
                      {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                      {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                        [[<<"X-Riak-Last-Modified">>|{1310,672109,470142}]],
                        [],[]}}},
                  <<"brother">>}],
             [{<<"bro!">>,{1,63477891309}}],
             {dict,1,16,16,8,80,48,
                 {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                 {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                   [[clean|true]],
                   []}}},
             undefined}},
     {sister,
         {r_object,<<"x">>,<<"4">>,
             [{r_content,
                  {dict,1,16,16,8,80,48,
                      {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                      {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                        [[<<"X-Riak-Last-Modified">>|{1310,672109,470148}]],
                        [],[]}}},
                  <<"sister">>}],
             [{<<"sis!">>,{1,63477891309}}],
             {dict,1,16,16,8,80,48,
                 {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                 {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                   [[clean|true]],
                   []}}},
             undefined}},
     {otherbrother,
         {r_object,<<"x">>,<<"4">>,
             [{r_content,
                  {dict,1,16,16,8,80,48,
                      {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                      {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                        [[<<"X-Riak-Last-Modified">>|{1310,672109,470157}]],
                        [],[]}}},
                  <<"otherbrother">>}],
             [{<<"bro2">>,{1,63477891309}}],
             {dict,1,16,16,8,80,48,
                 {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                 {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                   [[clean|true]],
                   []}}},
             undefined}},
     {current,
         {r_object,<<"x">>,<<"4">>,
             [{r_content,
                  {dict,1,16,16,8,80,48,
                      {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                      {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                        [[<<"X-Riak-Last-Modified">>|{1310,672109,470163}]],
                        [],[]}}},
                  <<"current">>}],
             [{<<"bro!">>,{1,63477891309}},
              {<<"bro2">>,{1,63477891309}},
              {<<"sis!">>,{1,63477891309}}],
             {dict,1,16,16,8,80,48,
                 {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                 {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                   [[clean|true]],
                   []}}},
             undefined}}],
    0,
    [{notfound,w,0,dw,0,primary}],
    [],false,[],[]}
BucketProps: [{n_val,1},
              {allow_mult,false},
              {chash_keyfun,{riak_core_util,chash_std_keyfun}},
              {pw,0},
              {w,quorum},
              {dw,quorum}]
VPutReplies = [{1,w},{1,{dw,notfound,notfound}}]
PrefList2: [{{1,<0.4073.0>},primary}]
N: 1 PW: quorum RealPW: 1 W:default RealW: 1 DW: one RealDW: 1 EffDW: 1 Pid: <0.4164.0>
Options: [{timeout,200},{dw,one},{w,default},{pw,quorum}]
Precommit: []
Postcommit: []
Object: {r_object,<<"x">>,<<"4">>,
                  [{r_content,{dict,1,16,16,8,80,48,
                                    {[],[],[],[],[],[],[],[],[],[],[],[],[],
                                     [],[],[]},
                                    {{[],[],[],[],[],[],[],[],[],[],[],[],[],
                                      [[<<"X-Riak-Last-Modified">>|
                                        {1310,672109,470163}]],
                                      [],[]}}},
                              <<"current">>}],
                  [{<<"bro!">>,{1,63477891309}},
                   {<<"bro2">>,{1,63477891309}},
                   {<<"sis!">>,{1,63477891309}}],
                  {dict,1,16,16,8,80,48,
                        {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                        {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                          [[clean|true]],
                          []}}},
                  <<"current">>}
Expected Object: noreply
History: []
Expected: {error,timeout} Res: timeout
PostCommits:
Expected: []
Got: []
result: failed
timeout /= {error,timeout}
*failed*
::error:{assertEqual_failed,
          [{module,put_fsm_eqc},
           {line,83},
           {expression,
               "quickcheck ( numtests ( 1000 , ? QC_OUT ( prop_basic_put ( ) ) ) )"},
           {expected,true},
           {value,false}]}
  in function put_fsm_eqc:'-eqc_test_/0-fun-13-'/1
```

src/riak_kv_vnode.erl
slfritchie: After this patch:

```
% ./dev/dev1/bin/riak attach
Attaching to /tmp//Users/fritchie/b/src/riak/dev/dev1/erlang.pipe.1 (^D to exit)

nodes().
['dev3@127.0.0.1','dev2@127.0.0.1']
(dev1@127.0.0.1)2> O1 = riak_object:new(<<"b">>, <<"k">>, <<"val master">>).
{r_object,<<"b">>,<<"k">>,
          [{r_content,{dict,0,16,16,8,80,48,
                            {[],[],[],[],[],[],[],[],[],[],[],[],[],[],...},
                            {{[],[],[],[],[],[],[],[],[],[],[],[],...}}},
                      <<"val master">>}],
          [],
          {dict,1,16,16,8,80,48,
                {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],...},
                {{[],[],[],[],[],[],[],[],[],[],[],[],[],...}}},
          undefined}
(dev1@127.0.0.1)3> {ok, C1} = riak:local_client().
{ok,{riak_client,'dev1@127.0.0.1',<<5,231,185,7>>}}
(dev1@127.0.0.1)4> C1:put(O1).
ok
(dev1@127.0.0.1)5> C1:put(O1).
ok
(dev1@127.0.0.1)6> C1:get(<<"b">>, <<"k">>).
{ok,{r_object,<<"b">>,<<"k">>,
              [{r_content,{dict,2,16,16,8,80,48,
                                {[],[],[],[],[],[],[],[],[],[],[],[],...},
                                {{[],[],[],[],[],[],[],[],[],[],...}}},
                          <<"val master">>}],
              [{'dev1@127.0.0.1',{2,63477892858}}],
              {dict,1,16,16,8,80,48,
                    {[],[],[],[],[],[],[],[],[],[],[],[],[],...},
                    {{[],[],[],[],[],[],[],[],[],[],[],...}}},
              undefined}}
(dev1@127.0.0.1)7> [Quit]
```

jonmeredith:
Hey Scott,

Pull the az303 branch for riak core to fix the put FSM test - I may need to look at the merge test.

Jon

slfritchie:
AZ303, right, nothing like reading the introduction.  Yup, that fixed it.  And the other
failing eunit test case is something that one of Bryan's comments mentions also.

The complexity of testing that's in ec_eqc.erl is ... big.  The must vs. have stuff is
necessary, but hopefully there's a better way than reimplementing a fraction of the
Erlang VM also.  Bah.  I'm thinking that's a good dragon to try to slay.

jonmeredith:
Replaced by https://github.com/basho/riak_kv/pull/157 - rebased version against master.
Did not want to risk losing detailed review comments by rebase/pushing

PR #128, opened 2011-07-14
AZ515 - Add riak_kv_wm_index for HTTP 2I Access

rustyio:
Implement `riak_kv_wm_index` module and update webmachine routes to allow 2I queries via HTTP.

Best described by example:

```
# Index some documents...

curl -v -X PUT \
-d 'data1' \
http://127.0.0.1:8098/riak/mybucket/mykey1

curl -v -X PUT \
-d 'data2' \
-H "x-riak-index-Field1_bin: val2" \
-H "x-riak-index-Field2_int: 1002" \
http://127.0.0.1:8098/riak/mybucket/mykey2

curl -v -X PUT \
-d 'data3' \
-H "X-RIAK-INDEX-FIELD1_BIN: val3" \
-H "X-RIAK-INDEX-FIELD2_INT: 1003" \
http://127.0.0.1:8098/riak/mybucket/mykey3

curl -v -X PUT \
-d 'data4' \
-H "x-riak-index-field1_bin: val4" \
-H "x-riak-index-field2_int: 1004" \
http://127.0.0.1:8098/riak/mybucket/mykey4

# Retrieve the documents...

curl -i http://127.0.0.1:8098/riak/mybucket/mykey1
curl -i http://127.0.0.1:8098/riak/mybucket/mykey2
curl -i http://127.0.0.1:8098/riak/mybucket/mykey3
curl -i http://127.0.0.1:8098/riak/mybucket/mykey4

# Query against a binary field...

curl http://localhost:8098/buckets/mybucket/index/field1_bin/eq/val1
curl http://localhost:8098/buckets/mybucket/index/field1_bin/lte/val2
curl http://localhost:8098/buckets/mybucket/index/field1_bin/range/val2/val4

# Query against an integer field...

curl http://localhost:8098/buckets/mybucket/index/field2_int/eq/1001
curl http://localhost:8098/buckets/mybucket/index/field2_int/lte/1002
curl http://localhost:8098/buckets/mybucket/index/field2_int/range/1002/1004

# Query against the speciel fields $bucket and $key

curl http://localhost:8098/buckets/mybucket/index/\$bucket/eq/mybucket
curl http://localhost:8098/buckets/mybucket/index/\$key/range/mykey2/mykey3
```

Other Changes:
- Update riak_index to treat all index fields and value as binaries.
- Normalize field names as lowercase.
- Remove float support.
- Add range query operator.
- Update riak_kv_wm_object.erl to store the parsed fields in the riak object, and update the index backend to use the parsed fields.


seancribbs:
Sorry to be persnickety about these changes, but expressing queries as URL paths (vs. using the query params) seems unintuitive.  Why not something like `?field1_bin.eq=val1`. Will this change when "RQL" comes around?

rustyio:
The goal was to keep it as simple and straightforward as possible. Not sure whether query params or url parts is more straightforward, from a user's standpoint. From a development standpoint, I think it's a little more straightforward to validate queries with our current approach.

Yes, once RQL arrives, the plan is to have a 'q' or 'query' parameter where the user can specify more complex queries. (URL encoded, of course.)

p.s. Upvoted for "persnickety".

src/riak_kv_wm_index.erl
```diff
@@ -0,0 +1,179 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Resource for listing Riak buckets over HTTP.
```
kellymclaughlin: This looks like c&p leftover from riak_kv_wm_buckets.

kellymclaughlin:
The field parsing process seems overly complex. Instead of calling `field_types` each time and building a function to match each type and then iterating over each type function in `parse_field`, it seems simpler to do the pattern match in `parse_field`. Something like this:

```
parse_field(Key, Value) ->
      Offset = size(Key) - 4,
      case Key of
        <<"$key">> ->
            parse_binary(Value);
        <<"$bucket">> ->
            parse_binary(Value);
        <<_Field:Offset/binary, "_bin">> ->
            parse_binary(Value);
        <<_Field:Offset/binary, "_int">> ->
            parse_integer(Value);
        _ ->
            %% No matching data types, return an error.                                                                                                                                                   
            {error, {unknown_field_type, Key}}
      end.
```

As we add types, we would have to add an offset calculation for each type length, but since the overall number of types will always be relatively small, it seems like an acceptable tradeoff.

kellymclaughlin:
Most of the files in this changeset are still using the edoc function specs instead of the erlang specs for dialyzer. I understand it may not be practical to do this now due to time constraints, but would be good to convert them at some point. A good portion of the riak_kv source files still use the edoc specs so it may easier to create a task to tackle them all at once.

kellymclaughlin:
All the index query operations work for me and all the unit tests are passing.

src/riak_index.erl
```diff
@@ -141,14 +153,29 @@ parse_fields(IndexFields) ->
...
     %% Run the regex to check if the key suffix matches this data
     %% type.
-    {MatchFunction, ParseFunction} = Type,
-    case MatchFunction(Key) of
+    {Suffix, ParseFunction} = Type,
+    Offset = size(Key) - size(Suffix),
+
+    %% Check if the current field is the special field $bucket or
+    %% $key, or if the field matches the suffix for this type.
+    IsMatch =
```
kellymclaughlin: This is just my opinion, but I think the match code would be better broken out into a separate matching function that takes Key, Offset, and Suffix as parameters. The big chunk of match code breaks up the flow of the function and I think could impede someone understanding it.

src/riak_index_mi_backend.erl
```diff
@@ -117,21 +117,24 @@ fold_index(State, Index, Query, SKFun, Acc, FinalFun) ->
     %% this will run more complicated queries, and will run them
     %% asynchronously.
     Results = case Query of
-                  [{eq, Field, Term}] ->
+                  [{eq, Field, [Term]}] ->
                       merge_index:lookup_sync(Pid, Index, Field, Term, FilterFun);
 
-                  [{lt, Field, Term}] ->
+                  [{lt, Field, [Term]}] ->
                       merge_index:range_sync(Pid, Index, Field, undefined, inc(Term, -1), ?SIZE, FilterFun);
```
kellymclaughlin, line 124: `lt` is not working. The 4th and 5th parameters need to be swapped.

src/riak_index_mi_backend.erl
```diff
@@ -117,21 +117,24 @@ fold_index(State, Index, Query, SKFun, Acc, FinalFun) ->
...
 
-                  [{gt, Field, Term}] ->
+                  [{gt, Field, [Term]}] ->
                       merge_index:range_sync(Pid, Index, Field, inc(Term, 1), undefined, ?SIZE, FilterFun);
 
-                  [{lte, Field, Term}] ->
+                  [{lte, Field, [Term]}] ->
                       merge_index:range_sync(Pid, Index, Field, undefined, Term, ?SIZE, FilterFun);
```
kellymclaughlin, line 130: `lte` is not working. The 4th and 5th parameters need to be swapped.

kellymclaughlin:
Ignore my last two comments. I was testing with the wrong version of merge_index.

kellymclaughlin:
Looks good. +1 to merge.

PR #130, opened 2011-07-22
Allow reloading of Erlang code in the add_paths env var.

seancribbs:
Adds a `riak_kv_console` function to iterate over the paths in `add_paths` and reload all the .beam files found there.  This aids in automated code deployment of map/reduce functions.  Needs application of the corresponding [pull request](https://github.com/basho/riak/pull/48) in the riak repository to add the `erl_reload` command to `riak-admin`.

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -33,7 +33,8 @@
          ringready/1,
          transfers/1,
          cluster_info/1,
-         down/1]).
+         down/1,
+         reload_code/1]).
 
 join([NodeStr]) ->
     try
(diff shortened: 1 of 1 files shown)


seancribbs:
Cheesy demo video: http://vimeo.com/26782465

beerriot:
I like the idea.  Two thoughts:

It would be nice to have some output for each reloaded module, to help with loading errors, and also just to have a record of what happened.

It may be a tiny bit safer to only purge+load things that have already been loaded.  That is, only if `code:loaded/1` returns `{file, Loaded}`.  Maybe `Loaded` should even be checked for matching `Path`?

seancribbs:
@beerriot Thanks for the input.  Yes, it makes sense to only purge things that were loaded, and new files will be loaded the first time they are used.  I'll fix.

kellymclaughlin:
I was initially concerned that using `purge` instead of `soft_purge` could result in some of the m/r processes being killed, but I am not seeing that behavior in my tests. I checked specifically for a process that was running a version of code marked as _old_ to be killed when the module was reloaded, but it was not and the m/r result came back as expected. 

If we wanted to be extra-safe we could change the use of `purge` to `soft_purge` and allow a `force` optional arg for the command that would use `purge`, but since I don't see any bad behavior currently I am fine with merging as-is.

kellymclaughlin:
+1 to merge.

PR #134, opened 2011-07-26
Remove special case formatting for vnode_gets.

seancribbs:
There's no apparent reason for this function clause and it screws up attempts to do automated parsing of the output.


kellymclaughlin:
Looks good, +1 to merge. I rebased for testing and there were no conflicts.

PR #135, opened 2011-07-28
Az528 integrate indexes with coverage

kellymclaughlin:
Change secondary indexes to use coverage instead of legacy key listing code.

### DIFF
--- a/src/riak_kv_index_backend.erl
+++ b/src/riak_kv_index_backend.erl
@@ -35,6 +35,7 @@
          list/1,
          list_bucket/2,
          fold/3,
+         fold_index/3,
          fold_keys/3,
          fold_bucket_keys/4,
          drop/1,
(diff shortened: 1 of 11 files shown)


src/riak_kv_index_backend.erl
```diff
@@ -177,41 +178,37 @@ fold_keys(State, Fun, Extra) ->
...
-fold_bucket_keys(State, {index_query, Bucket, Query}, _Fun, _Extra) ->
-    #state { index_mod = IndexMod, index_state = IndexState } = State,
-
+%% @doc Pass the fold_index request through to the Index backend.
+fold_index(#state {index_mod = IndexMod, 
+                   index_state = IndexState},
+           Bucket, Query) ->
     %% Update riak_kv_stat...
-    riak_kv_stat:update(vnode_index_read),
+    riak_kv_stat:update(vnode_index_read),
```
seancribbs: `M-x whitespace-cleanup` ;)

src/riak_kv_index_fsm.erl
```diff
@@ -0,0 +1,122 @@
...
+            %% occurred or the timeout has elapsed.
+            ClientPid ! {ReqId, Error}
+    end,
+    {stop, normal, StateData};
+finish(clean,
+       StateData=#state{from={raw, ReqId, ClientPid},
+                        client_type=ClientType}) ->
+    case ClientType of
+        mapred ->
+            luke_flow:finish_inputs(ClientPid);
```
rustyio, line 105: This will cause problems when mapreduce uses riak_pipe.

src/riak_kv_index_fsm.erl
```diff
@@ -0,0 +1,122 @@
...
+
+%% ===================================================================
+%% Internal functions
+%% ===================================================================
+
+process_query_results(plain, _Bucket, Results, ReqId, ClientPid) ->
+    ClientPid ! {ReqId, {results, Results}};
+process_query_results(mapred, Bucket, Results, _ReqId, ClientPid) ->
+    try
+        luke_flow:add_inputs(ClientPid, [{Bucket, Result} || Result <- Results])
```
rustyio, line 119: This will cause problems when mapreduce uses riak_pipe.

src/riak_kv_vnode.erl
```diff
@@ -261,6 +262,19 @@ handle_coverage(?KV_LISTKEYS_REQ{bucket=Bucket,
...
+                              item_filter=ItemFilter,
+                              qry=Query},
+                FilterVNodes,
+                Sender,
+                State=#state{idx=Index,
+                             mod=Mod,
+                             modstate=ModState}) ->
+    %% Construct the filter function
+    FilterVNode = proplists:get_value(Index, FilterVNodes),
+    Filter = riak_kv_coverage_filter:build_filter(Bucket, ItemFilter, FilterVNode),
```
rustyio, line 280: What is the form of ItemFilter here? 

Digging into riak_kv_coverage_filter, the allowed values for ItemFilter are 'none', an empty list, or a list of {Mod,Fun,Args} where Mod:Fun(Args) generates another function, but it's unclear whether the generated function should return a boolean (which seems to be what riak_kv_coverage_filter:compose_filter/3 expects) or whether it should return a value to be passed into other filter functions (which is what riak_kv_coverage_filter:compose/2 expects).

This may be a bug in riak_kv_coverage_filter? Difficult to follow.

jonmeredith:
Nice the see the %% HACK gone :)

Everything was functionally correct for me while taking nodes up/down.

Given it was difficult for Rusty to follow the filter function maybe some extra comments to explain why the filtering is necessary (to prevent result duplication) and that it is applied on the result bkeys from the index fold.  Once the backend interface refactor is in place, it should be possible to apply it as part of the fold_keys and reduce garbage from unneeded bkeys.

+1 merge once comments and Rusty's map/reduce concerns have been adressed. I'm not sure how pipe is intending to handle list keys, for the initial 2I release it should handle it the same way.

src/riak_kv_vnode.erl
kellymclaughlin, line 280: I've attempted to document this a little better. The item filter is just a generalization of the client specified key filters, but it also supports the list of `{Mod, Fun, Args}` tuples used by MapReduce filters. The MapReduce filters stuff is very ugly and needs some attention, but it was out of scope for the coverage code work so I just left it alone.

src/riak_kv_index_fsm.erl
kellymclaughlin, line 119: I think this might be ok because I think the pipe implementation of mapreduce currently uses a ClientType of `plain`, but I'll verify that with Bryan tomorrow.

kellymclaughlin, line 119: Followed up with Bryan and he confirmed that the pipe implementation does use the `plain` type currently.

jonmeredith, line 119: Good - it'll be nice to deprecate that stuff in a few releases.

kellymclaughlin:
I addressed the issue with key filtering that uses functions from `riak_kv_mapred_filters`. Thanks to Rusty for spotting this!

kellymclaughlin:
Derp. Ok, fixed that. Thx Jon.

PR #136, opened 2011-07-29
Use 60sec timeout for async get/tombstone harvest

slfritchie:
Subject says it all....


Vagabond:
Looks fine to me, although the use of that intermediate value doesn't really thrill me.

+1

slfritchie:
Andrew, I wanted an extra line to add some docs, and the extra var helps
a little bit I hope?

PR #137, opened 2011-08-01
AZ567 - Update 2i merge_index data_root setting.

rustyio:
With 2i, we now run an additional merge_index instances. This change sets the 2i instances to point to a different data_root than the search instances.


jonmeredith:
The test cases at the end should be updated to use the new app env. 

Other than that +1 merge - although the 2i suffix isn't pretty it hopefully won't be around for too long  as we deprecate them moving towards the backend refactor.

PR #138, opened 2011-08-01
Az573 use kv fsm for repl

rzezeski:
In order to use the KV put FSM in riak_repl an option to disable the pre/post commit hooks needs to be added.  This patch adds the `disable_hooks` option to do just that.


slfritchie:
Ryan: basho_expect's Run-systest-riak-repl-2x2 test passes, +1 to merge, thanks.

PR #139, opened 2011-08-03
Changes to riak_kv_console to support new join/claim approach in riak_core

jtuple:
Updated riak_kv_console (and therefore riak-admin) to use the new join/remove/leave functionality exposed in the new join/claim/membership code.


jonmeredith:
(copying comments from riak_core branch)
+1 reviewed.

Future tasks will address:

rolling upgrades
only singleton nodes can join clusters (must leave first - no merging clusters)
make sure node belongs to the cluster for partition ownership.
mechanism to mark nodes as down so that progress can take place.

PR #141, opened 2011-08-04
Add timeout around bad node ping in put_fsm_eqc.

kellymclaughlin:
The intentionally failing node ping in the put_fsm_eqc was timing out and causing the test suite to fail. Add a 60 second timeout to avoid this.


jonmeredith:
+1 merge

PR #142, opened 2011-08-04
Use pseudo-random node names to avoid problems with racing with epmd to r

slfritchie:
Use pseudo-random node names to avoid problems with racing with epmd to reuse a node name

### DIFF
--- a/src/riak_kv_delete.erl
+++ b/src/riak_kv_delete.erl
@@ -285,7 +285,8 @@ setup() ->
     error_logger:tty(false),
     error_logger:logfile({open, "riak_kv_delete_test.log"}),
     %% Start erlang node
-    TestNode = list_to_atom("testnode" ++ integer_to_list(element(3, now()))),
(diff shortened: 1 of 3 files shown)


jonmeredith:
Randomness in testing scares me a bit.  I'm going to try and reproduce and see if calling erl_epmd:stop() will fix instead.

slfritchie:
Notes since last update:
- The # of microseconds is used for the node name, so chance of collision
  is one in a million.  I could use all 2 or 3 of the now() results to make it
  impossible to make a collision, if it'd get a positive review.
- After all the EUnit day work on the master branch, I'm still seeing riak_kv_delete
  test errors due to net_kernel startup failures, both with and without a eunit-process-
  killing-at-the-end-of-a-run hacks to rebar (platform = 1 CPU ubuntu9 32 bit).
  When using the pseudo-random node name patches, the errors by riak_kv_delete's
  tests disappear.

jonmeredith:
The non-determinism still makes me sad - but we have bigger fish to fry than net_kernel right now, so a pragmatic +1.

PR #143, opened 2011-08-04
Rename eunit target to test in Makefile.

kellymclaughlin:


beerriot:
`test` also depends on `all`, because it needs the lager parse transform to compile. Should that change be added to this PR?

jaredmorrow:
az584 created to fix the compilation issue with parse transforms in lager

PR #148, opened 2011-08-07
Log error if `send_inputs` breaks

rzezeski:
az583

It's not very helpful to just see `bad_mapred_inputs` because you
can't distinguish between the two different places where this can occur.
Furthermore, even if you could it still tells you nothing as to
why it actually broke.


beerriot:
+1, good idea

PR #149, opened 2011-08-08
Make 'test' depends on 'all' so lager parse trasforms are compiled before using them in tests

jaredmorrow:


rzezeski:
When I run a `make distclean; make test` I get errors related to logging.

https://gist.github.com/f756e1c7b5beb45ef280

PR #150, opened 2011-08-11
Az543 backend refactoring phase1

kellymclaughlin:
This rather large set of changes does the following:
- Updates all backends to use a standard api
- Removes the gb_trees, cache, fs, ets, and dets backends
- Consolidates cache and ets backend functionality into a memory backend
- Updates `riak_kv_vnode` to use the new backend api
- Change bitcask backend to use a symlink to point to the active directory to facilitate drops.
- Add a module for buffered return of backend fold results. 
- Updates to relevant eqc and eunit tests.
- Specs for the backend api functions in all backend modules.


kellymclaughlin:
The `backend_eqc` tests will fail when run by `riak_kv_index_backend` until [this](https://github.com/basho/merge_index/pull/10) change is reviewed and merged to `merge_index`.

src/riak_kv_backend.erl
```diff
@@ -52,33 +69,86 @@ callback_after(Time, Ref, Msg) when is_integer(Time), is_reference(Ref) ->
 -ifdef(TEST).
 
 standard_test(BackendMod, Config) ->
```
Vagabond, line 72: I'd be kind of nice if this test wasn't testing everything at once, all interleaved like this. Its hard to keep track of what should be the result here. An eunit {setup, ...} fixture might be a good idea.

src/riak_kv_bitcask_backend.erl
```diff
@@ -51,134 +51,215 @@
...
+api_version() ->
+    {?API_VERSION, ?CAPABILITIES}.
 
+%% @doc Start the bitcask backend
+-spec start(integer(), config()) -> {ok, state()} | {error, term()}.
+start(Partition, Config) ->
+    %% Get the data root directory
+    case config_value(data_root, Config) of
+        undefined ->
+            error_logger:error_msg("Failed to create bitcask dir: data_root is not set\n"),
```
Vagabond: These should be calls to lager:error (you can omit the newline).

src/riak_kv_bitcask_backend.erl
```diff
@@ -51,134 +51,215 @@
...
-    {ok, FNs} = file:list_dir(BitcaskRoot),
-    [file:delete(filename:join(BitcaskRoot, FN)) || FN <- FNs],
-    file:del_dir(BitcaskRoot),
-    ok.
 
-is_empty({Ref, _}) ->
+    {MegaSecs, Secs, MicroSecs} = erlang:now(),
+
+    %% Setup new bitcask dir for this partition
+    NewDataFile = lists:flatten([LinkFile,
```
Vagabond: The flatten here is unnecessary, filename:join will handle a deep list OK:

2> filename:join(["/foo", [[$b, $a], $r]]).
"/foo/bar"

src/riak_kv_bitcask_backend.erl
```diff
@@ -51,134 +51,215 @@
...
+                                 "-",
+                                 integer_to_list(MegaSecs),
+                                 integer_to_list(Secs),
+                                 integer_to_list(MicroSecs)]),
+    NewDataPath = filename:join([DataRoot, NewDataFile]),
+
+    case filelib:ensure_dir(NewDataPath) of
+        ok ->
+            %% Move the link to point to the new data file
+            file:delete(LinkPath),
```
Vagabond: I know this is preexisting, but the use of linkpath/linkfile is really confusing. LinkFile is just the filename and LinkPath is the filename with the root_dir prefixed. Why do we need to store both and name them so confusingly?

src/riak_kv_bitcask_backend.erl
```diff
@@ -51,134 +51,215 @@
...
+            %% files will be cleaned up on the next drop.
+            %% The worst case is that the files hang
+            %% around and take up some disk space.
+            spawn(drop_data_cleanup(DataRoot, LinkFile, NewDataFile)),
+
+            %% Now open the bitcask and return an updated state
+            %% so this backend can continue processing.
+            case bitcask:open(NewDataPath, BitcaskOpts) of
+                Ref1 when is_reference(Ref1) ->
+                    {ok, State#state{ref=Ref1}};
```
Vagabond, line 232: link_file in the state doesn't need to be updated to NewDataFile here? Same for link_path/NewDataPath?

src/riak_kv_bitcask_backend.erl
Vagabond: Also, isn't this code a duplicate of check_symlink? Could you simply do the file:delete before a call to check_symlink and then spawn the drop_data?

src/riak_kv_eleveldb_backend.erl
```diff
@@ -73,137 +79,206 @@ start(Partition, Config) ->
...
+        Value ->
+            Value
+    end.
+
+%% @private
+%% Return a function to fold over the buckets on this backend
+fold_buckets_fun(FoldBucketsFun) ->
+    fun(BK, Acc) ->
+            {Bucket, _} = sext:decode(BK),
+            FoldBucketsFun(Bucket, Acc)
```
Vagabond: Won't this end up calling the FoldBucketsFun for every single key? Wouldn't keeping track of already-visited buckets be prudent?

src/riak_kv_index_backend.erl
```diff
@@ -53,134 +58,226 @@
...
-    ErrorResults = [X || X <- Results, X /= ok],
-    case ErrorResults of
-        [] -> ok;
-        _ -> {error, ErrorResults}
-    end.
+    #state{kv_mod = KVMod,
+           kv_state = KVState,
+           index_mod = IndexMod,
+           index_state = IndexState}=State,
+    KVMod:stop(KVState),
```
Vagabond, line 122: Errors returned here no longer bubble up. Either the documentation or the code should be corrected.

src/riak_kv_bitcask_backend.erl
kellymclaughlin: This is fixed.

src/riak_kv_bitcask_backend.erl
kellymclaughlin: Yep, I refactored this to use check_symlink.

src/riak_kv_bitcask_backend.erl
kellymclaughlin: You're right. Changed so it just uses LinkFile now.

src/riak_kv_bitcask_backend.erl
kellymclaughlin, line 232: Nope this just tracks the name of the linked file, not the actual data file. The only time the name of the actual file is needed is in drop during the cleanup and it just needs the new file name which is generated in the same function.

src/riak_kv_index_backend.erl
kellymclaughlin, line 122: Which documentation are you referring to? The function spec says it should return `ok`.

src/riak_kv_eleveldb_backend.erl
kellymclaughlin: Yes that is true. There may be a way to optimize this a bit with leveldb. I'll look at that a little bit.

src/riak_kv_index_backend.erl
Vagabond, line 122: "%% @doc Stop backends, bubble up any errors."

src/riak_kv_backend.erl
kellymclaughlin, line 72: I took a step in this direction, but maybe I should take it a step further and create a test case functions for the different things that are being tested. What do you think?

src/riak_kv_multi_backend.erl
```diff
@@ -60,231 +80,326 @@
...
+       true ->
+            {First, _, _} = hd(Defs),
+
+            %% Get the default
+            DefaultBackend = proplists:get_value(multi_backend_default, Config, First),
+            case lists:keymember(DefaultBackend, 1, Defs) of
+                true ->
+                    %% Start the backends
+                    Backends = [begin
+                                    {ok, State} = Module:start(Partition, SubConfig),
```
Vagabond: I know this is legacy, but a misconfigured backend will throw an obscure badmatch error here.

src/riak_kv_backend.erl
Vagabond, line 72: Yes, I'd like to see it broken up more cleanly into annotated test cases, with a common setup/teardown.

Vagabond:
Removing the ets backend breaks one of the riakc tests (when you have a riak node running).

kellymclaughlin:
Added [pr](https://github.com/basho/riak-erlang-client/pull/33) to update riak client tests to use the memory backend instead of ets.

Vagabond:
A test for a multi-backend startup failure would be good, to go with the new code.

kellymclaughlin:
Added test case for error on multi-backend startup.

Vagabond:
I regret that I have but +1 to give to this pull request.

src/riak_kv_vnode.erl
```diff
@@ -163,9 +163,20 @@ get_vclocks(Preflist, BKeyList) ->
 init([Index]) ->
     Mod = app_helper:get_env(riak_kv, storage_backend),
     Configuration = app_helper:get_env(riak_kv),
-    {ok, ModState} = Mod:start(Index, Configuration),
 
-    {ok, #state{idx=Index, mod=Mod, modstate=ModState, mrjobs=dict:new()}}.
+    case Mod:start(Index, Configuration) of
```
jonmeredith: could do `case catch Mod:start` and add {'EXIT', '} clauses  to make sure riak shuts down if you can't start the backend module.

src/riak_kv_vnode.erl
```diff
@@ -397,15 +408,15 @@ perform_put({false, Obj},#state{idx=Idx},#putargs{returnbody=true,reqid=ReqID})
 perform_put({false, _Obj}, #state{idx=Idx}, #putargs{returnbody=false,reqid=ReqId}) ->
     {dw, Idx, ReqId};
 perform_put({true, Obj}, #state{idx=Idx,mod=Mod,modstate=ModState},
-            #putargs{returnbody=RB, bkey=BKey, reqid=ReqID}) ->
+            #putargs{returnbody=RB, bkey={Bucket, Key}, reqid=ReqID}) ->
     Val = term_to_binary(Obj),
-    case Mod:put(ModState, BKey, Val) of
-        ok ->
+    case Mod:put(Bucket, Key, Val, ModState) of
+        {ok, _UpdModstate} ->
```
jonmeredith, line 449: The UpdModState isn't being re-set in the main #state{}.  This code will conflict with the az://303 merge so we can fix it then.

src/riak_kv_vnode.erl
```diff
@@ -475,116 +486,194 @@ do_mget({fsm, Sender}, BKeys, ReqId, State=#state{idx=Idx, mod=Mod, modstate=Mod
...
 
 %% @private
 list_buckets(Sender, Filter, Mod, ModState) ->
-    %% TODO: Decide if we want to continue to allow key filters
-    %% to be used to filter the list of buckets. I think it is
-    %% more useful to move all filtering out of the backend and
-    %% not have to force all backends to fold over all keys
-    %% to generate a list of buckets.
-    Buckets = Mod:list_bucket(ModState, '_'),
+    BufferSize = 1000,
```
jonmeredith: Maybe we should pluck this from the app env - it won't be called very often so no perf concerns there.

kellymclaughlin: Maybe the vnode should read the buffer size for key and bucket listing and secondary index buffers on startup and store it in state. You would have to restart the vnode for changes to take effect or add some message handling to update the values, but it seems like a good balance of the tradeoffs to me.

src/riak_kv_bitcask_backend.erl
```diff
@@ -253,19 +369,133 @@ schedule_sync(Ref, SyncIntervalMs) when is_reference(Ref) ->
...
+    {MegaSecs, Secs, MicroSecs} = erlang:now(),
+    DataFile = filename:join([filename:dirname(LinkFile),
+                              [filename:basename(LinkFile),
+                               "-",
+                               integer_to_list(MegaSecs),
+                               integer_to_list(Secs),
+                               integer_to_list(MicroSecs)]]),
+    case filelib:ensure_dir(DataFile) of
+        ok ->
+            case file:make_symlink(filename:absname(DataFile), filename:absname(LinkFile)) of
```
jonmeredith: This makes horrible symlinks on my system

```
lrwxr-xr-x  1 jmeredith  staff  133 Aug 26 08:46 981946412581700398168100746981252653831329677312 -> /Users/jmeredith/basho/work/az543-be1/riak_ee/dev/dev1/data/bitcask/981946412581700398168100746981252653831329677312-1314369968166336
```

What about doing file:make_symlink(filename:basename(DataFile), file:absname(LinkFile) instead - I think that'll make a relative link which would make it possible to move the bitcask data dir around without have to recreate all those links.

src/riak_kv_bitcask_backend.erl
```diff
@@ -221,12 +301,48 @@ key_counts(RootDir) ->
...
 %% @private
-%% Invoke bitcask:status/1 for a given directory
-status(Dir) ->
-    Ref = bitcask:open(Dir),
-    try bitcask:status(Ref)
-    after bitcask:close(Ref)
-    end.      
+%% Return a function to fold over the buckets on this backend
+fold_buckets_fun(FoldBucketsFun) ->
+    fun(#bitcask_entry{key=BK}, Acc) ->
```
jonmeredith: I think this function will visit each bucket name more than once.  Shouldn't you keep track of which buckets you've seen before calling FoldBucketsFun?

src/riak_kv_coverage_filter.erl
```diff
@@ -81,39 +81,22 @@ build_filter(Bucket, ItemFilterInput, FilterVNode) ->
...
 %% @private
 compose_filter(ItemFilter) ->
-    fun(Item, Acc) ->
-            case ItemFilter(Item) of
-                true ->
-                    [Item | Acc];
-                false ->
-                    Acc
-            end
+    fun(Item) ->
```
jonmeredith: Isn't this equivalent to just returning ItemFilter?  You're wrapping a fun/1 with a fun/1 and not doing anything extra  What am I missing?

src/riak_kv_coverage_filter.erl
```diff
@@ -81,39 +81,22 @@ build_filter(Bucket, ItemFilterInput, FilterVNode) ->
...
-    fun(Key, Acc) ->
-            case VNodeFilter(Key) of
-                true ->
-                    [Key|Acc];
-                false ->
-                    Acc
-            end
-
+    fun(Item) ->
+            VNodeFilter(Item)
```
jonmeredith: Another double-wrap

src/riak_kv_eleveldb_backend.erl
jonmeredith: I'll second Vagabonds comment.

src/riak_kv_multi_backend.erl
```diff
@@ -60,231 +80,356 @@
...
-    [F(X) || X <- State#state.backends],
-    ok.
+    Module:delete(Bucket, Key, SubState).
+
+%% @doc Fold over all the buckets
+-spec fold_buckets(riak_kv_backend:fold_buckets_fun(),
+                   any(),
+                   [],
+                   state()) -> {ok, any()} | {error, term()}.
+fold_buckets(FoldBucketsFun, Acc, Opts, #state{backends=Backends}) ->
```
jonmeredith, line 206: You could optimize here by folding over the named buckets in the multi config and just doing the backend bucket fold over the default.

src/riak_kv_multi_backend.erl
```diff
@@ -60,231 +80,356 @@
...
-    %% We have less work to do than fold/3 does because we're a
-    %% backend-aware manager.
-    %% Mimic riak_kv_vnode:do_list_keys() logic but with nested try/catch.
-    {_Name, Module, SubState} = get_backend(Bucket, State),
+%% @doc Fold over all the keys for one or all buckets.
+-spec fold_keys(riak_kv_backend:fold_keys_fun(),
+                any(),
+                [{atom(), term()}],
+                state()) -> {ok, term()} | {error, term()}.
+fold_keys(FoldKeysFun, Acc, Opts, #state{backends=Backends}) ->
```
jonmeredith: Optimization if bucket provided - just lookup that backend

src/riak_kv_multi_backend.erl
```diff
@@ -60,231 +80,356 @@
...
+                                                   SubState),
+                      case Result of
+                          {ok, Acc2} ->
+                              Acc2;
+                          {error, Reason} ->
+                              throw({error, {Module, Reason}})
+                      end
+              end,
+    try
+        Acc0 = lists:foldl(FoldFun, Acc, Backends),
```
jonmeredith: Same optimization if bucket provided - could just filter on Backends before the lists:foldl.

jonmeredith:
Completed code review 
- Minor issues commented on inline above.
- Although the BE State is being returned by calls now it isn't being returned the kv_vnode state.  The az://303 changes require some fiddling around with the put path.  I was part way through refactoring it when I realized they'd clash with the phase1/phase2 refactor.  Once phase2 is signed off/merged I'm happy to do the work as part of merging in the az://303 branch and it should be independent of the phase3 work around async folds.
- Once BEModState is being kept the gen_server in riak_kv_memory_backend can go away - the only purpose I can see for it is to keep track of used_memory.  That will remove an extra process/hop which will help with latency/garbage generation.
- We do expiry in bitcask and memory backends - perhaps we should define a common option for both and set it as a capability.  Same for bounding in-memory size - would be nice to have a way to specify maximum memory usage per-vnode.
- Did not review the changes to riak_kv_index_backend yet - I'll review as a whole on the phase2 pull request.

Now off to run unit tests and play around with it.

jonmeredith:
Joined 4 nodes together, got this when vnodes shutdown.

```
11:00:09.941 [error] gen_fsm <0.368.0> in state active terminated with reason: no match of right hand value {ok,{state,#Ref<0.0.0.3991>,[46,47,100,97,116,97,47,98,105,116,99,97,115,107,47,48],[{read_write,true},{included_applications,[]},{allow_strfun,false},{reduce_js_vm_count,6},{storage_backend,riak_kv_bitcask_backend},{legacy_keylisting,true},{pb_ip,[49,50,55,46,48,46,48,46,49]},{hook_js_vm_count,2},{mapred_name,[109,97,112,114,101,100]},{stats_urlpath,[115,116,97,116,115]},{js_thread_stack,16},{riak_kv_stat,true},{add_paths,[]},{http_url_encoding,on},{map_js_vm_count,8},{pb_port,8082},{mapred_system,pipe},{js_max_vm_mem,8}],[46,47,100,97,116,97,47,98,105,116,99,97,115,107]}} in riak_kv_vnode:delete/1
```

Looks like riak_kv_vnode:delete is matching on ok when it returns {ok, State}

src/riak_kv_coverage_filter.erl
kellymclaughlin: derp

src/riak_kv_multi_backend.erl
kellymclaughlin, line 206: Not sure I quite understand this. Where are buckets stored in the multi config?

src/riak_kv_vnode.erl
kellymclaughlin, line 449: So `perform_put` and its caller `do_put` need to be replumbed to return the updated state back to handle_command? That should be pretty simple if you want to handle it now.

src/riak_kv_multi_backend.erl
jonmeredith, line 206: Ignore me - they're config names.  Reading the ring, getting all bucket properties, making a minimal list of configs and going over them seems like a bit too much work for me.

src/riak_kv_bitcask_backend.erl
kellymclaughlin: So since we have to call `bitcask:fold_keys` to do this work we have to visit a particular bucket name more than once. We could check the `Acc` for the bucket name in the FoldFun before it calls the passed-in FoldBucketsFun, but then we have to make assumptions about `Acc` in the backend module. The benefit of the current approach is that the backend doesn't have to know anything about `Acc` and it can be a `riak_kv_fold_buffer` or a list or whatever. The best approach to me seems to be to change the FoldBucketsFun in `riak_kv_vnode` to check the `Acc` before doing anything else instead of uniquing the list as part of the buffering as it currently does.

jonmeredith: I'd rather deal with it locally in the backends - the function is supposed to fold over bucket names - it would be weird for fold_keys/fold_objects to return objects more than once.

You could do it by wrapping the Acc with a tuple containing a set of previously visited buckets and the user accumulator for bitcask, or for an ordered store like eleveldb you could track the previous bucket and call the fold function when it is different.

jonmeredith:
Handoff may be broken.
Wrote an object to an empty node.
Joined another node.  Waited for errors about delete to pass and noticed

```
12:53:04.758 [error] gen_fsm <0.418.0> in state active terminated with reason: call to undefined function riak_kv_bitcask_backend:fold/3 from riak_kv_vnode:handle_command/3
```

After that the object was notfound.

src/riak_kv_bitcask_backend.erl
kellymclaughlin: Fixed.

src/riak_kv_coverage_filter.erl
kellymclaughlin: Fixed.

src/riak_kv_coverage_filter.erl
kellymclaughlin: Fixed.

src/riak_kv_multi_backend.erl
kellymclaughlin: Fixed.

src/riak_kv_multi_backend.erl
kellymclaughlin: Fixed.

src/riak_kv_vnode.erl
kellymclaughlin: I added a clause for `{EXIT, Reason}`. Should I also add one for `_` in case start throws an arbitrary term?

Also, there's  bug in riak_core [here](https://github.com/basho/riak_core/blob/master/src/riak_core_vnode.erl#L279) that causes problems if the initial Reason is not the same as the reason returned from the call to `handle_exit`. I'll just open a pull request against master to fix that.

src/riak_kv_vnode.erl
jonmeredith: That would work ok - although I didn't think list_buckets operations happened often enough to worry about an app_env access (it's just an ETS table lookup I think).

jonmeredith:
Not sure if you expect handoff to be fixed by the last commit.  Do a make stagedevrel, then on dev1

```
(dev1@127.0.0.1)1> {ok,C}=riak:local_client().
{ok,{riak_client,'dev1@127.0.0.1',<<2,224,44,40>>}}
(dev1@127.0.0.1)2> C:get(<<"b">>,<<"k">>).
{error,notfound}
(dev1@127.0.0.1)3> C:put(riak_object:new(<<"b">>,<<"k">>,<<"v">>)).
ok
(dev1@127.0.0.1)4> C:get(<<"b">>,<<"k">>,[{r,all},details]).       
{ok,{r_object,<<"b">>,<<"k">>,
              [{r_content,{dict,3,16,16,8,80,48,
                                {[],[],[],[],[],[],[],[],[],[],[],[],...},
                                {{[],[],[],[],[],[],[],[],[],[],...}}},
                          <<"v">>}],
              [{<<2,224,44,40>>,{1,63481621079}}],
              {dict,1,16,16,8,80,48,
                    {[],[],[],[],[],[],[],[],[],[],[],[],[],...},
                    {{[],[],[],[],[],[],[],[],[],[],[],...}}},
              undefined},
    [{vnode_oks,3},{duration,1116}]}
```

Join dev2 and wait > 60secs for handoffs to complete (debug tracer on riak_kv_vnode to check it really was accessing dev1 and dev2)

```

(dev1@127.0.0.1)7> C:get(<<"b">>,<<"k">>,[{r,all},details]).
(<0.757.0>) call riak_kv_vnode:get([{639406966332270026714112114313373821099470487552,'dev1@127.0.0.1'},
 {662242929415565384811044689824565743281594433536,'dev2@127.0.0.1'},
 {685078892498860742907977265335757665463718379520,'dev1@127.0.0.1'}],{<<"b">>,<<"k">>},84137161)
(<0.757.0>) returned from riak_kv_vnode:get/3 -> ok
{error,notfound,[{vnode_oks,0},{duration,675}]}
(dev1@127.0.0.1)8> 
```

src/riak_kv_vnode.erl
jonmeredith, line 449: Exactly.  Once that's done the memory backend can be simplified.

kellymclaughlin:
Handoff itself is fixed, but I'm still seeing some issues with bitcask. I'll sort that out and finish up the rest of the tasks after a trip to the doctor.

kellymclaughlin:
I think I'd like to do the put replumbing in `riak_kv_vnode` as part of phase 2 because there are some other changes in there that will make it simpler to merge those changes in. What do you think?

kellymclaughlin:
Handoff should be working now for all backends.

jonmeredith:
I'm fine with replumbing the backend status as part of phase 2 - I'll give the changes a whirl.

jonmeredith:
And for today's fun - replication.  Configured 2x2node clusters

When fullsync started, had a vnode crash.

```
09:01:56.580 [info] Elected as replication leader
09:01:56.580 [info] Re-elected as replication leader
09:02:29.559 [info] Full-sync with site "from_c2" starting; 64 partitions.
09:02:29.560 [info] Full-sync with site "from_c2"; hashing partition 0 data
09:02:29.583 [error] gen_fsm <0.364.0> in state active terminated with reason: fun called with wrong arity of 4 instead of 3 in riak_kv_bitcask_backend:'-fold_objects_fun/2-fun-0-'/4
09:02:29.585 [error] CRASH REPORT Process [] with 0 neighbours crashed with reason: fun called with wrong arity of 4 instead of 3 in riak_kv_bitcask_backend:'-fold_objects_fun/2-fun-0-'/4
09:02:29.586 [error] Supervisor riak_core_vnode_sup had child at module undefined at <0.364.0> exit with reason fun called with wrong arity of 4 instead of 3 in riak_kv_bitcask_backend:'-fold_objects_fun/2-fun-0-'/4 in context child_terminated
```

src/riak_kv_memory_backend.erl
```diff
@@ -0,0 +1,536 @@
...
+         terminate/2,
+         code_change/3]).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-define(API_VERSION, 1).
+-define(CAPABILITIES, []).
+-define (DEFAULT_TTL, timer:hours(1)).
```
jonmeredith: TTL is applied in seconds in exceeds_ttl - timer:hours gives result in milliseconds.  Default TTL will be 1000 times longer than desired.

src/riak_kv_memory_backend.erl
```diff
@@ -0,0 +1,536 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc riak_kv_memory_backend is a Riak storage backend that uses ets
```
jonmeredith, line 23: Add some docs on config parameters and their units.

jonmeredith:
Tested handoff with data stored - configured as

```
{storage_backend, riak_kv_multi_backend},
{multi_backend_default, <<"bitcask">>},
{multi_backend, [
   {<<"bitcask">>, riak_kv_bitcask_backend, [
      {data_root, "data/bitcask"}
   ]},
   {<<"eleveldb">>, riak_kv_eleveldb_backend, [
      {data_root, "data/eleveldb"}
   ]},            
   {<<"index">>, riak_kv_index_backend, [
      {bitcask, [{data_root, "data/index-bitcask"}]},
      {merge_index, [{data_root_2i, "data/index-mergeindex"}]}
   ]},
   {<<"memory">>, riak_kv_memory_backend, [
   ]},
   {<<"memorymax1024">>, riak_kv_memory_backend, [
      {max_memory, 1024}
   ]},
   {<<"memoryttl10">>, riak_kv_memory_backend, [
      {ttl, 10}
   ]}
]},
```

then ran

```
riak_core_bucket:set_bucket(<<"bitcask">>, [{backend, <<"bitcask">>}]).
riak_core_bucket:set_bucket(<<"eleveldb">>, [{backend, <<"eleveldb">>}]).
riak_core_bucket:set_bucket(<<"index">>, [{backend, <<"index">>}]).
riak_core_bucket:set_bucket(<<"memory">>, [{backend, <<"memory">>}]).
riak_core_bucket:set_bucket(<<"memorymax1024">>, [{backend, <<"memorymax1024">>}]).
riak_core_bucket:set_bucket(<<"memoryttl10">>, [{backend, <<"memoryttl10">>}]).

Backends = [<<"bitcask">>,<<"eleveldb">>,<<"index">>,<<"memory">>,<<"memorymax1024">>,<<"memoryttl10">>],
[{ B, systest:write(1, 100, B, 3, 3), systest:read(1, 100, B, 3) } || B <- Backends].
[{ B, systest:read(1, 100, B, 3) } || B <- Backends].
```

Made dev2 leave the cluster

```
(dev2@127.0.0.1)3> riak_kv_console:leave([]).
 ok
```

and saw lots of this on dev1

```
11:22:07.767 [info] Handoff receiver for partition 1073290264914881830555831049026020342559825461248 exited after processing 28 objects
11:22:07.768 [error] riak_kv_vnode failed to store handoff obj: ok
```

jonmeredith:
Tried rerunning - this time cleared down dev1/dev2 completely - rejoined them/reconfigured bucket props.

Joined dev2 to dev1.
Set Bucket Props
Issued the backend write fragment above.
Issues leave on dev2

Saw this on dev1

```

12:26:56.740 [error] Handoff receiver for partition 844930634081928249586505293914101120738586001408 exited abnormally after processing 1 objects: {{{badrecord,state},[{riak_kv_multi_backend,stop,1},{riak_kv_vnode,terminate,2},{riak_core_vnode,terminate,3},{gen_fsm,terminate,7},{proc_lib,init_p_do_apply,3}]},{gen_fsm,sync_send_all_state_event,[<0.3250.0>,{handoff_data,<<[141,81,203,74,3,49,20,77,38,29,218,90,148,42,130,224,66,65,186,29,168,218,133,238,221,248,168,136,72,113,87,206,76,210,38,243,132,49,163,83,183,126,140,63,224,23,184,113,237,31,248,25,46,77,58,98,167,226,194,27,184,9,231,222,156,220,115,178,178,154,136,36,203,103,9,202,253,254,193,96,163,65,8,25,110,191,178,39,217,228,164,149,143,51,63,20,129,78,12,186,220,104,145,121,111,108,18,149,140,147,118,62,14,178,84,139,84,203,54,39,13,174,2,13,134,174,89,45,92,161,47,187,225,175,144,180,142,205,121,108,218,178,212,157,91,239,90,33,242,70,55,152,226,16,37,70,120,192,49,46,241,136,2,54,98,179,151,200,112,110,170,83,28,33,199,41,238,113,82,227,97,150,199,85,41,23,165,116,236,185,217,243,139,32,18,127,136,169,234,141,94,36,102,63,186,234,35,81,139,110,126,143,116,129,59,237,13,51,174,38,74,112,201,124,243,200,158,79,58,59,47,62,97,31,103,230,218,130,162,242,166,34,119,214,119,63,165,3,154,186,100,240,252,254,182,22,46,92,162,255,119,105,201,41,202,137,27,196,2,169,33,210,121,33,194,208,252,66,97,4,79,84,42,248,23]>>},60000]}}
```

jonmeredith:
+1 merge - only issue remaining is 

```
14:29:10.931 [error] CRASH REPORT Process [] with 3 neighbours crashed with reason: no match of right hand value ok in riak_kv_index_backend:callback/3
```

messages which Kelly says is fixed in the phase 2 code.
The plan is to merge phase2 branch into phase1 then merge to master, so it should be resolved by then and I will rerun the testing.

PR #152, opened 2011-08-16
Improve the error handling for the fsms that use the coverage code

kellymclaughlin:
Handle errors that may be returned from backend fold functions in the fsms for bucket and key listing and indexes.

### DIFF
--- a/src/riak_kv_buckets_fsm.erl
+++ b/src/riak_kv_buckets_fsm.erl
@@ -63,7 +63,7 @@ process_results({final_results, Buckets},
                 StateData=#state{buckets=BucketAcc}) ->
     {done, StateData#state{buckets=(Buckets ++ BucketAcc)}};
 process_results({error, Reason}, _State) ->
-    Reason.
+    {error, Reason}.
 
 finish({error, Error},
        StateData=#state{client_type=ClientType,
--- a/src/riak_kv_buckets_fsm.erl
+++ b/src/riak_kv_buckets_fsm.erl
@@ -61,7 +61,9 @@ process_results({results, Buckets},
     {ok, StateData#state{buckets=(Buckets ++ BucketAcc)}};
 process_results({final_results, Buckets},
                 StateData=#state{buckets=BucketAcc}) ->
-    {done, StateData#state{buckets=(Buckets ++ BucketAcc)}}.
+    {done, StateData#state{buckets=(Buckets ++ BucketAcc)}};
+process_results({error, Reason}, _State) ->
+    Reason.
 
 finish({error, Error},
        StateData=#state{client_type=ClientType,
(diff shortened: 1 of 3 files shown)


src/riak_kv_keys_fsm.erl
```diff
@@ -79,7 +79,9 @@ process_results({final_results, {Bucket, Keys}},
                 StateData=#state{client_type=ClientType,
                                  from={raw, ReqId, ClientPid}}) ->
     process_keys(ClientType, Bucket, Keys, ReqId, ClientPid),
-    {done, StateData}.
+    {done, StateData};
+process_results({error, Reason}, _State) ->
```
rzezeski, line 83: I don't see anywhere in the `riak_kv_vnode` or the `riak_core_coverage_fsm` that would cause this function head to be matched.

src/riak_kv_index_fsm.erl
```diff
@@ -80,7 +80,9 @@ process_results({final_results, {Bucket, Results}},
                 StateData=#state{client_type=ClientType,
                                  from={raw, ReqId, ClientPid}}) ->
     process_query_results(ClientType, Bucket, Results, ReqId, ClientPid),
-    {done, StateData}.
+    {done, StateData};
+process_results({error, Reason}, _State) ->
```
rzezeski, line 84: I don't see anywhere in the `riak_kv_vnode` or the `riak_core_coverage_fsm` that would cause this function head  be matched.

src/riak_kv_buckets_fsm.erl
```diff
@@ -61,7 +61,9 @@ process_results({results, Buckets},
     {ok, StateData#state{buckets=(Buckets ++ BucketAcc)}};
 process_results({final_results, Buckets},
                 StateData=#state{buckets=BucketAcc}) ->
-    {done, StateData#state{buckets=(Buckets ++ BucketAcc)}}.
+    {done, StateData#state{buckets=(Buckets ++ BucketAcc)}};
+process_results({error, Reason}, _State) ->
```
rzezeski, line 65: Same as other two comments, don't see what could cause this to match.

src/riak_kv_keys_fsm.erl
kellymclaughlin, line 83: You are correct, but this is anticipating the changes when this [pr](https://github.com/basho/riak_kv/pull/150) is merged.

rzezeski, line 83: Fair enough.  Looking at `riak_core_coverage_fsm` I notice that it passes the `Error` along as-is to the finish callback, i.e. `Mod:finish(Error, ModState)`.  However, the keys implementation is expecting `{error, Error}` in it's finish clause but you extract the `Error` from the tuple here so you'll end up with a `function_clause` error.  Note, I didn't test this, just tracing the code in my head.  Did I miss something?

This should be the same for the other two.

kellymclaughlin, line 83: You are right, that's definitely incorrect. Lemme fix.

rzezeski:
+1 to merge

PR #153, opened 2011-08-17
Add the bucket props to the key stream output

russelldb:
Current new api code doesn't return bucket props when keys are streamed


russelldb:
It did/does[1], now it is like the legacy API (including empty {} if props=false)

[1] https://github.com/basho/riak_kv/blob/d1e75e57109870693434cd44e83b640f85455612/src/riak_kv_wm_keylist.erl#L137

russelldb:
Duplicate of 146.

PR #156, opened 2011-08-25
Az596 backend refactoring phase2

kellymclaughlin:
These changes add support secondary indexing to the backend api and makes changes to `riak_kv_vnode` and the secondary index code to use these changes. These changes include a number of optimizations to improve the performance of the index backend including discontinuing the use of a proxy object to store the index information in `merge_index`. My benchmarking shows that standard kv performance (_i.e._ get, put, update, delete) is very similar to the performance using bitcask alone. Benchmarking the performance of index creation and querying remains an outstanding task.

### DIFF
--- a/src/riak_index_mi_backend.erl
+++ b/src/riak_index_mi_backend.erl
@@ -128,19 +128,6 @@ fold_index(FoldFun, Acc, Opts, State) ->
     Results = case Query of
                   [{eq, Field, Term}] ->
                       merge_index:lookup_sync(Pid, Bucket, Field, Term, FilterFun);
-
-                  [{lt, Field, Term}] ->
-                      merge_index:range_sync(Pid, Bucket, Field, undefined, inc(Term, -1), ?SIZE, FilterFun);
-
-                  [{gt, Field, Term}] ->
-                      merge_index:range_sync(Pid, Bucket, Field, inc(Term, 1), undefined, ?SIZE, FilterFun);
-
-                  [{lte, Field, Term}] ->
-                      merge_index:range_sync(Pid, Bucket, Field, undefined, Term, ?SIZE, FilterFun);
-
-                  [{gte, Field, Term}] ->
-                      merge_index:range_sync(Pid, Bucket, Field, Term, undefined, ?SIZE, FilterFun);
-
                   [{gte, Field, StartTerm}, {lte, Field, EndTerm}] ->
                       merge_index:range_sync(Pid, Bucket, Field, StartTerm, EndTerm, ?SIZE, FilterFun);
                   [{all, _Field}] ->
--- a/src/riak_index_mi_backend.erl
+++ b/src/riak_index_mi_backend.erl
@@ -12,7 +12,7 @@
          index/2,
          delete/2,
          lookup_sync/4,
-         fold_index/6,
+         fold_index/4,
          drop/1,
          callback/3
         ]).
@@ -106,9 +106,19 @@ lookup_sync(State, Index, Field, Term) ->
     FilterFun = fun(_Value, _Props) -> true end,
     merge_index:lookup_sync(Pid, Index, Field, Term, FilterFun).
 
-%% @spec fold_index(State :: state(), Bucket :: binary(), Query :: riak_index:query_elements(),
-%%                  SKFun :: function(), Acc :: term(), FinalFun :: function()) -> term().
-fold_index(State, Index, Query, SKFun, Acc, FinalFun) ->
+%% @spec fold_index(FoldFun :: function(), Acc :: term(),
+%%                   Opts :: [term()], State :: state()) -> term().
+fold_index(FoldFun, Acc, Opts, State) ->
+    case lists:keyfind(index, 1, Opts) of
+        false ->
+            %% At this point a check has already been done to verify
+            %% that a tuple of the form {index, Bucket, Query} or
+            %% {bucket, Bucket, Query} exists so we are safe in
+            %% assuming this call to keyfind will return the tuple.
+            {_, Bucket, Query} = lists:keyfind(bucket, 1, Opts);
+        {_, Bucket, Query} ->
+            ok
+    end,                
     Pid = State#state.pid,
     FilterFun = fun(_Value, _Props) -> true end,
 
@@ -116,45 +126,48 @@ fold_index(State, Index, Query, SKFun, Acc, FinalFun) ->
     %% this will run more complicated queries, and will run them
     %% asynchronously.
     Results = case Query of
-                  [{eq, Field, [Term]}] ->
-                      merge_index:lookup_sync(Pid, Index, Field, Term, FilterFun);
+                  [{eq, Field, Term}] ->
+                      merge_index:lookup_sync(Pid, Bucket, Field, Term, FilterFun);
 
-                  [{lt, Field, [Term]}] ->
-                      merge_index:range_sync(Pid, Index, Field, undefined, inc(Term, -1), ?SIZE, FilterFun);
+                  [{lt, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, undefined, inc(Term, -1), ?SIZE, FilterFun);
 
-                  [{gt, Field, [Term]}] ->
-                      merge_index:range_sync(Pid, Index, Field, inc(Term, 1), undefined, ?SIZE, FilterFun);
+                  [{gt, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, inc(Term, 1), undefined, ?SIZE, FilterFun);
 
-                  [{lte, Field, [Term]}] ->
-                      merge_index:range_sync(Pid, Index, Field, undefined, Term, ?SIZE, FilterFun);
+                  [{lte, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, undefined, Term, ?SIZE, FilterFun);
 
-                  [{gte, Field, [Term]}] ->
-                      merge_index:range_sync(Pid, Index, Field, Term, undefined, ?SIZE, FilterFun);
+                  [{gte, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, Term, undefined, ?SIZE, FilterFun);
 
-                  [{range, Field, [StartTerm, EndTerm]}] ->
-                      merge_index:range_sync(Pid, Index, Field, StartTerm, EndTerm, ?SIZE, FilterFun);
-
+                  [{gte, Field, StartTerm}, {lte, Field, EndTerm}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, StartTerm, EndTerm, ?SIZE, FilterFun);
+                  [{all, _Field}] ->
+                      %% Don't think merge_index currently supports
+                      %% a way to use a wildcard to collect all
+                      %% results.
+                      %% @TODO Decide what to do about returning
+                      %% all keys for a given index.
+                      {error, {unsupported_query_element, Query}};
                   _ ->
                       {error, {unknown_query_element, Query}}
               end,
 
-    %% If Results is a list of results, then call SKFun to accumulate
-    %% the results.
-    Results1 = case is_list(Results) of
-                   true -> SKFun({results, Results}, Acc);
-                   false -> {ok, Results}
-               end,
-
-    %% Call FinalFun on the results.
-    case Results1 of
-        {ok, NewAcc} ->
-            FinalFun(done, NewAcc);
+    case Results of 
         {error, Reason} ->
-            FinalFun({error, Reason}, Acc);
-        done ->
-            FinalFun(done, Acc);
-        Other ->
-            FinalFun({error, {unexpected_result, Other}}, Acc)
+            {error, Reason};
+        _ ->
+            FoldKeysFun = fold_keys_fun(FoldFun, Bucket),
+            FoldResults = lists:foldl(FoldKeysFun, Acc, Results),
+            {ok, FoldResults}
+    end.
+
+%% @private
+%% Return a function to fold over keys on this backend
+fold_keys_fun(FoldFun, Bucket) ->
+    fun({Key, _}, Acc) ->
+            FoldFun(Bucket, Key, Acc)
     end.
 
 %% @spec drop(State::state()) -> ok.
--- a/src/riak_index_mi_backend.erl
+++ b/src/riak_index_mi_backend.erl
@@ -168,28 +168,6 @@ drop(State) ->
 callback(_State, _Ref, _Msg) ->
     ok.
 
-
-%% @private
-%% @spec inc(term(), Amt :: integer()) -> term().
-%%
-%% @doc Increments or decrements an Erlang data type by one
-%%      position. Used during construction of exclusive range searches.
-inc(Term, Amt)  when is_integer(Term) ->
-    Term + Amt;
-inc(Term, _Amt) when is_float(Term) ->
-    Term; %% Doesn't make sense to have exclusive range on floats.
-inc(Term, Amt)  when is_list(Term) ->
-    NewTerm = inc(list_to_binary(Term), Amt),
-    binary_to_list(NewTerm);
-inc(Term, Amt)  when is_binary(Term) ->
-    Bits = size(Term) * 8,
-    <<Int:Bits/integer>> = Term,
-    NewInt = inc(Int, Amt),
-    <<NewInt:Bits/integer>>;
-inc(Term, _) ->
-    throw({unhandled_type, binary_inc, Term}).
-
-
 %% ===================================================================
 %% EUnit tests
 %% ===================================================================
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -48,6 +48,7 @@ behaviour_info(callbacks) ->
      {stop,1},        % (State)
      {get,3},         % (Bucket, Key, State)
      {put,4},         % (Bucket, Key, Val, State)
+     {put,5},         % (Bucket, Key, IndexSpecs, Val, State)
      {delete,3},      % (Bucket, Key, State)
      {drop,1},        % (State)
      {fold_buckets,4},% (FoldBucketsFun, Acc, Opts, State),
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -91,9 +91,9 @@ basic_store_and_fetch({Backend, State}) ->
      fun() ->
              [
               ?_assertMatch({ok, _},
-                            Backend:put(<<"b1">>, <<"k1">>, <<"v1">>, State)),
+                            Backend:put(<<"b1">>, <<"k1">>, [], <<"v1">>, State)),
               ?_assertMatch({ok, _},
-                            Backend:put(<<"b2">>, <<"k2">>, <<"v2">>, State)),
+                            Backend:put(<<"b2">>, <<"k2">>, [], <<"v2">>, State)),
               ?_assertMatch({ok,<<"v2">>, _},
                             Backend:get(<<"b2">>, <<"k2">>, State)),
               ?_assertMatch({error, not_found, _},
@@ -232,7 +232,7 @@ fold_objects({Backend, State}) ->
                                 lists:sort(Objects1)
                             end),
               ?_assertMatch({ok, _},
-                            Backend:put(<<"b3">>,<<"k3">>,<<"v3">>, State)),
+                            Backend:put(<<"b3">>, <<"k3">>, [], <<"v3">>, State)),
               ?_assertEqual([{{<<"b1">>,<<"k1">>},<<"v1">>},
                              {{<<"b3">>,<<"k3">>},<<"v3">>}],
                             begin
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -47,7 +47,6 @@ behaviour_info(callbacks) ->
      {start,2},       % (Partition, Config)
      {stop,1},        % (State)
      {get,3},         % (Bucket, Key, State)
-     {put,4},         % (Bucket, Key, Val, State)
      {put,5},         % (Bucket, Key, IndexSpecs, Val, State)
      {delete,3},      % (Bucket, Key, State)
      {drop,1},        % (State)
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -31,6 +31,7 @@
          stop/1,
          get/3,
          put/4,
+         put/5,
          delete/3,
          drop/1,
          fold_buckets/4,
@@ -136,6 +137,21 @@ get(Bucket, Key, #state{ref=Ref}=State) ->
                  {error, term(), state()}.
 put(Bucket, Key, Val, #state{ref=Ref}=State) ->
     BitcaskKey = term_to_binary({Bucket, Key}),
+    case bitcask:put(Ref, BitcaskKey, Val) of
+        ok ->
+            {ok, State};
+        {error, Reason} ->
+            {error, Reason, State}
+    end.
+
+%% @doc Insert an object with secondary index 
+%% information into the bitcask backend
+-type index_spec() :: {add, Index, SecondaryKey} | {remove, Index, SecondaryKey}.
+-spec put(riak_object:bucket(), riak_object:key(), [index_spec()], binary(), state()) ->
+                 {ok, state()} |
+                 {error, term(), state()}.
+put(Bucket, PrimaryKey, _IndexSpecs, Val, #state{ref=Ref}=State) ->
+    BitcaskKey = term_to_binary({Bucket, PrimaryKey}),
     case bitcask:put(Ref, BitcaskKey, Val) of
         ok ->
             {ok, State};
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -30,7 +30,6 @@
          start/2,
          stop/1,
          get/3,
-         put/4,
          put/5,
          delete/3,
          drop/1,
@@ -60,7 +59,7 @@
                 opts :: [{atom(), term()}],
                 root :: string()}).
 
--opaque(state() :: #state{}).
+-type state() :: #state{}.
 -type config() :: [{atom(), term()}].
 
 %% ===================================================================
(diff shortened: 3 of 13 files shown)


src/riak_object.erl
```diff
@@ -323,6 +329,99 @@ increment_vclock(Object=#r_object{}, ClientId) ->
...
+                               dict:fetch(?MD_INDEX, MD)
+                           end || MD <- MetaDatas, dict:is_key(?MD_INDEX, MD)]
+                         ),
+                    IndexResolver = 
+                        fun({Index, Value}, Acc) ->
+                             case orddict:is_key(Index, CurrentIndexes) of
+                                 true ->
+                                     CurrentValue =
+                                         orddict:fetch(Index, CurrentIndexes),
+                                     case Value < CurrentValue of
```
rustyio: Are you sure this should be a less than comparison? Shouldn't it be `Value /= CurrentValue`?

src/riak_object.erl
```diff
@@ -323,6 +329,99 @@ increment_vclock(Object=#r_object{}, ClientId) ->
...
     Object#r_object{vclock=vclock:increment(ClientId, Timestamp, Object#r_object.vclock)}.
 
+%% @doc Prepare a list of index specifications
+%% to pass to the backend. The `WithSpecialFields'
+%% parameter is used to include or exclude the special
+%% index fields for the bucket and key.
+-spec get_index_specs(riak_object(), index_op(), boolean()) ->
+                             [{index_op(), binary(), index_value()}].
+get_index_specs(RObj, IndexOp, WithSpecialFields) ->
+    case value_count(RObj) > 1 of
```
rustyio: Do we really need two different cases? The sibling case should also work when there are no siblings.

src/riak_kv_vnode.erl
```diff
@@ -447,23 +479,52 @@ select_newest_content(Mult) ->
...
             Obj0 = binary_to_term(Val0),
-            ResObj = riak_object:syntactic_merge(
-                       Obj0,Obj1,term_to_binary(ReqId), StartTime),
+            case IndexBackend of
+                true ->
+                    IndexSpecs = riak_object:resolve_index_specs(Obj0, Obj1);
+                false ->
+                    IndexSpecs = []
+            end,
+            ResObj = riak_object:syntactic_merge(Obj0,
```
rustyio: This is not right. The syntactic_merge is going to invalidate the indexes. We need another step that updates IndexSpecs with the result of the merge.

rustyio:
See comments above.

src/riak_object.erl
kellymclaughlin: Yeah there is probably little if any benefit to splitting this into two cases.

src/riak_kv_vnode.erl
```diff
@@ -475,23 +513,52 @@ select_newest_content(Mult) ->
...
 
 %% @private
+get_index_specs(Mod, ModState, {Bucket, Key}, Obj1) ->
+    case Mod:get(Bucket, Key, ModState) of
+        {error, not_found, _UpdModState} ->
+            riak_object:get_index_specs(Obj1, add, true);
+        {ok, Val0, _UpdModState} ->
+            NewIndexSpecs = riak_object:get_index_specs(Obj1, add, false),
+            Obj0 = binary_to_term(Val0),
+            OldIndexSpecs = riak_object:get_index_specs(Obj0, remove),
```
rustyio: riak_object:get_index_specs/2 doesn't exist.

src/riak_object.erl
```diff
@@ -323,6 +329,87 @@ increment_vclock(Object=#r_object{}, ClientId) ->
...
+                    %% are no siblings.
+                    []
+            end;
+        false ->
+            %% Current object has no siblings so we replace
+            %% the current indexes with the indexes from the
+            %% new object.
+            CurrentIndexes = get_index_specs(CurrentObj, remove, false),
+            NewIndexes = get_index_specs(NewObj, add, false),
+            CurrentIndexes ++ NewIndexes
```
rustyio: This may cause problems if we add and remove the same value within a batch of index updates. merge_index will likely treat the value as deleted, and leveldb doesn't define what happens. We should consolidate the index changes in advance to avoid this problem.

src/riak_kv_vnode.erl
```diff
@@ -475,23 +513,52 @@ select_newest_content(Mult) ->
...
+            OldIndexSpecs ++ NewIndexSpecs
```
rustyio: This may cause problems if we add and remove the same value within a batch of index updates. merge_index will likely treat the value as deleted, and leveldb doesn't define what happens. We should consolidate the index changes in advance to avoid this problem.

src/riak_object.erl
```diff
@@ -323,6 +329,87 @@ increment_vclock(Object=#r_object{}, ClientId) ->
...
+                             [{index_op(), binary(), index_value()}].
+resolve_index_specs(CurrentObj, NewObj) ->
+    case value_count(CurrentObj) > 1 of
+        true ->
+            NewMD = get_metadata(NewObj),
+            case dict:is_key(?MD_INDEX, NewMD) of
+                true ->
+                    NewIndexes = dict:fetch(?MD_INDEX, NewMD),
+                    MetaDatas = get_metadatas(CurrentObj),
+                    CurrentIndexes =
```
rustyio: This section duplicates code in get_index_specs/3, seems like we may be able to consolidate?

src/riak_object.erl
```diff
@@ -323,6 +329,87 @@ increment_vclock(Object=#r_object{}, ClientId) ->
...
+                                             [{remove, Index, CurrentValue},
+                                              {add, Index, Value}]  ++ Acc;
+                                         false ->
+                                             Acc
+                                     end;
+                                 false ->
+                                     [{add, Index, Value} | Acc]
+                             end
+                         end,
+                    lists:foldl(IndexResolver, [], NewIndexes);
```
rustyio: As noted elsewhere, we need to apply the IndexResolver logic wherever we combine old and new indexes. Also, any way that this can be simplified, perhaps something like:

``` erlang

%% OldIndexes is list of {Field, Value} tuples
%% NewIndexes is list of {Field, Value} tuples

%% Convert the index lists to ordered sets...
OldSet = ordsets:from_list(OldIndexes),
NewSet = ordsets:to_list(NewIndexes),

%% Add any fields that are in the New set, but not in the Old set...
AddSet = ordsets:subtract(NewSet, OldSet),
AddOps = [{add, Field, Value} || {Field, Value} <- ordsets:to_list(AddSet)],

%% Delete any fields that are in the Old set, but not in the New set...
DeleteSet = ordsets:subtract(OldSet, NewSet),
DeleteOps = [{delete, Field, Value} || {Field, Value} <- ordsets:to_list(DeleteSet)],

FinalIndexOps = AddOps ++ DeleteOps
```

Also, with this resolver in place everywhere, we could probably get rid of the variable that determines whether special fields are included. Just always include them, and the resolver will determine when to filter them out. (It looks like the only place we don't include them in the vnode is during a put with last-write-wins enabled. Individual backends may still filter them out, but that should be a backend thing.)

src/riak_index_mi_backend.erl
```diff
@@ -116,45 +126,48 @@ fold_index(State, Index, Query, SKFun, Acc, FinalFun) ->
...
+                  [{gt, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, inc(Term, 1), undefined, ?SIZE, FilterFun);
+
+                  [{lte, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, undefined, Term, ?SIZE, FilterFun);
+
+                  [{gte, Field, Term}] ->
+                      merge_index:range_sync(Pid, Bucket, Field, Term, undefined, ?SIZE, FilterFun);
+
+                  [{gte, Field, StartTerm}, {lte, Field, EndTerm}] ->
```
rustyio: If these terms are reversed, it will lead to an error. Also, a more recent checkin reduced available operations down to 'eq' and 'range'. Gets rid of lt/gt/lte/gte as a way of simplifying code.

src/riak_kv_vnode.erl
kellymclaughlin: Oh, I didn't realize that was the case.

src/riak_index_mi_backend.erl
kellymclaughlin: Do you think [this](https://github.com/basho/riak_kv/pull/156/files#L9L157) check is enough to ameliorate the problem of term reversal for the range?

kellymclaughlin:
I believe the changes I've made address all the concerns and feedback. My testing yields the behavior I expect in the case of objects with and without siblings.

kellymclaughlin:
Went ahead and refactored the memory backend to remove the use of `gen_server` and added the required extra plumbing to `riak_kv_vnode`.

src/riak_kv_vnode.erl
```diff
@@ -380,12 +399,11 @@ handle_exit(_Pid, Reason, State) ->
 
 %% old vnode helper functions
 
-
-%store_call(State=#state{mod=Mod, modstate=ModState}, Msg) ->
-%    Mod:call(ModState, Msg).
+%% store_call(State=#state{mod=Mod, modstate=ModState}, Msg) ->
```
jonmeredith: Seems like dead comments we could chop.

src/riak_kv_backend.erl
```diff
@@ -48,6 +48,7 @@ behaviour_info(callbacks) ->
      {stop,1},        % (State)
      {get,3},         % (Bucket, Key, State)
      {put,4},         % (Bucket, Key, Val, State)
```
jonmeredith: drop {put, 4} - only need {put, 5}.  Backends that don't support index data should still store the k/v obj ignoring SI entries and fail on index queries.  One put to rule them all eye-of-Mordor.

src/riak_kv_bitcask_backend.erl
```diff
@@ -31,6 +31,7 @@
          stop/1,
          get/3,
          put/4,
```
jonmeredith: kill put/4

src/riak_kv_eleveldb_backend.erl
```diff
@@ -29,6 +29,7 @@
          stop/1,
          get/3,
          put/4,
```
jonmeredith: kill put/4

src/riak_kv_index_backend.erl
```diff
@@ -32,6 +32,7 @@
          stop/1,
          get/3,
          put/4,
```
jonmeredith: kill put/4

src/riak_kv_index_backend.erl
```diff
@@ -195,9 +209,18 @@ fold_buckets(FoldBucketsFun, Acc, Opts, #state{kv_mod = KVMod,
                 any(),
                 [{atom(), term()}],
                 state()) -> {ok, term()} | {error, term()}.
-fold_keys(FoldKeysFun, Acc, Opts, #state{kv_mod = KVMod,
+fold_keys(FoldKeysFun, Acc, Opts, #state{index_mod = IndexMod,
+                                         index_state = IndexState,
+                                         kv_mod = KVMod,
                                          kv_state = KVState}) ->
-    KVMod:fold_keys(FoldKeysFun, Acc, Opts, KVState).
+    case lists:keymember(index, 1, Opts) orelse
```
jonmeredith, line 202: if we're using keymember/keyfind on options need to make  sure they are {Key,Val} options, not proplists.  Maybe dialyzer can help.

src/riak_kv_memory_backend.erl
```diff
@@ -82,14 +74,15 @@
...
+-record(state, {data_ref :: integer() | atom(),
                        time_ref :: integer() | atom(),
                        expiry_enabled :: boolean(),
                        memory_cap_enabled :: boolean(),
                        max_memory :: undefined | integer(),
                        used_memory=0 :: integer(),
                        ttl :: integer()}).
--type state() :: pid().
+
+-opaque(state() :: #state{}).
```
jonmeredith: Why is state opaque?

src/riak_kv_memory_backend.erl
```diff
@@ -82,14 +74,15 @@
 -define (DEFAULT_TTL, 3600). % 1 hour
 -define (DEFAULT_MEMORY,  100).  % 100MB
 
--record(server_state, {data_ref :: integer() | atom(),
```
jonmeredith, line 85: calling it something like bestate makes it easier to read crash dumps when you're digging through a vnode.

src/riak_kv_memory_backend.erl
```diff
@@ -107,55 +100,120 @@ api_version() ->
...
-        {ok, Pid} ->
-            {ok, Pid};
-        {error, Reason} ->
-            {error, Reason};
-        ignore ->
-            {error, ignore}
-    end.
+    ExpiryEnabled = config_value(object_expiry_enabled, Config, false),
+    TTL = config_value(ttl, Config, ?DEFAULT_TTL),
+    MemoryCapEnabled = config_value(memory_cap_enabled, Config, false),
```
jonmeredith: Do we need memory_cap_enabled and object_expiry_enabled?  Seems like specifying ttl should be enough for expirty and max_memory for bounding memory.  I'm not sure there's a sane default as it's app specific, that's all I can see the enabled giving us.

src/riak_kv_memory_backend.erl
```diff
@@ -107,55 +100,120 @@ api_version() ->
...
 
 %% @doc Insert an object into the memory backend
 -spec put(riak_object:bucket(), riak_object:key(), binary(), state()) ->
                  {ok, state()} |
                  {error, term(), state()}.
-put(Bucket, Key, Val, State) ->
-    case gen_server:call(State, {put, Bucket, Key, Val}) of
-        ok ->
-            {ok, State};
+put(Bucket, Key, Val, State=#state{data_ref=DataRef,
```
jonmeredith: No need for put/4 any more.  Just make this put/5 and ignore the index entries.

src/riak_kv_memory_backend.erl
```diff
@@ -107,55 +100,120 @@ api_version() ->
...
-    case gen_server:call(State, {delete, Bucket, Key}) of
-        ok ->
+delete(Bucket, Key, State=#state{data_ref=DataRef}) ->
+    case ets:delete(DataRef, {Bucket, Key}) of
+        true ->
             {ok, State};
-        {error, Reason} ->
-            {error, Reason, State}
+        _ ->
+            {error, ets_delte_failed, State}
```
jonmeredith: typo in error message, would be nice to pass the reason on.

src/riak_kv_memory_backend.erl
```diff
@@ -163,164 +221,69 @@ delete(Bucket, Key, State) ->
...
                    [{atom(), term()}],
                    state()) -> {ok, any()} | {error, term()}.
-fold_objects(FoldObjectsFun, Acc, Opts, State) ->
+fold_objects(FoldObjectsFun, Acc, Opts, #state{data_ref=DataRef}) ->
     Bucket =  proplists:get_value(bucket, Opts),
     FoldFun = fold_objects_fun(FoldObjectsFun, Bucket),
-    gen_server:call(State, {fold_objects, FoldFun, Bucket, Acc}).
+    case Bucket of
+        undefined ->
+            ObjectList = ets:match_object(DataRef, {{'_', '_'}, '_'});
```
jonmeredith: Would be better to use ets:foldl here or ets:match/3 and use the continuations rather than building a whopper of a list and then folding over that.  Any extra gc is bad for vnode latency.

src/riak_kv_memory_backend.erl
```diff
@@ -163,164 +221,69 @@ delete(Bucket, Key, State) ->
                    any(),
                    [],
                    state()) -> {ok, any()} | {error, term()}.
-fold_buckets(FoldBucketsFun, Acc, _Opts, State) ->
+fold_buckets(FoldBucketsFun, Acc, _Opts, #state{data_ref=DataRef}) ->
     FoldFun = fold_buckets_fun(FoldBucketsFun),
-    gen_server:call(State, {fold_buckets, FoldFun, Acc}).
+    BucketList = ets:match(DataRef, {{'$1', '_'}, '_'}),
```
jonmeredith: See L258

src/riak_kv_memory_backend.erl
```diff
@@ -163,164 +221,69 @@ delete(Bucket, Key, State) ->
...
                 [{atom(), term()}],
                 state()) -> {ok, term()} | {error, term()}.
-fold_keys(FoldKeysFun, Acc, Opts, State) ->
+fold_keys(FoldKeysFun, Acc, Opts, #state{data_ref=DataRef}) ->
     Bucket =  proplists:get_value(bucket, Opts),
     FoldFun = fold_keys_fun(FoldKeysFun, Bucket),
-    gen_server:call(State, {fold_keys, FoldFun, Bucket, Acc}).
+    case Bucket of
+        undefined ->
+            KeyList = ets:match(DataRef, {{'$1', '$2'}, '_'});
```
jonmeredith: See L258

src/riak_kv_memory_backend.erl
```diff
@@ -163,164 +221,69 @@ delete(Bucket, Key, State) ->
...
+    Acc0 = lists:foldl(FoldFun, Acc, ObjectList),
+    {ok, Acc0}.
 
 %% @doc Delete all objects from this memory backend
--spec drop(state()) -> {ok, state()} | {error, term(), state()}.
-drop(State) ->
-    gen_server:call(State, drop).
+-spec drop(state()) -> {ok, state()}.
+drop(State=#state{data_ref=DataRef}) ->
+    ets:delete_all_objects(DataRef),
```
jonmeredith, line 245: What about TimeRef?

src/riak_kv_memory_backend.erl
```diff
@@ -163,164 +221,69 @@ delete(Bucket, Key, State) ->
...
-    gen_server:call(State, is_empty).
+-spec is_empty(state()) -> boolean().
+is_empty(#state{data_ref=DataRef}) ->
+    ets:info(DataRef, size) =:= 0.
 
 %% @doc Get the status information for this memory backend
 -spec status(state()) -> [{atom(), term()}].
-status(State) ->
-    gen_server:call(State, status).
+status(#state{data_ref=DataRef}) ->
```
jonmeredith: What about TimeRef?

src/riak_kv_memory_backend.erl
```diff
@@ -107,55 +100,120 @@ api_version() ->
...
-        {error, Reason} ->
-            {error, Reason, State}
+get(Bucket, Key, State=#state{data_ref=DataRef,
+                              ttl=TTL}) ->
+    case ets:lookup(DataRef, {Bucket, Key}) of
+        [] -> {error, not_found, State};
+        [{{Bucket, Key}, {{ts, Timestamp}, Val}}] ->
+            case exceeds_ttl(Timestamp, TTL) of
+                true ->
+                    ets:delete(DataRef, {Bucket, Key}),
```
jonmeredith: You need to delete from TimeRef too - maybe it would be better to have a function that takes Bucket, Key, State and does the appropriate deletes and call that from get when TTL exceeded, put when memory is exceeded and delete.

src/riak_kv_multi_backend.erl
```diff
@@ -29,6 +29,7 @@
          stop/1,
          get/3,
          put/4,
```
jonmeredith: ditch put/4

src/riak_kv_vnode.erl
```diff
@@ -482,20 +542,29 @@ select_newest_content(Mult) ->
...
+    case riak_object:vclock(ResObj) =:= riak_object:vclock(Obj0) of
+        true ->
+            {oldobj, ResObj};
+        false ->
+            {newobj, ResObj}
+    end.
 
-syntactic_put_merge(Mod, ModState, {Bucket, Key}, Obj1, ReqId, StartTime) ->
+%% @private
+get_index_specs(Mod, ModState, {Bucket, Key}, Obj1) ->
```
jonmeredith: Can we consider renaming this to lookup_index_specs or something like that, or call the versions in riak_object index_specs/diff_index_specs as I mistakenly thought you were doing unnecessary gets.  I can see this is only called in the lww case.

src/riak_kv_vnode.erl
```diff
@@ -706,17 +775,44 @@ do_legacy_list_buckets(Caller,ReqId,Idx,Mod,ModState) ->
...
-index_query(Sender, Bucket, Query, Filter, Mod, ModState) ->
-    %% @TODO Should probably add some buffering like the key listing uses.
-    QueryResults = Mod:fold_index(ModState, Bucket, Query),
+index_query(Sender, Bucket, Query, Filter, Mod, ModState, BufferSize) ->
+    BufferFun = fun(Results) ->
+                        riak_core_vnode:reply(Sender,
+                                              {results, {Bucket, Results}})
+                end,
+    Buffer = riak_kv_fold_buffer:new(BufferSize, BufferFun),
     case Filter of
```
jonmeredith, line 752: This code looks the same as the one in list_keys - could we factor out the creation of FoldKeysFun/FlushFun from both?

src/riak_object.erl
```diff
@@ -323,6 +329,66 @@ increment_vclock(Object=#r_object{}, ClientId) ->
...
+
+%% @doc Prepare a list of index specifications to pass to the
+%% backend. If the object passed as the first parameter does not have
+%% siblings, the function will assemble specs to replace all of its
+%% indexes with the indexes specified in the object passed as the
+%% second parameter. If there are siblings only the unique new
+%% indexes are added.
+-spec get_index_specs(riak_object(), riak_object()) ->
+                             [{index_op(), binary(), index_value()}].
+get_index_specs(Obj, OldObj) ->
```
jonmeredith: Why are we treating siblings differently? Lots of ordsets calls when you could code up as a recursive function (rough example, untested, not drop in replacement)

```
get_index_specs(Obj, OldObj) ->
  Old = lists:usort(get_index_data(OldObj)),
  Upd = lists:usort(get_index_data(Obj)),
  diff(Old, Upd, []).

diff([], [], Acc) ->
  Acc;
diff([], Upd, Acc) ->
  [{X, add} || X <- Upd] ++ Acc;
diff(Old, [], Acc) ->
  [{X, remove} || X <- Old] ++ Acc;
diff([Spec|Old],[Spec|Upd], Acc) ->
  diff(Old, Upd, Acc);
diff([OSpec|Old],[USpec|Upd], Acc) when OSpec < USpec -> % OSpec not in new spec
  diff(Old, Upd, [{OSpec, remove}]);
diff([Old, [USpec|Upd], Acc) -> % Uspec must not be in ospec
  diff(Old, Upd, [{USpec, add}]).
```

This really needs comprehensive unit tests - preferably quickcheck ones maybe written with ordsets :)

jonmeredith:
I've added my blow-by-blow.  I've done a tiny bit of running the code but will hold off until you've had a chance to look at the review comments.  I did noticed that riak_kv_index_fsm is a bit fragile about failure.

```

(dev1@127.0.0.1)52> C:get_index(<<"b">>,[{gte, <<"idx1_bin">>, <<"b">>}]).
08:19:32.514 [error] gen_fsm <0.7740.0> in state waiting_results terminated with reason: no function clause matching riak_kv_index_fsm:process_results({error,{unknown_query_element,[{gte,<<105,100,120,49,95,98,105,110>>,<<98>>}]}}, {state,plain,{raw,67416687,<0.7345.0>}})
08:19:32.517 [error] CRASH REPORT Process [] with 0 neighbours crashed with reason: no function clause matching riak_kv_index_fsm:process_results({error,{unknown_query_element,[{gte,<<105,100,120,49,95,98,105,110>>,<<98>>}]}}, {state,plain,{raw,67416687,<0.7345.0>}})
08:19:32.518 [error] Supervisor riak_kv_index_fsm_sup had child at module undefined at <0.7740.0> exit with reason no function clause matching riak_kv_index_fsm:process_results({error,{unknown_query_element,[{gte,<<105,100,120,49,95,98,105,110>>,<<98>>}]}}, {state,plain,{raw,67416687,<0.7345.0>}}) in context child_terminated
{error,timeout,[]}
```

Also had a similar error when I ran with bitcask set as the backend - it crashed/hung for 60s instead of immediately telling me index operations were not supported.  The log message is fixed by a lager pull request so that should be nicer next time it crashes.

src/riak_kv_memory_backend.erl
kellymclaughlin: Not sure what's best to do here. `ets:delete` will only return `true` and dialyzer is not fooled by the error clause that will never be reached so maybe just get rid of it and amend the specs.

kellymclaughlin:
Removed `put/4` from all backends.

src/riak_kv_memory_backend.erl
kellymclaughlin: So you would say ditch the default values for ttl and memory cap and if ttl or max_memory is not defined then expiry or memory cap respectively are not enabled? That's the only reason the for `_enabled` options is to allow the use of those defaults that were carried over from the cache backend.

src/riak_kv_memory_backend.erl
kellymclaughlin, line 85: so `bestate` instead of `state` for all the backends?

src/riak_kv_memory_backend.erl
kellymclaughlin: changed to type

src/riak_kv_memory_backend.erl
kellymclaughlin: Done. Changed to use `ets:foldl` for all 3 fold functions.

src/riak_kv_memory_backend.erl
jonmeredith: agreed - get rid of it and just match on true/false - then if ets ever adds a new return and somebody updates the specs dialyzer will point out our ignorance.

src/riak_kv_memory_backend.erl
jonmeredith, line 85: That's what I like, but it's probably a bit late to make changes to all of the backends - I only mentioned it as you were renaming it here.

src/riak_kv_memory_backend.erl
jonmeredith: Yes - given we renamed the backend and config variables it'll need a person to adjust the app.config.  Seems like they're both parameters you can't guess good defaults for anyway.

src/riak_kv_memory_backend.erl
kellymclaughlin, line 85: Since all the other backends are using `state` I'll leave this for now and we can think about changing it later.

src/riak_kv_memory_backend.erl
```diff
@@ -105,222 +84,198 @@ api_version() ->
...
+            {ok, Val, State};
+        Error ->
+            {error, Error, State}
     end.
 
-%% @doc Insert an object into the memory backend
--spec put(riak_object:bucket(), riak_object:key(), binary(), state()) ->
+%% @doc Insert an object into the memory backend.
+%% NOTE: The memory backend does not currently support
+%% secondary indexing. This function is only here
```
jonmeredith: I'd just say memory backend does not support secondary indexing so _IndexSpecs are ignored.  It's the real-deal put function, not a dummy stub.

jonmeredith:
+1 merge.

PR #157, opened 2011-08-26
az://303 vclock enhancements redux

jonmeredith:
Final version of the node-based vclock work.  

The put FSM has been changed in the following ways
- Puts must run on a node in the preference list.  If not in the preflist it will be forwarded by the put FSM.
- Puts are written to the local vnode before being written to the remotes.
- Vclocks are no longer incremented by riak_client, they are done by the first vnode in the preference list. 
- The Id for the vector clock update is based on the node name and start time for the vnode.
- The changes should significantly reduce the number of entries present in the vclocks and make pruning a very unusual event.
- Because the vclock updates are now guaranteed to create vclocks that are frontiers the introduction of extra vclock ids in riak_kv_vnode:syntactic_put_merge has now been removed, getting rid of a class of vclock explosion problems.

This p/r replaces the original for az://303 and should be reviewed with an eye to the comments raised by Bryan/Scott being addressed.  It needs to be reviewed in conjunction with the riak_core az://303 pull request.

### DIFF
--- a/src/riak_kv_put_core.erl
+++ b/src/riak_kv_put_core.erl
@@ -102,8 +102,9 @@ enough(#putcore{w = W, num_w = NumW, dw = DW, num_dw = NumDW,
         (NumW < W andalso NumFail >= WFailThreshold).
 
 %% True if coordinator result has been returned - anything other than w.
-coord_result(CoordIdx, Results) ->
-    [Result || {Idx, Result} <- Results, Idx == CoordIdx] -- [w] /= [].
+coord_result(_CoordIdx, _Results) ->
+    true.
+%    [Result || {Idx, Result} <- Results, Idx == CoordIdx] -- [w] /= [].
 
 %% Get success/fail response once enough results received
 -spec response(putcore()) -> {reply(), putcore()}.
--- a/src/riak_kv_put_core.erl
+++ b/src/riak_kv_put_core.erl
@@ -20,9 +20,8 @@
 %%
 %% -------------------------------------------------------------------
 -module(riak_kv_put_core).
--export([init/7, coord_idx/2, add_result/2, enough/1, response/1, 
+-export([init/7, add_result/2, enough/1, response/1, 
          final/1]).
--export([coord_result/2]).%% DEBUG
 -export_type([putcore/0, result/0, reply/0]).
 
 -type vput_result() :: any().
@@ -44,7 +43,6 @@
                   dw_fail_threshold :: pos_integer(),
                   returnbody :: boolean(),
                   allowmult :: boolean(),
-                  coord_idx :: non_neg_integer(),
                   results = [] :: [idxresult()],
                   final_obj :: undefined | riak_object:riak_object(),
                   num_w = 0 :: non_neg_integer(),
@@ -65,9 +63,6 @@ init(N, W, DW, WFailThreshold, DWFailThreshold, AllowMult, ReturnBody) ->
              dw_fail_threshold = DWFailThreshold,
              allowmult = AllowMult,
              returnbody = ReturnBody}.
-
-coord_idx(Idx, PutCore) ->
-    PutCore#putcore{coord_idx = Idx}.
    
 %% Add a result from the vnode
 -spec add_result(vput_result(), putcore()) -> putcore().
@@ -95,16 +90,10 @@ add_result(_Other, PutCore = #putcore{num_fail = NumFail}) ->
 -spec enough(putcore()) -> boolean().
 enough(#putcore{w = W, num_w = NumW, dw = DW, num_dw = NumDW, 
                 num_fail = NumFail, w_fail_threshold = WFailThreshold,
-                dw_fail_threshold = DWFailThreshold,
-                coord_idx = CoordIdx, results = Results}) ->
-    (NumW >= W andalso NumDW >= DW andalso coord_result(CoordIdx, Results)) orelse
+                dw_fail_threshold = DWFailThreshold}) ->
+    (NumW >= W andalso NumDW >= DW) orelse
         (NumW >= W andalso NumFail >= DWFailThreshold) orelse
         (NumW < W andalso NumFail >= WFailThreshold).
-
-%% True if coordinator result has been returned - anything other than w.
-coord_result(_CoordIdx, _Results) ->
-    true.
-%    [Result || {Idx, Result} <- Results, Idx == CoordIdx] -- [w] /= [].
 
 %% Get success/fail response once enough results received
 -spec response(putcore()) -> {reply(), putcore()}.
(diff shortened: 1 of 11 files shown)


jonmeredith:
riak_core pull request: https://github.com/basho/riak_core/pull/72
original pull request: https://github.com/basho/riak_kv/pull/127

argv0:
i didn't do a thorough review, but I've done a significant amount of local testing with this PR and saw no errors.

src/riak_kv_put_core.erl
```diff
@@ -90,11 +95,17 @@ add_result(_Other, PutCore = #putcore{num_fail = NumFail}) ->
...
+                dw_fail_threshold = DWFailThreshold,
+                coord_idx = CoordIdx, results = Results}) ->
+    (NumW >= W andalso NumDW >= DW andalso coord_result(CoordIdx, Results)) orelse
         (NumW >= W andalso NumFail >= DWFailThreshold) orelse
         (NumW < W andalso NumFail >= WFailThreshold).
 
+%% True if coordinator result has been returned - anything other than w.
+coord_result(_CoordIdx, _Results) ->
+    true.
+%    [Result || {Idx, Result} <- Results, Idx == CoordIdx] -- [w] /= [].
```
beerriot: This looks like a debugging modification that was left in - is it?

src/riak_kv_put_fsm.erl
```diff
@@ -71,6 +75,7 @@
                 n :: pos_integer(),
                 w :: non_neg_integer(),
                 dw :: non_neg_integer(),
+                coord_idx_node,
```
beerriot: `:: atom()`?

src/riak_kv_put_fsm.erl
```diff
@@ -258,35 +289,51 @@ precommit(timeout, State = #state{precommit = [Hook | Rest], robj = RObj}) ->
...
-                                   timeout=Timeout, preflist2 = Preflist2, bkey=BKey,
-                                   vnode_options=VnodeOptions,
-                                   putcore=PutCore,
-                                   starttime = StartTime}) ->
+execute_local(timeout, StateData0=#state{robj=RObj, req_id = ReqId,
+                                         timeout=Timeout, bkey=BKey,
+                                         coord_idx_node = {_Index, _Node} = CoordIndexNode,
+                                         vnode_options=VnodeOptions,
+                                         starttime = StartTime}) ->
+    %RObj = riak_object:increment_vclock(RObj0, Node, StartTime),
```
beerriot: `<General call for not leaving unused code hanging around commented />`

src/riak_kv_stat.erl
```diff
@@ -410,7 +414,8 @@ node_stats(Moment, State=#state{node_gets_total=NGT,
      {node_put_fsm_time_95, PutNF},
      {node_put_fsm_time_99, PutNN},
      {node_put_fsm_time_100, PutH},
-     {read_repairs_total, RRT}].
+     {read_repairs_total, RRT},
+     {coord_redirs_total, CRT}].
```
beerriot, line 422: Please add to the module's doc, a description of what `coord_redirs_total` is.

src/riak_object.erl
```diff
@@ -101,29 +101,47 @@ ensure_robject(Obj = #r_object{}) -> Obj.
...
+            case Obj1#r_object.key =:= Obj2#r_object.key of
+                false -> false;
+                true -> equal1(Obj1,Obj2)
+            end
+    end.
+equal1(Obj1,Obj2) ->
+    case vclock:equal(vclock(Obj1),vclock(Obj2)) of
+        false -> false;
+        true -> equal2(Obj1,Obj2)
+    end.
```
beerriot: Why swap the short-circuiting boolean operations for more verbose case statements?

src/riak_object.erl
```diff
@@ -137,9 +155,9 @@ reconcile(Objects, AllowMultiple) ->
     AllContents = lists:flatten([O#r_object.contents || O <- RObjs]),
     Contents = case AllowMultiple of
         false ->
-            [most_recent_content(AllContents)];
+            [hd(lists:sort(fun compare_content_dates/2, AllContents))];
```
beerriot: Not that `hd(lists:sort(...` is hard to read, but wasn't `most_recent_content` more descriptive?

beerriot:
I've noticed two differences between the AZ303 branch and current master so far.  I've put code in a gist (https://gist.github.com/445ece2b530c15a06008) to demonstrate each of:
- writes using vector clocks that are ancestors to the current vclock are no longer rejected, and instead cause siblings (`az303:ignore_old_vclocks/2`)
- allow_mult is no longer enforced at write time, so siblings are stored, even if only one value will be returned on read (`az303:enforce_allow_mult_on_put/2`)

Each function returns `pass` on master, but `{fail, Description, Object}` on az303 (apologies for the pass/fail semantics if this is an intended change).

beerriot:
It should be well-documented that no node should have `vnode_vclocks=true` before every node is running the az303 branch code.  Otherwise, objects may be written without vclocks.  To reproduce, start a cluster with one node running az303 and `vnode_vclocks=true`, then add 3 nodes running master to that cluster.  Write some objects, read them back, and you'll find vclocks of `[]`.

src/riak_kv_put_core.erl
jonmeredith: Yes, was an early stab at an alternative method where the put request got sent to all nodes with one marked as a coordinator and if you discovered you weren't creating a frontier vclock then you resubmitted.  That method may still work but I hit some problems with it and decided to simplify.

src/riak_kv_put_fsm.erl
jonmeredith: ty - fixed.

src/riak_kv_put_fsm.erl
jonmeredith: :: {integer(), atom()} - it's actually the coordinator preflist entry so I've renamed it to coord_pl_entry instead

src/riak_object.erl
jonmeredith: These are rebase failure - I'll try and work out the best way to fix.

src/riak_object.erl
jonmeredith: Rebase failure - code should match the current master.

jonmeredith:
Thx for the code examples.

The put behavior with out of date vclocks is by design - the idea is that if a user has said they want to put data we should try and do that even if it has been subsequently updated.  I do however need to add a way so we can do repl puts without changing the vclocks.

The allow_mult behavior should be enforced - I'll address that when I rebase after backend phase2 review is completed.

jonmeredith:
Superceded by https://github.com/basho/riak_kv/pull/166

PR #159, opened 2011-08-31
Make sure MapReduce pipes are torn down if the clients waiting for their results disconnect early.

beerriot:
This branch will accumulate patches that shutdown MapReduce pipelines as soon as the clients that requested them disconnect.  There are more patches to come, but I thought I'd try out the "start a pull request early" workflow.

These first patches target the ProtocolBuffers socket.  The two big changes are:
- not `pause`ing the PB socket process while waiting for mapred results
- sending the inputs from a process other than the PB socket process

In order to support the former, it is now an error that will cause the socket to shutdown, to send any request to a socket before it returns the results of a pipe-based mapred request.  The alternative to this would be to queue any requests that arrive while we waiting for mapred results, and then work through them after sending the results to the client, but that raises complexity quite a bit, and is likely to lead to other ordering errors.

Commits yet to come:
- the same kind of immediate teardown for HTTP-based mapred queries
- removing the `pause` behavior from PB list-keys and legacy mapreduce

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -335,12 +335,19 @@ send_inputs_async(Pipe, Arg) ->
 send_inputs_async(Pipe, Arg, Timeout) ->
     spawn_monitor(
       fun() ->
+              %% tear this process down if the pipeline goes away;
+              %% also automatically tears down the pipeline if feeding
+              %% it inputs fails (which is what the users of this
+              %% function, riak_kv_pb_socket and riak_kv_wm_mapred, want)
+              erlang:link(Pipe#pipe.builder),
               case send_inputs(Pipe, Arg, Timeout) of
                   ok ->
                       %% monitoring process sees a 'normal' exit
+                      %% (and linked builder is left alone)
                       ok;
                   Error ->
                       %% monitoring process sees an 'error' exit
+                      %% (and linked builder dies)
                       exit(Error)
               end
       end).
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -106,6 +106,7 @@
          send_inputs_async/3,
          collect_outputs/2,
          collect_outputs/3,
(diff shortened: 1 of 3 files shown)


beerriot:
This PR now also contains the commits for AZ509 (pipe support for HTTP&PB mapred timeouts).  It turns out to not be possible to immediately tear down a pipe as soon as the HTTP client that started it disconnects, because Webmachine is not actively monitoring the socket.  Timeouts are well-respected now, so users can use that facility to time the death of their mapred queries and disconnects over HTTP.

beerriot:
In the interest of cleaning this up and merging for release, I'm going to back out of removing the use of `pause` in the PB support of listkeys and legacy mapreduce.

kellymclaughlin:
I got some strange results when using a mapred timeout of 1 with the pb client.

<pre><code>
** exception error: no match of right hand side value {error,{timeout,[]}}
(riak@127.0.0.1)36> 11:47:49.335 [error] Pipe worker startup failed:fitting was gone before startup
11:47:49.335 [error] Supervisor riak_pipe_builder_sup had child at module undefined at <0.615.0> exit with reason killed in context child_terminated
11:47:49.338 [error] Supervisor riak_pipe_fitting_sup had child at module undefined at <0.617.0> exit with reason killed in context child_terminated
11:47:49.338 [error] Supervisor riak_pipe_fitting_sup had child at module undefined at <0.618.0> exit with reason killed in context child_terminated
11:47:49.339 [error] Supervisor riak_pipe_fitting_sup had child at module undefined at <0.616.0> exit with reason killed in context child_terminated
</code></pre>


Using such a short timeout and it seems like the pipe mechanisms just haven't had time to get set up. I discussed with Bryan and decided to leave investigating this as a separate task since using such a short timeout is an edge case that no sane users will do.

kellymclaughlin:
+1 on the timeouts

slfritchie:
Can't seem to make it do bad things. +1 to merge, thanks!

PR #160, opened 2011-09-01
Support for specifying Timeout in Pipe-powered MapReduce queries.

beerriot:
(another experiment in opening the PR early in the process)

This branch will add correct support for specifying a timeout on MapReduce queries over HTTP and PB.  This first patch is the PB half.

Note that it's an extension of the az628-close-pipe-on-error branch.


beerriot:
I've merged this into the az628-close-pipe-on-error branch, since later commits improved things relevant to that ticket.

PR #161, opened 2011-09-02
support stfun in pipe-power mapreduce

beerriot:
just what it says on the tin


rzezeski:
Tested successfully for both raw source and bucket/key source using Erlang.

+1 to merge

PR #164, opened 2011-09-02
Az603 backend refactoring phase3

kellymclaughlin:
Adds support for asynchronous folds for the backends. The async behavior is used by default if it is available in the backend (this is the case for bitcask, eleveldb, and memory), but can be disabled using an entry in the `riak_kv` section of `app.config`. Setting `async_folds` to `false` will override the default.

This pull request also updates the `backend_eqc` module to test the async folds and bitcask, eleveldb, and memory backends now run quickcheck tests using both sync and async folds.

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -248,7 +248,31 @@ fold_keys(FoldKeysFun, Acc, Opts, #state{ref=Ref}) ->
 -spec fold_objects(riak_kv_backend:fold_objects_fun(),
                    any(),
                    [{atom(), term()}],
-                   state()) -> {ok, any()} | {error, term()}.
+                   state()) -> {ok, any()} | {async, fun()} | {error, term()}.
+fold_objects(FoldObjectsFun, Acc, Opts, #state{opts=BitcaskOpts,
+                                               data_file=DataFile,
+                                               root=DataRoot,
+                                               async_folds=true}) ->
+    Bucket =  proplists:get_value(bucket, Opts),
+    FoldFun = fold_objects_fun(FoldObjectsFun, Bucket),
+    case BitcaskOpts of
+        [{read_write, _} | Config] ->
+            ok;
+        Config ->
+            ok
+    end,
+    ReadOpts = [{read_only, true} | Config],
+    ObjectFolder =
+        fun() ->
+                case bitcask:open(filename:join(DataRoot, DataFile),
+                                  ReadOpts) of
+                    Ref when is_reference(Ref) ->
+                        bitcask:fold(Ref, FoldFun, Acc);
+                    {error, Reason} ->
+                        {error, Reason}
+                end
+        end,
+    {async, ObjectFolder};
 fold_objects(FoldObjectsFun, Acc, Opts, #state{ref=Ref}) ->
     Bucket =  proplists:get_value(bucket, Opts),
     FoldFun = fold_objects_fun(FoldObjectsFun, Bucket),
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -88,7 +88,7 @@ start(Partition, Config) ->
             %% of the file linked to by the existing link.
             case check_symlink(LinkFile, false) of
                 {ok, DataFile} ->
-                    BitcaskOpts = [{read_write, true} | Config],
+                    BitcaskOpts = set_mode(read_write, Config),
                     case bitcask:open(filename:join(DataRoot, DataFile), BitcaskOpts) of
                         Ref when is_reference(Ref) ->
                             schedule_merge(Ref),
@@ -175,13 +175,7 @@ fold_buckets(FoldBucketsFun, Acc, _Opts, #state{opts=BitcaskOpts,
                                                 root=DataRoot,
                                                 async_folds=true}) ->
     FoldFun = fold_buckets_fun(FoldBucketsFun),
-    case BitcaskOpts of
-        [{read_write, _} | Config] ->
-            ok;
-        Config ->
-            ok
-    end,
-    ReadOpts = [{read_only, true} | Config],
+    ReadOpts = set_mode(read_only, BitcaskOpts),
     BucketFolder =
         fun() ->
                 case bitcask:open(filename:join(DataRoot, DataFile),
@@ -215,13 +209,7 @@ fold_keys(FoldKeysFun, Acc, Opts, #state{opts=BitcaskOpts,
                                          async_folds=true}) ->
     Bucket =  proplists:get_value(bucket, Opts),
     FoldFun = fold_keys_fun(FoldKeysFun, Bucket),
-    case BitcaskOpts of
-        [{read_write, _} | Config] ->
-            ok;
-        Config ->
-            ok
-    end,
-    ReadOpts = [{read_only, true} | Config],
+    ReadOpts = set_mode(read_only, BitcaskOpts),
     KeyFolder =
         fun() ->
                 case bitcask:open(filename:join(DataRoot, DataFile),
@@ -255,13 +243,7 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{opts=BitcaskOpts,
                                                async_folds=true}) ->
     Bucket =  proplists:get_value(bucket, Opts),
     FoldFun = fold_objects_fun(FoldObjectsFun, Bucket),
-    case BitcaskOpts of
-        [{read_write, _} | Config] ->
-            ok;
-        Config ->
-            ok
-    end,
-    ReadOpts = [{read_only, true} | Config],
+    ReadOpts = set_mode(read_only, BitcaskOpts),
     ObjectFolder =
         fun() ->
                 case bitcask:open(filename:join(DataRoot, DataFile),
@@ -543,6 +525,14 @@ data_directory_cleanup(DirPath) ->
         _ ->
             ignore
     end.
+
+%% @private
+set_mode(read_only, Config) ->
+    Config1 = lists:keystore(read_only, 1, Config, {read_only, true}),
+    lists:keydelete(read_write, 1, Config1);
+set_mode(read_write, Config) ->
+    Config1 = lists:keystore(read_write, 1, Config, {read_write, true}),
+    lists:keydelete(read_only, 1, Config1).
 
 %% ===================================================================
 %% EUnit tests
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -52,12 +52,14 @@
 
 -define(MERGE_CHECK_INTERVAL, timer:minutes(3)).
 -define(API_VERSION, 1).
--define(CAPABILITIES, []).
+-define(CAPABILITIES, [async_fold]).
 
 -record(state, {ref :: reference(),
                 link_file :: string(),
+                data_file :: string(),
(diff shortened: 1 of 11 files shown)


src/riak_kv_bitcask_backend.erl
```diff
@@ -164,7 +169,30 @@ delete(Bucket, Key, #state{ref=Ref}=State) ->
...
                    [],
-                   state()) -> {ok, any()} | {error, term()}.
+                   state()) -> {ok, any()} | {async, fun()} | {error, term()}.
+fold_buckets(FoldBucketsFun, Acc, _Opts, #state{opts=BitcaskOpts,
+                                                data_file=DataFile,
+                                                root=DataRoot,
+                                                async_folds=true}) ->
+    FoldFun = fold_buckets_fun(FoldBucketsFun),
+    case BitcaskOpts of
+        [{read_write, _} | Config] ->
```
Vagabond: Are the options always going to have a guaranteed order so this will work? Maybe lists:keystore would be more appropriate than this case.

kellymclaughlin: The `read_write` option is always put on the head of the list [here](https://github.com/basho/riak_kv/blob/9313535d115cca3490851ac87e24e105337508f4/src/riak_kv_bitcask_backend.erl#L91) in the start function so it should be ok I think.

Vagabond: The coupling is very implicit, though, if that other code was  changed you'd get weird behavior over here.

Vagabond: Also, this same thing is done over and over again, getting bitcask options with read_only set to true. This should be a separate function that does it right.

kellymclaughlin: Yep you are correct. Changes to how those props are initialized would yank the rug out from under all of those cases. Will change.

src/riak_kv_bitcask_backend.erl
```diff
@@ -233,9 +309,10 @@ drop(#state{ref=Ref,
             %% so this backend can continue processing.
             case bitcask:open(filename:join(DataRoot, DataFile), BitcaskOpts) of
                 Ref1 when is_reference(Ref1) ->
-                    {ok, State#state{ref=Ref1}};
+                    {ok, State#state{data_file=DataFile,
```
Vagabond, line 294: Are these changes because of an oversight from the previous refactor?

kellymclaughlin, line 294: Trying to use the link file was causing problems in the eqc tests when it was doing a lot of drops so added the data_file to the state to avoid the issues. I didn't have time to fully explore why the link file wasn't adequate. I would like to just get rid of the link file altogether in the future or better yet make bitcask take care of itself.

src/riak_kv_bitcask_backend.erl
kellymclaughlin: Change pushed.

src/riak_kv_eleveldb_backend.erl
```diff
@@ -151,17 +160,26 @@ fold_buckets(FoldBucketsFun, Acc, _Opts, #state{fold_opts=FoldOpts,
...
+    {async, KeyFolder};
 fold_keys(FoldKeysFun, Acc, Opts, #state{fold_opts=FoldOpts1,
                                          ref=Ref}) ->
     Bucket =  proplists:get_value(bucket, Opts),
     FoldOpts = fold_opts(Bucket, FoldOpts1),
     FoldFun = fold_keys_fun(FoldKeysFun, Bucket),
-    try
-        Acc0 = eleveldb:fold_keys(Ref, FoldFun, Acc, FoldOpts),
-        {ok, Acc0}
+        try
```
Vagabond: I think you messed up the indentation here.

kellymclaughlin: Fixed.

src/riak_kv_fold_worker.erl
```diff
@@ -0,0 +1,140 @@
...
+
+%% @doc Perform the asynchronous fold operation.
+handle_work({Bucket, [FoldFun | RestFoldFuns]}, From, State) ->
+    FoldResults = try
+                      FoldFun()
+                  catch
+                      {break, AccFinal} ->
+                          AccFinal
+                  end,
+    case FoldResults of
```
Vagabond: Is this for final accumulator vs intermediate one or what?

kellymclaughlin: It's to account for the bucket folds which use the second tuple element to track previously seen buckets.

Vagabond: Using a named underscore variable would make this a little clearer then, but its not essential.

src/riak_kv_multi_backend.erl
```diff
@@ -203,23 +209,25 @@ delete(Bucket, Key, State) ->
...
+                              {Acc1, SyncResults, [Work | AsyncWork]};
                           {error, Reason} ->
                               throw({error, {Module, Reason}})
                       end
               end,
     try
-        Acc0 = lists:foldl(FoldFun, Acc, Backends),
-        {ok, Acc0}
+        {_, SyncResults, AsyncWork} = lists:foldl(FoldFun, {Acc, [], []}, Backends), 
+        {{sync, SyncResults}, {async, AsyncWork}}
```
Vagabond, line 238: Doesn't the spec say this should be {ok, {sync, ...}, {async, ...}}?

src/riak_kv_vnode.erl
```diff
@@ -181,20 +182,27 @@ init([Index]) ->
...
+    case catch Mod:start(Index, [{async_folds, AsyncFolding},
+                                 Configuration]) of
         {ok, ModState} ->
             %% Get the backend capabilities
             {_, Capabilities} = Mod:api_version(),
             IndexBackend = lists:member(indexes, Capabilities),
+            AsyncBackend = AsyncFolding andalso
+                lists:member(async_fold, Capabilities),
+            %% Create worker pool initialization tuple
+            FoldWorkerPool = {pool, riak_kv_fold_worker, 10, []},
```
Vagabond: Why create the worker pool if the backend isn't async capable, or enabled?

kellymclaughlin: Excellent observation. Fixed.

src/riak_kv_fold_worker.erl
kellymclaughlin: My only problem with naming it is determining something appropriate to use as a the name. What exactly it is depends on the backend. For eleveldb it's just the name of the last bucket, but for the bitcask and memory backends it's a set and for some future backend it could be who knows what. For that reason I think it's good just to leave it as `_`.

Vagabond:
I'm mostly satisfied with the code now (although some more comments wouldn't hurt). I'm concerned, however, with the code coverage. Parts of riak_kv_vnode, especially those that deal with the weird {{async, ...}, {sync, ...}} return value are not covered by either the test or the eunit code. Similarly, the fold worker code only covers the case where there's a single foldfun.

kellymclaughlin:
I broke out some of the repeated code dealing with folds into `riak_kv_fold_util`. It also has some better documentation I think. The only thing lacking in the module is eunit tests, but the code is indirectly tested with `backend_eqc`, but I do plan to add direct testing for it, but at this point I can't be sure it will be done before Tuesday morning.

I expanded the eqc testing done with the multi backend to test sync folding, async folding, and both together.

src/riak_kv_fold_util.erl
```diff
@@ -0,0 +1,172 @@
...
+%% @doc Take the appropriate action to finalize a fold operation based
+%% on the result from a call to a riak_kv backend fold function such
+%% as fold_buckets.
+-spec finish_fold(fold_result(), from()) -> term().
+finish_fold(FoldResult, Sender) ->
+    case FoldResult of
+        {{sync, [FirstSyncResult | RestSyncResults]}, {async, []}} ->
+            [riak_kv_fold_buffer:flush(SyncBuf, get_buffer_fun(false, Sender))
+             || SyncBuf <- RestSyncResults],
+            riak_kv_fold_buffer:flush(FirstSyncResult,
```
Vagabond, line 83: Why do you flush these out of order, or seperately at all?

Vagabond:
There's still big chunks of fold_worker and fold_util not being covered, even by the EQC tests.

src/riak_kv_fold_util.erl
Vagabond, line 83: Oh, it it just so one of them finishes the flush?

kellymclaughlin, line 83: That's right. There is no guarantee of order, but it should only send one indication of final results back to the initiator of the fold so the others are just flush without sending the end of fold indication.

jonmeredith:
Had an error with legacy list keys

```
09:45:59.764 [error] gen_fsm <0.426.0> in state active terminated with reason: no case clause matching {async,#Fun<riak_kv_bitcask_backend.1.41931593>} in riak_kv_vnode:do_legacy_list_keys/6
```

Created a cluster of 4 nodes - set different backend on each, bitcask, eleveldb, memory, multi (although probably not necessary to reproduce)

On all nodes,

```
riak:join('dev1@127.0.0.1').
```

On dev1

```
systest:write(1000).
systest:listkeys(1000).
```

Vagabond:
I've managed to get my cluster to report ever-changing # of keys when doing a list keys:

  (dev1@127.0.0.1)25> length(element(2, RC:list_keys(<<"test">>))).
  38963
  (dev1@127.0.0.1)26> length(element(2, RC:list_keys(<<"test">>))).
  38973
  (dev1@127.0.0.1)27> length(element(2, RC:list_keys(<<"test">>))).
  38967
  (dev1@127.0.0.1)28> length(element(2, RC:list_keys(<<"test">>))).
  38973

Vagabond:
Downing a node also changes the keycount:

(dev1@127.0.0.1)14> length(element(2, RC:list_keys(<<"test">>))).
30278
*\* dev4 stopped **
(dev1@127.0.0.1)15> length(element(2, RC:list_keys(<<"test">>))).
31059
*\* dev4 started **
(dev1@127.0.0.1)17> length(element(2, RC:list_keys(<<"test">>))).
30278

Vagabond:
Oh, my cluster is configured as following:

dev1 bitcask default
dev2 eleveldb default
dev3 memory default
dev4 bitcask with async folding disabled

The unstable keycount seemed to go away after I restarted the whole cluster.

Vagabond:
Ok, I have an unstable keycount again by just running a basho_bench against the above cluster config and aborting the basho bench halfway through (I was bored).

jtuple:
Just finished testing async handoff in a variety of scenarios and things look good from my end. In particular, async handoff while adding/removing nodes from the cluster while the cluster is simultaneously being written to / read from seemed to do the right thing.

The pre-existing vnode forward logic in vnode_handle_handoff command (used by riak_kv) correctly wrote changes locally as well as forwarded them to the handoff target during the async fold, therefore allowing the existing "delete all partition data" on handoff complete to remain a correct transition. Likewise, the new forwarding logic in vnode_command still kicked in and performed the necessary fowards to the new partition owner during the ownership transition period.

All looked good, data didn't go missing. No new not_found behaviors during handoff resurfaced. Of course, this was all done by hand. I'll work on writing some actual repeatable tests once we branch and ramp up 1.0 testing.

jonmeredith:
+1 for now - passes my simple tests.
Will need to refactor to make multi backend work more cleanly.

PR #166, opened 2011-09-02
az://303 vclock enhancements redux be2

jonmeredith:
az303 redux branch rebased against be phase 2, use in conjunction with the same named riak_core branch.

To get the new vclock behavior you must set {vnode_vclocks, true} before retrieving riak_client objects.

This should address final comments about allow_mult in https://github.com/basho/riak_kv/pull/157.

Bryan requested whoever reviewed double-checked deletes.

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -317,7 +317,7 @@ waiting_local_vnode(Result, StateData = #state{putcore = PutCore}) ->
             case Result of
                 {fail, _Idx, _ReqId} ->
                     %% Local vnode failure is enough to sink whole operation
-                    process_reply({error, fail}, StateData#state{putcore = UpdPutCore1});
+                    process_reply({error, too_many_fails}, StateData#state{putcore = UpdPutCore1});
                 {w, _Idx, _ReqId} ->
                     {next_state, waiting_local_vnode, StateData#state{putcore = UpdPutCore1}};
                 {dw, _Idx, PutObj, _ReqId} ->
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -183,12 +183,13 @@ prepare(timeout, StateData0 = #state{from = From, robj = RObj,
             %% This node is not in the preference list
             %% forward on to the first node
             [{{_Idx, CoordNode},_Type}|_] = Preflist2,
-            %% io:format(user, "Forwarding ~p to coordinator on ~p\n", [BKey, CoordNode]),
             case riak_kv_put_fsm_sup:start_put_fsm(CoordNode, [From, RObj, Options]) of
                 {ok, _Pid} ->
                     riak_kv_stat:update(coord_redir),
                     {stop, normal, StateData0};
                 {error, Reason} ->
+                    lager:error("Unable to forward put for ~p to ~p - ~p\n",
+                                [BKey, CoordNode, Reason]),
                     process_reply({error, {coord_handoff_failed, Reason}}, StateData0)
             end;
         [CoordPLEntry|_] ->
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -75,7 +75,7 @@
                 n :: pos_integer(),
                 w :: non_neg_integer(),
                 dw :: non_neg_integer(),
-                coord_idx_node,
+                coord_pl_entry :: {integer(), atom()},
                 preflist2 :: riak_core_apl:preflist2(),
                 bkey :: {riak_object:bucket(), riak_object:key()},
                 req_id :: pos_integer(),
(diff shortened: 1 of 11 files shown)


src/riak_kv_put_fsm.erl
```diff
@@ -162,23 +167,42 @@ init({test, Args, StateProps}) ->
...
-    new_state_timeout(validate, StateData).
-    
+    %% Check if this node is in the preference list so it can coordinate
+    case ([IndexNode || {{_Index, Node} = IndexNode, _Type} <- Preflist2,
+                        Node == node()]) of
+        [] ->
+            %% This node is not in the preference list
+            %% forward on to the first node
+            [{{_Idx, CoordNode},_Type}|_] = Preflist2,
+            %% io:format(user, "Forwarding ~p to coordinator on ~p\n", [BKey, CoordNode]),
```
kellymclaughlin: Get rid of debug output.

src/riak_kv_put_fsm.erl
```diff
@@ -382,6 +428,27 @@ process_reply(Reply, StateData = #state{postcommit = PostCommit,
             new_state_timeout(finish, StateData2)
     end.
 
+
+%% @private
+%% Send the put requests to any remote nodes if necessary and decided if 
+%% enough responses have been received yet (i.e. if W/DW=1)
+execute_remote(StateData=#state{robj=RObj, req_id = ReqId,
```
kellymclaughlin, line 436: What about moving `execute_remote` before `waiting_remote_vnode`. That way the states would be in the order of their normal transition. Doesn't matter functionally, but makes more sense for reading the code.

test/ec_eqc.erl
```diff
@@ -0,0 +1,955 @@
...
+test(N) ->
+    quickcheck(numtests(N, prop())).
+
+check() ->
+    check(prop(), current_counterexample()).
+
+-define(B, <<"b">>).
+-define(K, <<"k">>).
+
+                                                %-define(FINALDBG(Fmt,Args), io:format("= " ++ Fmt, Args)).
```
kellymclaughlin: Add an extra `%` here so the emacs formatting doesn't keep sending him way over there.

test/ec_eqc.erl
```diff
@@ -0,0 +1,955 @@
...
+gen_client_seeds() ->
+    ?LET({Clients, Handoff}, 
+         {non_empty(list(gen_regular_client())), gen_handoff_client()},
+         [Handoff | Clients]).
+
+prop() ->
+    ?FORALL({Pris, ClientSeeds, ParamsSeed},
+            {gen_pris(), gen_client_seeds(),gen_params()},
+            begin
+                %% io:format(user, "Pris: ~p\n", [Pris]),
```
kellymclaughlin, line 123: Not sure if this was left intentionally or not so just point it out in case you meant to remove.

kellymclaughlin:
Seeing a problem with indexes missing after an update with `allow_mult=false`. Steps to reproduce using index backend:
- curl -v -X PUT -H "Content-Type: application/json" \
  -d '{"props":{"allow_mult":true}}' \
  http://127.0.0.1:8098/riak/kitchen
- curl -v -X POST -H "Content-Type: application/json" \
  -d '{"dishes":11}' \
  http://127.0.0.1:8098/riak/kitchen/sink?returnbody=true
- curl -v -X PUT -H "Content-Type: application/json" \
  -H "x-riak-index-field1_bin: val2" \
  -H "x-riak-index-field2_int: 1003" \
  -d '{"dishes":9}' \
  http://127.0.0.1:8098/riak/kitchen/sink?returnbody=true

Then `curl -v localhost:8098/buckets/kitchen/index/field1_bin/eq/val2` should return `sink`, but I get an empty list.

src/riak_kv_put_fsm.erl
jonmeredith: Removed - added error log for forwarding failure.

src/riak_kv_put_fsm.erl
jonmeredith, line 436: I need to add a variant of the put FSM that just stores and does not increment for repl - I'll restructure then.  It does read better that way but it makes it seem like execute_remote is a state when it isn't.

test/ec_eqc.erl
jonmeredith: Fixed.

jonmeredith:
I made a mistake in the refactor to combine to the put paths - for a coordinated put the vnode has to return the object.  With it in place I get this result.  I think the same issue was affecting the keys_fsm_eqc test.

```
jons-macpro:riak_kv(az303-vclock-enhancements-redux-be2) jmeredith$ curl -v localhost:8091/buckets/kitchen/index/field1_bin/eq/val2
* About to connect() to localhost port 8091 (#0)
*   Trying 127.0.0.1... connected
* Connected to localhost (127.0.0.1) port 8091 (#0)
> GET /buckets/kitchen/index/field1_bin/eq/val2 HTTP/1.1
> User-Agent: curl/7.19.7 (universal-apple-darwin10.0) libcurl/7.19.7 OpenSSL/0.9.8r zlib/1.2.3
> Host: localhost:8091
> Accept: */*
> 
< HTTP/1.1 200 OK
< Vary: Accept-Encoding
< Server: MochiWeb/1.1 WebMachine/1.8.1 (participate in the frantic)
< Date: Sat, 03 Sep 2011 21:43:34 GMT
< Content-Length: 17
< 
* Connection #0 to host localhost left intact
* Closing connection #0
{"keys":["sink"]}
```

src/riak_kv_put_fsm.erl
kellymclaughlin, line 436: Nevermind, I didn't pay close enough attention and thought it was a state.

kellymclaughlin:
The keys_fsm eqc test is working consistently for me now and the index issue is resolved. I'll probably just make one more pass through the code before giving it the green light for merge.

jonmeredith:
Replaced by https://github.com/basho/riak_kv/pull/173 which is rebased off master post-cluster membership merge.

PR #170, opened 2011-09-04
AZ650: Java client mapred queries vs. Pipe

beerriot:
Just one fix in this branch so far: returning non-streamed HTTP results in the order they were received.  They're returned in reverse order, otherwise.

### DIFF
--- a/src/riak_kv_wm_mapred.erl
+++ b/src/riak_kv_wm_mapred.erl
@@ -179,7 +179,11 @@ pipe_mapred_nonchunked(RD, State, Pipe, NumKeeps, Sender) ->
         {error, timeout} ->
             %% destroying the pipe will tear down the linked sender
             riak_pipe:destroy(Pipe),
-            {{halt, 500}, send_error({error, timeout}, RD), State}
+            {{halt, 500}, send_error({error, timeout}, RD), State};
+        {error, {Error, _Input}} ->
+            %% TODO: jsonify Input for error message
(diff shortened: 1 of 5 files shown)


beerriot:
Two more fixes:
- it is not required to `keep` at least one MapReduce phase (in contrast to link-walking, which does enforce this)
- bubbling Javascript evaluation timeouts up to the client as soon as they happen

src/riak_kv_wm_mapred.erl
```diff
@@ -179,7 +179,11 @@ pipe_mapred_nonchunked(RD, State, Pipe, NumKeeps, Sender) ->
         {error, timeout} ->
             %% destroying the pipe will tear down the linked sender
             riak_pipe:destroy(Pipe),
-            {{halt, 500}, send_error({error, timeout}, RD), State}
+            {{halt, 500}, send_error({error, timeout}, RD), State};
+        {error, {Error, _Input}} ->
+            %% TODO: jsonify Input for error message
```
rzezeski, line 184: Did you plan on taking care of this now or in a future PR?

beerriot, line 184: Ah, sorry for leaving such a vague comment in.  Definitely a future PR, covering more error messages than just `timeout`, the one necessary to fix the Java client's tests.

rzezeski:
Other than the mentioned TODO things look fine to me.

+1 to merge

PR #171, opened 2011-09-05
Fixed delete/put in backend_eqc.

jonmeredith:
Couple of fixes to make the backend EQC test work correctly again.

Added the IndexEntries arg to delete and added IndexEntries to the postcondition for put/delete.


kellymclaughlin:
I don't think this is correct. Why did you change the call to delete to be arity 4 when the backend spec expects 3?

kellymclaughlin:
Ok disregard that last comment. I did not realize we had changed delete to have 4 parameters and I missed the `-membership` on the end of the new branch name.

kellymclaughlin:
Looks good. Thanks for catching my missed update on that put postcondition. +1.

PR #172, opened 2011-09-05
Made keys fsm EQC work with the new ring code.

jonmeredith:
Changes to add the test node as a member and assign partitions in a ring trans.


test/keys_fsm_eqc.erl
```diff
@@ -129,7 +140,7 @@ prop_basic_listkeys() ->
                end,
                %% Call start_link
                Keys = start_link(ReqId, Bucket, KeyFilter, Timeout, ClientType),
-               ?WHENFAIL(
+              ?WHENFAIL(
```
kellymclaughlin, line 143: Fix ident.

kellymclaughlin:
+1

PR #174, opened 2011-09-06
Changes to riak_client so that if legacy_keylisting is not specified in the config, the legacy code will be used.

kellymclaughlin:


Vagabond:
Why not default the value in the env section of the riak_kv.app file?

seancribbs:
@Vagabond Because you want new clusters to use the new setting?

Vagabond:
But it'll be in the app.config file defaulted to false for new clusters, the app.config file overides the app.src file.

kellymclaughlin:
Updated per your suggestion.

Vagabond:
For absolute safety, the second clause of the case where you test for legacy_keylisting should probably be _, not false.

kellymclaughlin:
Good point.

Vagabond:
+1

PR #175, opened 2011-09-07
Az655 osync warning

massung:
Requires branch az655-osync-warning from riak_core as well. On linux, checks for bitcask config {sync_strategy,o_sync} and warns (if on Linux) that it will fail.

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -319,7 +319,7 @@ check_fcntl() ->
 	{undefined,{ok,o_sync}} ->
 	    case riak_core_util:is_arch(linux) of
 		true ->
-		    lager:warning("fcntl(f,F_SETFL,O_SYNC) fails on Linux."),
+		    lager:warning("{sync_strategy,o_sync} fails on Linux"),
 		    application:set_env(riak_kv,o_sync_warning_logged,true);
 		_ ->
 		    ok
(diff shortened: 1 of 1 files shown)


src/riak_kv_bitcask_backend.erl
```diff
@@ -89,6 +89,7 @@ start(Partition, Config) ->
                     BitcaskOpts = [{read_write, true} | Config],
                     case bitcask:open(filename:join(DataRoot, DataFile), BitcaskOpts) of
                         Ref when is_reference(Ref) ->
+			    check_fcntl(),
```
rzezeski: Substitute hard tabs for spaces.

src/riak_kv_bitcask_backend.erl
```diff
@@ -309,6 +310,25 @@ key_counts(RootDir) ->
...
+%% On linux there is a kernel bug that won't allow fcntl to add O_SYNC
+%% to an already open file descriptor.
+check_fcntl() ->
+    Logged=application:get_env(riak_kv,o_sync_warning_logged),
+    Strategy=application:get_env(bitcask,sync_strategy),
+    case {Logged,Strategy} of
+	{undefined,{ok,o_sync}} ->
+	    case riak_core_util:is_arch(linux) of
+		true ->
+		    lager:warning("{sync_strategy,o_sync} fails on Linux"),
```
jonmeredith: I would say 'not implemented' rather than fails.

rzezeski:
So why not put this check in `riak_kv_app:start/2` instead and avoid the "check if I already logged this" code?

jonmeredith:
Because it's a pluggable backend.  It's in the right place.

rzezeski:
+1 to merge

PR #176, opened 2011-09-07
Correct keyfilter evaluation in non-legacy keylister.

beerriot:
In the legacy keylisting system, keyfilters were defined as chains of
filters, where the output of one was in the input to the next.  This
patch restores that behavior for the new keylisting system.

This bug was found via the Java client integration tests.

This patch should be merged to both 1.0 and master.


kellymclaughlin:
+1 to merge.

PR #178, opened 2011-09-07
bz1175 add client_id to state of pb socket to emulate legacy API

russelldb:
Get client Id requests to the PB interface resulted in crash if the local client's id was undefined. This change adds a client_id value to the pb socket's state so that it can behave as before for legacy clients


seancribbs:
@russelldb You'll still need to set the client if `vnode_vclocks` (or whatever it's called) is false.

russelldb:
Doh! You're right. I rushed it. Starting again.

jonmeredith:
Initial code review looks good - can you add defaults to the get_env's of false as it is deliberately nod set in the .app file so rolling upgrades work.

jonmeredith:
+1 merge to master and 1.0 branches.

Code tests with vnode_vclocks, true, false and unset.

PR #180, opened 2011-09-07
Only run EQC if it's installed

rzezeski:


Vagabond:
+1, seems to solve the issue.

PR #181, opened 2011-09-07
AZ665 - Add variability to write_buffer_size to reduce concurrent compaction.

rustyio:
During benchmark tests, adding some variability to the write_buffer_size smoothed out latencies and sped up average write speeds during heavy loads by reducing the probability that all vnodes would kick off a compaction at the same time.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -55,7 +55,8 @@
                 read_opts = [],
                 write_opts = [],
                 fold_opts = [{fill_cache, false}],
-                async_folds :: boolean()}).
+                async_folds :: boolean(),
+                write_buffer_size :: integer()}).
 
 -type state() :: #state{}.
 -type config() :: [{atom(), term()}].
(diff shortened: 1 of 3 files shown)


src/riak_kv_eleveldb_backend.erl
```diff
@@ -78,11 +79,27 @@ start(Partition, Config) ->
...
     AsyncFolds = config_value(async_folds, Config, true),
-    case eleveldb:open(DataDir, [{create_if_missing, true}]) of
+
+    %% Use a variable write buffer size in order to reduce the number
+    %% of vnodes that try to kick off compaction at the same time
+    %% under heavy uniform load...
+    WriteBufferMin = config_value(write_buffer_size_min, Config, 3 * 1024 * 1024),
+    WriteBufferMax = config_value(write_buffer_size_max, Config, 6 * 1024 * 1024),
+    random:seed(now()),
+    WriteBufferSize = WriteBufferMin + random:uniform(WriteBufferMax - WriteBufferMin),
```
rzezeski: what if someone sets max and min to same value?

rzezeski:
+1 to merge

I ran the before/after code thru a 1 hour benchmark using [basho_bench](http://wiki.basho.com/Benchmarking-with-Basho-Bench.html) along with internal tool to graph IO/CPU.
- 3 physical machines in cluster
- 4th machine as load gen using 32 concurrent workers in `max` mode
- duration of 1hr
- 8KB value size
- 6 indexes on each object (3 bin, 3 int)
- binary index size of 20 bytes
- default min/max fuzzing values of 3MB/6MB

It was hard to differentiate the bench graphs but looking at the IO/CPU graphs it was very clear that fuzzing produces a much more consistent IO & CPU load across the run whereas before it was very spiky with hills and valleys.

PR #183, opened 2011-09-08
Using platform_data_dir to build the default path to data/kv_vnode.

massung:
If platform_data_dir is not defined in riak_core, then it will default back to ./data.


slfritchie:
Jeff, sorry, I didn't notice that you'd been working on this also.
Could you take a peek at #184, and if you like it, consider it
review and close #183?  I'll do the merges.

jonmeredith:
reviewed ok - but is superseded by #184.

PR #184, opened 2011-09-08
Fix BZ 1176: bad assumptions about cwd writablility

slfritchie:
Use app.config's 'platform_data_dir' variable (in the 'riak_core' app section)
for figuring out where a platform's data dir really is.  Assuming that
the current working directory is writable is not a good choice for Riak
Linux packages, at the very least.


slfritchie:
Oops, bad Scott, that should be BZ 1173.

slfritchie:
Jon & Jeff reviewed.  Merged into master (no problem) and into 1.0 branch (conflict, merged via patch),
closing.

PR #185, opened 2011-09-09
Az653 vnode fold refactor

kellymclaughlin:
These changes simplify the handling of backend folds in `riak_kv_vnode`. The three separate functions formerly used to do bucket listing, key listing, and index queries have been reduced to a single `list` function and a set of smaller supporting functions. The readability is much improved. 

Also the multi backend has been refactored so that it conforms to the orignal backend API specification for the fold functions in particular. This comes at the price of slightly more complexity in the multi backend, but the tradeoff seems to be worth it to keep the vnode and async worker code simple and clean.


jonmeredith:
Should this be az663?

PR #192, opened 2011-09-09
bz1177: remove bitcask symlinks and improve startup checks for existing data directories

kellymclaughlin:
Remove the symbolic links from the partition name to the actual active data
directory for the bitcask backend. The original intent was to facilitate
drop calls with bitcask and avoid file handle problems, but the links
are no longer used.

Also improve checking for existing data directories to use. This includes
a check for a directory with just the bare partition name which facilitate
rolling upgrades. Each new data directory is the name of the partition with a
timestamp appended. When the bitcask backend is started if no directory exists
with the bare partition name, the directory with the most recent time stamp
appended is opened and used as the current data directory. A log message
listing any unused directories for the partition is also generated.

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -518,7 +518,7 @@ log_unused_partition_dirs(Partition, PartitionDirs) ->
             ok;
         _ ->
             %% Inform the user in case they want to do some cleanup.
-            lager:info("Unused data directories exist for partition ~p: ~p",
+            lager:notice("Unused data directories exist for partition ~p: ~p",
                        [Partition, PartitionDirs])
     end.
 
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -475,8 +475,9 @@ config_value(Key, Config, Default) ->
     end.
 
 %% @private
-get_data_dir(DataRoot, Partition) ->
-    PartitionPath = filename:join([DataRoot, integer_to_list(Partition)]),
+get_data_dir(DataRoot, PartitionI) ->
+    Partition = integer_to_list(PartitionI),
+    PartitionPath = filename:join([DataRoot, Partition]),
     case filelib:is_dir(PartitionPath) of
         true ->
             {ok, Partition};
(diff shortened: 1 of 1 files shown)


src/riak_kv_bitcask_backend.erl
```diff
@@ -476,64 +476,84 @@ config_value(Key, Config, Default) ->
...
-            Error
+get_data_dir(DataRoot, PartitionDir) ->
+    case filelib:is_dir(PartitionDir) of
+        true ->
+            {ok, PartitionDir};
+        false ->
+            %% Check for any existing directories for the partition
+            %% and select the most recent as the active data directory
+            %% if any exist.
+            case file:list_dir(DataRoot) of
```
slfritchie: Kelly, if the sort order returned by filelib:wildcard/1 or /2 is sufficient for finding the
last timestamp'ed dir (or the only untimestamped dir), I recommend using it instead of
doing it manually here.

src/riak_kv_bitcask_backend.erl
```diff
@@ -476,64 +476,84 @@ config_value(Key, Config, Default) ->
...
 
 %% @private
-make_symlink(LinkFile) ->
+log_unused_partition_dirs(Partition, PartitionDirs) ->
+    case PartitionDirs of
+        [] ->
+            ok;
+        _ ->
+            %% Inform the user in case they want to do some cleanup.
+            lager:info("Unused data directories exist for partition ~p: ~p",
```
slfritchie: Since we don't yet have a what-priority-level-to-use-for-what-situation policy,
I'm thinking that it would be good to have these messages stand out from the
noise of other info-level messages in the log.

Borrowing from the syslog man page:

```
     LOG_NOTICE    Conditions that are not error conditions, but should possibly be handled specially.
```

src/riak_kv_bitcask_backend.erl
kellymclaughlin: Great idea. Thanks!

slfritchie:
Hrm, I've found another odd thing.  If app.config's value for bitcask's `data_root`
is an absolute path, things seem to work as I expect.  But if `data_root` is a
relative path, then after doing these steps:
1. make stage
2. rm -rf rel/riak/data/bitcask/*
3. Start and stop Riak
4. Start and stop Riak again
5. Run `find rel/riak/data -name bitcask`

The output that I see is:

```
rel/riak/data/bitcask
rel/riak/data/bitcask/data/bitcask
```

I need to relocate back home, sorry I don't have more concrete suggestion
for code to twiddle.

slfritchie:
Kelly, commit 2fa63e5 is the last fix.  If you like it, please merge to
all applicable branches, thanks!

PR #193, opened 2011-09-10
AZ668 Improve performance of Pipe-powered MR with bucket±filter inputs

beerriot:
The initial naive method of getting the keys for a whole-bucket (or filtered-bucket) MapReduce into a pipe (by having all keys delivered to one process that enqueued them as they arrived) was found to be a major bottleneck.  This patch removes that central process, and instead uses the new `riak_pipe_qcover_fsm`  behavior (https://github.com/basho/riak_pipe/pull/28) to list and enqueue each KV vnode's keys separately, and in parallel.

Performance improves, in general, at all bucket cardinalities, but the biggest benefit is in buckets with large numbers of keys.  The naive method would cause the central keylist->pipe shim process to grow a very large heap, to the point where `large_heap`, `long_gc` and `process_memory_high_watermark` warnings were common.  This new distributed method fixes these memory issues, as each worker deals with far fewer keys (only 1/22 of them, in the case of 64 partitions an N=3, for example).

This patch also means that all Pipe MapReduce queries always use the "new" (non-legacy) keylisting method (not `riak_kv_keys_fsm` exactly, but the same coverage computation and `riak_kv_vnode` calls).  I wager this is acceptable, as the legacy keylisting method and pipe are grossly mismatched (due to the centralized key result processing).

When testing this change, it is recommended to use large buckets (5 million keys showed useful results on the den-test-\* machines; more may be required on beefier hardware) and long timeouts (minutes).  No special tweaks are necessary: just ensure that `{mapred_system, pipe}` is set, and then use whatever client you like (remember to use `riak_kv_mrc_pipe` and not `riak_client` on the riak console).

Whole-bucket Example:

```
{"inputs":"lk","query":[{"map":{"language":"erlang","module":"riak_kv_mapreduce","function":"map_object_value"}},{"reduce":{"language":"erlang","module":"riak_kv_mapreduce","function":"reduce_sum"}}],"timeout":600000}
^D
```

Filtered-bucket Example:

```
{"inputs":{"bucket":"lk","key_filters":[["starts_with","1"]]},"query":[{"map":{"language":"erlang","module":"riak_kv_mapreduce","function":"map_object_value"}},{"reduce":{"language":"erlang","module":"riak_kv_mapreduce","function":"reduce_sum"}}],"timeout":600000}
^D
```


src/riak_kv_pipe_listkeys.erl
```diff
@@ -0,0 +1,154 @@
...
+
+%% Convenience
+
+-type bucket_or_filter() :: binary() | {binary(), list()}.
+
+%% @doc List the keys from a bucket, and send them as inputs to the
+%%      given pipe.  This starts a new pipe with one fitting
+%%      (`riak_kv_pipe_listkeys'), with its sink pointed at the
+%%      destination pipe.  The `riak_pipe_qcover_fsm' module is used
+%%      to triger keylisting on the appropriate vnodes.  The `eoi'
```
kellymclaughlin, line 116: typo: _triger_

src/riak_kv_mrc_pipe.erl
```diff
@@ -411,6 +409,20 @@ send_key_list(Pipe, Bucket, ReqId) ->
             ok
     end.
 
+build_key_filter(Exprs) ->
```
kellymclaughlin: I hate to duplicate this function from `riak_client`. Is there a suitable place to put this that it could be shared by this module and `riak_client`? My first thought was to stick it in `riak_kv_mapreduce`, but I don't know that it is a good fit there. Or maybe the answer at this time is to wait for a better occasion and make that part of a larger M/R refactor. Thoughts?

src/riak_kv_pipe_listkeys.erl
```diff
@@ -0,0 +1,154 @@
...
+                       item_filter=Filters},
+      {Partition, node()},
+      FilterVNodes,
+      {raw, ReqId, self()},
+      riak_kv_vnode_master),
+    keysend_loop(ReqId, Partition, FittingDetails),
+    {ok, State}.
+
+keysend_loop(ReqId, Partition, FittingDetails) ->
+    receive
```
kellymclaughlin, line 89: As an aside, due to some refactoring of key listing to streamline the code paths for sync and async options, this will need to change in the near future. If [this](https://github.com/basho/riak_kv/pull/194) pull request gets merged first, the simplest path might be to rebase and then make this change; otherwise, just something to be aware of so we can make sure this doesn't get broken and lead to a goose chase. The new code will need to look like this:

<pre><code>
keysend_loop(ReqId, Partition, FittingDetails) ->
    receive
        {ReqId, {Bucket, Keys}} ->
            keysend(Bucket, Keys, Partition, FittingDetails),
            keysend_loop(ReqId, Partition, FittingDetails);
        {ReqId, done} ->
            ok
    end.
</code></pre>

src/riak_kv_mrc_pipe.erl
beerriot: Actually, I always wondered why it wasn't in `riak_kv_mapred_filters`, since that's the where all the names are looked up.  I'll move them both there unless I hear argument.

src/riak_kv_pipe_listkeys.erl
beerriot, line 89: Merge race!  Er … I mean, noted.  I'll keep an eye out.

kellymclaughlin, line 89: I just merged those changes. Do you want to rebase and make that change and I can test it out?

beerriot:
I have rebased and sent a fresh pull request to make the changes for the new keylist interface: https://github.com/basho/riak_kv/pull/202

PR #194, opened 2011-09-12
Az663 vnode fold refactor

kellymclaughlin:
These changes simplify the handling of backend folds in riak_kv_vnode. The three separate functions formerly used to do bucket listing, key listing, and index queries have been reduced to a single list function and a set of smaller supporting functions. The readability is much improved.

Also the multi backend has been refactored so that it conforms to the orignal backend API specification for the fold functions in particular. This comes at the price of slightly more complexity in the multi backend, but the tradeoff seems to be worth it to keep the vnode and async worker code simple and clean.


jonmeredith:
Hey Kelly, here's my notes after our call

Major issues:
- Abuse of Sender argument for worker pool stuff.
- Async option on fold call, would remove duplication.
- Multi backend split into sync/async workers seems unecessary if you do async option.  Why not pass Acc in as an argument to the async func instead, rather than building the function and running it instantly?
- Acc not treated as opaque in memory backend.

For large numbers of buckets - for 10,000 entries, here's the difference between ordsets and sets (in usecs)

```
(dev1@127.0.0.1)12> OSF = fun() -> ordsets:is_element(5000, OS) end.
#Fun<erl_eval.20.21881191>
(dev1@127.0.0.1)13> OSF().
true
(dev1@127.0.0.1)14> SF = fun() -> sets:is_element(5000, S) end.     
#Fun<erl_eval.20.21881191>
(dev1@127.0.0.1)15> timefunc:func(OSF).
103.474
(dev1@127.0.0.1)16> timefunc:func(SF). 
3.5408
```

kellymclaughlin:
Ready for more review.

jonmeredith:
+1 merge.

PR #195, opened 2011-09-12
AZ680 increased documentation of pipe MR emulation

beerriot:
I've added many docs and type/specs to riak_kv_mrc_pipe.erl, riak_kv_pipe_get.erl, riak_kv_mrc_map.erl, and riak_kv_w_reduce.erl, trying to capture more of why some things are done the way they're done.

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -20,22 +20,27 @@
 
 %% @doc Riak KV MapReduce / Riak Pipe Compatibility
 %%
-%% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job
+%% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job ==
 %%
-%% An uncommonly-used option for Riak KV MapReduce is the option to
-%% use a fourth method of specifying inputs to the beginning of the
-%% MapReduce workflow.  All four methods are:
+%% The four methods of specifying input for a MapReduce job are:
 %%
 %% <ol>
 %% <li> Specify a bucket name (to emit all bucket/key pairs for that
-%%  bucket) </li>
-%% <li> Specify a bucket name and keyfilter spec, `{Bucket, KeyFilter}' </li>
+%%      bucket) </li>
+%% <li> Specify a bucket name and keyfilter spec, `{Bucket,
+%%      KeyFilter}' </li>
 %% <li> Specify an explicit list of bucket/key pairs </li>
-%% <li> Specify `{modfun, Mod, Fun, Arg}' to generate the raw input data
-%% for the rest of the workflow </li>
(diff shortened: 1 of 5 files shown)


src/riak_kv_mrc_pipe.erl
```diff
@@ -20,22 +20,27 @@
 
 %% @doc Riak KV MapReduce / Riak Pipe Compatibility
 %%
-%% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job
+%% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job ==
 %%
-%% An uncommonly-used option for Riak KV MapReduce is the option to
-%% use a fourth method of specifying inputs to the beginning of the
-%% MapReduce workflow.  All four methods are:
+%% The four methods of specifying input for a MapReduce job are:
```
massung, line 25: Looks like there are six (6) methods now?

src/riak_kv_mrc_pipe.erl
```diff
@@ -297,30 +462,57 @@ link_xform_compat(Bucket, Tag, _Arg) ->
...
+%% the pipeline shutsdown before sending inputs finishes, the process
+%% will be torn down automatically.  This also means that an error
+%% sending inputs will automatically tear down the pipe (because the
+%% process will exit abnormally).
+%%
+%% It's a good idea to prefer sending inputs and receiving outputs in
+%% different processes, especially if you're both sending a large
+%% number of inputs (a large bucket list, for instance) and expecting
+%% to receive a large number of outputs.  The mailbox for a process
+%% doing both is likely to be a point of contention, otherwise.
```
massung, line 512: Just curious, is it possible to deadlock with them in the same process?

beerriot, line 512: Not generally, no, because the output sent to the sink is non-blocking.  That's the source of the contention: that mailbox may grow large with pending outputs while the process is waiting on an input confirmation.

src/riak_kv_w_reduce.erl
```diff
@@ -18,9 +18,66 @@
...
+%% ```
+%% X = Fun([B,A], Arg),
+%% Y = Fun([C,X], Arg)
+%% '''
+%% The maximum batch size allowed is controlled by the riak_kv
+%% application environment variable `mapred_reduce_phase_batch_size'
+%%
+%% The other knob to control batching behavior is known as
+%% `reduce_phase_only_1'.  If this option is set in the `Arg'
+%% proplist, the reduce function will be evaluated at most once.  That
```
massung, line 63: Can this only be set via Erlang code or by a user using HTTP/JavaScript in the reduce function? I only ask because this reads like it merely needs to exist at all, yet where it's checked it assumes it's either undefined or 'true'. Is it possible for someone to send {reduce_phase_only_1,false}? Sorry, this comment isn't really about the documentation so much as wondering if there's an unhandled match possible.

massung:
+1 to merge

PR #196, opened 2011-09-13
Make the /ping endpoint check the status of the local node

chids:
The /ping endpoint doesn't really do us much good as a health check in our load balancer since it statically replies "OK" (unless riak or webmachine blows up of course). I suggest the /ping endpoint be rewamped to consider the status of the local node and only reply with a 200 OK when the node is in the "valid" state.

### DIFF
--- a/src/riak_kv_wm_ping.erl
+++ b/src/riak_kv_wm_ping.erl
@@ -28,13 +28,32 @@
 %% webmachine resource exports
 -export([
          init/1,
+         service_available/2,
          to_html/2
         ]).
 
 -include_lib("webmachine/include/webmachine.hrl").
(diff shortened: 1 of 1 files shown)


chids:
Any thoughts on this, is it a good idea in general and does the implementation make sense?

kellymclaughlin:
I like this idea, but I am concerned that `ping` is not the correct endpoint for this. I also worry about deviating from the semantics of the `riak-admin` tool's `ping` command which is also just a check if the riak node is available. I think the right solution is to create a new endpoint that exposes the information about the node status. For now I will close this pull request and I will make a ticket for this on our internal tracking system and we will address it in a future release. Thanks for the idea!

PR #197, opened 2011-09-13
BZ1186 Default reduce batch size to 20.

beerriot:
This way the number of reduce evaulations is comparable to the number
that would have been performed by the legacy system.  Still not exactly
equal, due to the legacy system's phase-output batching, but closer.


massung:
+1 to merge

PR #199, opened 2011-09-13
Get FSM defaults

jonmeredith:
Update default bprops for basic_quorum=false, notfound_ok=true and removes duplicate pr entry.


kellymclaughlin:
+1 to merge.

PR #201, opened 2011-09-13
Az688 phase only 1 fix

massung:
Don't explicitly test for undefined since it can be passed into an erlang reduce function with false or any other term. Not true == undefined.


beerriot:
+1 merge, thanks

PR #204, opened 2011-09-14
Bz1191 preserve reduce order

seancribbs:
Addresses http://issues.basho.com/1191. Before the patch, a second reduce would appear to have its inputs reversed from the order of the outputs of the first reduce. (redo, based on 1.0 branch)

### DIFF
--- a/src/riak_kv_w_reduce.erl
+++ b/src/riak_kv_w_reduce.erl
@@ -96,6 +96,7 @@
 -include("riak_kv_js_pools.hrl").
 
 -record(state, {acc :: list(),
+                inacc :: list(),
                 delay :: integer(),
                 delay_max :: integer(),
                 p :: riak_pipe_vnode:partition(),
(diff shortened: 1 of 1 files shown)


src/riak_kv_w_reduce.erl
```diff
@@ -164,7 +165,7 @@ handoff(HandoffAcc, #state{acc=Acc}=State) ->
 
 -spec handoff_acc([term()], [term()], state()) -> [term()].
 handoff_acc(HandoffAcc, LocalAcc, State) ->
-    InAcc = HandoffAcc++LocalAcc,
+    InAcc = LocalAcc++HandoffAcc,
```
beerriot, line 168: This does mean that order isn't guaranteed to be preserved during handoff, but since the order that inputs arrive at two different vnodes is undefined anyway, handling it in this simple manner is probably worth getting the other behavior sorted.

seancribbs, line 168: Yeah, actually ordering can't be preserved in any sensible way during handoff. This line should probably be reverted.

PR #206, opened 2011-09-14
BZ1189 HTTP/JS-safe reduce batch controls

beerriot:
This patch allows the new reduce batch size controls to be specified as JSON (or mochijson2) structures, to allow them to be controlled over HTTP, as well as safely used with Javascript-implemented phases.

To test, set `{allow_strfun, true}` in the `riak_kv` section of `app.config`, load 1000 objects into bucket `lk', then:

```
{"inputs":"lk","query":[{"reduce":{"language":"erlang","source":"fun(_,_) -> io:format(\"R\"), [] end.","arg":{"reduce_phase_batch_size":100}}}]}
```

This should print 20 `R`s on the Riak console.

```
{"inputs":"lk","query":[{"reduce":{"language":"erlang","source":"fun(_,_) -> io:format(\"R\"), [] end.","arg":{"reduce_phase_only_1":true}}}]}
```

This should print only 1 `R` on the Riak console.


dreverri:
Is there a concern about overloading the "arg" term?
If a user has an existing reduce function that depends on the "arg" from the phase definition they may not be able to access this reduce batch functionality while also controlling their custom reduce phase.

What happens when the "arg" is over specified?

``` javascript
"arg":{"reduce_phase_only_1":true, "reduce_phase_batch_size":10}
```

Would it be possible to add a new field to the phase specification to control this behavior, such as "batch_size"?

``` javascript
{"reduce":{"language":"erlang","source":"fun(_,_) -> io:format(\"R\"), [] end.","batch_size":100}}
```

A string value could be used to turn off batching:

``` javascript
{"reduce":{"language":"erlang","source":"fun(_,_) -> io:format(\"R\"), [] end.","batch_size":"infinity"}}
```

beerriot:
@dreverri Yes, we thought it was unfortunate as well:

https://github.com/basho/riak_kv/blob/1592c63eb0cbbb758198b8953288c419832282c7/src/riak_kv_mrc_pipe.erl#L93

But, the "arg" is the only place to fit it into the current specification, at the Erlang level (`{Phase, Fun, Arg, Keep}`), so overloading it here is just exposing that difficulty.  I like your proposal, and I think we should consider it for the next version.

When "arg" is overspecified, `reduce_phase_only_1` takes precedence:

https://github.com/basho/riak_kv/blob/1592c63eb0cbbb758198b8953288c419832282c7/src/riak_kv_w_reduce.erl#L280

beerriot:
I've resubmitted this request as https://github.com/basho/riak_kv/pull/207 with a better pull target.

PR #207, opened 2011-09-15
BZ1189 HTTP/JS-safe reduce batch controls (redo)

beerriot:
_This PR is a replacement for https://github.com/basho/riak_kv/pull/206 targetted for 1.0 to make it easier to see changes_

The first patch allows the new reduce batch size controls to be specified as JSON (or mochijson2) structures, to allow them to be controlled over HTTP, as well as safely used with Javascript-implemented phases.

To test, set `{allow_strfun, true}` in the `riak_kv` section of `app.config`, load 1000 objects into bucket `lk', then:

```
{"inputs":"lk","query":[{"reduce":{"language":"erlang","source":"fun(_,_) -> io:format(\"R\"), [] end.","arg":{"reduce_phase_batch_size":100}}}]}
```

This should print 20 `R`s on the Riak console.

```
{"inputs":"lk","query":[{"reduce":{"language":"erlang","source":"fun(_,_) -> io:format(\"R\"), [] end.","arg":{"reduce_phase_only_1":true}}}]}
```

This should print only 1 `R` on the Riak console.

The second patch prevents accidental JS failures caused by using the Erlang reduce batch size knob format, instead of the JSON format.  The atoms or tuples are simply filtered from the `Arg' list prior to evaluating the function.  This can be tested from the Riak console via:

```
 riak_kv_mrc_pipe:mapred(<<"lk">>, [{map, {modfun, riak_kv_mapreduce, map_object_value}, none, true},{reduce, {jsanon, <<"function() { ejsLog('/tmp/reduce_batch_test.log', '1'); return []; }">>}, [{reduce_phase_batch_size, 100}], true}]).
```

This should print ten `1`s to the file `/tmp/reduce_batch_test.log`.

```
 riak_kv_mrc_pipe:mapred(<<"lk">>, [{map, {modfun, riak_kv_mapreduce, map_object_value}, none, true},{reduce, {jsanon, <<"function() { ejsLog('/tmp/reduce_batch_test.log', '2'); return []; }">>}, [reduce_phase_only_1}], true}]).
```

This should print one `2` to the file `/tmp/reduce_batch_test.log`.

### DIFF
--- a/src/riak_kv_w_reduce.erl
+++ b/src/riak_kv_w_reduce.erl
@@ -322,16 +322,14 @@ calc_delay_max(#fitting_details{arg = {rct, _ReduceFun, ReduceArg}}) ->
 %% atoms, since the HTTP interface has no way to send atoms natively
 -spec delay_props_from_json(list()) -> [{atom(), term()}].
 delay_props_from_json(JsonProps) ->
-    Only1 = case lists:keyfind(<<"reduce_phase_only_1">>, 1, JsonProps) of
-                {_, Only1V} ->
-                    [{reduce_phase_only_1, Only1V}];
-                false ->
-                    []
-            end,
-    Batch = case lists:keyfind(<<"reduce_phase_batch_size">>, 1, JsonProps) of
-                {_, BatchV} ->
-                    [{reduce_phase_batch_size, BatchV}];
-                false ->
-                    []
-            end,
+    Only1 = extract_json_prop(reduce_phase_only_1, JsonProps),
+    Batch = extract_json_prop(reduce_phase_batch_size, JsonProps),
     Only1 ++ Batch.
+
+extract_json_prop (Key, JsonProps) ->
+    case lists:keyfind(atom_to_binary(Key, latin1), 1, JsonProps) of
+        {_,Value} ->
+            [{Key,Value}];
+        false ->
+            []
+    end.
--- a/src/riak_kv_w_reduce.erl
+++ b/src/riak_kv_w_reduce.erl
@@ -267,9 +267,10 @@ stored_source(Bucket, Key) ->
          fun( (list(), term()) -> list() ).
 js_runner(JS) ->
     fun(Inputs, Arg) ->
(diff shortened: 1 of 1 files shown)


src/riak_kv_w_reduce.erl
```diff
@@ -285,3 +317,21 @@ calc_delay_max(#fitting_details{arg = {rct, _ReduceFun, ReduceArg}}) ->
...
+                false ->
+                    []
+            end,
+    Batch = case lists:keyfind(<<"reduce_phase_batch_size">>, 1, JsonProps) of
+                {_, BatchV} ->
+                    [{reduce_phase_batch_size, BatchV}];
+                false ->
+                    []
+            end,
+    Only1 ++ Batch.
```
massung, line 327: Similar to before, but given that "writing something twice is once too many times..." How about:

``` erlang
delay_props_from_json (JsonProps) ->
    Only1 = extract_json_prop(reduce_phase_only_1,JsonProps),
    Batch = extract_json_prop(reduce_phase_batch_size,JsonProps),
    Only1 ++ Batch.

extract_json_prop (Key, JsonProps) ->
    case lists:keyfind(list_to_binary(Key),1,JsonProps) of
        {_,Value} ->
            [{Key,Value}];
        false ->
            []
    end.
```

massung, line 327: The first test to print 20 R's to the console will only print 15 for me. I loaded the 'lk' bucket with the google data in the fast track (1,438 objects) if that would make a difference, but if it did I would expect it to be more R's - not less. Other than that difference, all the test pass just fine and I was able to run other queries fine as well.

beerriot, line 327: I believe you missed the " load 1000 objects into bucket `lk'" instruction in the testing example.  And, I made a mistake: that should have said "will print 10`R`s", not 20 ( 1000 inputs / 100 per batch = 10 runs).  Your result of 15 is correct for the 1438 objects you used.

massung:
+1 to merge

PR #210, opened 2011-09-15
AZ727 - Detect Siblings in 2I Hook

rustyio:
Make the 2i pre-commit hook detect when the user attempts to write an object with siblings, and fail. THIS WILL NOT HAPPEN UNDER NORMAL USAGE. The HTTP and PB interfaces require a user to resolve any conflicts. This will only happen if the user tries to get clever with riak_object. 

To test this out:

``` erlang
{ok, M} = riak:client_connect('riak@127.0.0.1').
O1 = riak_object:new(<<"testbucket">>, <<"key">>, <<"value">>).
O2 = riak_object:new(<<"testbucket">>, <<"key">>, <<"value2">>).
O3 = riak_object:reconcile([O1, O2], true).
M:put(O3).
```

This is related to http://issues.basho.com/1196.


massung:
+1 to merge

PR #211, opened 2011-09-15
Fix repeated bitcask merge.

kellymclaughlin:
`riak_kv_bitcask_backend` is not passing the correct path to the bitcask
merge worker and as a result the hint and data files are created in
the wrong place. Bitcask then repeatedly tries to merge even though
from the end user perspective it should not need to.  

This change corrects the problem by passing the correct path to merge worker.

See the bugzilla [here](https://issues.basho.com/show_bug.cgi?id=1200) for steps to reproduce the problem and verify that the fix corrects the problem.

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -338,9 +338,11 @@ callback(Ref,
 callback(Ref,
          merge_check,
          #state{ref=Ref,
-                root=BitcaskRoot}=State) when is_reference(Ref) ->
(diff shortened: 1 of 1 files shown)


beerriot:
Looks fixed to me.  +1 merge.

It might be nice to clean up terminology in that file post-1.0.  Sometimes `data_dir` is captured as `DataDir`, and other times as `DataFile`.  Line 288 ignores that record field completely and explicitly calls `integer_to_list(Partition)`.  I'll file a card.

PR #212, opened 2011-09-16
Better error messages from Pipe-based MapReduce

beerriot:
Before this commit, when Pipe-based MapReduce failed to validate a phase function specification (for example, when a module or function was undefined), the client would receive a very generic error, possibly including a mostly useless stacktrace.  With this change, and https://github.com/basho/riak_pipe/pull/29, a much more descriptive error message is returned.

``` erlang
> riakc_pb_socket:mapred(S, <<"foo">>, [{map, {modfun, doesnot, exist}, none, true}]).                      
{error,<<"Phase 0: invalid module named in PhaseSpec function:\n must be a valid module name (failed to load doesnot: n"...>>}
```

``` bash
$ curl -i -X POST -H "content-type:application/json" http://localhost:8098/mapred?chunked=true --data @-
{"inputs":"foo","query":[{"map":{"language":"erlang","module":"does_not","function":"exist"}}]}
^D
HTTP/1.1 400 Bad Request

{"phase":0,"error":"invalid module named in PhaseSpec function:\n must be a valid module name (failed to load does_not: nofile)"}
```

### DIFF
--- a/src/riak_kv_pb_socket.erl
+++ b/src/riak_kv_pb_socket.erl
@@ -462,18 +462,22 @@ process_message(#rpbmapredreq{request=MrReq, content_type=ContentType}=Req,
     end.
 
 pipe_mapreduce(Req, State, Inputs, Query, Timeout) ->
-    {{ok, Pipe}, _NumKeeps} =
-        riak_kv_mrc_pipe:mapred_stream(Query),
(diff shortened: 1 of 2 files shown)


massung:
+1 to merge

PR #215, opened 2011-09-16
Expose vnode status via riak_kv_console.

kellymclaughlin:
This change adds a `vnode_status` function to `riak_kv_console`. It also adds code to `riak_kv_vnode` to handle a request for vnode status information. Finally, a `vnode_status` function is added to `riak_kv_status` to determine the local vnodes, send status requests to each one, and accumulate the results. 

Currently the only piece of vnode status information that is available is the backend status.

### DIFF
--- a/src/riak_kv_status.erl
+++ b/src/riak_kv_status.erl
@@ -23,7 +23,10 @@
 
 -export([statistics/0,
          ringready/0,
-         transfers/0]).
+         transfers/0,
+         vnode_status/0]).
+
+-include("riak_kv_vnode.hrl").
 
 %% ===================================================================
 %% Public API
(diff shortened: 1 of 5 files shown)


src/riak_kv_status.erl
```diff
@@ -43,3 +46,34 @@ ringready() ->
...
 transfers() ->
     riak_core_status:transfers().
+
+%% @doc Get status information about the node local vnodes.
+-spec vnode_status() -> [{atom(), term()}].
+vnode_status() ->
+    %% Get the kv vnode indexes and the associated pids for the node.
+    IndexPids = riak_core_vnode_master:all_index_pid(riak_kv_vnode),
+    %% Get the status of each vnode
+    [riak_core_vnode_master:command({Index, Pid},
```
jonmeredith: This would probably be better as a function in riak_kv_vnode to keep the message formats in one place, something like
riak_kv_vnode:vnode_status(PrefList) - and IndexPids is a match for an [{Idx,Pid}] preflist.

jonmeredith:
+1 merge after writing a wrapper around the vnode command in riak_kv_vnode.

PR #216, opened 2011-09-16
Fix eleveldb bucket listing with 2I

kellymclaughlin:
The exception handling to catch when eleveldb exits a fold early was omitted from the fold_buckets function, but with the addition of the secondary index capability to the backend, bucket listing can break out of fold_buckets early. This change adds that exception handling to the fold_buckets function.

See the bugzilla ticket [here](https://issues.basho.com/show_bug.cgi?id=1202) for steps to reproduce the problem and verify the fix.


massung:
+1 to merge

PR #217, opened 2011-09-22
Add PBC connects stat under new stats

jonmeredith:
Add the pbc_connects stat under the new stats system.

It's absence was breaking basho_jmx.


rzezeski:
+1 to merge

jonmeredith:
Rebased, then merged to 1.0 and master.

PR #219, opened 2011-09-22
Fix fold function arity problem with handoff and replication.

kellymclaughlin:
The function used for folding objects during handoff or replication needs to be wrapped since the `riak_kv` backends expect a fold function of arity 4, but `riak_core` presents a function of arity 3. This wrapping was formerly handled in the `handle_handoff_command` and `fold` functions, but this can cause problems when handoff and fullsync replication are occurring simultaneously. This change moves that function wrapping the `handle_command` clause that handles object fold requests in the case of both handoff and replication and alleviates the problems.

The best way to reproduce this is to setup a multi-node primary cluster and a replication cluster of 1 or more nodes. First add the replication listener to the primary cluster. Then write some data to the primary cluster. Next have a node leave the primary cluster so that it begins to handoff its data and while it is handing off add the secondary site so that the replication synchronization will begin.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -175,16 +175,9 @@ list_keys(Preflist, ReqId, Caller, Bucket) ->
                                    riak_kv_vnode_master).
 
 fold(Preflist, Fun, Acc0) ->
-    %% The function used for object folding expects the
-    %% bucket and key pair to be passed as the first parameter, but in
-    %% riak_kv the bucket and key have been separated. This function
-    %% wrapper is to address this mismatch.
-    FoldFun = fun(Bucket, Key, Value, Acc) ->
-                      Fun({Bucket, Key}, Value, Acc)
-              end,
     riak_core_vnode_master:sync_spawn_command(Preflist,
                                               ?FOLD_REQ{
-                                                 foldfun=FoldFun,
+                                                 foldfun=Fun,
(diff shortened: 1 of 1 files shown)


src/riak_kv_vnode.erl
```diff
@@ -327,8 +320,16 @@ handle_command(?KV_DELETE_REQ{bkey=BKey, req_id=ReqId}, _Sender, State) ->
...
     {reply, do_get_vclocks(BKeys, State), State};
-handle_command(?FOLD_REQ{foldfun=Fun, acc0=Acc}, Sender, State) ->
-    do_fold(Fun, Acc, Sender, State);
+handle_command(?FOLD_REQ{foldfun=FoldFun, acc0=Acc0}, Sender, State) ->
+    %% The function in riak_core used for object folding
+    %% during handoff expects the bucket and key pair to be
+    %% passed as the first parameter, but in riak_kv the bucket
+    %% and key have been separated. This function wrapper is
+    %% to address this mismatch.
+    HandoffFun = fun(Bucket, Key, Value, Acc) ->
```
Vagabond: HandoffFun seems like a misnomer now.

Vagabond:
+1, pending the variable naming review.

I tested this before and was able to reproduce the error, after applying the patch however, I wasn't able to trigger the error anymore.

PR #221, opened 2011-09-23
AZ754 - Execute 2i PreCommit Hook Last 

rustyio:
Move index precommit hook to the end of all precommit hooks. 

This makes sure that the system correctly parses and validates any index entries that a user has added in a custom Pre-Commit hook.


Vagabond:
+1 to merge

PR #224, opened 2011-09-27
prevent connection reuse after pipe mapred error (BZ1222)

beerriot:
Because of the way pipe sends results, sending the response to the client
before eoi means that the request process may still have messages in its
mailbox.  If mochiweb tries to keep the connection alive, its next loop will
receive one of these messages instead of HTTP/TCP data, which will cause it
to send a 400 to the client.  The interim fix here is to just tear down the
connection, unconditionally, after the initial mapred error is sent, forcing
the client to reconnect for its next request.

I think there may still be one corner case where a similar receive->400 might happen after a successful mapred, but we haven't yet seen that in testing, and that fix will be much more invasive.

### DIFF
--- a/src/riak_kv_wm_mapred.erl
+++ b/src/riak_kv_wm_mapred.erl
@@ -286,6 +286,12 @@ pipe_stream_mapred_results(RD, Pipe,
 %% This is used to workaround an issue in mochiweb, where the loop
 %% waiting for new TCP data receives a latent pipe message instead,
 %% and blows up, sending a 400 to the requester.
+%%
+%% WARNING: This uses an undocumented feature of mochiweb that exists
+%% in 1.5.1 (the version planned to ship with Riak 1.0).  The feature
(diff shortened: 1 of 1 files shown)


src/riak_kv_wm_mapred.erl
```diff
@@ -268,14 +271,25 @@ pipe_stream_mapred_results(RD, Pipe,
...
         {error, {sender_error, Error}} ->
+            prevent_keepalive(),
             {format_error(Error), done};
         {error, {Error, _Input}} ->
             riak_pipe:destroy(Pipe),
+            prevent_keepalive(),
             {format_error({error, Error}), done}
     end.
 
+%% @doc Prevent this socket from being used for another HTTP request.
```
Vagabond, line 285: I'd be nice if you documented that this was using a scary undocumented internal variable to make the magic happen, so future generations can beware.

Vagabond:
Otherwise, seems to be the simplest solution to the problem. +1.

PR #225, opened 2011-09-28
Bz1226 vtag from preinc vclock

jonmeredith:
Generate vtag from incoming vclock+md+value as vclock may not be updated yet.

For 1.0.0 with vnode_vclocks enabled the vclock will not be updated
until it is processed by the first vnode, so we cannot calculate the
vtag from the vclock.  Sibling hashes could be computed on get instead
of on put so we wouldn't even need to store them.
bz://1226 was reported too close to final 1.0.0 to make changes like
that, so instead include the MD/Value in the vtag hash which should
generate unique hashes for an update.

Fixes: bz://1226


reiddraper:
+1

PR #226, opened 2011-09-29
Add delete_mode to control when backend deletion takes effect.

jonmeredith:
With the vnode-based vclock scheme all update/delete cycles against
a single object will use the vnode ids.  If one of the vnodes is
running slow there is a small window where the tombstone has been
removed from some nodes but not others.  If an object is created
on a vnode where the object has been removed then the vclock entry
for that vnode will be less than the tombstone and the write would
be ignored.

This patch adds a delete_mode app env that controls when the tombstone
object is removed from the backend.  If 'keep' then tombstones
are always kept, which is the safest option, but an object will always
consume disk space once created.  'immediate' mirrors the 0.14.2
behavior of removing the tombstone once the vnode delete request
is received.  The middle ground is to set a delay time which
makes the vnode wait that long and only remove the object if it
has not been modified in that time.  Setting to 1000ms was sufficient
to pass the bz://260 delete_bug script.

If a delay is used, the vnode uses erlang:send_after to trigger the
deletion.  For a system that is under very high delete load
there could be many timers created, however they should be fairly
cheap.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -52,7 +52,8 @@
          handoff_finished/2,
          handle_handoff_data/2,
          encode_handoff_item/2,
-         handle_exit/3]).
+         handle_exit/3,
+         handle_info/2]).
 
 -include_lib("riak_kv_vnode.hrl").
 -include_lib("riak_kv_map_phase.hrl").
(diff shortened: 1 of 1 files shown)


src/riak_kv_vnode.erl
```diff
@@ -849,21 +890,20 @@ do_delete(BKey, ReqId, State) ->
...
-
-                    %% Calculate the index specs to remove...
-                    IndexSpecs = riak_object:diff_index_specs(undefined, RObj),
-
-                    %% Do the delete...
-                    {Bucket, Key} = BKey,
-                    case Mod:delete(Bucket, Key, IndexSpecs, ModState) of
-                        {ok, UpdModState} ->
-                            UpdState = State#state {modstate = UpdModState },
+                    case DeleteMode of
```
Vagabond, line 893: Just an observation, but if the option gets typoed in the config file, you'll get a case_clause error here.

Vagabond:
I like this patch, and it solves the problem. +1 to merge it.

jonmeredith:
I'll take any ideas I can get on deletion.  Synchronizing tombstone removal
is the hard part :)

Jon.

jonmeredith:
Too my pre-releasey fun going on here - you wouldn't have my full attention.

Say hi to the airport for me.

Jon

PR #227, opened 2011-09-29
add 2i capability to multi_backend

krestenkrab:
Here is a basic patch to make multi backend work with secondary indexing.  

I have only tested with this simple positive case:

```
        {storage_backend, riak_kv_multi_backend},                                                                                                                                                     
        {multi_backend_default, <<"leveldb">>},                                                                                                                                                       
        {multi_backend, [                                                                                                                                                                             
           {<<"leveldb">>, riak_kv_eleveldb_backend, [                                                                                                                                                
             {data_root, "data/leveldb"}                                                                                                                                                              
           ]},                                                                                                                                                                                        
           {<<"bitcask">>, riak_kv_bitcask_backend, [                                                                                                                                                 
             {data_root, "data/bitcask"}                                                                                                                                                              
           ]}                                                                                                                                                                                         
        ]},       

krab$ curl -v -X PUT -d 'payload' -H 'X-Riak-Index-num_int: 2' http://127.0.0.1:8091/riak/test/doc2
krab$ curl -v http://127.0.0.1:8091/buckets/test/index/num_int/1


krab$ curl -v http://127.0.0.1:809/buckets/test/index/num_int/2
{"keys":["doc2"]}
```

Obviously, this would need to go through your rigorous testing procedures; ... consider it a suggestion.

### DIFF
--- a/src/riak_kv_multi_backend.erl
+++ b/src/riak_kv_multi_backend.erl
@@ -230,13 +230,24 @@ delete(Bucket, Key, IndexSpecs, State) ->
 fold_buckets(FoldBucketsFun, Acc, Opts, State) ->
     fold(undefined, fold_buckets, FoldBucketsFun, Acc, Opts, State).
 
+%% @doc Given `FoldOptions', determine bucket (if any) that it is constrained to.
+get_fold_bucket([]) ->
+    undefined;
+get_fold_bucket([{bucket, Bucket}|_]) ->
+    Bucket;
+get_fold_bucket([{index, Bucket, _}|_]) ->
+    Bucket;
+get_fold_bucket([_|T]) ->
+    get_fold_bucket(T).
+
+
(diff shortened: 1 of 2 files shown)


PSUdaemon:
I think this would be a valuable addition.

krestenkrab:
We will need something like this working in order to be able to move forward on our next project.  
What should I do to push on this?

rustyio:
Hi Kresten,

We took a different approach to solving this problem than the one you suggested. Please see http://github.com/basho/riak_kv/pull/258/. The key difference is that we wanted to fix the capabilities system to prepare us for other capabilities that differ by bucket.

The end result is that multi_backend **will** support 2i in the next major release, assuming that my code makes it past my colleagues' review successfully. :)

Thanks again for your patch, your persistence, and your patience.

Best,
Rusty

PR #228, opened 2011-09-30
Non-legacy stats use update1 instead of update, which did not increment p

massung:
Non-legacy stats use update1 instead of update, which did not increment pbc_connects_total.


russelldb:
Superceded by folsom stats in riak 1.2.

PR #231, opened 2011-10-05
Pass [asis] option to put FSM if riak_client has clientid.

jonmeredith:
In a mixed cluster, if a 1.0.0 node is coordinating a put that it
is not in the preference list for and decides to forward to a pre-1.0.0
node it will crash as riak_kv_put_fsm_sup does not exist on the remote
node.

With the asis option the put FSM will not attempt to forward and
write to the vnodes in the preference list directly.

Fixes: bz://1234

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -292,15 +292,15 @@ put(RObj) -> THIS:put(RObj, []).
 %%       {error, Err :: term(), details()}
 %% @doc Store RObj in the cluster.
 put(RObj, Options) when is_list(Options) ->
-    UpdObj = case ClientId of
-                 undefined ->
-                     RObj;
-                 _ ->
-                     riak_object:increment_vclock(RObj, ClientId)
-             end,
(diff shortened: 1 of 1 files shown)


jonmeredith:
Check by spinning up a 4-node 0.14.2 cluster, upgrade one node to 1.0.0 and start with vnode_vclocks disabled then make sure you can put a an object from the 1.0.0 node that resides on the 0.14.2 nodes.  

You can use redbug:start("riak_core_apl:get_apl_ann -> return") to check the preflist being written to is on the 0.14.2 nodes.
I'd just do something like [C:put(riak_object:new(<<"b">>,<<K:32>>,<<"v">>)) || K <- lists:seq(1,10)].

massung:
Verified the node crash w/o the patch. Recompiled the beam with the patch and the puts return [ok,ok,...,ok]. +1 to merge.

jtuple:
Verified working as well. Going to merge over.

PR #233, opened 2011-10-05
Fix bitcask fd leak

kellymclaughlin:
Make sure the asynchronous bucket, key, and object folding functions properly close the bitcask reference they open after completing the fold operation to ensure prompt cleanup of the associated system resources.

The bugzilla ticket is [here](https://issues.basho.com/show_bug.cgi?id=1237) and describes a method to reproduce the error as well as a command line script Scott created to spit out file descriptor usage.


slfritchie:
As I mentioned in https://issues.basho.com/show_bug.cgi?id=1237#c2 the worst
is indeed fixed.  Thanks in advance for opening a second ticket for the deleted
descriptor leak.  +1 to merge.

PR #234, opened 2011-10-06
Fix keys_fsm_eqc test after recent delete changes.

kellymclaughlin:
The `keys_fsm_eqc` test was broken by commit 0f7f23a43063aff436ddb363375fa1be444f1bf4 which was the delete on timer change. This change fixes the test by setting `delete_mode` to `immediate`. It also specifies `legacy_keylisting` as false to make the test more efficient.


slfritchie:
Yup, the QC test works now.  +1 to merge, thanks!

PR #236, opened 2011-10-10
Always use legacy stats until new stats are fixed.

massung:


src/riak_kv_stat.erl
```diff
@@ -221,9 +221,12 @@ init([]) ->
         {ok, true} ->
```
jonmeredith: I think it would be better to write the case statement so you only get the error if you've specifically disabled legacy stats, and the legacy_init can be refactored outside the warning.

``` erlang
    case application:get_env(riak_kv, legacy_stats) of
        {ok, false} ->
            lager:warning("Overriding user-setting and using legacy stats.");
        _ ->
            ok
    end,
    legacy_init().
```

You could also add to the message "Set riak_kv {legacy_stats, true} to remove this message." to give the user a hand to clean up.

jonmeredith:
+1 merge

PR #238, opened 2011-10-12
BZ1243: Normalize riak_object JSON form.

seancribbs:
- Verifies that round-trips to/from JSON do not diverge.
- Handles special-case for indexes: they are allowed to have multiple values.

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -497,9 +497,9 @@ dejsonify_meta_value({struct, PList}) ->
     lists:foldl(fun({Key, List}, Acc) when is_list(List) ->
                         %% This reverses the {k,v},{k,v2} pattern that
                         %% is possible in multi-valued indexes.
-                        [{Key, L} || L <- List] ++ Acc;
+                        [{Key, dejsonify_meta_value(L)} || L <- List] ++ Acc;
                     ({Key, V}, Acc) ->
-                        [{Key,V}|Acc]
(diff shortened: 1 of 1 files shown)


src/riak_object.erl
```diff
@@ -444,17 +444,32 @@ jsonify_metadata_list(List) ->
...
+                                                  true -> Key
+                                               end,
+                                     JSONVal = if is_list(Value) -> jsonify_metadata_list(Value);
+                                                  true -> Value
+                                               end,
+                                     case dict:find(JSONKey, Dict) of
+                                         {ok, ListVal} when is_list(ListVal) ->
+                                             dict:append(JSONKey, JSONVal, Dict);
+                                         {ok, Other} ->
+                                             dict:store(JSONKey, [Other,JSONVal], Dict);
```
rustyio, line 467: This line looks odd, specifically `[Other,JSONVal]`. `Other` is going to be a binary, and `JSONVal` is going to be either a binary, or a list. Do we always want to create a new list out of this? It will work for the first two values, but after that I think this might create nested lists.

Edit: N/M... just realized what line 465 is doing.

seancribbs, line 467: The previous clause handles the case where the key is already a list, whereas this clause upgrades the previous value into a list. Does that make sense?

rustyio, line 467: Yep, dug into the code a little further after writing that comment and realized what was going on. Make sense.

rustyio:
+1 merge

PR #239, opened 2011-10-14
Fix backup/restore bug caused by misuse of fixed-up ring

jtuple:
Change riak_kv_backup from using get_my_ring to get_raw_ring so that the rings passed into the equal_rings check don't needlessly fail due to local (fixed-up) ring modifications. Without this change, backup/restore fail on clusters that engage the ring fix-up logic.

Fixes BZ1249


slfritchie:
Yup, the bad behavior goes away.  +1 to merge, thanks!

PR #240, opened 2011-10-14
Fix bitcask backend to delete pre-1.0 data directories

jtuple:
Upgrading 0.14.2 cluster to 1.0 cluster would result in two naming conventions for bitcask data directories: <index> and <index>-<timestamp> for 0.14.2 and 1.0 respectively. The bitcask backend would not clean-up the older <index> named directories, and therefore data would remain around after handoff. This pull-request fixes that.

Fixes bz1250


ian-plosker:
I've tested this patch on a cluster that experienced BZ1250. It appears to work, although I did have to prod the nodes to handoff with the following:

```
[ riak_core_vnode_master:get_vnode_pid(P, riak_kv_vnode) || {P, _} <- riak_core_ring:all_owners(Ring) ], 
riak_core_vnode_manager:force_handoffs().
```

kellymclaughlin:
+1 to merge.

PR #243, opened 2011-10-27
Add siblings and objsize stats (AZ895)

rustyio:
## Overview

This code adds Riak stats for `node_get_fsm_siblings` and
`node_get_fsm_objsize`, and supports hot upgrades (no need to restart
the node.)

**IMPORTANT**: Relies on https://github.com/basho/riak_core/pull/104
## Hot Upgrade

On a running Riak node:
1. Back up the following files:
   - `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_stat.beam`
   - `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_get_fsm.beam`
   - `/usr/lib/riak/lib/riak_core-1.0.1/ebin/slide.beam`
2. Replace old beam files with new beam files:
   - Copy `riak_kv_stat.beam` to `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_stat.beam`
   - Copy `riak_kv_get_fsm.beam` to `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_get_fsm.beam`
   - Copy `slide.beam` to `/usr/lib/riak/lib/riak_core-1.0.1/ebin/slide.beam`
3. Run `riak attach` to connect to the Erlang shell and paste the
   following commands. After running the commands, you will see an
   error message `[error] Supervisor riak_kv_sup had child
   riak_kv_stat... killed in context child_terminated`. This is
   expected and can be ignored.

``` erlang
c:l(slide).
sys:suspend(riak_kv_stat).
c:l(riak_kv_stat).
erlang:exit(erlang:whereis(riak_kv_stat), kill).
sys:resume(riak_kv_stat).
c:l(riak_kv_get_fsm).
```
1. Press Control-D to disconnect from the Erlang shell.
2. To verify installation, run `curl localhost:8098/stats` and look
   for a stat named `node_get_fsm_siblings_total`.
## Hot Downgrade

On a running Riak node:
1. Restore the files you backed up in Step #1 of  **Hot Upgrade**.
2. Run `riak attach` to connect to the Erlang shell and paste the
   following commands. After running the commands, you will see an
   error message `[error] Supervisor riak_kv_sup had child
   riak_kv_stat... killed in context child_terminated`. This is
   expected and can be ignored.

``` erlang
c:l(riak_kv_get_fsm).
sys:suspend(riak_kv_stat).
c:l(riak_kv_stat).
erlang:exit(erlang:whereis(riak_kv_stat), kill).
sys:resume(riak_kv_stat).
c:l(slide).
```
1. Press Control-D to disconnect from the Erlang shell.
2. To verify, run `curl localhost:8098/stats` and ensure that the stat
   named `node_get_fsm_siblings_total` does not exist.
## Rolling Upgrade / Downgrade

On a running Riak node:
1. Back up the following files:
   - `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_stat.beam`
   - `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_get_fsm.beam`
   - `/usr/lib/riak/lib/riak_core-1.0.1/ebin/slide.beam`
2. Replace old beam files with new beam files:
   - Copy `riak_kv_stat.beam` to `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_stat.beam`
   - Copy `riak_kv_get_fsm.beam` to `/usr/lib/riak/lib/riak_kv-1.0.1/ebin/riak_kv_get_fsm.beam`
   - Copy `slide.beam` to `/usr/lib/riak/lib/riak_core-1.0.1/ebin/slide.beam`
3. Restart the Riak node.

To downgrade, restore the previously backed up .beam files.
## Notes
- Updated the `riak_kv_stat` module to support `node_get_fsm_siblings` and
  `node_get_fsm_objsize` stats, tracking number of siblings and object
  size, respectively. Also updated stats module `code_changed/N`
  function to enable hot upgrades. (No need to take down the VM if
  upgrading from Riak 1.0.1.)
- Updated the `riak_kv_get_fsm` module to send stats about # of siblings
  and object size to the stats module.
- Noticed that sometimes 99% stats are larger than 100% stats. Assuming
  that this is due to a small sample size plus the fact that 99% stats
  being estimated while 100% stats are the actual max.
- Calculating object size by summing the bucket size, key size, using
  `term_to_binary/1` on the vector clock, and the metadata and value
  sibling. This isn't as exact of a calculation because it doesn't
  capture some of the overhead (ie: it doesn't count the size of the
  tuple that wraps the metadata and value.) But from what I've seen it's
  within 10% of the object size.
- Tested performance of old and new code via basho_bench. Adding the new
  stats had no discernible affect on performance.
## Test Code

``` erlang
%% Some rough smoke-testing code. Just copy this code and paste into
%% an Erlang shell. Creates siblings of different sizes, prints out
%% the stats, then pauses for 5 seconds.

%% The CreateSiblings/1 function creates creates an object with the
%% specified number of siblings.
CreateSiblings = fun(N) ->
    {ok, Client} = riak:local_client(),
    Client:set_bucket(<<"mybucket">>, [{allow_mult, true}]),
    Key = erlang:md5(term_to_binary(now())),
    Obj = riak_object:new(<<"mybucket">>, Key, <<"1">>),
    [Client:put(Obj) || _X <- lists:seq(1, N)],
    Client:get(<<"mybucket">>, Key),
    ok
end.

%% Print `fsm_siblings` and `fsm_objsize` stats.
PrintStats = fun() ->
    %% Use apply so that we don't hold a reference to old code.
    Stats = erlang:apply(riak_kv_stat, get_stats, []),
    io:format("---~n"),
    io:format("~p~n", [lists:keyfind(node_get_fsm_siblings_mean,   1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_siblings_median, 1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_siblings_95,     1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_siblings_99,     1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_siblings_100,    1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_objsize_mean,    1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_objsize_median,  1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_objsize_95,      1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_objsize_99,      1, Stats)]),
    io:format("~p~n", [lists:keyfind(node_get_fsm_objsize_100,     1, Stats)]),
    ok
end.

CreateSiblings(1).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(10).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(100).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(2).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(20).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(200).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(3).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(30).
PrintStats().
timer:sleep(5 * 1000).

CreateSiblings(300).
PrintStats().
```

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -516,17 +516,9 @@ spiral_minute(_Moment, Elt, State) ->
 %%      element of the state tuple.
 %%      If Count is 0, then all other elements will be the atom
 %%      'undefined'.
-slide_minute(Moment, Elt, State, Min, Max, Bins) ->
-    {Count, Mean, {Median, NF, NN, H}} = slide:mean_and_nines(element(Elt, State), Moment, Min, Max, Bins),
-
-    %% Massage the stats so that we never return a Median, 95th, or
-    %% 99th percentile that's higher than the max.
-    {Count, Mean, {
-              lists:min([Median, H]),
-              lists:min([NF, H]),
-              lists:min([NN, H]),
-              H
-             }}.
+slide_minute(Moment, Elt, State, Min, Max, Bins, RoundingMode) ->
+    {Count, Mean, Nines} = slide:mean_and_nines(element(Elt, State), Moment, Min, Max, Bins, RoundingMode),
+    {Count, Mean, Nines}.
 
 metric_stats({meter, M}) ->
     basho_metrics_nifs:meter_stats(M);
@@ -591,13 +583,13 @@ node_stats(Moment, State=#state{node_gets_total=NGT,
                                 coord_redirs_total=CRT,
                                 legacy=true}) ->
     {Gets, GetMean, {GetMedian, GetNF, GetNN, GetH}} =
-        slide_minute(Moment, #state.get_fsm_time, State, 0, 5000000, 20000),
+        slide_minute(Moment, #state.get_fsm_time, State, 0, 5000000, 20000, down),
     {Puts, PutMean, {PutMedian, PutNF, PutNN, PutH}} =
-        slide_minute(Moment, #state.put_fsm_time, State, 0, 5000000, 20000),
+        slide_minute(Moment, #state.put_fsm_time, State, 0, 5000000, 20000, down),
     {_Siblings, SiblingsMean, {SiblingsMedian, SiblingsNF, SiblingsNN, SiblingsH}} =
-        slide_minute(Moment, #state.node_get_fsm_siblings, State, 0, 1000, 1000),
+        slide_minute(Moment, #state.node_get_fsm_siblings, State, 0, 1000, 1000, up),
     {_ObjSize, ObjSizeMean, {ObjSizeMedian, ObjSizeNF, ObjSizeNN, ObjSizeH}} =
-        slide_minute(Moment, #state.node_get_fsm_objsize, State, 0, 16 * 1024 * 1024, 16 * 1024),
+        slide_minute(Moment, #state.node_get_fsm_objsize, State, 0, 16 * 1024 * 1024, 16 * 1024, down),
     [{node_gets, Gets},
      {node_gets_total, NGT},
      {node_get_fsm_time_mean, GetMean},
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -349,106 +349,7 @@ terminate(_Reason, _State) ->
     ok.
 
 %% @private
-code_change({up, siblings_and_objsize_stats},
-            {state,
-             VNodeGets, VNodePuts,
-             VNodeGetsTotal, VNodePutsTotal,
-             VNodeIndexReads, VNodeIndexReadsTotal,
-             VNodeIndexWrites, VNodeIndexWritesTotal,
-             VNodeIndexWritesPostings, VNodeIndexWritesPostingsTotal,
-             VNodeIndexDeletes, VNodeIndexDeletesTotal,
-             VNodeIndexDeletesPostings, VNodeIndexDeletesPostingsTotal,
-             NodeGetsTotal, NodePutsTotal,
-             GetFSMTime, PutFSMTime,
-             PBCConnects, PBCConnectsTotal, PBCActive,
-             ReadRepairs, ReadRepairsTotal,
-             CoordRedirs, CoordRedirsTotal, MapperCount,
-             GetMeter, PutMeter, Legacy},
-            _Extra) ->
-
-    %% Upgrade from the previous version of riak_kv_stat.
-    io:format("~nAdding siblings and objsize stats to riak_kv_stat process.~n~n"),
-    {ok, #state{
-       vnode_gets=VNodeGets,
-       vnode_gets_total=VNodeGetsTotal,
-       vnode_puts=VNodePuts,
-       vnode_puts_total=VNodePutsTotal,
-       vnode_index_reads=VNodeIndexReads,
-       vnode_index_reads_total=VNodeIndexReadsTotal,
-       vnode_index_writes=VNodeIndexWrites,
-       vnode_index_writes_total=VNodeIndexWritesTotal,
-       vnode_index_writes_postings=VNodeIndexWritesPostings,
-       vnode_index_writes_postings_total=VNodeIndexWritesPostingsTotal,
-       vnode_index_deletes=VNodeIndexDeletes,
-       vnode_index_deletes_total=VNodeIndexDeletesTotal,
-       vnode_index_deletes_postings=VNodeIndexDeletesPostings,
-       vnode_index_deletes_postings_total=VNodeIndexDeletesPostingsTotal,
-       node_gets_total=NodeGetsTotal,
-       node_puts_total=NodePutsTotal,
-       node_get_fsm_siblings=slide:fresh(),
-       node_get_fsm_objsize=slide:fresh(),
-       get_fsm_time=GetFSMTime,
-       put_fsm_time=PutFSMTime,
-       pbc_connects=PBCConnects,
-       pbc_connects_total=PBCConnectsTotal,
-       pbc_active=PBCActive,
-       read_repairs=ReadRepairs,
-       read_repairs_total=ReadRepairsTotal,
-       coord_redirs=CoordRedirs,
-       coord_redirs_total=CoordRedirsTotal,
-       mapper_count=MapperCount,
-       get_meter=GetMeter,
-       put_meter=PutMeter,
-       legacy=Legacy}};
-code_change({down, siblings_and_objsize_stats},
-            State = #state {}, _Extra) ->
-    io:format("~nCODE CHANGE: Removing siblings and objsize stats from riak_kv_stat process.~n~n"),
-    #state {
-            vnode_gets=VNodeGets,
-            vnode_gets_total=VNodeGetsTotal,
-            vnode_puts=VNodePuts,
-            vnode_puts_total=VNodePutsTotal,
-            vnode_index_reads=VNodeIndexReads,
-            vnode_index_reads_total=VNodeIndexReadsTotal,
-            vnode_index_writes=VNodeIndexWrites,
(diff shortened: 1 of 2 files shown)


jonmeredith:
Thanks for dropping everything and getting this done.  Here's my notes of our IM for review

1) We can drop the _total stats.
2) Can you fix the max/banding for the quantile calculations as we're not in microseconds any more.  Bin size of 1, max of 1000 for siblings should be sufficient and bin size of 16 bytes/max size of 16Mb should be sufficient for objsize.
3) The hot code upgrade is very cool, but I overstated the requirements a bit.  We just need to be able to restart the stats process after a code upgrade rather than keep stats.  I just wanted the stats module to handle old/new format get FSM stats so that it wouldn't have issues.  Simplest is best.

Thx!
Jon

src/riak_kv_stat.erl
```diff
@@ -494,9 +516,17 @@ spiral_minute(_Moment, Elt, State) ->
...
 %%      If Count is 0, then all other elements will be the atom
 %%      'undefined'.
-slide_minute(Moment, Elt, State) ->
-    {Count, Mean, Nines} = slide:mean_and_nines(element(Elt, State), Moment),
-    {Count, Mean, Nines}.
+slide_minute(Moment, Elt, State, Min, Max, Bins) ->
+    {Count, Mean, {Median, NF, NN, H}} = slide:mean_and_nines(element(Elt, State), Moment, Min, Max, Bins),
+
+    %% Massage the stats so that we never return a Median, 95th, or
+    %% 99th percentile that's higher than the max.
```
jonmeredith: Should explain in the comment that the higher values are due to estimates from banding in the slide histogram

src/riak_kv_stat.erl
```diff
@@ -358,6 +371,13 @@ update({vnode_index_write, Postings}, Moment, State=#state{vnode_index_writes_to
 update({vnode_index_delete, Postings}, Moment, State=#state{vnode_index_deletes_total=VID, vnode_index_deletes_postings_total=VIDP}) ->
     NewState = spiral_incr(#state.vnode_index_deletes, Moment, State#state{vnode_index_deletes_total=VID+1}),
     spiral_incr(#state.vnode_index_deletes_postings, Postings, Moment, NewState#state{vnode_index_deletes_postings_total=VIDP+Postings});
+update({get_fsm, _Bucket, Microsecs, NumSiblings, ObjSize}, Moment, State) ->
+    NGT = State#state.node_gets_total,
+    NewState1 = State#state { node_gets_total=NGT+1 },
+    NewState2 = slide_incr(#state.get_fsm_time, Microsecs, Moment, NewState1),
+    NewState3 = slide_incr(#state.node_get_fsm_siblings, NumSiblings, Moment, NewState2),
```
jonmeredith, line 385: NumSiblings/ObjSize should only be updated for non-zero stats, otherwise notfounds will give misleadingly low counts for number of siblings/obj sizes.

jonmeredith:
I've done some more tinkering around to check values of the siblings quantiles and they are being under-reported.  Although the bin size is now 1, basho_stats_histogram is providing a floating point estimate of the quantile within the bucket, so you get a floating point value.

```
(riak@127.0.0.1)5> basho_stats_histogram:quantile(0.5,basho_stats_histogram:update_all(lists:duplicate(1000, 1), basho_stats_histogram:new(0, 1000, 1000))).
0.5
```

Looking at the bin selection code, with the 0, 1000, 1000 arguments it uses
Bin 0 is Sibs <= 1,
Bin 1 is 1 <= Sibs < 2
Bin 2 is 2 <= Sibs < 3

As we're not going to record 0 siblings, it's safe to use Ceil on the quantile result (http://schemecookbook.org/Erlang/NumberRounding) to get a beter integer estimate (given the bin size of 1), however slide currently just uses erlang:trunc on the returned quantiles.

```
Ceil = fun(X) ->
  T = erlang:trunc(X),
    case (X - T) of
        Neg when Neg < 0 -> T;
     Pos when Pos > 0 -> T + 1;
      _ -> T
   end 
end.

%% 0.5,0.95,0.99 quantiles for 1000 sibling=1 values - all give 1
(riak@127.0.0.1)8> Ceil(basho_stats_histogram:quantile(0.5,basho_stats_histogram:update_all(lists:duplicate(1000, 1), basho_stats_histogram:new(0, 1000, 1000)))).
1
(riak@127.0.0.1)9> Ceil(basho_stats_histogram:quantile(0.95,basho_stats_histogram:update_all(lists:duplicate(1000, 1), basho_stats_histogram:new(0, 1000, 1000)))). 
1
(riak@127.0.0.1)10> Ceil(basho_stats_histogram:quantile(0.99,basho_stats_histogram:update_all(lists:duplicate(1000, 1), basho_stats_histogram:new(0, 1000, 1000)))). 
1

%% Testing siblings [1,2,3,4...100] also gives correct answers
(riak@127.0.0.1)11> Ceil(basho_stats_histogram:quantile(0.5,basho_stats_histogram:update_all(lists:seq(1,100), basho_stats_histogram:new(0, 1000, 1000)))).         
50
(riak@127.0.0.1)12> Ceil(basho_stats_histogram:quantile(0.95,basho_stats_histogram:update_all(lists:seq(1,100), basho_stats_histogram:new(0, 1000, 1000)))).
95
(riak@127.0.0.1)13> Ceil(basho_stats_histogram:quantile(0.99,basho_stats_histogram:update_all(lists:seq(1,100), basho_stats_histogram:new(0, 1000, 1000)))).
99

```

We'll get the correct results if slide:compute_quantiles does a Ceil instead - but I think that has to be controllable per slide so we don't change the way our existing stats compute (and the times with large bins are much less sensitive).

src/riak_kv_stat.erl
rustyio, line 385: Fixed.

src/riak_kv_stat.erl
```diff
@@ -44,32 +44,40 @@
...
-%%</dd><dd> update({index_delete, Postings})
+%%</dd><dd> update({index_deletes, Postings})
 %%
 %%</dd><dt> node_gets
 %%</dt><dd> Number of gets coordinated by this node in the last
 %%          minute.
 %%</dd><dd> update({get_fsm_time, Microseconds})
 %%
+%%</dd><dt> node_get_fsm_siblings
+%%</dt><dd> Stats about number of siblings per object in the last minute.
```
jonmeredith, line 74: The info about how to update is a little confusing int the comments now.  perhaps combine all of the node_gets_time/siblings/objsize under one entry with the new update statement? (or you could drop the update lines from siblings / objsize).

src/riak_kv_stat.erl
```diff
@@ -44,32 +44,40 @@
...
+%%</dd><dd> update({node_get_fsm_siblings, NumSiblings})
+%%
+%%</dd><dt> node_get_fsm_objsize
+%%</dt><dd> Stats about object size over the last minute.
```
jonmeredith: Should explain objsize is an estimate

jonmeredith:
+1 merge for code.

Upgrade/downgrade instructions worked for me.  We should probably make a couple of additions.
1) Beam paths were wrong for me in /usr/lib64 on my ubuntu64 release
2) Before/after load we should do an m(Module) to check the build date to ensure the right module has been loaded (first time I tried the upgrade I miscopied the beams so they weren't in ebin)
3) There is no need to sys:resume the stats process - the original one that was suspended got killed.

PR #245, opened 2011-11-01
Expose modfuns for future pipe use (AZ561)

beerriot:
After finding anonymous functions to be quite brittle during rolling upgrades, Riak Pipe is migrating to modfun tuples.  This patch exposes the functions that will be referenced as modfuns in the 1.1.0 version of Riak.  They are exposed here such that Riak 1.0.2 will know what to do with modfuns during the rolling upgrade to 1.1.0.


kellymclaughlin:
+1 to merge.

PR #246, opened 2011-11-01
Convert MapReduce to use modfun tuples in Pipe definition (AZ561)

beerriot:
This is the second half of the Pipe modfun tuple transition for Riak KV.  Because changes were required to `riak_kv_mrc_pipe`, all anonymous functions used there to define Pipe required emulation.  The `compat_fun/1` function exported from that module knows how to pull apart funs created on Riak 1.0.2, which should be the same as the funs created by Riak 1.0.1.

This PR depends on functionality in the Riak Pipe PR https://github.com/basho/riak_pipe/pull/32


kellymclaughlin:
+1 to merge.

PR #247, opened 2011-11-01
Bitcask data cleanup

kellymclaughlin:
- Change the bitcask data directory naming scheme so that timestamp are **no longer** appended.
- Check for existing timestamp-appended data directories and if any exist make a determination of the most recent data directory and make it the current data directory. Any others are moved to a `manual_cleanup` directory for safe keeping in case the automatic determination was incorrect or later manual deletion. If there is inconsistency between the most recently modified data directory and the directory with the most recent timestamp, the most recently modified directory is chosen and a log message is issued.
- The `drop` function now moves the current data directory to an `auto_cleanup` directory and a process is spawned that will cleanup any files in the directory and delete it. This should make sure  to avoid the race condition where a data directory created by a new instance of the vnode starting a bitcask process is deleted by the reaper process.
- `try/after` blocks have been added around asynchronous fold functions.
- The call to `ensure_dir` has been fixed so that it creates the intended directory and not just the parent directory.
- Add eunit test to verify the behavior on startup with respect to handling multiple data directories.
- Add eunit test for `drop` function.
- Add eunit tests for supporting functions `get_data_dir` and `existing_data_dirs`.
- Refactor `backend_eqc` to use different partitions for each generated command set. This is a step towards making the test more similar to the actual way the backends are used by Riak.

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -83,7 +83,8 @@ start(Partition, Config) ->
             {error, data_root_unset};
         DataRoot ->
             %% Check if a directory exists for the partition
-            case get_data_dir(DataRoot, Partition) of
+            PartitionStr = integer_to_list(Partition),
+            case get_data_dir(DataRoot, PartitionStr) of
                 {ok, DataDir} ->
                     BitcaskOpts = set_mode(read_write, Config),
                     case bitcask:open(filename:join(DataRoot, DataDir), BitcaskOpts) of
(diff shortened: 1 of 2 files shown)


jonmeredith:
+1 merge.

Tested 0.14.2 -> 1.0.0 -> 1.0.1 -> 1.0.1 with patch applied under OSX.
Verified that creating a file in a directory updated the modified time.
Checked that creating extra directories name <partnum>-blah got moved to the manual cleanup dir and logged.

PR #248, opened 2011-11-03
Properly sort results for comparison in bitcask startup_dir_test.

kellymclaughlin:
For the test to behave properly the results from the call to `file:list_dir` must be sorted and reversed.


jaredmorrow:
Merging for test.   Will close after tests pass.

jaredmorrow:
Tests pass now.

PR #249, opened 2011-11-14
parallelize 2i input to pipe mr (AZ798)

beerriot:
As with the earlier change to how bucket-key-lists are injected into a
MapReduce pipe, this patch makes the same change for secondary index
inputs: a short pipe is started, and the riak_pipe_qcover_fsm is used to
generate inputs for the MR pipe.

The appenv variable 'mapred_2i_pipe' is provided, and defaulted to 'false'
to provide rolling-upgrade compatibility with 1.0 clusters (which do not
have the riak_kv_pipe_index module).

This patch does introduce a bit of duplication with riak_kv_pipe_listkeys.  This duplication keeps the changeset and upgrade clean.  A merge could be possible if it annoys too much.


kellymclaughlin:
+1 to merge

PR #252, opened 2011-11-29
Improve MR Error Messages (AZ935)

beerriot:
This PR "improves" MapReduce error messages.

At the moment, "improves" means, "gets more information out in the HTTP response."  I'm interested in feedback about whether or not this is the right information to expose, and whether or not it's exposed in a useful format.  If both aspects are +1, similar changes should also be made to the PB response.

Side note: There was early discussion about translating errors like `bad_encoding` to more verbose English like "The input contained bytes that were not valid UTF-8 characters."  This was dropped in favor of just changing the error atom to `bad_utf8_character_code` to keep things simpler.  The new atom should be descriptive, yet unique enough to describe further in other Google-reachable contexts, like the wiki.

### DIFF
--- a/src/riak_kv_js_vm.erl
+++ b/src/riak_kv_js_vm.erl
@@ -233,10 +233,14 @@ code_change(_OldVsn, State, _Extra) ->
 
 %% Internal functions
 define_invoke_anon_js(JS, Args, #state{ctx=Ctx}=State) ->
-    JSFun =  define_anon_js(JS, Args),
-    case invoke_js(Ctx, JSFun) of
-        {ok, R} ->
-            {{ok, R}, State};
+    case define_anon_js(JS, Args) of
+        {ok, JSFun} ->
+            case invoke_js(Ctx, JSFun) of
+                {ok, R} ->
+                    {{ok, R}, State};
+                Error ->
+                    {Error, State}
+            end;
         Error ->
             {Error, State}
     end.
@@ -280,7 +284,7 @@ invoke_js(Ctx, Js, Args) ->
 define_anon_js(JS, Args) ->
     try
         ArgList = build_arg_list(Args, []),
-        iolist_to_binary([JS, <<"(">>, ArgList, <<");">>])
+        {ok, iolist_to_binary([JS, <<"(">>, ArgList, <<");">>])}
     catch
         exit: {ucs, {bad_utf8_character_code}} ->
             lager:error("Error JSON encoding arguments: ~p", [Args]),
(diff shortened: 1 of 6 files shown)


beerriot:
I suppose some hints for triggering errors that render these messages might have been helpful…

This one triggers an error in an erlang function if the given bucket/key does not exist:

```
curl -X POST -H "Content-type: application/json" http://localhost:8098/mapred --data @-
{"inputs":[["none","here"]],"query":[{"map":{"language":"erlang","module":"riak_kv_mapreduce","function":"map_object_value"}}]}
```

---

Triggering an error in JSON encoding, due to bad UTF8, takes two steps. First, upload a file of random, likely invalid-UTF8 bytes:

```
dd count=1 if=/dev/urandom | curl -X PUT -H "content-type: application/octet-stream" http://localhost:8098/riak/bin/rand --data-binary @-
```

Then run a JS map on it:

```
curl -X POST -H "Content-type: application/json" http://localhost:8098/mapred --data @-
{"inputs":[["bin","rand"]],"query":[{"map":{"language":"javascript","source":"function(v) { return [v]; }"}}]}
```

beerriot:
This small addition allowed all of this new error-returning stuff to work for reduce phases as well.  Previously, when a reduce phase encountered an error, it would try reprocessing its inputs when it received more inputs.  If the error still happened on the last input, the phase would end up passing its inputs onward as outputs, rather than raising the error to the user (https://issues.basho.com/show_bug.cgi?id=1185).  The new behavior raises the error and fails immediately.

russelldb:
Pulled, ran. tested, played the az935 game, read, understood, +1'd

russelldb:
Pulled the latest, ran, read, +1ing

tovbinm:
How to solve "bad_utf8_character_code" error returned by Riak MR?

beerriot:
@tovbinm Javascript phases require that all inputs to them be valid UTF8.  For map phases, this means that the value stored in your Riak object, as well as its bucket and key, must be valid UTF8.  If you're getting a `bad_utf8_character_code` then the data you're passing to Javascript contains a string or binary that is not valid UTF8.

tovbinm:
Got it. 

What about "application/binary" values? 

Why do I get the same error even in case when my map function is simple like this "function map () { return [1]; }" ? 
I was hoping is this case values will not be read.

beerriot:
Could we please move this discussion to the mailing list?  There are more people watching, ready to help, and interested in the discussion there.

tovbinm:
Sure. Thanks for your help!

PR #253, opened 2011-11-30
Added memory % used total and memory % used by erlang to kv stats.

massung:


massung:
Some whitespace changes thanks to Emacs. Use ?w=1 in the PR url to ignore them.

dreverri:
How does this differ from get_memory_data()?
https://github.com/basho/riak_kv/blob/master/src/riak_kv_stat.erl#L663

massung:
Good question. On the surface, they appear to be the same. In fact, using get_memory_data() instead, it looks like I could cut out a few proplists:get_value calls and a 1.0 - operation, so I'll make that change. 

As for why I can't use mem_stats() to get the same data, I need this to be output with the regular stats query, and sadly, those use the same tuple keys as what the erlang memory stats use (memory_total would have dual meanings). 

Do you think non-% numbers in the output would be more useful?

massung:
Nevermind. I see that mem_stats() is actually output along with everything else as well in the stats. Closing this PR as everything I need is already there.

Thanks for pointing it out, Dan!

PR #254, opened 2011-12-02
Return ETag and Last-Modified with returnbody=true and siblings

seancribbs:
Associated bugs:
- https://issues.basho.com/show_bug.cgi?id=1020
- https://issues.basho.com/show_bug.cgi?id=1297

Different but related issues to our capability for conditional requests, most specifically `PUT`s:
1. The client is unable to cache or validate previous responses when siblings are present because neither `ETag` nor `Last-Modified` are included in the response.
2. When using `returnbody=true` even in the case where siblings are not generated, neither `Last-Modified` nor `ETag` are present in response, even though we claim it has the same semantics as the response of a `GET` following the `PUT`.

Gotcha:

In the case of siblings, we should likely be returning "weak" entity tags (two available representations that are not the same content have the same ETag), but there are some other cases where we do not.

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -837,35 +837,45 @@ delete_resource(RD, Ctx=#ctx{bucket=B, key=K, client=C, rw=RW, r=R, w=W,
 %% @spec generate_etag(reqdata(), context()) ->
 %%          {undefined|string(), reqdata(), context()}
 %% @doc Get the etag for this resource.
-%%      Documents will have an etag equal to their vtag.  No etag will be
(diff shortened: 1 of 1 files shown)


atl:
Thanks for following this through, Sean!

atl:
This was a "pet" bug of mine, and I was disappointed to discover the code did not make it into the 1.1.0 release.

russelldb:
Works for me, tested it by hand and indeed `ETag` and `Last-Modified` are returned with both sibling `GET`s, and `returnbody` `PUT`s and `POST`s.

Ran the RJC integration test suite against it, and nothing broke, so it is backwards compatible, too (obv. but I had to check.)

PR #257, opened 2011-12-09
Jdm vnodeq stats

jonmeredith:
The stats webmachine resource will now include riak_core_stat:get_stats() which includes vnode queue stats and gossip/ring convergence stats.


reiddraper:
+1 to merge

PR #260, opened 2011-12-12
Sort directory listings in eunit assert to fix tests

jaredmorrow:
Fixes: az1008

This test was passing on my OSX box, but was failing on some build machines because the list order was random so it failed the match.


kellymclaughlin:
+1 to merge

PR #262, opened 2011-12-14
add map_object_key and map_object_bucket_key

sa2ajj:
```
* these functions help to produce a list of keys found by, e.g., secondary
  indices query or search query of Riak
* the former is the most useful for secondary indices
* the latter is the most useful for search queries
* 2nd and 3rd arguments for /3 functions are kept according to other
  helpers
```

### DIFF
--- a/src/riak_kv_mapreduce.erl
+++ b/src/riak_kv_mapreduce.erl
@@ -25,6 +25,8 @@
 
 %% phase spec producers
 -export([map_identity/1,
+         map_object_key/1,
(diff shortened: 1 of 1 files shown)


seancribbs:
@sa2ajj ...but you don't need to fetch the object in a map phase to know what the bucket or key is.

sa2ajj:
@seancribbs, it looked nice at the time.  Today I'd say it's not really necessary :)

PR #264, opened 2011-12-14
BZ1306 — get_fsm crashes when riak_object value is not a binary

ian-plosker:
Fixes issue in `riak_kv_get_fsm:update_stats/2` calling `erlang:size/1` on a `riak_object` value which isn't a binary.

https://issues.basho.com/show_bug.cgi?id=1306


ian-plosker:
Rebasing against 1.0 and resubmitting.

PR #269, opened 2012-01-06
wait for riak_pipe and _kv node services, and unload apps on fullstop (AZ1033 … related)

beerriot:
Not waiting for the node watcher to declare riak_pipe and riak_kv
services as up before running the test was causing preflist exhaustion
errors. Not unloading apps after stopping them was causing re-starting
them for the next test to fail.

### DIFF
--- a/test/mapred_test.erl
+++ b/test/mapred_test.erl
@@ -27,6 +27,7 @@
 dep_apps() ->
     DelMe = "./EUnit-SASL.log",
     DataDir = "./EUnit-datadir",
+    os:cmd("rm -rf " ++ DataDir),
     os:cmd("mkdir " ++ DataDir),
     KillDamnFilterProc = fun() ->
                                  catch exit(whereis(riak_sysmon_filter), kill),
@@ -44,7 +45,7 @@ dep_apps() ->
                    {reduce_js_vm_count, 3}],
     [
      fun(start) ->
-             net_kernel:start([mapred_test@localhost]),
+             net_kernel:start([mapred_test@localhost, shortnames]),
              timer:sleep(50),
              _ = application:stop(sasl),
              _ = application:load(sasl),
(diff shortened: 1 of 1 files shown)


rzezeski:
FWIW I'm seeing 23 failures on this branch.

beerriot:
Yeah, my `unload`ing was too aggressive, and I there may still be a conflict with a `riak_core` test anyway. Do Not Merge.

rzezeski:
Here are some snippets of what I see

```
mapred_test:199: compat_basic1_test_...*failed*
::error:{badmatch,{error,all_nodes_down}}
  in function mapred_test:'-compat_basic1_test_/0-fun-1-'/0


mapred_test:213: compat_basic1_test_...*failed*
::error:{badmatch,{error,{error,insufficient_vnodes_available},{ok,[]}}}
  in function mapred_test:'-compat_basic1_test_/0-fun-2-'/0
```

I also see a lot of the following, not sure if it's hurting anything (yes, my hostname is `crackpot`).

```
=ERROR REPORT==== 11-Jan-2012::11:38:31 ===
** System running to use fully qualified hostnames **
** Hostname crackpot is illegal **
```

rzezeski:
If there is anything I can do to assist let me know.

rzezeski:
Remove the `unload` calls gets me down to 13 errors.

rzezeski:
Just discovered the lists of nodes reported by `riak_core_node_watcher:nodes/1` and what's in the ring do not match up.  This means that `get_apl_ann` and `riak_core_apl:check_up` will fail to find any partitions that are considered available.

test/mapred_test.erl
```diff
@@ -98,8 +99,7 @@ dep_apps() ->
      inets,
      mochiweb,
      fun(start) ->
-             Ring = riak_core_ring:fresh(16, node()),
-             riak_core_ring_manager:set_ring_global(Ring),
+             riak_core_ring:fresh(16, node()),
```
jonmeredith: this change seems veyr strange - the output of riak_core_ring:fresh() is ignored, before it was being used to set the ring at 16 partitions.

rzezeski: You're right, I actually didn't mean to remove the set_ring_global but it worked just fine.  I think this wasn't needed in first place and that's why it works.

rzezeski:
I switched to shortnames cause I saw a lot of errors (that don't cause test to fail) about invalide hostnames and looking at keys_fsm_eqc I saw it uses shortnames as well.

rzezeski:
Fixed, merged PR https://github.com/basho/riak_kv/pull/275

PR #270, opened 2012-01-10
Az1064 postcommit error

rzezeski:
If a post-commit fails but there is no evidence in the logs did it actually
fail?  I dunno, but we should log post-commit failures.


src/riak_kv_put_fsm.erl
```diff
@@ -649,6 +652,22 @@ decode_precommit({js, JSName, Result}) ->
 decode_precommit({error, Reason}) ->
     {fail, Reason}.
 
+decode_postcommit({erlang, {M,F}, Res}) ->
+    case Res of
+        fail -> lager:error("Post-commit hook failed, no resaon given");
```
seancribbs: "resaon" -> "reason"

seancribbs: Also, perhaps it could still be useful to log the MFA.

rzezeski:
A PR to address same issue but in slightly different manner (after feedback from Jon): https://github.com/basho/riak_kv/pull/272

PR #273, opened 2012-01-11
1.0.3 eunit fixes

jonmeredith:
Make the mapred test wait for the riak_kv service to be up.  1.0.3 changed riak_kv service to be marked up after primary vnodes are initialized and running, not after application was started.


jonmeredith:
Replaced by pull request 275

PR #277, opened 2012-01-13
Kill vnode workers after finishing their fold.

massung:
Terminate the process instead of continuing with the state machine. Do this because if there was a lot of memory generated during the fold (e.g. list keys) then the memory won't be cleaned up immediately. Instead the process will go back to 'sleep' in the pool until it's needed again. Once needed it will GC, but that means there can potentially be N processes running all eating up unused memory. We could issue an erlang:collect_garbage() call here instead, but that would eat up more processor time than just killing and restarting the worker.

Note: when testing, please try and run a LOT of worker requests that will be short-lived. We want to make sure that the restart frequency isn't triggered. Also be sure and test riak_kv async_folds true and false.

### DIFF
--- a/src/riak_kv_worker.erl
+++ b/src/riak_kv_worker.erl
@@ -49,4 +49,17 @@ handle_work({fold, FoldFun, FinishFun}, _Sender, State) ->
     catch
         receiver_down -> ok
     end,
-    {noreply, State}.
+
+    %% Here we're going to terminate the process instead of
+    %% continuing with the state machine. We do this because
+    %% if there was a lot of memory generated during the fold
+    %% (e.g. list keys) then the memory won't be cleaned up
+    %% immediately. Instead the process will go back to 'sleep'
+    %% in the pool until it's needed again. Once needed it will
+    %% GC, but that means there can potentially be N processes
+    %% running all eating up unused memory. We could issue an
+    %% erlang:collect_garbage() call here instead, but that
+    %% would eat up more processor time than just killing and
+    %% restarting the worker.
+
+    exit(normal).


jaredmorrow:
According to buildbot, it doesn't compile.

http://buildbot-master.den.basho:8010/builders/test-riak_kv/builds/66/steps/shell_3/logs/stdio

slfritchie:
Riak has drifted enough to invalidate this old PR, sorry!

```
test/backend_eqc.erl:129:<0.1131.0>: Cmds: [{stopped,{backend_eqc,init_backend,3}},
       {running,{riak_kv_bitcask_backend,fold_buckets,4}},
       {running,{riak_kv_bitcask_backend,put,5}},
       {running,{riak_kv_bitcask_backend,get,3}},
       {running,{riak_kv_bitcask_backend,fold_buckets,4}}]

test/backend_eqc.erl:132:<0.1131.0>: Result: {postcondition,
            {'EXIT',
                {badarg,
                    [{erlang,'++',
                         [{error,
                              {worker_crash,normal,
                                  {fold,
                                      #Fun<riak_kv_bitcask_backend.6.36002965>,
                                      #Fun<backend_eqc.9.90135916>}}},
                          [<<"b2">>]],
                         []},
                     {backend_eqc,receive_fold_results,1,
                         [{file,"test/backend_eqc.erl"},{line,345}]},
                     {backend_eqc,postcondition,5,
                         [{file,"test/backend_eqc.erl"},{line,471}]},
                     {eqc_statem,run_commands,2,
                         [{file,"../src/eqc_statem.erl"},{line,704}]},
                     {eqc_fsm,run_commands,2,
                         [{file,"../src/eqc_fsm.erl"},{line,658}]},
                     {backend_eqc,'-prop_backend/4-fun-2-',2,
                         [{file,"test/backend_eqc.erl"},{line,124}]}]}}}
```

PR #280, opened 2012-01-31
Fix eunit timeout for eleveldb EQC test

Vagabond:


jonmeredith:
+1 merge

times on my system are
  riak_kv_eleveldb_backend:522: eqc_test_...[2.581 s] ok
  riak_kv_eleveldb_backend:528: eqc_test_...[1.559 s] ok

which is near enough 5s on a loaded vmbox to cause problems.

jtuple:
+1 merge

PR #281, opened 2012-02-02
Fix starting riak_core now lager is a dep and make buildbot happy

Vagabond:
The put_fsm_eqc test was including headers in a way that assumed riak_kv
was checked out to a directory called 'riak_kv', which isn't the case
whem buildbot runs a per-repo eunit run. This patch fixes the includes
to always work regardless of the checkout directory.


jaredmorrow:
+1

PR #282, opened 2012-02-03
Raise logging level for pre/post commit hooks

alinpopa:
Raise logging level for post and pre commit hooks from debug to error
only for situations when invoking the hook will fail.

When logging for these messages was 'debug', there is no way in production environments to know the reason why the hooks execution failed (you could find out that it failed if using riak-admin status, but it wouldn't say the reason)

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -626,7 +626,7 @@ decode_precommit({erlang, {Mod, Fun}, Result}) ->
             Result;
         {'EXIT',  Mod, Fun, Class, Exception} ->
             riak_kv_stat:update(precommit_fail),
-            lager:debug("Problem invoking pre-commit hook ~p:~p -> ~p:~p~n~p",
+            lager:error("Problem invoking pre-commit hook ~p:~p -> ~p:~p~n~p",
                         [Mod,Fun,Class,Exception, erlang:get_stacktrace()]),
             {fail, {hook_crashed, {Mod, Fun, Class, Exception}}};
         Obj ->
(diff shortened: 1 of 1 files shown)


alinpopa:
Per Andrew Thompson: "The reason we put it at the debug level is to avoid logging flood when
you break a postcommit hook in production and all your postcommits start
to fail. If you're doing thousands of writes per second, that would
obviously be bad."

Please disregard this pull request if this is the case.

Alin

jonmeredith:
Hi Alin,

Your comment is right - we decided not to set the default configuration so that setting a bad hook would spam the logs.  You can change the debug level on the fly for debugging.  Currently it applies to all log messages so there may be information from other subsystems too, we indent to make it finer grained in future.

To change the log level to the console (if you're attached) and send a test message to verify you see it.

```
(riak@127.0.0.1)3> lager:set_loglevel(lager_console_backend, debug).
(riak@127.0.0.1)4> lager:log(debug, "Test ~s", ["Msg"]).
10:18:39.429 [debug] Msg
ok
```

to revert

```
(riak@127.0.0.1)3> lager:set_loglevel(lager_console_backend, info).
```

Similarly you can set it for the disk logs

```
(riak@127.0.0.1)3> lager:set_loglevel(lager_file_backend, "./log/error.log", debug).
```

alinpopa:
Ok, going to use that when the problem comes up.

Thanks Jon.

PharkMillups:
@jonmeredith is this tip about/code for logging worth adding to the wiki?

jonmeredith:
Sure - have at it Phark, that'd be very helpful.

On Sun, Feb 5, 2012 at 1:06 PM, Mark Phillips <
reply@reply.github.com

> wrote:
> 
> @jonmeredith is this tip about/code for logging worth adding to the wiki?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/basho/riak_kv/pull/282#issuecomment-3820263

## 

Jon Meredith
Platform Engineering Manager
Basho Technologies, Inc.
jmeredith@basho.com

alinpopa:
@PharkMillups If I may jump in, I'd say that it would help for sure, as the wiki is also outdated saying that the problems will appear in the sasl-error.log file which is not the case anymore.

PharkMillups:
@alinpopa excellent. we would love your help. I've opened an issue on the wiki repo that we can work from. Feel free to add info to this and start crafting content. 

https://github.com/basho/riak_wiki/issues/254

PR #285, opened 2012-02-10
making riak-admin status and /stats output consistent.

fygrave:
After looking at the stats code, I figured out that apparently the reason why disk stats are excluded from /stats output is because it breaks mochijson2 json encoder. I made a quick workaround to convert the disk space data into lists, so it looks appropriate in json output.

### DIFF
--- a/src/riak_kv_wm_stats.erl
+++ b/src/riak_kv_wm_stats.erl
@@ -87,7 +87,7 @@ pretty_print(RD1, C1=#ctx{}) ->
     {json_pp:print(binary_to_list(list_to_binary(Json))), RD2, C2}.
 
 get_stats() ->
-    proplists:delete(disk, riak_kv_stat:get_stats()) ++
-        riak_core_stat:get_stats().
-
-
+    D = riak_kv_stat:get_stats(),
+    Disk = proplists:get_value(disk, D),
+    DiskFlat = lists:flatmap(fun(X) -> {Ad,Bd,Cd} = X,  [[ list_to_binary(Ad), Bd, Cd]] end, Disk),
+    lists:append([proplists:delete(disk, D), [{disk, DiskFlat}], riak_core_stat:get_stats()]).


src/riak_kv_wm_stats.erl
```diff
@@ -87,7 +87,7 @@ pretty_print(RD1, C1=#ctx{}) ->
     {json_pp:print(binary_to_list(list_to_binary(Json))), RD2, C2}.
 
 get_stats() ->
-    proplists:delete(disk, riak_kv_stat:get_stats()) ++
-        riak_core_stat:get_stats().
-
-
+    D = riak_kv_stat:get_stats(),
+    Disk = proplists:get_value(disk, D),
```
russelldb, line 91: I'd prefer `{value, Disk, Stats} = lists:keytake(disk, 1, riak_kv_stat:get_stats()),` saves deleting in the final list:append, and names the resultant list (Stats) more meaningfully than `D`

russelldb:
Also, running this on a multidisk system, I'm not sure the output is correct:

```
"disk": [
    [
        "/", 
        83301036, 
        58
    ], 
    [
        "/dev", 
        16452708, 
        1
    ], 
    [
        "/run", 
        6584708, 
        1
    ], 
    [
        "/run/lock", 
        5120, 
        0
    ], 
    [
        "/run/shm", 
        16461764, 
        0
    ], 
    [
        "/boot", 
        233335, 
        12
    ], 
    [
        "/data", 
        117050432, 
        2
    ], 
    [
        "/bulk", 
        1951642368, 
        19
    ]
], 
```

I think each disk should be a json object, not an array. Probably better structured like

```
"disk": [

        {"id": "/", 
          "size": 83301036, 
           "percent_used": 58
        }
]
```

russelldb:
Apart from that, really appreciate the work and agree it needs doing. If you don't have time to make the changes and re-submit the PR, let me know, and I'll raise an issue and address it as soon as I can.

russelldb:
I'm going to close, but have credited you with the fix at #383

PR #288, opened 2012-02-23
Make JS VMs monitor the processes that check them out

beerriot:
This is to prevent VMs from leaking out of the pool when their owner
dies before triggering their checkin.  For example, if a map process
dies after requesting a vm reservation, but before dispatching any js
evaluations.

For @rzezeski to ponder.


rzezeski:
I added to this PR in https://github.com/basho/riak_kv/pull/294

PR #289, opened 2012-02-23
Stop listing keys into a pipe sooner after it closes

beerriot:
When listing keys into an MR pipe, the target pipe may close, but the
listkeys pipe may live on for some time.  This is an attempt to shutdown
the listkeys pipe sooner by stopping if _none_ of the keys it was
sending could be succesfully enqueued.

The worker sending the keys would be killed after its vnode notices that its fitting has vanished, which happens after that fitting notices that the downstream fitting has vanished.  Unfortunately, these can be delayed due to the high message volume of a keylist stream, and further delayed by the `2000` shutdown method set in `riak_pipe_vnode_worker_sup`.  Thus, this attempt to speed things up.

More for @rzezeski to ponder.


jonmeredith:
Went with #298 in the end.  With this PR the worker would still succeed if any of the send_output succeeded, rather than all of them.

PR #291, opened 2012-02-23
Add the riak_kv_yessir_backend.erl storage manager for simulation purposes

slfritchie:
%% @doc riak_kv_yessir_backend is a backend for benchmarking Riak without
%%      any disk I/O or RAM constraints.
%%
%% Riak: "Store this key/value pair."
%% Backend: "Yes, sir!"
%% Riak: "Get me that key/value pair."
%% Backend: "Yes, sir!"
%%
%% This backend uses zero disk resources and uses constant memory.
%%
%% \* All put requests are immediately acknowledged 'ok'.  No
%%   data about the put request is stored.
%% \* All get requests are fulfilled by creating a constant binary for
%%   the value.  No attempt is made to correlate get keys with
%%   previously-put keys or to correlate get values with previously-put
%%   values.
%%   - Get operation keys that are formatted in with the convention
%%     <<"yessir.{integer}.anything">> will use integer as the returned binary's
%%     Size.
%%
%% This backend is the Riak storage manager equivalent of:
%%
%% \* cat > /dev/null
%% \* cat < /dev/zero

### DIFF
--- a/src/riak_kv_yessir_backend.erl
+++ b/src/riak_kv_yessir_backend.erl
@@ -51,6 +51,10 @@
 %%   - Anything non-zero would trigger read-repair, which could be useful
 %%     for some simulations.
 %% * Is there a need for simulations for get to return different vclocks?
+%% * Add variable latency before responding.  This callback API is
+%%   synchronous, but adding constant- & uniform- & pareto-distributed
+%%   delays would simulate disk I/O latencies because all other backend
+%%   APIs are also synchronous.
 
 -module(riak_kv_yessir_backend).
 -behavior(riak_kv_backend).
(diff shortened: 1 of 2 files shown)


src/riak_kv_yessir_backend.erl
```diff
@@ -0,0 +1,224 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_fs_backend: storage engine based on basic filesystem access
```
seancribbs: copy/paste-o

slfritchie: Fixed.  Turns out that that back end was so old that it didn't have the proper callbacks
or args in the correct order or -spec's or ... silly me.

buddhisthead:
+1 Confirmed operation with get/put/update/fold_keys/fold_objects and new configuration variables.

Issue #287, opened 2012-02-15
Race Condition causes Javascript VMs to never be returned to VM Pool


ian-plosker:
There is a race condition in the `riak_kv_js_manager` wherein:
1. A process (e.g. a pipe fitter) calls `riak_kv_js_manager:blocking_dispatch/4`
2. `riak_kv_js_manager:blocking_dispatch` function calls `reserve_vm`, a `gen_server:call` to the `riak_kv_js_manager`, which reserves a VM
3. The calling process (e.g. a pipe fitting) is killed before calling `riak_kv_js_vm:blocking_dispatch/3`, a `gen_server:call` to the `riak_kv_js_manager`, which normally would return the JS VM to the VM pool after successfully running the provided JS

In this situation, the reserved JS VM will never be returned to the JS VM pool.

The offending code is as follows:

``` erlang
blocking_dispatch(Name, JSCall, MaxCount, Count) ->
    case reserve_vm(Name) of %% reserve_vm is a gen_server:call to the riak_kv_js_manager
        {ok, VM} ->
            JobId = {VM, make_ref()},
            riak_kv_js_vm:blocking_dispatch(VM, JobId, JSCall); %% riak_kv_js_vm:blocking_dispatch is a gen_server:call to the riak_kv_js_vm
        {error, no_vms} ->
            back_off(MaxCount, Count),
            blocking_dispatch(Name, JSCall, MaxCount, Count - 1)
    end.
```

The simple solution would appear to be linking the calling process to the JS VM before allowing it to be checkout of the pool.

ian-plosker:
Bugzilla mirror: https://issues.basho.com/show_bug.cgi?id=1336

rzezeski:
Also related: https://issues.basho.com/show_bug.cgi?id=1151

rzezeski:
Fixed by https://github.com/basho/riak_kv/pull/294

PR #294, opened 2012-02-25
Rz js vm reserve

rzezeski:
This patch addresses https://github.com/basho/riak_kv/issues/287 (#287)

It prevents dead pids from owning JS VMs and thus leaking them from the pool.

### DIFF
--- a/src/riak_kv_js_manager.erl
+++ b/src/riak_kv_js_manager.erl
@@ -92,7 +92,7 @@ blocking_dispatch(Name, JSCall, Tries) ->
     blocking_dispatch(Name, JSCall, Tries, Tries).
 
 reserve_vm(Name) ->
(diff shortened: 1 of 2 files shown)


rzezeski:
I ran both manual and basho_bench tests for before and after this patch.  I tested against Erlang R14B04 on both Ubuntu and OS X.  I confirmed that before the patch JS VMs leak and after they do not.  After the patch if the cluster is overloaded you will see `All VMs are busy` on server and `preflist exhausted` on client.  This is expected behavior given that JS VMs are a limited resource.  I also verified that loaded but not overloaded JS map reduce correctly returns results.

jonmeredith:
+1 merge

rzezeski:
This fixes issue #287

PR #295, opened 2012-02-28
Make kvget and map workers return errors from send_output.

jonmeredith:
Previously the return was being ignored so if the send_output had
an error (e.g. worker_limit_reached) the inputs would be dropped
with no way for the client to know.

Tested by configuring a very low number of partitions (4) and workers (1 or 2), populating with a small test set and then running

```
pbmr(Inputs, MapReduce) ->
    {ok, S} = riakc_pb_socket:start_link({127,0,0,1},8081),
    try
        riakc_pb_socket:mapred(S, Inputs, MapReduce, 9000, 10000)
    after
        riakc_pb_socket:stop(S)
    end.


pbmr_count(Inputs) ->
    pbmr(Inputs,
         [{map, {modfun, riak_kv_mapreduce, map_object_value}, none, false},
          {reduce, {modfun, riak_kv_mapreduce, reduce_count_inputs}, none, true}]).


```

HTTP interface needs to be checked it propagates the failure.

### DIFF
--- a/src/riak_kv_mrc_map.erl
+++ b/src/riak_kv_mrc_map.erl
@@ -140,8 +140,7 @@ process(Input, _Last,
     case map(Phase, Arg, Input) of
         {ok, Results} when is_list(Results) ->
             ?T(_FittingDetails, [map], {produced, Results}),
-            send_results(Results, State),
-            {ok, State};
+            send_results(Results, State);
         {ok, _NonListResults} ->
             ?T(_FittingDetails, [map, error],
                {error, {non_list_result, Input}}),
@@ -227,9 +226,15 @@ bucket_linkfun(Bucket) ->
 
 %% @doc Send results to the next fitting.
 -spec send_results([term()], state()) -> ok.
-send_results(Results, #state{p=P, fd=FD}) ->
-    [ riak_pipe_vnode_worker:send_output(R, P, FD) || R <- Results],
-    ok.
+send_results([], State) ->
+    {ok, State};
+send_results([Result | Results], #state{p=P, fd=FD} = State) ->
+    case riak_pipe_vnode_worker:send_output(Result, P, FD) of
+        ok ->
+            send_results(Results, State);
+        ER ->
+            {ER, State}
+    end.
 
 %% @doc Unused.
 -spec done(state()) -> ok.
--- a/src/riak_kv_pipe_get.erl
+++ b/src/riak_kv_pipe_get.erl
@@ -54,6 +54,8 @@
          bkey_nval/1]).
 
 -include("riak_kv_vnode.hrl").
+-include_lib("riak_pipe/include/riak_pipe.hrl").
+-include_lib("riak_pipe/include/riak_pipe_log.hrl").
 
 -export_type([input/0]).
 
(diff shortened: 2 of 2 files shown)


src/riak_kv_mrc_map.erl
```diff
@@ -140,8 +140,7 @@ process(Input, _Last,
     case map(Phase, Arg, Input) of
         {ok, Results} when is_list(Results) ->
             ?T(_FittingDetails, [map], {produced, Results}),
-            send_results(Results, State),
-            {ok, State};
+            send_results(Results, State);
```
rzezeski, line 143: So if this errors out then it will return `{{error, Reason}, State}` but I don't see that in the spec for this function.

src/riak_kv_pipe_get.erl
```diff
@@ -87,15 +89,27 @@ process(Input, Last, #state{partition=Partition, fd=FittingDetails}=State) ->
...
         {ReqId, {r, {ok, Obj}, _, _}} ->
-            riak_pipe_vnode_worker:send_output(
-              {ok, Obj, keydata(Input)}, Partition, FittingDetails),
-            {ok, State};
+            case riak_pipe_vnode_worker:send_output(
+                   {ok, Obj, keydata(Input)}, Partition, FittingDetails) of
+                ok ->
+                    ?T(FittingDetails, [kvget], {got, Input}),
+                    {ok, State};
+                ER ->
```
rzezeski, line 97: Update spec to indicate it can return `{{error, Reason}, State}`

src/riak_kv_pipe_get.erl
```diff
@@ -87,15 +89,27 @@ process(Input, Last, #state{partition=Partition, fd=FittingDetails}=State) ->
...
-                      Partition, FittingDetails),
-                    {ok, State};
+                    case riak_pipe_vnode_worker:send_output(
+                           {Error, bkey(Input), keydata(Input)},
+                           Partition, FittingDetails) of
+                        ok ->
+                            ?T(FittingDetails, [kvget], {notfound, Input}),
+                            {ok, State};
+                        ER ->
+                            ?T(FittingDetails, [kvget], ER),
```
rzezeski, line 110: This makes the assumption that error traces are of the form `{error, Reason}`, which they currently are, but that's an implementation detail that could change.  It's nit picky but thought I'd point it out.  There is also the macro `T_ERR`.

rzezeski:
Ran multiple bench runs using increasing number of concurrent workers and tested both JS and Erlang mapreduce.  Unlike before I was unable to produce any incorrect results.  It wasn't until I used 20 concurrent workers that I saw errors on the client.  They were expected errors such as `worker_limit_reached`, `timeout` and `Error sending inputs`.  All these are better than returning incorrect results like before.

Make sure to update the `-specs` and make a new PR against 1.1.

+1 to merge

src/riak_kv_pipe_get.erl
beerriot, line 110: The difference between this use of the `?T` macro and `?T_ERR` is in the trace types.  The `?T_ERR` macro includes `error` as a type, which will trigger the error handling code in `riak_kv_wm_mapred` and `riak_kv_pb_socket`, causing each to shutdown the query.  The use of `?T` here will only be useful for debugging during development, as MapReduce pipes do not enable the `kvget` trace type.  The intention here seems to be to raise the error to `riak_pipe_vnode_word` by returning and error instead of `ok`, and let it send the error trace. What say you @jonmeredith?

rzezeski, line 110: Interesting, I missed the `error` atom inserted into the list by the macro.

jonmeredith, line 110: Use of ?T for the ER case was just about me debugging. I wanted to use the same mechanism as everything else for returning the errors.  When the fitting returns {error, Reason} that's what will shutdown the pipe, unless it's better to do it this way.

beerriot, line 110: No, I think you've chosen correctly, @jonmeredith.

jonmeredith:
Merged the 1.1 based branch instead - Jared has offered to merge 1.1 to master, so this should close then.

jaredmorrow:
1.1 merged to master

closing this pr

PR #298, opened 2012-02-29
Change listkeys pipe to fail when send_output fails

jtuple:
The listkeys pipe that sends listkeys results into downstream pipes did not check if send_output failed. This change checks for failed send_output, as well as propagates the failure up the pipe call-chain. This pull-request also ensures that the listkeys pipe correctly sets up error tracing when calling `riak_pipe:exec`.

For error propagation to work correctly, this must be used with the following changes to riak_pipe:
https://github.com/basho/riak_pipe/pull/43

The easiest way to test these changes is use the following branch that generates send_output failures for the listkeys fitting:
https://github.com/basho/riak_pipe/tree/jdb-pipe-listkeys-test


jonmeredith:
+1 merge

PR #303, opened 2012-03-14
Perfomance improvement.

ddosia:
Function length/1 was used to check for list emptiness. Every time
when chunk of keys had arrived into riak_kv_map_phase:handle_input/3 this
function call was executed on it.

I checked riak_kv_map_phase:handle_input/3 and it is called when i do mapred vi erlang client.
As @seancribbs noticed, this function is called only in legacy modue for {mapred_system, legacy}, but in my app.config i got {mapred_system, pipe}, so could anyone review this behaviour and this commit?


seancribbs:
Thanks! Sorry it took so long to review this.

PR #305, opened 2012-03-20
Instrument KV get latency in the worker

seancribbs:
This adds some latency timing traces to the `riak_kv_pipe_get` vnode worker so that pipes that have large workloads can detect when fetching the KV object is the bottleneck. Errors (like `notfound`) are instrumented as well as successful fetches.

### DIFF
--- a/src/riak_kv_pipe_get.erl
+++ b/src/riak_kv_pipe_get.erl
@@ -82,7 +82,7 @@ init(Partition, FittingDetails) ->
          -> {ok | forward_preflist | {error, term()}, state()}.
 process(Input, Last, #state{partition=Partition, fd=FittingDetails}=State) ->
     ReqId = make_req_id(),
-    Start = erlang:now(),
+    Start = os:timestamp(),
     riak_core_vnode_master:command(
       {Partition, node()}, %% assume local chashfun was used
       ?KV_GET_REQ{bkey=bkey(Input), req_id=ReqId},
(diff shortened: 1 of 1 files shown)


src/riak_kv_pipe_get.erl
```diff
@@ -89,6 +90,7 @@ process(Input, Last, #state{partition=Partition, fd=FittingDetails}=State) ->
       riak_kv_vnode_master),
     receive
         {ReqId, {r, {ok, Obj}, _, _}} ->
+            ?T(FittingDetails, [kv_get], [{ok, timer:now_diff(erlang:now(), Start)}]),
```
kellymclaughlin: I think we might want to expand the message to be more indicative of what is being logged. When I look at the lager output of one of these messages it looks like this:

```
{kvget_map,0}: {trace,[error],[{ok,26270}]}
```

which doesn't really give a lot of context for gathering any meaning from it. If the only `Type` specified is `kv_get` it's not a problem, but if there a re a lot of other trace types enabled it could be difficult to understand. 

Something more like this example from the `riak_pipe_vnode` module would be useful I think:

```
?T(Worker#worker.details, [queue], {vnode, {dequeue, Partition}})
```

That statement results in the following lager output and give some better context:

```
{kvget_map,0}: {trace,[error],{vnode,{dequeue,1324485858831130769622089379649131486563188867072}}}
```

seancribbs:
@kellymclaughlin I have improved the trace format and switched to `os:timestamp()` for timings. Review again?

kellymclaughlin:
Looks good, thanks! +1 to merge.

PR #315, opened 2012-04-17
Add QuickCheck coverage for 2I features.

seancribbs:
We didn't have decent test coverage of 2I for backends. This gives a standard suite to run against the backends that currently support it, including LevelDB and Memory (and Multi if configured properly), thanks to QuickCheck.

### DIFF
--- a/test/backend_eqc.erl
+++ b/test/backend_eqc.erl
@@ -56,7 +56,7 @@
          delete/5,
          init_backend/3]).
 
--define(TEST_ITERATIONS, 50).
+-define(TEST_ITERATIONS, 250).
 
 -record(qcst, {backend, % Backend module under test
                volatile, % Indicates if backend is volatile
@@ -316,7 +316,7 @@ update_indexes(Bucket, Key, [{remove, Index, SKey}|T], Indexes) ->
 
 remove_indexes(Bucket, Key, Indexes) ->
     ordsets:filter(fun({B,_,_,K}) ->
-                           B /= Bucket andalso K /= Key
+                           B /= Bucket orelse K /= Key
                    end, Indexes).
 
 %%====================================================================
--- a/test/backend_eqc.erl
+++ b/test/backend_eqc.erl
@@ -40,6 +40,7 @@
 -export([initial_state/0,
          initial_state_data/0,
          next_state_data/5,
+         dynamic_precondition/4,
          precondition/4,
          postcondition/5]).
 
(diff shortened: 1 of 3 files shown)


test/backend_eqc.erl
```diff
@@ -271,17 +351,20 @@ stopped(#qcst{backend=Backend,
...
-     {history, {call, Backend, fold_keys, [fold_keys_fun(), get_fold_buffer(), g_opts(), State]}},
+     {history, {call, Backend, fold_keys, [fold_keys_fun(), get_fold_buffer(), fold_keys_opts(), State]}},
      {history, {call, Backend, fold_objects, [fold_objects_fun(), get_fold_buffer(), g_opts(), State]}},
      {history, {call, Backend, is_empty, [State]}},
      {history, {call, ?MODULE, drop, [Backend, State]}},
      {stopped, {call, Backend, stop, [State]}}
     ].
 
+precondition(_From,_To,#qcst{backend=Backend},{call, _M, fold_keys, [_FoldFun, _Acc, [{index, Bucket, _}], BeState]}) ->
+    {ok, Capabilities} = Backend:capabilities(Bucket, BeState),
```
Vagabond, line 381: This seems to call Backend:capabilities/2 with the symbolic state, and thus break the test.

Vagabond:
So, this test fails for me on both multi and eleveldb backends, is this expected? Memory seems to actually pass.

seancribbs:
@Vagabond Thanks for looking, I will debug tomorrow.

seancribbs:
@Vagabond Another look-see? Thanks!

Vagabond:
I haz counterexample:

https://gist.github.com/5d46a79ee2af2390c6fe

This is eleveldb without async folds enabled.

Vagabond:
Found another with async folds on.

Vagabond:
You need to run quite a few tests before you'll hit one, though. The default of 50 is nowhere near what is needed.

seancribbs:
I will double-check the postconditions tomorrow, what it shrank to doesn't seem erroneous.

seancribbs:
Yes, but I have run all backends with 250 or more and not provoked that example. Needs more time I guess.

Sean Cribbs

Vagabond:
Perhaps your preconditions aren't strict enough to let it shrink correctly. Although that first countexample doesn't look like a bad shrink to me...

Vagabond:
+1 to merge, now.

PR #319, opened 2012-04-28
Update to use new riak_core capability system

jtuple:
This pull-request changes several kv configuration options to use the new `riak_core` capabilities system. See the related `riak_core` pull-request: basho/riak_core#171

Associated feature request: basho/riak_core#169

### DIFF
--- a/src/riak.erl
+++ b/src/riak.erl
@@ -107,8 +107,7 @@ client_connect(Node, ClientId= <<_:32>>) ->
             %% or the new vnode based vclocks should be used.
             %% N.B. all nodes must be upgraded to 1.0 before
             %% this can be enabled.
-            case rpc:call(Node, app_helper, get_env,
-                          [riak_kv, vnode_vclocks, false]) of
+            case vnode_vclocks(Node) of
                 {badrpc, _Reason} ->
                     {error, {could_not_reach_node, Node}};
                 true ->
@@ -121,6 +120,16 @@ client_connect(Node, undefined) ->
     client_connect(Node, riak_core_util:mkclientid(Node));
 client_connect(Node, Other) ->
     client_connect(Node, <<(erlang:phash2(Other)):32>>).
+
+vnode_vclocks(Node) ->
+    case rpc:call(Node, riak_core_capability, get,
+                  [{riak_kv, vnode_vclocks}]) of
+        {badrpc, {'EXIT', {undef, _}}} ->
+            rpc:call(Node, app_helper, get_env,
+                     [riak_kv, vnode_vclocks, false]);
+        Result ->
+            Result
+    end.
 
 %%
 %% @doc Validate that a specified node is accessible and functional.
--- a/src/riak_kv_util.erl
+++ b/src/riak_kv_util.erl
@@ -153,7 +153,7 @@ normalize_rw_value(_, _) -> error.
 %%      application's environment.
 -spec mapred_system() -> pipe | legacy.
 mapred_system() ->
-    riak_core_capability:get({riak_kv, mapred_system}).
+    riak_core_capability:get({riak_kv, mapred_system}, legacy).
 
 
 %% ===================================================================
--- a/src/riak_kv_util.erl
+++ b/src/riak_kv_util.erl
@@ -153,19 +153,7 @@ normalize_rw_value(_, _) -> error.
 %%      application's environment.
 -spec mapred_system() -> pipe | legacy.
 mapred_system() ->
-    case app_helper:get_env(riak_kv, mapred_system, legacy) of
-        pipe   -> pipe;
-        legacy -> legacy;
-        Other  ->
-            error_logger:warning_msg(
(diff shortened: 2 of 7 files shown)


src/riak.erl
```diff
@@ -122,6 +121,16 @@ client_connect(Node, undefined) ->
 client_connect(Node, Other) ->
     client_connect(Node, <<(erlang:phash2(Other)):32>>).
 
+vnode_vclocks(Node) ->
```
rzezeski, line 124: Do you invision more remote calls like this being made in the future?  If so it would be nice to add a `riak_core_capability:get/2` for remote calls that does all this work and returns `true | false | {error, Reason}`.

src/riak_kv_util.erl
```diff
@@ -153,19 +153,7 @@ normalize_rw_value(_, _) -> error.
 %%      application's environment.
 -spec mapred_system() -> pipe | legacy.
 mapred_system() ->
-    case app_helper:get_env(riak_kv, mapred_system, legacy) of
-        pipe   -> pipe;
-        legacy -> legacy;
-        Other  ->
```
rzezeski, line 159: Shouldn't we still match and display this error msg in case someone botches the config value?

rzezeski, line 159: Building on what I said above.  If the value is bogus then `riak_core_capability:query_capability/4` will _silently_ return the default.  I feel like we should still log something to indicate that the user provided an invalid value.

rzezeski, line 159: Finally, unless I'm missing something, neither negotiate nor renegotiate will warn about an invalid value for `use`/`prefer`.

jtuple, line 159: I'll re-add the error message here since we already had it in the first place. I'm not convinced that this is the best approach in general though, and suggest would suggest most cases take advantage of the new `riak_core_capability:get/2` function to have a sane default.

The only place a user can provide a capability is by using the override settings. There really isn't an "invalid value" for `use/prefer`. An acceptable use-case of the override system is to force the cluster into a mode that isn't registered as supported by default. For example, we may have a code path that matches on three modes: `version1, version2, experimental-version3`. However, we only register `version1, version2` as supported capabilities for the given release, so the experimental mode is never negotiated. Yet, users (or our support staff) could use the proper app config override settings to force a node to use the "unsupported" version.

rzezeski, line 159: Well my point was more about the user fat-fingering it.  There should be some msg indicating to the user "hey, you fat-fingered this."

jtuple:
Pushed out a commit that adds defaults to all mode queries that were converted to the capability system. This ensures that a reasonable default is always chosen.

The user can still override settings and trigger bad matches. Current reality is "Bad user. Don't do that". No easy way to change things without more significant changes.

I did work on adding code to warn the user if they specified a use/prefer mode that was not one of the registered modes, but the message would print every time negotiation was triggered -- which was every gossip. I imagine I could add additional logic to keep track of which capabilities we've already warned about, and reset it in certain scenarios, but that seemed like too much additional logic/state to add this close to the code freeze. For now, I'm assuming people won't go around overriding capabilities often enough for it to matter, and there's plenty of other areas where people can fat finger things in Riak where we don't have graceful operation either.

In general, I doubt this will be a major issue. Before it was _required_ for users to muck with app.config settings and set proper modes. Now, this is all automatic. Only users that specifically go in and override things will potentially run into problems, and they really shouldn't do that. I have no intention of documenting the override process for this release, and reserve it for support / internal use for the time being.

src/riak_kv_util.erl
jtuple, line 159: Turns out it's not easy to re-add this message, and it relied on override the application variable in order to prevent the message from being printed over and over again. There is no equivalent for the capability system. A general-purpose "fat finger" warning built into the capability system itself would be a good future addition, but I feel it's safe enough to skip that until we have more time in the future.

src/riak.erl
jtuple, line 124: I don't see this being common. The `vnode_vclocks` code is kinda bizarre in that it changes the actual request sent to a node based on that node's operating mode. Everything else in Riak sends a request to Riak, and then the Riak node determines what to do with it -- ie. legacy vs pipe m/r, legacy vs proxy vnode routing, etc.

If this ever comes up again down the road, then we should refactor into a general purpose utility func. Or, better yet, come up with a different approach that isn't based on RPC'ing to Riak everytime a client creates a connection. I just went this route because it mirrored the pre-existing RPC query of the app var.

src/riak_kv_util.erl
rzezeski, line 159: Yea, you make good points that this should be less of an issue given the more automatic nature of capabilities.  +1 for leaving things as they are.

src/riak.erl
rzezeski, line 124: Make sense.

rzezeski:
+1 to merge

PR #321, opened 2012-05-10
Extract Protocol Buffers interface from KV

seancribbs:
This is part of a series of pull-requests to extract the Protocol Buffers interface from riak_kv and make it available to all applications in an unobtrusive manner.

Related: basho/riak#138

This pull-request extracts `riak_kv_pb_socket` functionality into 3 individual services implementing the `riak_api_pb_service` behaviour -- `riak_kv_pb_object`, `riak_kv_pb_bucket`, and `riak_kv_pb_mapred`. The "ping" and "server_info" requests were moved to the `riak_api` application.

Additionally, review of the `riak_api` application and its design is requested.


kellymclaughlin:
_Editorial comment_: really awesome to see the monstrosity of `riak_kv_pb_socket` broken up

kellymclaughlin:
What about some lager output [here](https://github.com/basho/riak_api/blob/master/src/riak_api_pb_listener.erl#L36) in the case that a user is upgrading and uses an old config. I actually did just that and the error when the `riak_api` section is missing is non-obvious.

src/riak_kv_pb_bucket.erl
```diff
@@ -0,0 +1,133 @@
...
+%% 22 - RpbSetBucketResp
+%% </pre>
+%%
+%% <p>The semantics are unchanged from their original
+%% implementations.</p>
+%% @end
+
+-module(riak_kv_pb_bucket).
+
+-include_lib("riak_pb/include/riak_kv_pb.hrl").
```
kellymclaughlin, line 48: Is it worth a comment here indicating that this is a generated file to avoid any confusion? I'm not certain it is, just a thought.

seancribbs, line 48: In the past we haven't, but the deps should work out so that the header is generated before riak_kv is compiled.

src/riak_kv_pb_bucket.erl
```diff
@@ -0,0 +1,133 @@
...
+         decode/2,
+         encode/1,
+         process/2,
+         process_stream/3]).
+
+-record(?MODULE, {client,    % local client
+                  req,       % current request (for multi-message requests like list keys)
+                  req_ctx}). % context to go along with request (partial results, request ids etc)
+
+-define(state, #?MODULE).
```
rzezeski: Once again, the state/macro thing.  Since you changed it in other places probably want to change it here too.

src/riak_kv_pb_mapred.erl
```diff
@@ -0,0 +1,278 @@
...
+-export([init/0,
+         decode/2,
+         encode/1,
+         process/2,
+         process_stream/3]).
+
+-record(?MODULE, {req,
+                  req_ctx,
+                  client}).
+-define(state, #?MODULE).
```
rzezeski: State thing.

src/riak_kv_pb_mapred.erl
```diff
@@ -0,0 +1,277 @@
...
+    {ok, C} = riak:local_client(),
+    #state{client=C}.
+
+decode(Code, Bin) ->
+    {ok, riak_pb_codec:decode(Code, Bin)}.
+
+encode(Message) ->
+    {ok, riak_pb_codec:encode(Message)}.
+
+%% Start map/reduce job - results will be processed in handle_info
```
rzezeski, line 70: You mean by `process_stream`?

src/riak_kv_pb_mapred.erl
```diff
@@ -0,0 +1,277 @@
...
+                pipe ->
+                    pipe_mapreduce(Req, State, Inputs, Query, Timeout);
+                legacy ->
+                    legacy_mapreduce(Req, State, Inputs, Query, Timeout)
+            end
+    end.
+
+
+process_stream(#pipe_eoi{ref=ReqId}, ReqId,
+               State=#state{req=#rpbmapredreq{},
```
rzezeski, line 87: nitpick: No need to match ReqId three times.  Although, I guess it makes certain the state is for the correct request.  So yea, ignore my nitpicking.

rzezeski:
Is it just me, or did the `rpbgetserverinforeq` disappear?

seancribbs:
@rzezeski It has moved to riak_api.

src/riak_kv_pb_object.erl
```diff
@@ -0,0 +1,303 @@
...
+%% bucket by the get/put FSMs.
+make_option(_, undefined) ->
+    [];
+make_option(_, default) ->
+    [];
+make_option(K, V) ->
+    [{K, V}].
+
+%% Convert a vector clock to erlang
+erlify_rpbvc(undefined) ->
```
rzezeski, line 291: Maybe call this `decode_vc` to be more consistent with other changes?  (and `pbify_rpbvc` to `encode_vc`)

src/riak_kv_pb_object.erl
```diff
@@ -0,0 +1,303 @@
...
+erlify_rpbvc(<<>>) ->
+    vclock:fresh();
+erlify_rpbvc(PbVc) ->
+    binary_to_term(zlib:unzip(PbVc)).
+
+%% Convert a vector clock to protocol buffers
+pbify_rpbvc(Vc) ->
+    zlib:zip(term_to_binary(Vc)).
+
+default_timeout() ->
```
rzezeski, line 302: While we're in here should we potentially do something about this hardcoded timeout?  Or maybe file an issue about it?

rzezeski:
I see a lot of failures with the Java client integration tests on a devrel.  The entries in the crash.log aren't helping much as they just show the incoming message and `{error, function_clause}`.  Currently master of riak is broken because the PB work is half merged (which I take full responsibility for).  I'd like to get this wrapped up today so master will build again.

rzezeski:
I discovered what the problem is after tracing at debug level on the `riak_api_pb_server`.  I don't like that we catch errors and then only show stack if debug level is set.  This means that when a customer encounters an error like this we have to show them how to use debug level tracing so we an see the actual exception.  I say all those debugs should be logged with at error.  Or at least return the stacktrace along with the reason so the logs aren't useless.

rzezeski:
The new `decode_quorum` slightly breaks old behavior of `normalize_mumble`.  The old function did not have a guard to check that argument is integer.  Please revert to old logic for now so things don't break.

seancribbs:
@rzezeski Early feedback from @jonmeredith indicated that a frequently failing client would spam the logs, hence the level was lowered to debug from info. Also, the client should receive the error message in an rpberrorresp, although i admit it's not as useful as a full backtrace. Again, the desire here is to warn the client before stopping abruptly so the error can at least go through the client error handling logic.

rzezeski:
The problem is that what is returned to the client is **useless**.  Either the stack needs to be formatted with it or we need to change from debug->error (it's an error, lets log it as one).

```
catch
    %% Tell the client we errored before closing the connection.
    Type:Failure ->
        lager:debug("Message processing error: ~p:~p ~p", [Type, Failure, erlang:get_stacktrace()]),
        send_error("Error processing incoming message: ~p:~p", [Type, Failure], State),
        {stop, {Type, Failure}, State}
end
```

rzezeski:
I've reviewed this code, compiled it, ran unit tests, and _most_ of the Java client integration tests are running.  I'm going to  merge this in master so that it builds again and file issues for remaining failures.

PR #323, opened 2012-05-14
Klarna extend erlang riak client

rmeritz:
This change allows allows you to make a status request equivalent to riak-admin status from the the protocall buffer interface.

This change is dependent on this pull request that changes basho/riak-erlang-client.


seancribbs:
@rmeritz We're in the process of refactoring both stats and protobuffs. Stay tuned.

seancribbs:
@rmeritz Closing, this will not be compatible going forward.

PR #324, opened 2012-05-16
Refactor how we process config options so that ALL valid config values

dizzyd:
make it through.


jonmeredith:
The code tests as advertised - config options are passed through.  I think we should strongly consider warnings, even if it will be painfully repeated once for each vnode.

jonmeredith:
Much cleaner and nicer.  I think the options message needs to be downgraded to debug, it's too spammy on startup right now.

I'd say it's good enough to merge, especially if we patch eleveldb to write options to the LOG on opening.  There's still the risk of typos in options going unnoticed, but it's probably more work than needed right now and can be done on the next backend interface refactor.

The only final thought I had was what to do for people that have the current version of eleveldb configured correctly with read_options/write_options/fold_options set, or we could just add release notes.  Advantage of this is that you could still set read opts/fold opts separately.

```
    %% Parse out the open/read/write options
    {OpenOpts, _BadOpenOpts} = eleveldb:validate_options(open FinalConfig),
    ReadOpts0 = proplists:get_value(read_options, FinalConfig, FinalConfig),
    {ReadOpts, _BadReadOpts} = eleveldb:validate_options(read, ReadOpts0),
    WriteOpts0 = proplists:get_value(write_options, FinalConfig, FinalConfig),
    {WriteOpts, _BadWriteOpts} = eleveldb:validate_options(write, WriteOpts0),
    FoldOpts0 = lists:keystore(fill_cache, 1, ReadOpts, {fill_cache, false}),
    FoldOpts = proplists:get_value(read_options, FoldOpts0)
```

matthewvon:
+1 would like it merged

dizzyd:
Merged.

PR #325, opened 2012-05-17
Repair Command

rzezeski:
Add the ability to repair a KV partition.  If a partition's data is
lost or corrupted this command may be used to rebuild it from adjacent
partitions.  This is preferable over a list-keys + read-repair since
it is more targeted, should complete more quickly and should put less
strain on the cluster.

The command can currently only be run by attaching to Riak.

```
riak_kv_vnode:repair(Partition).
```


jtuple:
All looks good. My only suggestion has already been noted in basho/riak_core#175:
https://github.com/basho/riak_core/pull/175#issuecomment-6074139

I'm +1 for this after all the `riak_core` changes land.

PR #328, opened 2012-05-22
Use folsom stats for riak_kv

russelldb:
Use the folsom stats lib for all riak_kv stats
Add backwards compatibility code for get_stats
Fix a bug in riak_kv_vnode that reported index_writes when there were none
Deprecate and remove the option to switch stats off.

Depends on https://github.com/basho/riak_core/pull/179


buddhisthead:
+1 to merge

Issue #326, opened 2012-05-18
Typo in riak_client


seancribbs:
riak_client.erl has a reference to a non-existent module riak_kv_mapred_filter.
(without "s")

YAMAMOTO Takashi

http://lists.basho.com/pipermail/riak-users_lists.basho.com/2012-May/008374.html

PR #329, opened 2012-05-22
Fixes Issue #326

michajlo:
Add 's' to riak_kv_mapred_filter to fix bug when running mapreduce with key filters


seancribbs:
:+1: thanks @michajlo!

PR #332, opened 2012-05-25
Extend riak_kv_console to support staged joins

jtuple:
Update `riak_kv_console` to support staged joins used by the new staged clustering system discussed in basho/riak_core#177.

Related to pull-request that implements the new system: basho/riak_core#181

See also the pull-request that changes `riak-admin` to call staged join: basho/riak#148


Vagabond:
+1 on this.

PR #333, opened 2012-05-25
Speed up accumulation of results for multi-keep MR queries

beerriot:
A similar solution to what was suggested in PR #331 is applied here.  In fact, the branch is based on #331's branch, so this PR is targeted there to make the diff more obvious.

Result lists are now built in reverse order to take advantage of the cheaper cons operation instead of the more expensive append.  It is assumed that the number of phases producing results is small enough that a linear search through a list of phases will not be expensive.

Testing can be done in a similar manner:

```
# Load 50k items into bucket fmrg2
for i in {1..10000}; do curl -X PUT -H "content-type: text/plain" http://localhost:8091/riak/fmrg2/$i --data "hello"; done

# Two-reduce query
time curl -s -X POST -H "content-type: application/json" http://localhost:8091/mapred --data "{\"inputs\":\"fmrg2\",\"query\":[{\"reduce\":{\"language\":\"erlang\",\"module\":\"riak_kv_mapreduce\",\"function\":\"reduce_identity\",\"keep\":true}},{\"reduce\":{\"language\":\"erlang\",\"module\":\"riak_kv_mapreduce\",\"function\":\"reduce_identity\",\"keep\":true}}]}" > /dev/null

# Two-map query
time curl -s -X POST -H "content-type: application/json" http://localhost:8091/mapred --data "{\"inputs\":\"fmrg2\",\"query\":[{\"map\":{\"language\":\"javascript\",\"source\":\"function(v) { return [ [v.bucket, v.key] ]; }\",\"keep\":true}},{\"map\":{\"language\":\"javascript\",\"source\":\"function(v) { return [v.key]; }\",\"keep\":true}}]}" > /dev/null
```

These two queries test important different behaviors: two reduces in a row mean that the sink will likely read all of one phase's results and then all of the others, while two maps in a row is likely to yield a mixed order of each phases results.  This could have an affect on the efficiency of the aggregation process.

On a Mac laptop with the eleveldb backend with this 50k set, I see the two-reduce query drop from ~72 seconds to ~28.5 seconds. On the same system, I see the two-map query drop from ~95 seconds to ~40 seconds.  It may also be worth noting that the 10k input described in #331 drops from 3.5 seconds to ~2.5 seconds for two-reduce, and from 9.5 seconds to ~8.2 seconds for two-map, reinforcing that this improvement is about growth of result counts.

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -636,15 +636,19 @@ group_outputs(Outputs, NumKeeps) when NumKeeps < 2 -> % 0 or 1
     %% when length(Outputs) is large
     [ O || {_, O} <- Outputs ];
 group_outputs(Outputs, _NumKeeps) ->
-    Merged = lists:foldl(fun({I,O}, Acc) ->
-                                 dict:append(I, O, Acc)
-                         end,
-                         dict:new(),
-                         Outputs),
-    %% though this looks very similar to the other group_outputs/2
-    %% clause, note that each O is a *list* of outputs here, while
-    %% each O was *an* output there
-    [ O || {_, O} <- lists:keysort(1, dict:to_list(Merged)) ].
+    Group = fun({I,O}, Acc) ->
+                    %% it is assumed that the number of phases
+                    %% producing outputs is small, so a linear search
+                    %% through phases we've seen is not too taxing
+                    case lists:keytake(I, 1, Acc) of
+                        {value, {I, IAcc}, RAcc} ->
+                            [{I,[O|IAcc]}|RAcc];
+                        false ->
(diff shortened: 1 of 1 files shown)


Vagabond:
How do I confirm that the M/R outputs are the same? At the very least the keys aren't coming back in the same order...

Vagabond:
Ok, so it looks like the sublists have the same contents, just in a different order.

beerriot:
Ah, yes, they will come back in deferent orders, depending on how fast each vnode lists its keys, how fast the objects can be read off disk, etc. It should be the case that sorting the sublists causes the before and after results to compare equal.

Vagabond:
+1 to merge this and #331.

beerriot:
Merged.

PR #339, opened 2012-06-04
Remove whereis check for riak_kv_stat gen_server from riak_kv_status

russelldb:
Update riak_kv_stat docs


buddhisthead:
+1 to merge - confirmed fix and "riak-admin status" works again. Thanks.

PR #340, opened 2012-06-06
Track timers throughout lifecycle of the MR processing; fixes #338

### LINKED ISSUE #338: Spurious 400 after HTTP MapReduce
History:
- https://issues.basho.com/show_bug.cgi?id=1222
- https://issues.basho.com/show_bug.cgi?id=1324

The `riak_kv_wm_mapred` module, which provided the MapReduce HTTP interface, implements its timeout mechanism by scheduling a message to be delivered to its process after the timeout has occurred:

https://github.com/basho/riak_kv/blob/master/src/riak_kv_wm_mapred.erl#L148

When a MapReduce query finishes before that message is received, however, mochiweb is likely to receive it instead:

https://github.com/mochi/mochiweb/blob/master/src/mochiweb_http.erl#L69-70

The `mochiweb_http` module expects _only_ tcp socket messages, so if this unknown `pipe_timeout` message is received, mochiweb immediately sends a 400 error message to the client.  The message is safe to ignore, but it can be confusing to prove that it came from this bit of code, and not somewhere else.

What `riak_kv_wm_mapred` should do instead is keep the timer reference around, and cancel the timer before returning control to mochiweb.  Alternatively, all pipe and timeout messages should be handled by a process other than the mochiweb/webmachine/socket-holding process, which would also make it possible to remove the use of the `mochiweb_request_force_close` hack:

https://github.com/basho/riak_kv/blob/master/src/riak_kv_wm_mapred.erl#L299-300

dizzyd:


src/riak_kv_wm_mapred.erl
```diff
@@ -178,12 +191,14 @@ pipe_mapred_nonchunked(RD, State, Pipe, NumKeeps, Sender) ->
...
             HasMRQuery = State#state.mrquery /= [],
             JSONResults1 = riak_kv_mapred_json:jsonify_bkeys(JSONResults, HasMRQuery),
+            cleanup_timer(PipeTref),
             {true,
              wrq:set_resp_body(mochijson2:encode(JSONResults1), RD),
              State};
         {error, {sender_error, Error}} ->
             %% the sender links to the builder, so the builder has
             %% already been torn down
+            cleanup_timer(PipeTref),
```
beerriot, line 201: These calls just before the `prevent_keepalive/0` calls are not necessary. `prevent_keepalive/0` sets a flag to tell Mochiweb to close the socket at the end of this request, instead of attempting to try to read another request.

…on the other hand, we will need them once we implement a better fix than using that flag, which is an undocumented Mochiweb feature.

beerriot:
+1 this fixes the tests I have run via telnet:

```
$ telnet localhost 8098
POST /mapred HTTP/1.1
Content-type: application/json
Content-length: 42

{"inputs":"foo","query":[],"timeout":5000}
```

_wait at least five seconds_ (also swap `/mapred` for `/mapred?chunked=true`).  You'll see a 400 on master, but nothing on this branch.

Also confirmed that timeouts still work:

```
$ telnet localhost 8098
POST /mapred?chunked=true HTTP/1.1
Content-type: application/json
Content-length: 39

{"inputs":"foo","query":[],"timeout":1}
```

As long as your pipe isn't _super_ fast, this responds correctly with `{"error":"timeout"}` on both branches, and then closes the connection.

dizzyd:
Another reproducer/verifier: https://gist.github.com/2879412

PR #346, opened 2012-06-15
check json ctype on mr query

reiddraper:
Verify that map reduce POST requests are sent with content-type of `application/json` (possible including a subtype parameter). So far I've successfully tested the Java and Ruby clients with this patch. I'm working to get my python environment back ready for testing that client...


dizzyd:
+1 w/ caveat that we discuss the style of having check_body in a sep. function on next mumble. :)

PR #348, opened 2012-06-19
Add Referer header validation

dizzyd:


slfritchie:
Pulling PRs out of the freezer ... @seancribbs or @broach any opinions?

seancribbs:
@slfritchie I think this is already handled. https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_utils.erl#L260

PR #349, opened 2012-06-20
spawn writes to folsom, cache reads via riak core stat cache

russelldb:
After testing, reading and writing stats concurrently slowed Riak's general throughput. Short of a process per stat (ideal) or re-introducing the gen_server (possible), spawning for each update seems like a reasonable compromise.

Also, use new stat cache provided by core.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -175,23 +175,60 @@
 %%
 -module(riak_kv_stat).
 
-%% API
--export([get_stats/0, update/1, register_stats/0, produce_stats/0]).
+-behaviour(gen_server).
 
+%% API
+-export([start_link/0, get_stats/0,
+         update/1, register_stats/0, produce_stats/0]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(SERVER, ?MODULE).
 -define(APP, riak_kv).
 
-%% @spec get_stats() -> proplist()
-%% @doc Get the current aggregation of stats.
-get_stats() ->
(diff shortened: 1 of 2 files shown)


beerriot:
It looks like some set of tests needs to start `riak_core_stat_cache` now. I'm seeing several of these print during eunit:

```
*** context setup failed ***
::{noproc,{gen_server,call,
                    [riak_core_stat_cache,
                     {register,riak_kv,{riak_kv_stat,produce_stats,[]},5}]}}
```

russelldb:
I just don't see these errors…I tried with master, then a I tried again with the rdb-stat-cache branch.

To be sure I did a clean checkout of riak_kv, ran `rebar get-deps compile` then `./rebar eunit skip-deps=true`. I got some setup failures saying mochiweb was not started.

I checked out `rdb-stat-cache` and ran `./rebar delete-deps get-deps compile` followed by `./rebar eunit skip-deps=true` and the same setup failures. No mention of `riak_core_stat_cache`.

There are some broken windows that need mending. I'm going to try and get _all_ the tests passing in the morning.

russelldb:
Hmm, maybe this commit https://github.com/basho/riak_kv/commit/e069497fbe9a312178790d7233ab38c16487d91e is the reason I don't see these errors.

russelldb:
Further, the tests that are broken still seem to be mapred_test. The problems (I fixed) then saw that they're addressed already by #353 which prompted Sean to add #354.

Please re-run this branch and see if you still get the same failures.

beerriot:
Yep, I must have had a version mismatch somewhere. Re-running as you did on just a KV checkout, and also on a fresh top-level riak checkout, shows no `riak_core_stat_cache` errors. +1 mege

russelldb:
Thanks!

PR #350, opened 2012-06-20
Start folsom so stat system will work during tests

rzezeski:
Now that our underlying stats system is based on folsom we need to
make sure the application is started and the stats are registered.


beerriot:
+1 fixes five tests for me, please merge, thanks.

PR #353, opened 2012-06-24
Fix assorted test sadness.

jerith:
Most of the fixes are just cleaning up resources created by various tests.
Some stuff in test/mapred_test.erl starts riak_kv_app multiple times, which
causes problems with the return value of the riak_api_pb_service:register()
call. Therefore, this is no longer required to be 'ok'.

I'm not very familiar with the Riak codebase, so I'm not hugely confident that I've done the right thing here, especially in `riak_kv_app:start()`. Also, any comments on style or idiom would be appreciated.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -129,7 +129,7 @@ start(_Type, _StartArgs) ->
                 {bucket_validator, riak_kv_bucket}
             ]),
 
-            riak_api_pb_service:register([{riak_kv_pb_object, 3, 6}, %% ClientID stuff
+            ok = riak_api_pb_service:register([{riak_kv_pb_object, 3, 6}, %% ClientID stuff
                                                {riak_kv_pb_object, 9, 14}, %% Object requests
                                                {riak_kv_pb_bucket, 15, 22}, %% Bucket requests
                                                {riak_kv_pb_mapred, 23, 24}, %% MapReduce requests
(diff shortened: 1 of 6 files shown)


src/riak_kv_app.erl
```diff
@@ -129,7 +129,7 @@ start(_Type, _StartArgs) ->
                 {bucket_validator, riak_kv_bucket}
             ]),
 
-            ok = riak_api_pb_service:register([{riak_kv_pb_object, 3, 6}, %% ClientID stuff
```
seancribbs, line 132: Instead of removing this match, shouldn't it deregister at the end of tests? Or maybe `riak_api_pb_service` should have an `unregister` function.

jerith, line 132: I couldn't find an obvious way to deregister. I'd be much happier doing it that way than removing the match.

seancribbs, line 132: I'll make a PR for `riak_api` and the other apps that adds a `deregister` function, and then calls it in the `stop` application callback.

jerith, line 132: Should I put the match back in, then? The tests will fail with the "already registered" error until the `deregister` magic is in, and then they'll be happy again.

seancribbs, line 132: Please leave the match in, see also basho/riak_api#4 and #354.

jerith, line 132: Cool, I'll do that as soon as I get home. Thanks for your assistance.

jerith:
The match is back in. Is there anything else I should do?

seancribbs:
@jerith This has now been addressed by other PRs. Closing.

PR #354, opened 2012-06-24
Add PB service deregistration on stop.

seancribbs:
This helps address the lingering service registration problem in #353 while maintaining correctness. Requires merge of basho/riak_api#10.


seancribbs:
basho/riak_api#4 has been closed because its implementation is race-prone. I will redo, but the registration/deregistration API should probably stay the same.

joedevivo:
+1

PR #359, opened 2012-07-02
Use a sliding one minute window not a EWMA meter

russelldb:
Depends on https://github.com/boundary/folsom/pull/27

---

Remove dep on basho_metrics


rzezeski:
+1 to merge

PR #360, opened 2012-07-03
Fix eunit failures

seancribbs:
This fixes a bunch of failures and setup problems in `riak_kv`'s eunit/EQC suite. It is currently [passing on Travis](http://travis-ci.org/#!/basho/riak_kv/builds/1770489).


rzezeski:
I haven't reviewed the code but it did fix eunit failures.

src/riak_kv_test_util.erl
```diff
@@ -96,4 +125,101 @@ wait_for_children(PPid) ->
...
+        end,
+
+    [sasl, Silencer, crypto, public_key, ssl, riak_sysmon, os_mon,
+     runtime_tools, erlang_js, inets, mochiweb, webmachine, luke,
+     basho_stats, bitcask, compiler, syntax_tools, lager, folsom,
+     riak_core, riak_pipe, riak_kv, DefaultSetupFun, Extra].
+
+
+do_dep_apps(StartStop, Apps) ->
+    lists:map(fun(A) when is_atom(A) -> application:StartStop(A);
```
beerriot: `application:unload/1` has been known to do strange things in the past. But, I think that may have been before we started running each app's eunits as a separate rebar invocation. If it passes all platforms now, we can return to its debugging when/if necessary.

src/riak_kv_test_util.erl
```diff
@@ -96,4 +125,101 @@ wait_for_children(PPid) ->
...
+setup(TestName, SetupFun) ->
+    %% Cleanup in case a previous test did not
+    cleanup(TestName, SetupFun, setup),
+
+    %% Load application environments
+    Deps = dep_apps(TestName, SetupFun),
+    do_dep_apps(load, Deps),
+
+    %% Start erlang node
+    {ok, Hostname} = inet:gethostname(),
```
beerriot, line 160: This used to be problematic on some platforms that didn't set their DNS/hostname properly, most often after DHCP (you'd get a splitbrain where one component saw the configured hostname, and another saw the DHCP/reverse-dns hostname). Something to watch for on our testing platforms.

src/riak_kv_test_util.erl
```diff
@@ -96,4 +125,101 @@ wait_for_children(PPid) ->
             ok
     end.
 
+%% Performs setup for a test
```
beerriot: These are pretty easy functions to read, but it might be worth a little english describing what atoms `SetupFun` and `CleanupFun` are called with, and when. Eunit setup/cleanup is confusing enough on its own without an extra layer to read through. ;)

beerriot:
Running locally (OS X 10.8, Erlang R15B01), `make test` hangs during the setup of the `riak_kv_delete` tests. After adding debugging printouts, it appears to be stuck at `riak_core:wait_for_application(riak_kv)` in `riak_kv_test_util:setup/2`.

src/riak_kv_test_util.erl
seancribbs, line 160: Does the call actually throw an exception or do you just get different hostnames every time?

src/riak_kv_test_util.erl
seancribbs: I think my goal was to remove all application env entries for the app, and then load them fresh on the next run from the .app file. If unload doesn't do that, then it can easily be removed.

src/riak_kv_test_util.erl
beerriot, line 160: Different hostnames. It always manifested as a generic inability to find/connect-to a node in some later operation. I forgot to note that the last time we monkeyed with this was back in the R13 days, so it's even possible that this is no longer an issue.

src/riak_kv_test_util.erl
beerriot: This, I must admit to cargo-culting (i.e. I'm not sure what else it does). @slfritchie has warned me off when I've snuck in calls to unload before.

slfritchie: application:unload() can occasionally fail with an exception that cascades to the (IIRC) the application controller and then crashes the OTP kernel app, which then stops the VM. Tests interpret the VM shutting down as a failure, which I think is a good feature.  I forget the details of how it works, but it does happen, at least with R14B04.  I haven't tried risking the ire of R15B01.

seancribbs: @slfritchie Is it only certain apps? I could understand why unloading kernel or stdlib would do that.

src/riak_kv_test_util.erl
seancribbs, line 160: Actually this shouldn't matter that much since we have a clean ring every time, as long as the test uses the common setup/cleanup functions. Of course, that assumes there's no blip DNS/DHCP between when we detect the hostname and when the net_kernel starts.  Still something to watch for on our testing platforms, as you say.

seancribbs:
I have rebased this from the 1.2 branch. @beerriot @rzezeski I would appreciate a final review.

src/riak_kv_test_util.erl
slfritchie: I couldn't find a pattern, but it wasn't a matter of unloading apps with sticky dirs like kernel.  It can happen for any of the Riak dependency apps.

seancribbs: Either way, i removed the unloading bit. Not having it doesn't seem to affect any of our existing tests.

src/riak_kv_test_util.erl
```diff
@@ -96,4 +132,160 @@ wait_for_children(PPid) ->
...
+%%
+%% The given `SetupFun' will be called first with the argument `stop'
+%% before other applications are stopped (to cleanup leaky test
+%% state), `load' after all other applications are loaded, and then
+%% `start' after all other applications are started. It is generally
+%% good practice to use the same function in the `SetupFun' as the
+%% `CleanupFun' given to `cleanup/3'.
+%%
+%% @see common_setup/2, dep_apps/2, do_dep_apps/2
+-spec setup(TestName::string(), fun((load|start|stop) -> any()) -> ok.
```
rzezeski: There is no need to put the `TestName` in the spec.  Since sometime in release 14 edoc is smart enough to match specs and function arguments up automatically.

rzezeski:
Why not put `riak_kv_test_util` into the test dir?

Also, that module failed to compile for me:

```
/Users/rzezeski/tmp/riak_kv/src/riak_kv_test_util.erl:51: syntax error before: '->'
/Users/rzezeski/tmp/riak_kv/src/riak_kv_test_util.erl:150: syntax error before: '->'
/Users/rzezeski/tmp/riak_kv/src/riak_kv_test_util.erl:219: syntax error before: ':'
```

rzezeski:
The tests pass and based on the output seem to run more cleanly as well.  Nice job @seancribbs.

+1 to merge.

PR #361, opened 2012-07-07
Track node gets and puts with a spiral stat

russelldb:
When using the `slide_uniform` sample type for the node puts / gets
histogram, the one minute count will be inaccurate for very small
and somewhat large numbers of readings. `Folsom` (actually `bear`)
only calculates the histogram when there are 5 or more readings.
The `slide_uniform` sample type holds a maximum number of 1028
readings per second by default, so nodes with greater throughput
than that will show lower than actual one minute counts. Using
`spiral` for the get / put count solves these issues.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -245,17 +245,17 @@ update1({vnode_index_delete, Postings}) ->
     folsom_metrics:notify_existing_metric({?APP, vnode_index_deletes}, Postings, spiral),
     folsom_metrics:notify_existing_metric({?APP, vnode_index_deletes_postings}, Postings, spiral);
 update1({get_fsm, Bucket, Microsecs, undefined, undefined, PerBucket}) ->
-    folsom_metrics:notify_existing_metric({?APP, node_gets_total}, {inc, 1}, counter),
+    folsom_metrics:notify_existing_metric({?APP, node_gets}, 1, spiral),
     folsom_metrics:notify_existing_metric({?APP, node_get_fsm_time}, Microsecs, histogram),
     do_get_bucket(PerBucket, {Bucket, Microsecs, undefined, undefined});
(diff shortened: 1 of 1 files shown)


src/riak_kv_stat.erl
```diff
@@ -245,17 +245,17 @@ update1({vnode_index_delete, Postings}) ->
     folsom_metrics:notify_existing_metric({?APP, vnode_index_deletes}, Postings, spiral),
     folsom_metrics:notify_existing_metric({?APP, vnode_index_deletes_postings}, Postings, spiral);
 update1({get_fsm, Bucket, Microsecs, undefined, undefined, PerBucket}) ->
-    folsom_metrics:notify_existing_metric({?APP, node_gets_total}, {inc, 1}, counter),
+    folsom_metrics:notify_existing_metric({?APP, node_gets}, 1, spiral),
```
rzezeski, line 248: Did you mean to change the stat name? From `node_gets_total` to `node_gets`.

src/riak_kv_stat.erl
```diff
@@ -340,14 +340,6 @@ backwards_compat(pbc_connects_active, counter, Stats) ->
     {pbc_active, Stats};
 backwards_compat(Name, counter, Stats) ->
     {Name, Stats};
-backwards_compat(node_get_fsm_time, histogram, Stats) ->
```
rzezeski, line 343: Is this no longer needed now that a spiral type is being used?

russelldb, line 343: Correct. The `spiral` makes the code cleaner / simpler since we don't need to calculate the sliding total from the readings.

src/riak_kv_stat.erl
russelldb, line 248: yes, node_gets_total was a counter of the total, and node_gets was assembled from the histogram readings. Now this `spiral` type has 2 fields, a total and sliding 1 minute count. The rename reflects that.

rzezeski:
+1 to merge

PR #363, opened 2012-07-10
When converting riak_object to JSON, treat usermeta as a struct even if it's empty.

seancribbs:
This prevents the case in the Java client where a "map" is expected for the usermeta, but instead the JSON form contains an empty list. Along with basho/riak_pb#10, this resolves the regression in the Java client integration suite.


russelldb:
WFM, +1

PR #364, opened 2012-07-11
Use the pb stats from riak_api

russelldb:
Riak_api does not depend on riak_kv so couldn't call
riak_kv_stat:update/1. Riak_api has its own stats, use them
and make them backwards compatible.

Depends on https://github.com/basho/riak_api/pull/5


seancribbs:
<del>Has riak_api actually been updated to use its own stats?</del> Disregard, I found it.

russelldb:
Did you look at  https://github.com/basho/riak_api/pull/5?

seancribbs:
:+1: works, merge

PR #365, opened 2012-07-11
Add check for "null" origin header

dizzyd:


cmeiklejohn:
+1, tested with riak-ruby-client and curl.  verified origin of null returns 403, and missing or different origin works.

PR #368, opened 2012-07-18
Call eleveldb:close before destroy

jonmeredith:
Requires https://github.com/basho/eleveldb/pull/33

Should fix the MANIFEST missing race, and may help with occasional memory free coredumps we've been seeing.


chardan:
+1

PR #371, opened 2012-07-23
Corrected host token order during referer validation.

glickbot:
The host token order of 'OriginTuple' within 'is_valid_referer(RD)' was backwards. It was checking '1.0.0.127' against '127.0.0.1', or 'com.basho.www' against 'www.basho.com', and thus always failing.


reiddraper:
Looks like this should be fixed in webmachine, not riak_kv. [PR here](https://github.com/basho/webmachine/pull/35).

PR #374, opened 2012-07-25
Put riak_kv_stat under supervision of riak_core

russelldb:
riak_core now supervisior folsom as an included app, stat modules
need to be part of the same supervision tree for consistency.

Remove check of app_env for stats, as stats are always on now


russelldb:
Depends on https://github.com/basho/riak_core/pull/217

argv0:
+1, tested kv stats work before / after crash.

PR #375, opened 2012-07-27
Fixed referer validation false negatives

glickbot:
Effect of bug: All http requests to riak which pass a referer header get a 403 response due to the referer validation failing.

This also affects 1.1.4, as it was introduced with the XSS changes.

Detail: 
The host token order of 'OriginTuple' within 'is_valid_referer(RD)' was backwards. It was checking '1.0.0.127' against '127.0.0.1', or 'com.basho.www' against 'www.basho.com', and thus always failing.


jonmeredith:
Resolves the issue with Rekon, double checking a few last things before merge.

PR #377, opened 2012-07-31
Catch noproc exit on pipe listkeys/index link

beerriot:
This patch is related to basho/riak_pipe#48.

It is possible that the coverage fsm handling the index query at the head of a pipe
will finish (or otherwise exit) before we try to link to it. This patch wraps the
link in a try-catch to prevent a 'noproc' exit (if the coverage fsm did finish
successfully faster than expected, we want to pick up the 'done' message from our
inbox).

The listkeys pipe coverage fitting had the same potential problem (which this patch also addresses).

This was tested by temporarily adding a sleep before the call to erlang:link/1 in riak_kv_pipe_index, to make sure that the coverage FSM had exited before linking to it.


seancribbs:
:+1: This removes those messages from the logs.

PR #378, opened 2012-08-02
Remove the async_folds option passed into the backend at startup.

kellymclaughlin:
The original design for async folding support had the backend make the decision to use async folding at start time, but we changed to having the decision be made per fold call. This change removes some leftover cruft from that original path that is now obsolete.
## Testing

The backend tests should pass with and without this change. I was unable to verify that the entire `riak_kv` test suite passes at the moment because there appear to be unrelated test failures on `master` at the moment.


beerriot:
:+1: yay for removing cruft

PR #379, opened 2012-08-06
Resolve 2I timeout error from case clause

seancribbs:
Neither the HTTP nor the PB interfaces match the `{error, timeout, Acc}` return value of `riak_client:wait_for_query_results/3` that results from a receive timeout. This case is difficult to reach except in the case of a heavily loaded cluster, which is why it has not been hit before.

This changes the return value on timeout to be simply `{error, timeout}`. Partial results from a timed out request could be of dubious value and correctness, so the simplest solution is not to return them.


reiddraper:
Welp, I'm having all sorts of fun running the kv eunit tests, but this looks good to me. +1

PR #381, opened 2012-08-10
Simplify the logging on Travis so it's less spammy.

seancribbs:
At a later date, we can automate uploading the logs somewhere if the build fails.


seancribbs:
Ignore this, use #382.

PR #382, opened 2012-08-10
Use regular logging on Travis for now so we can actually read the output

seancribbs:
At a later date we can automate uploading the logs on failure.


rzezeski:
+1 to merge

PR #366, opened 2012-08-20
Spurious #pipe_log messages in logs as "Unrecognized message"

seancribbs:
When the `riak_kv_pb_mapred` service destroys a pipe (from timeout or simply eoi), it does not drain its message queue of any pipe related messages before exiting streaming mode. This means the `riak_api_pb_server` receives and logs a number of messages it doesn't recognize, including one like below:

```
18:02:25.136 [error] Unrecognized message {pipe_log,#Ref<0.0.1.169795>,0,{trace,[error],{vnode,{fitting_died,1050454301831586472458898473514828420377701515264}}}}
```

If the mapred service would instead drain the mailbox of pipe-related messages, the server will not log them.  Something like the below might work:

``` erlang
drain_pipe_messages(Pipe) ->
    case riak_pipe:receive_result(Pipe, 0) of
        timeout -> ok;
        _Ignore -> drain_pipe_messages(Pipe)
    end.
```

### DIFF
--- a/src/riak_kv_pb_mapred.erl
+++ b/src/riak_kv_pb_mapred.erl
@@ -86,8 +86,10 @@ process(#rpbmapredreq{request=MrReq, content_type=ContentType}=Req,
 process_stream(#pipe_eoi{ref=ReqId}, ReqId,
                State=#state{req=#rpbmapredreq{},
                             req_ctx=#pipe_ctx{ref=ReqId,
-                                              timer=Timer}}) ->
+                                              timer=Timer,
+                                              pipe=Pipe}}) ->
     erlang:cancel_timer(Timer),
+    drain_pipe_messages(Pipe),
     {done, #rpbmapredresp{done = 1}, State#state{req = undefined, req_ctx = undefined}};
 
 process_stream(#pipe_result{ref=ReqId, from=PhaseId, result=Res},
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
In testing this by setting a very low timeout for a MR job, I am observing a race condition where `drain_pipe_messages` is called and the 0 ms timeout fires prior to the messages arriving. This results in the `Unrecognized message` errors still being output. Adding a 5 ms timeout to the `riak_pipe:receive_result` call or a lager statement added sufficient delay for the messages to arrive and be handled as expected.

beerriot:
Thinking about this a bit more, the true fix is one that we also need to implement in `riak_kv_wm_mapred` eventually anyway. That fix is to separate this socket-handling process from the process acting as the pipe sink. Right now they're the same process, which leads to situations like this where we've stopped caring what the pipe is doing, but it's still sending us messages. Here, we see it as log spam; on the webmachine side, we saw it as mochiweb errors (because it expects only `inets` messages): https://github.com/basho/riak_kv/pull/224

beerriot:
I believe that @joedevivo has just seen a more problematic version of this bug. If a `#rpblistkeysreq{}` comes in before the spew of messages finishes, then `riak_kv_pb_bucket:process_stream/2` might be called with those `#pipe_log{}` messages. That function does not contain a catch-all clause, so the process dies with a function clause error. Thanks, `riak_test`!

I am currently hacking on a middleman process to solve this issue for good.

seancribbs:
@beerriot That feels like the best way to go, and will also avoid putting the functionality in too many places.

seancribbs:
Closing in favor of #416

PR #380, opened 2012-08-20
Improper match on w_val_unsatisfied error in HTTP

seancribbs:
On put requests, the `riak_kv_put_core` responds to the client layer with `{error, {w_val_unsatisfied, NumW, NumDW, W, DW}}` if either W or DW is not met. In `riak_kv_wm_object`, the error is attempted to be matched as `{error, {w_val_unsatisfied, Requested, Returned}}`, which causes the response to be 500 instead of 503.

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -1032,12 +1032,12 @@ handle_common_error(Reason, RD, Ctx) ->
                             [Returned, Requested]),
                         RD)),
                 Ctx};
-        {error, {w_val_unsatisfied, Requested, Returned}} ->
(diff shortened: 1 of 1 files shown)


russelldb:
:thumbsup: WFM (tested by changing `riak_kv_put_fsm` to always return `{error, {w_val_unsatisfied, 1, 1, 3 3}}`)

seancribbs:
@russelldb @rzezeski Any objection to merging to 1.2 as well?

russelldb:
No problem for me.

seancribbs:
Cherry-picked to 1.2, didn't want to inadvertantly pull in master commits.

PR #389, opened 2012-09-07
Make vnode worker pool size configurable.

kellymclaughlin:
The size was previously hardcoded to 10. Allow it to be configured using the `worker_pool_size` configuration parameter.


reiddraper:
Do we want to consider an overflow size parameter as well?

kellymclaughlin:
I think that's worth considering, but it requires a change in `riak_core` to make that happen so I'd say defer to a separate PR.

seancribbs:
This is a straightforward change. :+1:

PR #390, opened 2012-09-08
erlang:now() -> os:timestamp() in all the places it is safe

Vagabond:
There are 2 holdouts in kv_lru and kv_mrc_pipe that are less clear if
they're safe to change, so I left them.


jonmeredith:
+1 merge

PR #395, opened 2012-09-14
Add retry on eleveldb lock errors during open for up to 1 minute.

jonmeredith:
If a vnode crashes, the eleveldb NIF will close the leveldb database cleanly, writing out any pending data.  However, the vnode supervisor can restart the crashed vnode faster than the close process and trigger a lock error which fails the vnode start and takes riak down.

This patch adds a retry every two seconds, waiting up to a minute for the lock to clear.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -368,10 +368,31 @@ init_state(DataRoot, Config) ->
              config = FinalConfig }.
 
 %% @private
-open_db(State0) ->
+open_db(State) ->
+    RetriesLeft = app_helper:get_env(riak_kv, eleveldb_open_retries, 30),
+    open_db(State, max(1, RetriesLeft), undefined).
+
+open_db(_State0, 0, LastError) ->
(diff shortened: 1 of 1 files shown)


jonmeredith:
Here's an on the console way to play with it...

```
lager:set_loglevel(lager_console_backend, debug).
Bytes = crypto:rand_bytes(1000000).
{ok,C}=riak:local_client().
F = fun(X) -> C:put(riak_object:new(<<"b">>,<<"k">>,Bytes)), X(X) end. 
[spawn(fun() -> F(F) end) || I <- lists:seq(1,10)].
timer:sleep(10000),
rpc:multicall(erlang,
                 apply,
                 [fun() ->
                         [ 
                          exit(Pid, kill) ||
                              {_,_,Pid} <-
                                  riak_core_vnode_manager:all_vnodes(riak_kv_vnode),
                              is_pid(Pid)
                         ]
                  end,
                  []]).

```

reiddraper:
Nice tests. Looks good other than the two comments above.

PR #399, opened 2012-09-27
2I backpressure

kellymclaughlin:
Use the same mechanism used for key listing to provide backpressure for secondary index requests.


beerriot:
I think we'll have to do some capability negotiation and handling of both varieties of messages, or mixed-version clusters won't work.

kellymclaughlin:
Quite right. I will add that in.

beerriot:
I think that should do it. Tested with a mixed cluster, one HEAD node and a 1.2 node, and also with two HEAD nodes. Have a look, @kellymclaughlin?

PR #401, opened 2012-10-02
Quick fix to prevent mapred_test from hanging

kuenishi:
TODO: This cleanup is just a quick fix to run the whole test. Defining API dispatcher by application:set_env/3 may cause race conditions of mis-deregistering and application stop.

And this patch will cause 5 test failure; better than test hangs :'(

### DIFF
--- a/src/riak_kv_test_util.erl
+++ b/src/riak_kv_test_util.erl
@@ -242,14 +242,11 @@ dep_apps(Test, Extra) ->
                 application:set_env(lager, crash_log, Test ++ "/log/crash.log");
            (stop) ->
                 %% TODO: This cleanup is just a quick fix to run the whole test.
-		%% Defining API dispatcher by application:set_env/3 may cause
-		%% race conditions of mis-deregistering and application stop.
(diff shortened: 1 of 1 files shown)


src/riak_kv_test_util.erl
```diff
@@ -241,24 +241,22 @@ dep_apps(Test, Extra) ->
...
-                %% dispatch table are the ones registered by calling
-                %% riak_api_pb_service:register/1 in
-                %% riak_kv_app:start/2.
-                Services = riak_api_pb_service:dispatch_table(),
-                NewServices = lists:foldl(
-                                fun dict:erase/2,
-                                Services,
-                                lists:flatten([lists:seq(3,6), lists:seq(9,26)])),
-                application:set_env(riak_api, services, NewServices);
+                %% TODO: This cleanup is just a quick fix to run the whole test.
```
russelldb, line 244: Indentation looks incorrect here

russelldb:
Apart from the indentation, this does indeed stop tests from hanging. +1

PR #404, opened 2012-10-16
Spawn remote vnodes using start_link rather than start.

jonmeredith:
Mixed clusters do not have riak_kv_put_fsm:start()


kellymclaughlin:
+1 to merge

PR #405, opened 2012-10-17
Randoming kv_put forwardee node

lordnull:
Instead of always choosing the first node in the preferred node list, randomly choose a node from that list.


macshonle:
+1

PR #406, opened 2012-10-18
Make riak_client work on non-riak nodes after get/put FSM startup change

jonmeredith:
Resolves issue with running riak-admin test


jaredmorrow:
+1 via testing

PR #407, opened 2012-10-19
separate the socket handler and sink processes in MR endpoints

beerriot:
I think this is the correct fix for #366.

Prior to this commit, the process holding the HTTP/PB socket was also the process acting as the Pipe sink for MapReduce requests. This led to many problems, including overwhelming mailboxes with error logging, and extra unexpected messages that had to be ignored if the socket was kept open to handle other requests.

This commit creates a new process to act as the sink, and sets up a request-response interface between them. The sink process spools pipe output until the socket-handling process asks for them.

This removes our dependence on the undocumented `mochiweb_request_force_close` feature, which was used to forcibly close an HTTP connection after a MapReduce failure to avoid incorrect 400 errors later (introduced in #224)

I've made this PR against 1.2, in case we want to include it in a point release, but the patch should be easily rebase-able onto master.

Some why's (structure):
- _Why start these sinks under a supervisor?_ I agree, it seems like a little bit of overkill. Mostly, we want to make sure they're tied to the app's supervision tree. They could just be linked directly to the processes handling the sockets, but then those processes would have to trap exits in order to be able to provide error messages to their clients. However, the mapred modules are not the only handlers in those processes, so others would have to care about whether exits were trapped as well (or we'd have to be super careful about cleaning up). Putting the sinks under a supervisor also makes it super easy to check for leaks: `supervisor:which_children(riak_kv_mrc_sink_sup)`.
- _What's with the `self()`-messaging `#sink_part{}` in `riak_kv_pb_mapred`?_ The module has to operate asynchronously, in order to be able to watch for a timeout message. The design of `riak_api` only allows one message to be sent to the client per erlang message to the process. The `#rbpmapredresp{}` record only allows one phase's outputs per message. The `riak_kv_mrc_sink` fsm might send results for multiple phases in one message. Thus, we must break them down to allow multiple messages to be sent to the client.
- _If the point was to get rid of extra messages, what's with the speculative receives in the cleanup?_ It is still possible that a reply from `riak_kv_mrc_sink` might arrive after our timeout message arrives, for example. We can't let it slip. The difference between this speculative receiving and other attempts, is that we know that there is at most one message waiting, and that it will have been delivered before we check here, if it will be delivered at all.
- _Why am I still getting "fitting was gone before startup" messages in the log?_ This patch does not target those message. This one is aimed at the "unrecognized message" errors coming out of `riak_api`.

Some what's (testing, mostly):
- _What has been tested already?_ I've been running the Java client unit tests via `riak_test` against this branch, as they're good at finding these kinds of errors. There are loadgen tests in issues attached to #366 that should be tried as well.
- _How does this affect performance?_ I don't know. It does introduce an extra process for results to hop through, so copying them could make things slower. The `basho_bench` tests for mapred should be run.
- _Are there any compatibility concerns?_ Because of the buffering that `riak_kv_mrc_sink` is doing, it's possible that clients will once again see multiple results per streaming message, as they did before the transition to pipe. This _should not_ affect compatibility, but it's something to watch for. (It also might improve performance from some perspectives.)

### DIFF
--- a/include/riak_kv_mrc_sink.hrl
+++ b/include/riak_kv_mrc_sink.hrl
@@ -1,0 +1,9 @@
+%% used to communicate from riak_kv_mrc_sink to riak_kv_wm_mapred and
+%% riak_kv_pb_mapred
+-record(kv_mrc_sink,
+        {
+          ref :: reference(), % the pipe ref
+          results :: [{PhaseId::integer(), Result::term()}],
+          logs :: [{PhaseId::integer(), Message::term()}],
+          done :: boolean()
+        }).
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -111,6 +111,7 @@
          mapred/2,
          mapred/3,
          mapred_stream/1,
+         mapred_stream/2,
          send_inputs/2,
          send_inputs/3,
          send_inputs_async/2,
@@ -118,6 +119,7 @@
          collect_outputs/2,
          collect_outputs/3,
          group_outputs/2,
+         error_exists/1,
          mapred_plan/1,
          mapred_plan/2,
          compile_string/1,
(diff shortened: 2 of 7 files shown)


seancribbs:
So far, I love this refactoring. Would it simplify things for 1.3 to allow riak_api to reply with multiple protocol messages? The change would be straightforward, and would not break existing clients.

beerriot:
The `riak_kv_wm_link_walker` module should use this new sink process as well, to keep pipe messages from crashing mochiweb loops, just like `riak_kv_wm_mapred`.

beerriot:
@seancribbs Yes, being able to reply with a list of PB messages would eliminate the need for sending `#sink_part{}` messages to `self()` in this module. That would collapse four `process_stream/3` clauses into one, and remove one speculative receive from `cleanup_pipe/1`.

seancribbs:
@beerriot Great, I'll do that pronto.

beerriot:
I am closing this PR unmerged, in favor of #416, which is a rebase of this PR onto master with rewrites to take advantage of basho/riak_api#14.

Issue #290, opened 2012-02-23
riak_kv_pipe_get's failover has trouble with fast-input+slow-output rates


beerriot:
A problem was uncovered with `riak_kv_pipe_get`'s failover strategy.
## Description

When the KV vnode that the pipe worker talks to returns an error, "not
found" for example, the worker attempts to have another worker for
this fitting try on a different vnode by returning the
`forward_preflist` result.  The last worker in the fallback list is
then supposed to forward the error to the next phase if it also fails.

This strategy does not work when the kvget workers are unable to keep
up with the rate of incoming inputs.  The `forward_preflist` strategy
has to use a non-blocking enqueue to prevent workers in a fitting from
mutually blocking on one another.  If the workers' queues are already
full of other inputs, forwarding retry inputs to them will never
succeed.  When forwarding retry inputs fails, the worker does not get
a chance to send the error to the next phase; Pipe's internal code
sends an `error` trace message and drops the input instead.

This bug manifests itself as MapReduce queries failing with a
timeout-like error before their configured timeout, since both the
HTTP and PB mapreduce endpoints treat error trace messages as reasons
to terminate the query.  The issue was raised on the mailing list:
http://lists.basho.com/pipermail/riak-users_lists.basho.com/2012-February/007590.html
## Possible fixes

One quick and incomplete fix may be to simply set the `q_limit` much
higher on the `riak_kv_pipe_get` fittings in the `riak_kv_mrc_pipe`
module.  This would raise the level/rate of inputs necessary to
trigger the bug.  Raising the following `riak_kv_mrc_map` fitting's
`q_limit` may help as well.

One long-term fix is likely to change `riak_kv_pipe_get` such that it
does not use the `forward_preflist` return value.  Either it will need
to talk directly to the fallback KV vnodes without forwarding the
input, or it will need to be able to call
`riak_pipe_vnode_worker:recurse_input/5` directly (which will also
require passing the used-preflist to the worker) to be able to handle
with full-queue errors.

Another long-term fix might be to alter the behavior of
`forward_preflist` such that instead of producing a trace error, it
forwards some standard message to the downstream fitting.
## Reproduction

Below is a basic method for reproducing the problem.  The method is to
follow a `riak_kv_pipe_get` fitting with a fitting that takes forever
to process its first input.  This in turn causes the
`riak_kv_pipe_get` fitting's queues to back up.  The `nval`, `q_limit`
and `chashfun` parameters have been tuned to provide a minimal case.
[...]

PR #413, opened 2012-10-24
do not use forward_preflist in riak_kv_pipe_get

beerriot:
As described in basho/riak_kv#290, using forward_preflist makes a MapReduce
pipeline unnecessarily brittle when map or reduce phases are too slow to
keep up with kvget_map fittings. When kvget_map's queues get full,
forward_preflist fails in a way that the fitting is unable to handle.

This patch instead talks directly to another (non-local) KV vnode in the
object's primary preflist to attempt to find the value (instead of
moving the pipe input to a different pipe worker local to that other KV
vnode).

Talking directly to another non-local KV vnode has the downside of
pulling object data across the cluster's internal network if the object
is found. This should be a temporary situation that will correct itself
when the replica is repaired.

Talking to KV vnodes in series, instead of in parallel as a get_fsm
would do, will adversely effect tail latency, but no worse than waiting
for an opening on a different pipe vnode would have.

Directly using riak_kv_get_fsm would have been another option, but that
would also risk incurring other overhead like read-repair. While
read-repair has been a desired feature of MapReduce requests in the
past, introducing it as a surprising side effect of this fix seems
unwise.

### DIFF
--- a/src/riak_kv_pipe_get.erl
+++ b/src/riak_kv_pipe_get.erl
@@ -34,13 +34,16 @@
 %% partition number as the Pipe vnode owning this worker.  For this
 %% reason, it is important to use a `chashfun' for this fitting that
 %% gives the same answer as the consistent hashing function for the KV
-%% object.
+%% object. If the object is not found at the local KV vnode, each KV
+%% vnode in the remainder of the object's primary preflist is tried in
+%% sequence.
 %%
 %% If the object is found, the tuple `{ok, Object, Keydata}' is sent
 %% as output.  If an error occurs looking up the object, and the
 %% preflist has been exhausted, the tuple `{Error, {Bucket, Key},
-%% KeyData}' is sent as output (where `Error' is usually `not_found').
-%% The atom `undefined' is used as `KeyData' if none is specified.
+%% KeyData}' is sent as output (where `Error' is usually `{error,
+%% notfound}').  The atom `undefined' is used as `KeyData' if none is
+%% specified.
 
 -module(riak_kv_pipe_get).
 -behaviour(riak_pipe_vnode_worker).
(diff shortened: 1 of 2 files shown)


src/riak_kv_pipe_get.erl
```diff
@@ -79,39 +82,65 @@ init(Partition, FittingDetails) ->
 %% @doc Lookup the bucket/key pair on the Riak KV vnode, and send it
 %% downstream.
 -spec process(riak_kv_mrc_pipe:key_input(), boolean(), state())
-         -> {ok | forward_preflist | {error, term()}, state()}.
+         -> {ok | {error, term()}, state()}.
 process(Input, Last, #state{partition=Partition, fd=FittingDetails}=State) ->
+    %% assume local chashfun was used for initial attempt
+    case try_partition(Input, {Partition, node()}, FittingDetails) of
+        {error, _} when Last == false ->
```
russelldb, line 89: I don't know the meaning of `Last`, and I see that is the only subtle difference between this fun and `try_preflist/3`. My first response when reading the code is that `process/3` could just call `try_preflist/3` and you'd never have to remove the current node from the preflist, making this whole mod shorter and easier to read still.

So, I guess I'm asking what `Last` means, and why it means don't try other vnodes?

russelldb:
Using the code from https://github.com/basho/riak_kv/issues/290 the outcome is no longer a timeout with this code. 

Test just hung when I ran it without rebasing onto master. I'll look into that when I get back.

Code reads really well, documentation is clear, apart from my question above +1, once I get the test to run.

src/riak_kv_pipe_get.erl
beerriot, line 89: Agreed, I spent a while considering just that as well.

`Last` is a signal that Pipe gives workers to let them know that they are the final entry in the preflist for this `Input`, so no more attempts to process it will be made if this worker decides it's unable to do so itself. Put another way, returning `forward_preflist` when `Last==true` will raise a `preflist_exhausted` error.

I used `Last` here to short-circuit the expected case where failures at other Pipe vnodes means that we should expect those KV vnodes to be dead as well. Checking `Last` here means we can skip preflist generation and such.

We could skip the `Last` check and just call `try_preflist/2`, but we'd want to do a little mucking with the generated preflist anyway, because we always want to attempt to talk to the local KV vnode first, to avoid using network bandwidth if we don't have to. That's as simple as `[{P, node()}|RestPreflist]` between lines 114 and 115, though.

beerriot:
I had the test hang when I was in an awkward deps state as well. I think something changed about `riak_api` startup during `riak_kv` tests maybe. It worked again when I reset that dep, anyway.

src/riak_kv_pipe_get.erl
russelldb, line 89: Ok, understood.

beerriot:
I've just rebased this PR onto master to allow tests to be run after the messy merge in `tests/mapred_test.erl`.

russelldb:
Tests pass for me now. Big fat +1.

Issue #411, opened 2012-10-23
system_info(global_heaps_size) not supported in R15B02


vinoski:
The `riak_kv_stat:system_stats/0` function calls `erlang:system_info(global_heaps_size)` but that call results in a `badarg` exception under Erlang R15B02. The OTP team removed support for `global_heaps_size` in [commit 9cfeeb9b](https://github.com/erlang/otp/commit/9cfeeb9b7e66a3345cbc0e5b4b634cb5965635fc) in the erlang/otp repo.

russelldb:
Got it, will remove/substitute in a backwards compatible way for 1.3. Thanks.

russelldb:
See https://github.com/basho/riak_kv/pull/415 for proposed fix. Welcome review / comments.

russelldb:
Fixed with https://github.com/basho/riak_kv/pull/415

PR #415, opened 2012-10-25
Fix stats for r15b02

russelldb:
OTP team removed the erlang:system_info(global_heap_size) stat
and so riak-admin status and riak stats fail hard.

This commit wraps the call to said stat in a try catch. On < 15b02
you get global heapsize. Otherwise you get the atom 'deprecated'.

Addresses https://github.com/basho/riak_kv/issues/411


russelldb:
/stats - https://gist.github.com/29b7a1dcf831ac816104
riak-admin status https://gist.github.com/f4cc51ee9c224c005aea
jmx is unaffected, that stat is not exposed
snmp is unaffected, that stat is not exposed

PR #416, opened 2012-10-25
separate the socket handler and sink processes in MR endpoints

beerriot:
I think this is the correct fix for #366.

**Dependencies**: This PR depends on basho/riak_api#14

**History**: This is a rebase of #407 onto master, with rewrites to use new functionality introduced by basho/riak_api#14.

Prior to this commit, the process holding the HTTP/PB socket was also the process acting as the Pipe sink for MapReduce requests. This led to many problems, including overwhelming mailboxes with error logging, and extra unexpected messages that had to be ignored if the socket was kept open to handle other requests.

This commit creates a new process to act as the sink, and sets up a request-response interface between them. The sink process spools pipe output until the socket-handling process asks for them.

This removes our dependence on the undocumented `mochiweb_request_force_close` feature, which was used to forcibly close an HTTP connection after a MapReduce failure to avoid incorrect 400 errors later (introduced in #224)

Some why's (structure):
- _Why start these sinks under a supervisor?_ I agree, it seems like a little bit of overkill. Mostly, we want to make sure they're tied to the app's supervision tree. They could just be linked directly to the processes handling the sockets, but then those processes would have to trap exits in order to be able to provide error messages to their clients. However, the mapred modules are not the only handlers in those processes, so others would have to care about whether exits were trapped as well (or we'd have to be super careful about cleaning up). Putting the sinks under a supervisor also makes it super easy to check for leaks: `supervisor:which_children(riak_kv_mrc_sink_sup)`.
- _If the point was to get rid of extra messages, what's with the speculative receives in the cleanup?_ It is still possible that a reply from `riak_kv_mrc_sink` might arrive after our timeout message arrives, for example. We can't let it slip. The difference between this speculative receiving and other attempts, is that we know that there is at most one message waiting, and that it will have been delivered before we check here, if it will be delivered at all.
- _Why am I still getting "fitting was gone before startup" messages in the log?_ This patch does not target those message. This one is aimed at the "unrecognized message" errors coming out of `riak_api`.

Some what's (testing, mostly):
- _What has been tested already?_ I've been running the Java client unit tests via `riak_test` against this branch, as they're good at finding these kinds of errors. There are loadgen tests in issues attached to #366 that should be tried as well.
- _How does this affect performance?_ I don't know. It does introduce an extra process for results to hop through, so copying them could make things slower. The `basho_bench` tests for mapred should be run.
- _Are there any compatibility concerns?_ Because of the buffering that `riak_kv_mrc_sink` is doing, it's possible that clients will once again see multiple results per streaming message, as they did before the transition to pipe. This _should not_ affect compatibility, but it's something to watch for. (It also might improve performance from some perspectives.)

### DIFF
--- a/src/riak_kv_mrc_sink.erl
+++ b/src/riak_kv_mrc_sink.erl
@@ -175,10 +175,10 @@ handle_info(#pipe_log{ref=Ref, from=PhaseId, msg=Msg},
 handle_info(#pipe_eoi{ref=Ref},
             StateName, #state{ref=Ref}=State) ->
     info_response(StateName, State#state{done=true});
-handle_info({'DOWN', _, process, Pid, Reason}, _,
+handle_info({'DOWN', _, process, Pid, _Reason}, _,
             #state{owner=Pid}=State) ->
     %% exit as soon as the owner dies
-    {stop, Reason, State};
+    {stop, normal, State};
 handle_info({'DOWN', _, process, Pid, Reason}, _,
             #state{builder=Pid}=State) when Reason /= normal ->
     %% don't stop when the builder exits 'normal', because that's
--- a/src/riak_kv_mrc_sink.erl
+++ b/src/riak_kv_mrc_sink.erl
@@ -1,0 +1,234 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_mrc_sink: A simple process to act as a Pipe sink for
+%% MapReduce queries
+%%
+%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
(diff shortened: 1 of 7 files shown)


src/riak_kv_wm_mapred.erl
```diff
@@ -210,6 +221,14 @@ cleanup_sender({SenderPid, SenderRef}) ->
     erlang:demonitor(SenderRef, [flush]),
     exit(SenderPid, kill).
 
+cleanup_pipe({Sink,SinkMon}, Timer) ->
```
joedevivo, line 224: Same function, different module: https://github.com/basho/riak_kv/pull/416/files#L4R202

src/riak_kv_mrc_sink.erl
```diff
@@ -0,0 +1,234 @@
...
+handle_info(#pipe_log{ref=Ref, from=PhaseId, msg=Msg},
+            StateName, #state{ref=Ref, logs=Acc}=State) ->
+    info_response(StateName, State#state{logs=[{PhaseId, Msg}|Acc]});
+handle_info(#pipe_eoi{ref=Ref},
+            StateName, #state{ref=Ref}=State) ->
+    info_response(StateName, State#state{done=true});
+handle_info({'DOWN', _, process, Pid, Reason}, _,
+            #state{owner=Pid}=State) ->
+    %% exit as soon as the owner dies
+    {stop, Reason, State};
```
beerriot: This line and 187 should be `{stop, normal, State}`, not `{stop, Reason, State}`. We don't want the sink supervisor to print errors about this process dying, and all the other processes that care about when this one dies are monitoring it, not linking to it, so exiting `normal` is fine.

joedevivo:
Reviewed and tested repeatedly using the java_client tests.

I found no "Unexpected messages" in the logs, and the java unit tests passed on 10 consecutive runs.

+1

I'm Joe DeVivo, and I approve this pull request.

beerriot:
Merged to master as squashed commit ffc0d29

PR #367, opened 2012-10-26
riak 1.2rc1 - memory backend issue with 2i & $key/$bucket

seancribbs:
the memory backend returns different results from the leveldb backend when using $key with a range query where the range begin and range end are equal.

(see https://github.com/basho/riak-ruby-client/issues/46#issuecomment-7041161)

simple code to reproduce the difference can be found in [this gist](https://gist.github.com/3130528).


joedevivo:
+1

Ran all eunit and eqc tests on kv, and riak_test for secondary indexes for memory and leveldb. everything passes

PR #418, opened 2012-10-30
Add timings to vnode_get/put stats

russelldb:
Add per partition stats for get/put. Test as with the other stat PRs. Do some gets / puts against a node, attach to it and run:

```
[{Stat, folsom_metrics:get_metric_value(Stat)} || 
 Stat <- folsom_metrics:get_metrics(), 
             is_tuple(Stat),
             element(1, Stat) == riak_kv,
             element(2, Stat) == vnode_gets].  
```

Or

```
[{Stat, folsom_metrics:get_metric_value(Stat)} || 
 Stat <- folsom_metrics:get_metrics(), 
             is_tuple(Stat),
             element(1, Stat) == riak_kv,
             element(2, Stat) == vnode_puts].  
```

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -280,9 +280,11 @@ handle_command(?KV_PUT_REQ{bkey=BKey,
                            start_time=StartTime,
                            options=Options},
                Sender, State=#state{idx=Idx}) ->
+    StartTS = os:timestamp(),
     riak_kv_mapred_cache:eject(BKey),
     riak_core_vnode:reply(Sender, {w, Idx, ReqId}),
     UpdState = do_put(Sender, BKey,  Object, ReqId, StartTime, Options, State),
+    update_vnode_stats(vnode_put, Idx, StartTS),
     {noreply, UpdState};
 
 handle_command(?KV_GET_REQ{bkey=BKey,req_id=ReqId},Sender,State) ->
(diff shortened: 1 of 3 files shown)


coderoshi:
+1

src/riak_kv_vnode.erl
```diff
@@ -280,9 +280,11 @@ handle_command(?KV_PUT_REQ{bkey=BKey,
                            start_time=StartTime,
                            options=Options},
                Sender, State=#state{idx=Idx}) ->
+    StartTS = os:timestamp(),
     riak_kv_mapred_cache:eject(BKey),
```
beerriot, line 284: A note to PR #433: merging with this commit may create a conflict that tries to resurrect the call to `riak_kv_mapred_cache`.

beerriot:
It's quite interesting that per-vnode stats are not expunged when the vnode shuts down. I think that's a good thing, because you won't "just miss" a vnode's final stats after it hands off. And, given riak's on-demand vnode startup, "zero" spirals for non-running vnodes are technically correct.

It might be a good idea to head off some possible worry and questions by explaining this in our user docs, though: why some nodes show stats for all vnodes, others only some; why it's okay that many show zero forever; etc.

src/riak_kv_vnode.erl
russelldb, line 284: Noted, thank you.

russelldb:
Good point. I don't believe we have _any_ docs about stats. The release notes is one place that this can be mentioned. I'll make sure it is (once this PR is merged.)

That aside, is this a +1?

beerriot:
Sounds great, +1.

PR #419, opened 2012-10-30
add knob to disable referer check

joecaswell:


bsparrow435:
+1, should help some customers where the referrer thing gets in the way and they already have adequete security.

src/riak_kv_wm_utils.erl
```diff
@@ -235,7 +235,8 @@ any_to_bool(V) when is_boolean(V) ->
     V.
 
 is_forbidden(RD) ->
-    is_null_origin(RD) or not is_valid_referer(RD).
+    is_null_origin(RD) or 
+    (app_helper:get_env(riak_kv,secure_referer_check,true) and not is_valid_referer(RD)).
```
reiddraper, line 239: pretty sure you mean `orelse` and `andalso` here and not `or` and `and`

joecaswell, line 239: You are absolutely correct.

Do you have an opinion on whether this setting should also disable the null
origin check?

reiddraper, line 239: I'd lean toward yes.

PR #420, opened 2012-10-30
use short-circuit boolean in refercheck setting

joecaswell:


reiddraper:
Looks good. What do you think about adding in a default in the app.config for this, so that it's easier for people to find?

joecaswell:
That is an excellent idea

reiddraper:
I was thinking in the [app.config template](https://github.com/basho/riak/blob/master/rel/files/app.config), not in app.src.

reiddraper:
had an offline convo, +1 to merge

PR #421, opened 2012-11-02
Active Anti-Entropy

jtuple:
Had trouble merging and rebasing, so I ended up squashing into a single commit. So, here's AAE in all it's a glory, as a single super commit. Given how much this has been rewritten/reorganized, the old commit history doesn't help much anyway.

I'm extending/polishing my riak_tests some more, but will have thoses pushed soon and linked to this pull-request.

Hopefully all documentation and included -specs make this straightforward to review, despite being a massive pull-request. Also, everything from my AAE screencast that I previously distributed is still valid:
http://coffee.jtuple.com/video/AAE.html

I'm assuming @rzezeski and @Vagabond will be the two reviewing this, but others can jump on as well. However, we should have at least 2 reviewers on this one.

AAE is currently disabled by default, you'll need to enable it in the riak_kv section in app.config.

``` erlang
{anti_entropy, {on, []}},
%% or
{anti_entropy, {on, [debug]}},
%% or
{anti_entropy, {manual, []}},
%% or
{anti_entropy, {manual, [debug]}},
```

Also, additional settings you can set/change as desired (times are in milliseconds):

``` erlang
            {anti_entropy_timeout, 20000},
            {anti_entropy_tick, 1000},
            {anti_entropy_concurrency, 2},
            {anti_entropy_expire, 10000},
```

(cabal: basho/cabal#1)

---

Active anti-entropy is a process through which multiple replicas periodically sync with each other and automatically repair any keys that are missing or have divergent data. For example, a user could delete all data from a node and Riak would automatically regenerate the data from the other replicas.

This implementation uses hash trees / Merkle trees to perform lightweight key exchange, with work proportional to the number of divergent keys rather than the size of the overall keyspace.

This implementation meets several design goals:
1. The underlying hash trees are on-disk structures, and can scale to billions+ keys. This is in contrast to in-memory trees that require significant RAM to support massive keyspaces.
2. The underlying hash trees are persistent. Riak nodes can be restarted without fear of hash tree data being lost and needing to be rebuilt.
3. As new data is written to Riak, the hash trees associated with the various partitions are kept up-to-date. Each write in Riak triggers an asynchronous write to one or more hash trees. Combined with 2, this enables trees to be built once through a scan over existing data, and then maintained in real-time. In reality, trees are expired over time and rebuilt to ensure the hash tree and backend data stay in sync, and also to identify bit rot / disk failure.
4. The entire implementation is designed to be non-blocking. For example, a snapshot of a hash tree is generated before performing an exchange with other replicas, therefore allowing concurrent inserts of new key/hash pairs as new writes occur.

The current implementation triggers read repair for each key difference identified through the hash exchange. This is a reasonable approach as read repair is stable, production-tested mechanism in Riak. However, the read repair approach leads to slower replica repair in cases where there are a large number of key differences. This is an area for future improvement.


src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+            end
+    end.
+
+-spec all_pairwise_exchanges(index(), riak_core_ring())
+                            -> [exchange()].
+all_pairwise_exchanges(Index, Ring) ->
+    LocalIndexN = riak_kv_util:responsible_preflists(Index, Ring),
+    Sibs = riak_kv_util:preflist_siblings(Index),
+    lists:flatmap(
+      fun(RemoteIdx) when RemoteIdx == Index ->
```
rzezeski: Why not just have `preflist_siblings` not include the index itself?  It removes this filter case, which confused me the first time I reviewed this code.  It seems to me an index shouldn't be a sibling to itself.

src/riak_kv_util.erl
```diff
@@ -157,6 +165,92 @@ mapred_system() ->
...
+                          responsible_preflists_n(RevIndices, N)
+                  end, AllN).
+
+-spec responsible_preflists_n([index()], pos_integer()) -> [index_n()].
+responsible_preflists_n(RevIndices, N) ->
+    {Pred, _} = lists:split(N, RevIndices),
+    [{Idx, N} || Idx <- lists:reverse(Pred)].
+
+-spec determine_max_n(riak_core_ring()) -> pos_integer().
+determine_max_n(Ring) ->
```
rzezeski, line 230: How about calling `determine_all_n` and using `lists:max`?

src/riak_kv_sup.erl
```diff
@@ -118,6 +123,7 @@ init([]) ->
         KeysFsmSup,
         IndexFsmSup,
         LegacyKeysFsmSup,
+        ?IF(EntropyEnabled, EntropyManager, []),
```
rzezeski: So if a user decides they want to enable entropy now they need to restart Riak for that to take effect.  Should we add a function to entropy mgr to allow dynamic adding of child?

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+%% @doc Acquire an exchange concurrency lock if available, and associate
+%%      the lock with the calling process.
+-spec get_lock(any()) -> ok | max_concurrency.
+get_lock(Type) ->
+    get_lock(Type, self()).
+
+%% @doc Acquire an exchange concurrency lock if available, and associate
+%%      the lock with the calling process.
```
rzezeski: s/with the calling process/with the `Pid'/

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+set_debug(Enabled) ->
+    Modules = [riak_kv_index_hashtree,
+               riak_kv_entropy_manager,
+               riak_kv_exchange_fsm],
+    case Enabled of
+        true ->
+            [lager:trace_console([{module, Mod}]) || Mod <- Modules];
+        false ->
+            [begin
+                 {ok, Trace} = lager:trace_console([{module, Mod}]),
```
rzezeski, line 144: I guess these have to be started first or else `lager:stop_trace` will error?

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+-spec settings() -> {boolean(), proplists:proplist()}.
+settings() ->
+    case app_helper:get_env(riak_kv, anti_entropy, {off, []}) of
+        {on, Opts} ->
+            {true, Opts};
+        {manual, Opts} ->
```
rzezeski: I think allowing `manual` as an entry for the first element in the tuple is a bit confusing.  Entropy mgr has the notion of automatic or manual.  That makes sense.  Either mode requires that anti entropy be "on."  To me it makes more sense if the first atom is `on | off` and manual mode is invoked by include the option `manual` or `manual_mode`.  As it stands now the user could invoke manual mode in two ways `{on, [manual]}` or `{manual, []}`.  Why have two different ways of putting the system in the same state?

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+
+%%%===================================================================
+%%% gen_server callbacks
+%%%===================================================================
+
+-spec init([]) -> {'ok',state()}.
+init([]) ->
+    schedule_tick(),
+    {_, Opts} = settings(),
+    Mode = case lists:member(manual, Opts) of
```
rzezeski: Nitpicky...why not use proplists for the options to make the intention more clear.

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+                       {index(), index_n()} |
+                       {index(), index(), index_n()}, state()) -> state().
+trigger_exchange(E={Index, _RemoteIdx, _IndexN}, State) ->
+    %% Verify that the exchange is valid
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    Exchanges = all_pairwise_exchanges(Index, Ring),
+    case lists:member(E, Exchanges) of
+        true ->
+            trigger_exchanges([E], State);
+        false ->
```
rzezeski, line 499: I feel like we shouldn't be quiet about invalid exchanges in
`trigger_exchange`.  I assume that this can only happen when a user is
triggering exchange manually.  In that case it would be helpful to log
something to let the user know it's invalid.  Also, it would be a good
safeguard just in case the situation ever arises where invalid
exchanges are trigger during automatic mode.

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+    Trees2 = orddict:from_list(Trees ++ L),
+    State2 = State#state{trees=Trees2},
+    State3 = lists:foldl(fun({Idx,Pid}, StateAcc) ->
+                                 monitor(process, Pid),
+                                 add_index_exchanges(Idx, StateAcc)
+                         end, State2, L),
+    State3.
+
+
+-spec do_get_lock(any(),pid(),state()) -> {'max_concurrency',state()}.
```
rzezeski: s/'max_concurrency'/'ok | max_concurrency/

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+%% queried.
+-spec reload_hashtrees(state()) -> state().
+reload_hashtrees(State=#state{trees=Trees}) ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    Indices = riak_core_ring:my_indices(Ring),
+    Existing = dict:from_list(Trees),
+    MissingIdx = [Idx || Idx <- Indices,
+                         not dict:is_key(Idx, Existing)],
+    L = [{Idx, Pid} || Idx <- MissingIdx,
+                       {ok,Pid} <- [riak_kv_vnode:hashtree_pid(Idx)],
```
rzezeski: If a vnode is backed up then `riak_kv_vnode:hashtree_pid` could take a
long time to respond.  This would block `reload_hashtrees` which would
also block the tick.  Do we want to block the entropy manager because
of one backed-up vnode?  Why not use a small timeout and if the vnode
doesn't return in time then it will be tried again on the next tick.

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,547 @@
...
+-spec all_exchanges(node(), riak_core_ring(), state())
+                   -> [exchange()].
+all_exchanges(_Node, Ring, #state{trees=Trees}) ->
+    Indices = orddict:fetch_keys(Trees),
+    lists:flatmap(fun(Index) ->
+                          all_pairwise_exchanges(Index, Ring)
+                  end, Indices).
+
+-spec add_index_exchanges(index(), state()) -> state().
+add_index_exchanges(_Index, State) when State#state.mode == manual ->
```
rzezeski, line 588: Why have `add_index_exchanges` and `trigger_exchange`?  I feel like
`add_index_exchange` makes more sense because exchanges are being
added to the exchange queue (so perhaps `queue_exchange` would make
most sense).  The word "trigger" makes me thing it's happening
immediately.

rzezeski:
I think we are going to need to give AAE more sensible defaults for when it is turned on.  I ran read/update benchmark against a local devrel and turning AAE on with the defaults caused throughput to drop by more than half.  95th gets shot up by a factor of 3 and updates by a factor of 6.  There was also more variance with AAE on than off.  I'm going to do some runs while playing with the various knobs and see what I find.

EDIT: This was against a healthy cluster.  No attempt was made to actually cause divergence.  I simply wanted to see how the overhead introduced by AAE would affect a get/update load.

rzezeski:
If a node is down I see constant chatter about timeout on exchange.

```
012-11-06 18:18:03.876 [info] <0.5232.6>@riak_kv_exchange_fsm:do_timeout:255 Timeout during exchange between (local) {1004782375664995756265033322492444576013453623296,'dev1@127.0.0.1'} and (remote) {1050454301831586472458898473514828420377701515264,'dev3@127.0.0.1'}, (preflist) {1004782375664995756265033322492444576013453623296,3}
2012-11-06 18:18:03.876 [info] <0.5238.6>@riak_kv_exchange_fsm:do_timeout:255 Timeout during exchange between (local) {1187470080331358621040493926581979953470445191168,'dev1@127.0.0.1'} and (remote) {1233142006497949337234359077604363797834693083136,'dev3@127.0.0.1'}, (preflist) {1187470080331358621040493926581979953470445191168,3}
```

rzezeski:
> Repaired 17 keys through active anti-entropy

I'm watching these fly by with no idea what is being repaired.  It would be nice to know the nodes/index/{index,n} involved

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+-spec update(index_n(), pid()) -> ok | not_responsible.
+update(Id, Tree) ->
+    put(calling, ?LINE),
+    gen_server:call(Tree, {update_tree, Id}, infinity).
+
+%% @doc Return a hash bucket from the tree identified by the given tree id
+%%      that is managed by the provided index_hashtree.
+-spec exchange_bucket(index_n(), integer(), integer(), pid()) -> orddict().
+exchange_bucket(Id, Level, Bucket, Tree) ->
+    put(calling, ?LINE),
```
rzezeski: What is this proc dict entry for?  Something to do with logging I take it?

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+-spec destroy(pid()) -> ok.
+destroy(Tree) ->
+    gen_server:call(Tree, destroy).
+
+%%%===================================================================
+%%% gen_server callbacks
+%%%===================================================================
+
+init([Index, IndexNs]) ->
+    Root = "data/anti",
```
rzezeski: This needs to be configurable like our other data dirs.

src/riak_kv_vnode.erl
```diff
@@ -258,7 +332,8 @@ init([Index]) ->
                 true ->
                     %% Create worker pool initialization tuple
                     FoldWorkerPool = {pool, riak_kv_worker, WorkerPoolSize, []},
-                    {ok, State, [FoldWorkerPool]};
+                    State2 = maybe_create_hashtrees(State),
```
rzezeski, line 342: The hashtree is linked to the vnode.  If the hashtree dies then the
vnode dies.  This seems wrong.  I feel like the HTs should be under
their own supervisor.  The vnode would tell the supervisor to start a
new child and then restart it as needed.

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+    Path = filename:join(Root, integer_to_list(Index)),
+
+    State = #state{index=Index,
+                   trees=orddict:new(),
+                   built=false,
+                   path=Path},
+    State2 = init_trees(IndexNs, State),
+    {ok, State2}.
+
+handle_call({new_tree, Id}, _From, State) ->
```
rzezeski, line 213: This looks like an unused code path.

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+                      hashtree:new({Index,IdBin}, [{segment_path, Path}]);
+                  [{_,Other}|_] ->
+                      hashtree:new({Index,IdBin}, Other)
+              end,
+    Trees2 = orddict:store(Id, NewTree, Trees),
+    State#state{trees=Trees2}.
+
+-spec do_get_lock(any(), pid(), state()) -> {not_built | ok | already_locked, state()}.
+do_get_lock(_, _, State) when State#state.built /= true ->
+    lager:debug("Not built: ~p :: ~p", [State#state.index, State#state.built]),
```
rzezeski, line 410: The `state.built` in this msg is redundant since it starts with `Not built...`.

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+    {Reply, State2} = do_get_lock(Type, Pid, State),
+    {reply, Reply, State2};
+
+handle_call({start_exchange_remote, FsmPid, _IndexN}, _From, State) ->
+    case riak_kv_entropy_manager:get_lock(exchange_remote, FsmPid) of
+        max_concurrency ->
+            {reply, max_concurrency, State};
+        ok ->
+            case do_get_lock(remote_fsm, FsmPid, State) of
+                {ok, State2} ->
```
rzezeski: Why bother doing an explicit match against `ok` here when the 2nd match does the same thing for any reply.  I.e. nothing different is done between these two matches.

``` erlang
{Reply, State2} = do_get_lock(...),
{reply, Reply, State2}
```

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+maybe_build(State=#state{built=false}) ->
+    Self = self(),
+    Pid = spawn_link(fun() ->
+                             case riak_kv_entropy_manager:get_lock(build) of
+                                 max_concurrency ->
+                                     gen_server:cast(Self, build_failed);
+                                 ok ->
+                                     build_or_rehash(Self, State)
+                             end
+                     end),
```
rzezeski, line 602: The building of the hashtree in `maybe_build` links the building
process to the index hashtree.  If this build crashes then the index
hashtree crashes.  Since the index hashtree is linked to the vnode it
then will also cause the vnode to crash.  Do we want this behavior?
Also, since this is a spawn-linked process it does not sit under a
supervisor.  Have you considered having a simple one-for-one that
these pids can live under?  That would make it easier to see all the
hashtree builds that are running and kill them if necessary.

src/riak_kv_index_hashtree.erl
```diff
@@ -0,0 +1,592 @@
...
+%% the key does not already exist in the tree. This allows real-time updates
+%% to the hashtree to occur concurrently with the fold. For example, if an
+%% incoming write triggers a real-time insert of a key/hash pair for an object
+%% before the fold reaches the now out-of-date version of the object, the old
+%% key/hash pair will be ignored.
+-spec fold_keys(index(), pid()) -> ok.
+fold_keys(Partition, Tree) ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    Req = ?FOLD_REQ{foldfun=fun(BKey={Bucket,Key}, RObj, _) ->
+                                    IndexN = riak_kv_util:get_index_n({Bucket, Key}, Ring),
```
rzezeski, line 379: I think the `IndexN` calculation could possibly be made more efficient
by making use of the techniques found in `riak_core_repair`.  It would
avoid a `div` and `rem` operation per key which could maybe make a
difference in the tight loop of vnode folds.  Anyways, I realize the
focus right now is to make it work first, worry about optimization
later.  If I have time I might try taking a closer look at this.

src/hashtree.erl
```diff
@@ -0,0 +1,1031 @@
...
+new_segment_store(Opts, State) ->
+    DataDir = case proplists:get_value(segment_path, Opts) of
+                  undefined ->
+                      Root = "/tmp/anti/level",
+                      <<P:128/integer>> = crypto:md5(term_to_binary(erlang:now())),
+                      filename:join(Root, integer_to_list(P));
+                  SegmentPath ->
+                      SegmentPath
+              end,
+    Options = [{create_if_missing, true},
```
rzezeski, line 433: In the future we may want to merge these with `Opts` in case we want to tweak the level options used for AAE.

src/hashtree.erl
```diff
@@ -0,0 +1,1031 @@
...
+        false ->
+            set_disk_bucket(Level, Bucket, Val, State)
+    end.
+
+-spec new_segment_store(proplist(), hashtree()) -> hashtree().
+new_segment_store(Opts, State) ->
+    DataDir = case proplists:get_value(segment_path, Opts) of
+                  undefined ->
+                      Root = "/tmp/anti/level",
+                      <<P:128/integer>> = crypto:md5(term_to_binary(erlang:now())),
```
rzezeski, line 428: I realize this code path isn't hit by AAE but should it instead default to using the `Index` as the segment path?

src/hashtree.erl
```diff
@@ -0,0 +1,1031 @@
...
+    State2.
+
+-spec destroy(hashtree()) -> hashtree().
+destroy(State) ->
+    eleveldb:close(State#state.ref),
+    ok = eleveldb:destroy(State#state.path, []),
+    State.
+
+-spec insert(binary(), binary(), hashtree()) -> hashtree().
+insert(Key, ObjHash, State) ->
```
rzezeski, line 246: nitpick: The hash could be of anything, doesn't have to be an `ObjHash`.  Naming is hard.

jtuple:
Pretty sure I addressed all raised issues, plus several other things that came up along the way. Organized the work such that this should be easy to review commit-by-commit.

rzezeski:
We still have both `add_index_exchanges` and `enqueue_exchanges` which are doing the same work.  It seems we could replace the call to `add_index_excahnges` with `enqueue_exchanges`.  Not a blocker, but thought I'd throw it out there.

NOTE: Well `add_index_exchanges` checks the mode but I don't see why that couldn't be added to `enqueue_exchanges`.

src/riak_kv_entropy_manager.erl
```diff
@@ -0,0 +1,671 @@
...
+-type index_n() :: {index(), pos_integer()}.
+-type vnode() :: {index(), node()}.
+-type exchange() :: {index(), index(), index_n()}.
+-type riak_core_ring() :: riak_core_ring:riak_core_ring().
+
+-record(state, {mode           = automatic :: automatic | manual,
+                trees          = []        :: [{index(), pid()}],
+                tree_queue     = []        :: [{index(), pid()}],
+                locks          = []        :: [{pid(), reference()}],
+                build_tokens   = 0         :: pos_integer(),
```
rzezeski: Dialyzer bitches about this.  Use non_neg_integer or integer.

rzezeski:
When we talked in the back-channel you said this would be enabled by default.  I think it should be.  We need a Riak PR for that I guess.

rzezeski:
+1 to merge
- Reviewed code multiple times
- Compiled, ran unit tests, ran quick check
- Used new interceptor code to try emulate dropped puts and slow tree updates, AAE still worked
- Ran almost all riak tests with AAE enabled and they all passed

jtuple:
We'll want to do a separate pull-request to Riak to bring AAE app.config defaults into play. Take built-in timer defaults and put them into app.config as documentation, plus (obviously) enable AAE default (which is already in riak_kv.app so it'll be enabled if missing), plus anti_entropy_data_dir. Right now, we use fallback:

```
[warning] Config riak_kv/anti_entropy_data_dir is missing. Defaulting to: "./data/anti_entropy
```

All easy to change in app.config though. Might even consider adding them riak_kv.app as well, but that's up for discussion. Merging now.

docs/hashtree.md
```diff
@@ -0,0 +1,64 @@
...
+updated bottom-up as necessary. Only paths along which hashes have
+been changed are re-computed.
+
+The current implementation uses LevelDB for the heavy lifting. Rather
+than reading/writing the on-disk segments as a unit, `(key, hash)`
+pairs are written to LevelDB as simple key-value pairs. The LevelDB
+key written is the binary `<<$s, SegmentId:64/integer,
+Key/binary>>`. Thus, inserting a new key-value hash is nothing more
+than a single LevelDB write. Likewise, key-hash pairs for a segment
+are laided on sequentially on-disk based on key sorting. An in-memory
```
DeadZen, line 26: This should read... "are laid sequentially on-disk"

docs/hashtree.md
```diff
@@ -0,0 +1,64 @@
...
+bitvector is used to track dirty segments, although a `gb_sets` was
+formerly used.
+
+When updating the segment hashes, a LevelDB iterator is used to access
+the segment keys in-order. The iterator seeks to the beginning of the
+segment and then iterators through all of the key-hash pairs. As an
```
DeadZen, line 32: This should read... "and then iterates"

docs/hashtree.md
```diff
@@ -0,0 +1,64 @@
...
+optimization, the iteration process is designed to read in multiple
+segments when possible. For example, if the list of dirty segments was
+`[1, 2, 3, 5, 6, 10]`, the code will seek an iterator to the beginning
+of segment 1, iterator through all of its keys, compute the
```
DeadZen, line 36: This should read... "iterate through"

docs/hashtree.md
```diff
@@ -0,0 +1,64 @@
...
+(all writes), followed by an in-order traversal of the LevelDB
+database (all reads).
+
+Trees are compared using standard hash tree approach, comparing the
+hash at each level, and recursing to the next level down when
+different. After reaching the leaf nodes, any differing hashes results
+in a key exchange of the keys in the associated differing segments.
+
+By default, the hash tree itself is entirely in-memory. However, the
+code provides a `MEM_LEVEL` paramemter that specifics that levels
```
DeadZen, line 52: Typo on "parameter", this should read... "parameter that specifies that"

docs/hashtree.md
```diff
@@ -0,0 +1,64 @@
...
+greater than the parameter should be stored on-disk instead. These
+buckets are simply stored on disk in the same LevelDB structure as
+`{$b, Level, Bucket} -> orddict(Key, Hash)}` objects.
+
+The default settings use `1024*1024` segments with a tree width of
+`1024`. Thus, the resulting tree is only 3 levels deep. And there
+are only `1+1024+1024*1024` hashs stored in memory -- so, a few
```
DeadZen, line 59: Typo on "hashes"

PR #357, opened 2012-06-28
Add bucket property deletion to Erlang client and HTTP API.

jerith:
This is an implementation of the `riak_kv` functionality required for basho/riak_core#197. It depends on basho/riak_core#202.

I have implement and tested the HTTP API, but haven't done anything on the PB side yet.

PR #422, opened 2012-11-02
reset bucket props via http

jrwest:
This PR exposes the functionality in https://github.com/basho/riak_core/pull/248 over the version 2 HTTP API.

Depends on: https://github.com/basho/riak_core/pull/248
Original Implementation: https://github.com/basho/riak_kv/pull/357

### DIFF
--- a/src/riak_kv_wm_props.erl
+++ b/src/riak_kv_wm_props.erl
@@ -44,6 +44,10 @@
 %%     {"props":{Prop:Val}}
 %%   Where the "props" object takes the same form as returned from
 %%   a GET of the same resource.
+%% 
+%% DELETE /buckets/Bucket/props
+%%   Reset bucket properties back to the default settings
+%%   not supporteed by the OLD API
 
 -module(riak_kv_wm_props).
 
@@ -59,7 +63,8 @@
          content_types_accepted/2,
          produce_bucket_body/2,
          accept_bucket_body/2,
-         get_bucket_props_json/2
+         get_bucket_props_json/2,
+         delete_resource/2
         ]).
 
 %% @type context() = term()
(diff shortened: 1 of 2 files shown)


src/riak_kv_wm_props.erl
```diff
@@ -44,6 +44,10 @@
 %%     {"props":{Prop:Val}}
 %%   Where the "props" object takes the same form as returned from
 %%   a GET of the same resource.
+%% 
+%% DELETE /buckets/Bucket/props
+%%   Reset bucket properties back to the default settings
+%%   not supporteed by the OLD API
```
seancribbs: Typo

src/riak_kv_wm_props.erl
```diff
@@ -208,6 +217,12 @@ accept_bucket_body(RD, Ctx=#ctx{bucket=B, client=C, bucketprops=Props}) ->
             {{halt, 400}, RD2, Ctx}
     end.
 
+%% @spec delete_resource(reqdata(), context()) -> {boolean, reqdata(), context()}
```
seancribbs, line 220: FYI I realize it's out-of-scope for this PR but in all other places we tend to use the newer `-spec` rather than the `%% @spec` form. The origins of this file are old enough to have used the comment form, so let's consider that a janitorial task for later.

jrwest, line 220: yea I noticed @rzezeski made a similar comment on the original implementation. Figured it best to not change it over as part of this PR like you said.

jrwest:
merging (+1 [here](https://github.com/basho/riak_core/issues/197#issuecomment-10336715) for reference since gh wasn't able to make the association)

PR #426, opened 2012-11-03
Since "Try again" is not relevant to a self-join, make that error explicit

macintux:
Tiny, tiny patch.  Noticed that when trying to join a node to itself that the error message was unhelpful.


seancribbs:
Thanks! Merging.

PR #428, opened 2012-11-06
Use 'tree' style stat names, and riak_core_stat_q

russelldb:
Use riak core stat query for riak_kv_stat get_stats
Add new read repair dimensioned stat aggregations to stat blob

Depends on riak_core pr - https://github.com/basho/riak_core/pull/252

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -152,17 +152,16 @@ do_per_index(Op, Idx, USecs) ->
 do_get_bucket(false, _) ->
     ok;
 do_get_bucket(true, {Bucket, Microsecs, Stages, NumSiblings, ObjSize}=Args) ->
-    BucketAtom = binary_to_atom(Bucket, latin1),
-    case (catch folsom_metrics:notify_existing_metric({?APP, node, gets, BucketAtom}, 1, spiral)) of
+    case (catch folsom_metrics:notify_existing_metric({?APP, node, gets, Bucket}, 1, spiral)) of
         ok ->
-            [folsom_metrics:notify_existing_metric({?APP, node, gets, Dimension, BucketAtom}, Arg, histogram)
+            [folsom_metrics:notify_existing_metric({?APP, node, gets, Dimension, Bucket}, Arg, histogram)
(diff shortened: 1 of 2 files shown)


src/riak_kv_stat_bc.erl
```diff
@@ -0,0 +1,384 @@
...
+                        orddict:store(Key, [{count, C+C0}, {one, O+O0}], Acc)
+                end,
+                orddict:new(),
+                Stats).
+
+key_from_fields(Name, Fields) ->
+    Key = [element(N, Name) || N <- Fields],
+    join(Key).
+
+recursive_join(BaseName, Aggregates) ->
```
beerriot: The name `recursive_join` was a little vague for me. Maybe a comment explaining how it's flattening the nested stat tree, creating a flat list of stats with names prepended by `BaseName_`?

src/riak_kv_stat_bc.erl
```diff
@@ -0,0 +1,384 @@
...
+    L = orddict:fold(fun(K, V, Acc) when not is_list(V) ->
+                             [{join([BaseName, K]), V}|Acc];
+                        (K, V, Acc)  ->
+                             [recursive_join(join([BaseName, K]), V)|Acc]
+                     end,
+                     [],
+                     Aggregates),
+    lists:flatten(L).
```
beerriot: It seems like it shouldn't be necessary to call `lists:flatten/1` on every recursion of this function … it should only be needed once at the top level (in `aggregate/3`?).

src/riak_kv_stat.erl
```diff
@@ -443,71 +263,16 @@ register_stat(Name, histogram) ->
     {SampleType, SampleArgs} = get_sample_type(Name),
     folsom_metrics:new_histogram(Name, SampleType, SampleArgs).
 
+%% @doc the histogram sample type may be set in app.config
+%% use key `stat_sample_type` in the `riak_kv` section. Or the
```
beerriot: These are markdown-style literal escapes, not edoc-style.

russelldb:
Pushed a commit that addresses review comments.

beerriot:
Looks great, +1!

PR #429, opened 2012-11-06
Apply backpressure from the MR sink

beerriot:
This is a re-opening of #425. That PR closed when I deleted the branch for #416, which 425 targeted. This now targets master.

This PR uses the new ability of basho/riak_pipe#59 to provide backpressure to the MR pipe in the new MR sink introduced by #416. This PR is targeted at the bwf-sep-sink-2 branch to more clearly demonstrate the changes made here until #416 is merged to master.

In addition to using the fsm_sync sink type to synchronously receive pipe results/logs, the sink now has a capped-size buffer. Once that buffer is full, the sink delays acks to result/log senders until the sink's owner empties its buffer. This is to keep the problem of the expanding mailbox from becoming the problem of the expanding heap. The size of this buffer is configurable globally as the riak_kv app env mrc_sink_buffer or as a startup option for the FSM, buffer.

The changes to src/riak_kv_mrc_pipe.erl look more dramatic than they really are because enforcing the use of riak_kv_mrc_sink in all MR pipelines has cleaned up the logic of output collection and sorting.

This branch has not yet been tested for performance changes. It is likely that latency will take a hit, due to the extra synchronicity. The safety of not blowing up memory consumption in a sink process is likely desirable enough to counter worries. There is also work underway (in the bwf-pmi branch) to enable delivery of multiple outputs in a single message.

There are no "capabilities" or other "upgrade" facilities included here. Nodes running earlier versions of Riak will simply ignore the sink_type option, and send regular messages, as they do now, to the MR sink. This is fine - the sink will accept them, even though it cannot provide backpressure to those senders. Nodes running the new version will similarly send regular messages if they are working for a Pipe that was started by an older node (and therefore had not set sink_type).

### DIFF
--- a/src/riak_kv_pb_mapred.erl
+++ b/src/riak_kv_pb_mapred.erl
@@ -52,9 +52,8 @@
                 req_ctx,
                 client}).
 
--record(pipe_ctx, {pipe,     % pipe handling mapred request
-                   ref,      % easier-access ref/reqid
-                   timer,    % ref() for timeout send_after
+-record(pipe_ctx, {ref,      % easier-access ref/reqid
+                   mrc,      % #mrc_ctx{}
                    sender,   % {pid(), monitor()} of process sending inputs
                    sink,     % {pid(), monitor()} of process collecting outputs
                    has_mr_query}). % true if the request contains a query.
@@ -91,7 +90,7 @@ process_stream(#kv_mrc_sink{ref=ReqId,
                ReqId,
                State=#state{req=#rpbmapredreq{},
                             req_ctx=#pipe_ctx{ref=ReqId,
-                                              sink={Sink,_}}=PipeCtx}) ->
+                                              mrc=Mrc}=PipeCtx}) ->
     case riak_kv_mrc_pipe:error_exists(Logs) of
         false ->
             case msgs_for_results(Results, State) of
@@ -107,6 +106,7 @@ process_stream(#kv_mrc_sink{ref=ReqId,
                              Msgs++[#rpbmapredresp{done=1}],
                              clear_state_req(State)};
                        true ->
+                            {Sink, _} = Mrc#mrc_ctx.sink,
                             riak_kv_mrc_sink:next(Sink),
                             {reply, Msgs, State}
                     end;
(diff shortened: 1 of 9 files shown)


src/riak_kv_pb_mapred.erl
```diff
@@ -52,9 +52,8 @@
                 req_ctx,
                 client}).
 
--record(pipe_ctx, {pipe,     % pipe handling mapred request
-                   ref,      % easier-access ref/reqid
-                   timer,    % ref() for timeout send_after
+-record(pipe_ctx, {ref,      % easier-access ref/reqid
```
joedevivo, line 55: ref, sink, and sender are all included in mrc_ctx{}.

src/riak_kv_wm_mapred.erl
```diff
@@ -180,73 +180,27 @@ pipe_mapred(RD,
...
     end.
 
-%% Destroying the pipe via riak_pipe_builder:destroy/1 does not kill
-%% the sender immediately, because it causes the builder to exit with
-%% reason `normal', so no exit signal is sent. The sender will
-%% eventually receive `worker_startup_error's from vnodes that can no
-%% longer find the fittings, but to help the process along, we kill
-%% them immediately here.
-cleanup_sender({SenderPid, SenderRef}) ->
-    erlang:demonitor(SenderRef, [flush]),
```
joedevivo, line 221: :)

src/riak_kv_pb_mapred.erl
beerriot, line 55: Indeed. I duplicated them to make the matches in `process_stream/3` less nesty. There's a comment in `pipe_mapreduce/5` about it.

src/riak_kv_wm_mapred.erl
beerriot, line 221: feelsgoodman

joedevivo:
+1!

beerriot:
Simple performance check 1: a single MR query, input is listkeys of 1-million item bucket (8-byte object values), query is `[{map, {modfun, riak_kv_mapreduce, map_identity}, none, true}]`. Looks like up to 10-20% overhead (i.e. this sink-backpressure branch takes 10-20% longer to finish than master does). Master prints lots of the following messages, while this sink-backpressure branch prints none:

```
2012-11-09 09:50:50.087 [info] <0.63.0>@riak_core_sysmon_handler:handle_event:95 monitor large_heap <0.5848.48> [{initial_call,{riak_api_pb_server,init,1}},{almost_current_function,{prim_inet,send,3}},{message_queue_len,2}] [{old_heap_block_size,0},{heap_block_size,47828850},{mbuf_size,0},{stack_size,37},{old_heap_size,0},{heap_size,23898616}]
2012-11-09 09:50:50.088 [info] <0.63.0>@riak_core_sysmon_handler:handle_event:95 monitor long_gc <0.5848.48> [{initial_call,{riak_api_pb_server,init,1}},{almost_current_function,{prim_inet,send,3}},{message_queue_len,2}] [{timeout,146},{old_heap_block_size,0},{heap_block_size,47828850},{mbuf_size,0},{stack_size,37},{old_heap_size,0},{heap_size,15618308}]
```

Next up is a concurrent basho_bench run to see what effect that improvement of memory-constraint has in that case (and to double-check the overhead estimate).

beerriot:
Further testing, in real distribution, with basho_bench and some concurrency showed that acking every message was, indeed, too much overhead. I made modifications to basho/riak_pipe#59 to allow specification of how often messages should be acked. Using that facility to send 10 un-acked messages before sending an acked message improved performance, such that the overhead is now only 1-2%.

Furthermore, the frequency of acked message is tweakable by setting the `riak_kv` application environment variable `mrc_sink_sync_period`. Its default is 10, for the ten un-acked messages described above. Setting it higher will cause more un-acked messages to be used between acked messages. Setting it to the atom `infinity` will cause all messages to be sent to the sink un-acked, which may be useful in the case that all MR queries produce very few results, and the overhead of even a few acks is too expensive.

joedevivo:
+1

beerriot:
Squashed to master as e639737.

PR #430, opened 2012-11-07
do not call riak_api functions removed in riak_api#15

beerriot:
Saving/clearing/restoring the riak_api dispatch table is no longer
necessary as of basho/riak_api#15 merging into master. In fact, the
dispatch_table function no longer exists, so calling it is removed here
to fix the tests.


joedevivo:
eunit tests look good when run locally.

+1

PR #431, opened 2012-11-09
Add leveldb ReadBlockError counter to riak_kv_stats

russelldb:
Ad hoc adding of one off leveldb stat for `ReadBlockError` on request of MvM.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -300,7 +300,8 @@ is_empty(#state{ref=Ref}) ->
 -spec status(state()) -> [{atom(), term()}].
 status(State) ->
     {ok, Stats} = eleveldb:status(State#state.ref, <<"leveldb.stats">>),
-    [{stats, Stats}].
+    {ok, ReadBlockError} = eleveldb:status(State#state.ref, <<"leveldb.ReadBlockError">>),
+    [{stats, Stats}, {read_block_error, ReadBlockError}].
 
 %% @doc Register an asynchronous callback
 -spec callback(reference(), any(), state()) -> {ok, state()}.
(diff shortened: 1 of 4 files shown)


jonmeredith:
Adding leveldb stats directly into riak_kv_stat breaks our backend abstraction, but otoh I'd like to freeze soon so it doesn't need to be undone.

Perhaps we need an additional per-backend level stats abstraction next release.

russelldb:
I can leave it out. I did this specifically at MvM's request. It's clearly a temporary bolt on. I have no feelings on it either way.

jonmeredith:
I really like the stats - it's the abstraction breaking that I'd like to
fix.

Pragmatism, not idealism.

On Fri, Nov 9, 2012 at 12:51 PM, Russell Brown notifications@github.comwrote:

> I can leave it out. I did this specifically at MvM's request. It's clearly
> a temporary bolt on. I have no feelings on it either way.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/basho/riak_kv/pull/431#issuecomment-10242221.

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

russelldb:
hmm, so if I renamed "backend stats" and generalised the code?

jonmeredith:
We can live with it for this release. I just wanted to bring it up for the future.

russelldb:
Noted.

russelldb:
So, is that this reviewed, or just commented?

russelldb:
Added commit to use `{function, M, F}` rather than anon fun, as per comment here https://github.com/basho/riak_api/pull/18

beerriot:
I've just verified that:
1. using bitcask backend, this stat measures `undefined`
2. using eleveldb backend, without errors, this stat measures `0`
3. using multibackend, without errors, this stat measures `0`

I was also able to verify that the stat increased when errors happen by overwriting one SST file and then doing a bunch of reads, some of which hit the vnode for that SST (thanks for the tip, @matthewvon). The stat did increase. Repeating the reads did not further increase the stat, I assume because read-repair fixed their storage on the corrupted vnode (the read-repair stat did increase during the first reads). Repeating the test by overwriting another vnode's SST and performing more reads did increase the read-error stat again. Using redbug to trace calls to `riak_kv_eleveldb_backend:status/1` shows that all vnodes are reporting the same `read_block_error`, not just the one I corrupted.

So, +1 merge!

russelldb:
Merged, many thanks for all the reviews on this stats epic.

PR #432, opened 2012-11-10
Location header returned on object creation updated to new URL format.

clr:
On POST creation of new object, the header 'Location:' was returning /riak/Bucket/Key and now returns /buckets/Bucket/keys/Key.


seancribbs:
We still support the old scheme, and older clients might be unable to understand the new format, so we need a deprecation path. Can you refactor it so that it checks the 'api_version' given in the state record to determine which format to generate?

machizhao:
132

beerriot:
This issue was resolved by #487

PR #433, opened 2012-11-12
Remove Luke usage

beerriot:
riak_pipe powers MapReduce queries now


beerriot:
In reviewing other PRs, I was reminded that we can also remove the "mget" call from riak_kv_vnode. Luke-based MapReduce was the only thing that used it.

slfritchie:
Bryan, if you'd create a new issue to remove the `mget` call, I'm fine with that.

No new failing EUnit failures, no significant changes to Dialyzer warnings.  +1 to merge all 5 related PRs, many thanks!

seancribbs:
finksplosion

kellymclaughlin:
Can [this](https://github.com/basho/riak_kv/blob/bwf-no-luke/src/riak_kv_app.erl#L118) be removed as part of this pr also?

beerriot:
Good catch, @kellymclaughlin! Unfortunately, it can't be remove the `{riak_kv, mapred_system}` capability just yet. If we remove it, and then there is a cluster membership change before all nodes are running the version of Riak with it removed, a renegotiation will flip those old nodes back to `legacy`, which will then cause them to fail at MapReduce queries.

However, new nodes also shouldn't lie about what they're capable of. I've just pushed a patch to remove `legacy` from what new nodes announce. Eventually we'll get all nodes announcing `pipe` by default, and then we can remove it.

kellymclaughlin:
Great point, I need to do that for my legacy keylisting pr as well.

kellymclaughlin:
+1 to merge after last commit

PR #437, opened 2012-11-16
Mw get put stats redux

lordnull:
- Adds a get/put monitor process to act as a point of folsom stat updating. Process announce their existence to the monitor saying if they are put or get fsm's
- Updates the get/put fsm's to announce their precess to the monitor.
- Adds options to allow get fsm's to skip read_repair under certain read loads. Appenv options 'riak, read_repair_skip_soft_cap' and 'riak, read_repair_skip_hard_cap' provide a range where read_repairs will occasionally skipped.
- Adds tests for above.


russelldb:
How do customers read the stats?

src/riak_kv_get_put_monitor.erl
```diff
@@ -0,0 +1,196 @@
...
+%% ------------------------------------------------------------------
+%% gen_server Function Definitions
+%% ------------------------------------------------------------------
+
+%% @private
+init([]) ->
+    [begin
+        folsom_metrics:Func(?COUNTER(FsmType, DataPoint))
+    end || {Func, FsmType, DataPoint} <- ?STATTYPES],
+%    folsom_metrics:new_counter(put_fsm_in_progress),
```
russelldb: Could we name these stats in line with the naming scheme I've been using (a tuple as a name, starting with the app name, e.g. {riak_kv, node, puts, fsm, in_progress})

src/riak_kv_get_put_monitor.erl
```diff
@@ -0,0 +1,196 @@
...
+%    folsom_metrics:new_counter(get_fsm_in_progress),
+%    folsom_metrics:new_spiral(put_fsm_errors_minute),
+%    folsom_metrics:new_spiral(get_fsm_errors_minute),
+%    folsom_metrics:new_counter(put_fsm_errors_since_start),
```
russelldb: DO you need a spiral and a counter? The spiral tracks the total and the minute, with two fields on read (count, one)

src/riak_kv_get_put_monitor.erl
lordnull: This was actually done, though the format is a bit different.  The ?COUNTER macro takes the type and data point and expands it to a tuple of {riak_kv, node(), Type :: 'get' | 'put', DataPoint :: 'in_progess' | 'errors'}.  I've missed the fsm bit obviously, so I can add that in if you think it's needed.

src/riak_kv_get_put_monitor.erl
lordnull: Nope, I don't, but you're looking at commented out code.  The ?STATTYPE macro has the list of stats:

-define(STATTYPES, [
    {new_counter, put, in_progess},
    {new_counter, get, in_progess},
    {new_spiral, put, errors},
    {new_spiral, get, errors}
]).

I'll remove the commented out code.  A function call might be better than a macro, I just wanted to avoid typing the same data out in two places.

lordnull:
> > How do customers read the stats?

A very good question, and one I do not have a good answer for at the moment.  the all_stats/0 function isn't going to cut it.  The riak_kv_stat module looks like the place to put the new stats, but requires editing the module, rather can calling into it.  It seems like if I follow the pattern of the stats currently laid half of the code for get_put monitor ends up in riak_kv_stat.  Instead of the get_put_monitor being dependant on the _stat module, suddenly it's both ways.

I'm likely misunderstanding something or some purpose.

russelldb:
So. There is no general query method for stats exposed yet.

If you want the stats in the blob, then add them via `riak_kv_stat_bc`. Use `riak_core_stat_q` to pull the stats from folsom, or have `riak_kv_stat_bc` call your `all_stats/0` fun and spooge your stats into the big stats blob so there available over `/stats` and `riak-admin status`.

If these stats are just for CSEs / engineers / support then they can attach and use your `all_stats/0` fun.

src/riak_kv_get_fsm.erl
```diff
@@ -355,6 +360,48 @@ maybe_delete(_StateData=#state{n = N, preflist2=Sent,
             nop
     end.
 
+%% based on what the get_put_monitor stats say, and a random roll, potentially
+%% skip read-repriar
```
russelldb, line 364: typo of 'repair'

russelldb:
When I checkout this branch, start riak and try and get stats I get the following error:

```
{error,{function_clause,[{riak_kv_stat_bc,'-get_put_stats/2-lc$^0/1-0-',
                                      [{error,{riak_kv,node,puts,in_progess},nonexistent_metric},
                                       {riak_kv,node,puts,in_progess}],
                                      [{file,"src/riak_kv_stat_bc.erl"},{line,362}]},
                     {riak_kv_stat_bc,get_put_stats,2,
                                      [{file,"src/riak_kv_stat_bc.erl"},{line,362}]},
                     {riak_kv_stat_bc,produce_stats,0,
                                      [{file,"src/riak_kv_stat_bc.erl"},{line,148}]},
                     {timer,tc,3,[{file,"timer.erl"},{line,194}]},
                     {folsom_metrics,histogram_timed_update,4,
                                     [{file,"src/folsom_metrics.erl"},{line,163}]},
                     {riak_core_stat_cache,'-do_get_stats/2-fun-0-',4,
                                           [{file,"src/riak_core_stat_cache.erl"},{line,182}]}]}}
```

So I tried `riak_kv_get_put_monitor:all_stats()` and got the same error:

```
[{{riak_kv,node,puts,in_progess},
  {error,{riak_kv,node,puts,in_progess},nonexistent_metric}},
 {{riak_kv,node,gets,in_progess},
  {error,{riak_kv,node,gets,in_progess},nonexistent_metric}},
 {{riak_kv,node,puts,errors},
   {error,{riak_kv,node,puts,errors},nonexistent_metric}},
 {{riak_kv,node,gets,errors},
  {error,{riak_kv,node,gets,errors},nonexistent_metric}}]
```

Running `whereis(riak_kv_get_put_monitor)` gets me a `Pid` so the gen_server is running.
Attaching to the node and running `riak_core_stat_q:names_and_types([riak_kv, node, puts, '_']).` and
`riak_core_stat_q:names_and_types([riak_kv, node, gets, '_']).` shows the metrics aren't registered in folsom.

Is there some other dependancy I'm missing? This is a clean clone from master, but with the riak_kv dep checked out to `mw_get_put_stats_redux`

---

So I tried killing the `riak_kv_get_put_monitor` and when the sup restarted it everything worked as expected.

---

Another thing, is the monitor _just_ for stats? In that case, shouldn't it be started by the stats sup, like other stats modules? That way when folsom crashes it gets restarted and re-registers stats. (maybe this is the problem, is `folsom` up when this gen_server attempts to register stats?)

---

Lastly, please can someone provide some context as to the actual _purpose_ of this PR? What problem does it solve? I heard mention of a performance improvement, if that is the case, where are the numbers? How do I validate / verify that improvement?

---

All that said, after the kill / restart of the `riak_kv_get_put_monitor` it all works as advertised, and the code is very readable and rather excellent.

jonmeredith:
This PR is about tracking the changes we made against 1.2-perf taking supervisors off the critical path.  It may make more sense to submit it against 1.2-perf then merge 1.2-perf into master.

lordnull:
The stats missing error is because register stats happens after the get_put_monitor starts.  The register stats deletes all the folsom metrics, making them missing.  I'm working on a fix.

lordnull:
@russelldb The recent pushes address the major issue of the stats missing. I didn't see a stats sup when I made the module. I've (unfortunately) got the folsom metric definition code in two places, the stat module and the monitor module.

@jonmeredith I'm working on getting a clean rebase for 1.2-perf created.

lordnull:
@jonmeredith https://github.com/basho/riak_kv/commits/mw_get_put_stats_1.2-perf is a clean(ish) application of the commits in this branch to 1.2-perf.  Tests pass, stats work, riak-admin status works.  If a review finds that branch more palatable than this, this pr can just be closed, and a proper one opened on the other branch I suppose.

russelldb:
```
@russelldb The recent pushes address the major issue of the stats missing. 
I didn't see a stats sup when I made the module. I've (unfortunately) got the
folsom metric definition code in two places, the stat module and the monitor module.
```

Is this so that riak_kv_stat mod does the create / destroy for you at start up / crash time? I'm just wondering if it wouldn't make more sense to have this module registered as a `stat_mod` with riak_core like the other stat modules. Maybe some small change would be needed to riak_core to allow multiple stat_mods per app.

russelldb:
That aside, this works, and if expedience is key, then I'm +1 (as long as you create an issue for riak_kv that says something along the lines of 'Include riak_kv_get_put_monitor in future stats refactoring work')

lordnull:
It is in the stat module because I had no other choice but to have it handle stat creation at start up.  It comes down to this line in master of riak_kv (https://github.com/basho/riak_kv/blob/master/src/riak_kv_stat.erl#L57):

```
    [(catch folsom_metrics:delete_metric(Stat)) || Stat <- folsom_metrics:get_metrics(),
```

The 1.2-perf version doesn't have that line, meaning the only potion of the stats there is a bit of formatting magic so there's proper output for riak-admin status.

I'm not certain if this is a good fit for being supervised by risk_core:stats_sup as this needs to be running with riak_kv (not optional) and does monitoring.  I'm all for the issue creation.

I'd like to hear from @jonmeredith on if we should merge this branch or use the 1.2-perf based branch as he mentioned.

jonmeredith:
I had a chance to play with this 1.2-perf based branch yesterday and it tested well still - 10% improvement on a simple load.  The code reviewed well against 1.2-perf.

get_put_monitor logically belongs with in the KV sup as it's an integral part of KV as it is also controlling backpressure on read repairs now.

My recommendation would be to merge against 1.2-perf then merge 1.2-perf into master and resolve stats conflicts there.  I'd expect a minor conflict with the luke removal pull requests around mapred_cache being removed.

If I had one last ask I'd rename the 'in_progress' to 'active' but that's probably just personal preference.

Jon

jonmeredith:
Also, I'm not sure how the changes interact with the extra infromation on read repairs Russell added - it would be very valuable to have a count of skipped read repairs due to overload as well as the other new categories.

lordnull:
The new pull request for these features into 1.2-perf is at:  https://github.com/basho/riak_kv/pull/438

I'll be closing this and putting a link to this discussion on that issue.

russelldb:
```
I'm not certain if this is a good fit for being supervised by risk_core:stats_sup as this needs to be running with riak_kv (not optional) and does monitoring. I'm all for the issue creation.
```

@lordnull did an issue get created? I'm very late to reply as I have been travelling. I would say the problem is your conflating the stats gathering and what you use them for. Folsom and the stats sub-system is owned by riak_core, and I think it is a mistake to try and work against that. You can still have your gen_server that monitors and reads the stats to make read_repair choices. It seems that this is getting merged regardless…I just wanted to make sure my opinion is recorded somewhere.

lordnull:
@russelldb An issue hasn't been created yet.  In 1.2-perf, the stats module doesn't remove all stats from folsom, so the issue was resolved there. The stats are still used for both stat gathering and read_repair choices, and I can see how splitting those up may make it cleaner; I'd want to check performance impact too though.

Edit:  In merging 1.2 into master, this is an issue that we'll need to resolve at some point.  Once I'm done w/ the merge and pushed, an issue should be opened on the new master to make the stat tracking separate from the fsm monitoring.

PR #438, opened 2012-11-20
get put stats -> 1.2-perf

lordnull:
Addes a get_put monitor and some stat tracking, as well as skipping read-repair on over loaded nodes.

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -129,6 +129,7 @@ init([From, Bucket, Key, Options]) ->
                        options = Options,
                        bkey = {Bucket, Key},
                        startnow = StartNow},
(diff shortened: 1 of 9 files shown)


jonmeredith:
I tried out the read repair limit code tonight and pushed a couple of suggested changes on this branch  mw_get_put_stats-1.2-perf-jdm

Thinking about the intent of the read repair limit, the soft cap/hard cap is applied to the total amount of gets in the system, not the amount of read repair which is a subtly different thing ... what should you sent the soft/hard caps on to limit the amount of read repair when you don' t know how many of the active get FSMs are coordinating requests and how many are in read repair.

Being able to limit in this way is far better than what we have - so given freeze is very close upon us, this may be where we can get to.

PR #440, opened 2012-11-25
Re-instate code from #415 that was lost in later changes

russelldb:
The bulk of the riak_kv_stat code was moved to a to-be deprecated
module riak_kv_stat_bc. The change that fixes memory stats for
erlang r15b02 was lost. See riak_kv PR #415 for details.


buddhisthead:
+1

Tested with R15B02 without this PR and observed failure in folsom stats.
Tested again with PR applied and saw no failures.
Yay!

russelldb:
Thanks!

PR #442, opened 2012-11-28
Merge 1.2-perf into master

lordnull:


src/riak_client.erl
```diff
@@ -137,10 +143,22 @@ put(RObj, Options) when is_list(Options) ->
     ReqId = mk_reqid(),
     case ClientId of
         undefined ->
-            riak_kv_put_fsm_sup:start_put_fsm(Node, [{raw, ReqId, Me}, RObj, Options]);
+            case node() of
+                Node ->
+                    riak_kv_put_fsm:start_link({raw, ReqId, Me}, RObj, Options);
+                _ ->
+                    proc_lib:spawn_link(Node, riak_kv_put_fsm, start_link,
```
russelldb, line 150: what was the observed improvement spawning directly rather than using the sup?

src/riak_kv_get_fsm.erl
```diff
@@ -355,6 +360,53 @@ maybe_delete(_StateData=#state{n = N, preflist2=Sent,
             nop
     end.
 
+%% based on what the get_put_monitor stats say, and a random roll, potentially
```
russelldb, line 363: Can I get some background on this, please? (reason for, observed benefit.)

I'd really like the comment here to say why we do this rather than describe the mechanism by which we do it. Otherwise it will be lost to time.

src/riak_kv_get_fsm.erl
```diff
@@ -355,6 +360,53 @@ maybe_delete(_StateData=#state{n = N, preflist2=Sent,
...
+    Actual = riak_kv_get_put_monitor:gets_active(),
+    determine_do_read_repair(SoftCap, HardCap, Actual).
+
+determine_do_read_repair(undefined, HardCap, Actual) ->
+    determine_do_read_repair(HardCap, HardCap, Actual);
+determine_do_read_repair(_SoftCap, HardCap, Actual) when HardCap =< Actual ->
+    false;
+determine_do_read_repair(SoftCap, _HardCap, Actual) when Actual =< SoftCap ->
+    true;
+%determine_do_read_repair(HardCap, HardCap, _Actual) ->
```
russelldb: I think it is better to delete than leave commented out code in, git can give us the history.

src/riak_kv_get_fsm.erl
```diff
@@ -355,6 +360,53 @@ maybe_delete(_StateData=#state{n = N, preflist2=Sent,
             nop
     end.
 
+%% based on what the get_put_monitor stats say, and a random roll, potentially
+%% skip read-repriar
```
russelldb, line 364: s/repriar/repair

src/riak_kv_get_fsm.erl
```diff
@@ -355,6 +360,53 @@ maybe_delete(_StateData=#state{n = N, preflist2=Sent,
...
+            read_repair(Indices, RepairObj, UpdStateData);
+        true ->
+            riak_kv_stat:update(skipped_read_repairs),
+            skipping
+    end.
+
+determine_do_read_repair(_SoftCap, HardCap) when HardCap == undefined ->
+    true;
+determine_do_read_repair(SoftCap, HardCap) ->
+    Actual = riak_kv_get_put_monitor:gets_active(),
```
russelldb, line 384: These are the active gets, not the currently active read repairs. Is there a chance, with enough gets starting at once, that no actual read repairs occur?

Say hard limit is 100, and 1000 gets start at the same time, by the time the first get decides to read repair, the `Actual` maybe way above the hard limit. Should we be using a measure of the active read repairs here? Is it OK since we have AAE now, as well?

Sorry if this has already been discussed somewhere, I don't know where to look for further context.

src/riak_kv_get_put_monitor.erl
```diff
@@ -0,0 +1,194 @@
...
+%% @private
+handle_call(dump_state, _From, State) ->
+    {reply, State, State};
+
+handle_call(_Request, _From, State) ->
+    {reply, ok, State}.
+
+%% @private
+handle_cast({get_fsm_spawned, Pid}, State) ->
+    #state{monitor_list = List} = State,
```
russelldb, line 123: It took me a second glance to see that your just matching of the `monitor_list` here. I know that this is perfectly valid. In most other gen_servers I've seen in riak the match is done in the function head. I'm just mentioning out of interest: do you prefer this style for a reason?

src/riak_kv_get_put_monitor.erl
```diff
@@ -0,0 +1,194 @@
...
+    folsom_metrics:notify({?COUNTER(Type, errors), 1}).
+
+orddict_get_erase(Key, Orddict) ->
+    case orddict:find(Key, Orddict) of
+        undefined ->
+            undefined;
+        {ok, Value} ->
+            Orddict2 = orddict:erase(Key, Orddict),
+            {Value, Orddict2}
+    end.
```
russelldb, line 194: Really nice, readable, well documented and commented code. I wish I wrote like this.

russelldb:
Merged pretty clean to master (just a conflict with riak_kv_sup over AAE and riak_kv_get_put_monitor.)

Tests pass, some poking from the console and it all looks good.

Should I benchmark master against this merged (or has that been done, are there numbers?)

Apart from comments above +1 from me.

src/riak_client.erl
jonmeredith, line 150: General waffle about performance near the end.

jonmeredith:
For performance - I've lost individual figures on improvements.  Under heavy load testing the get/put FSM supervisors would build up message queues of several hundred requests.  Once the FSMs were started up directly on the client, the 'rex' remote execution server shot to the top of etop and that became a bottleneck - it does extra work around making references and monitoring that we don't require.

I retook some measurements when I looked at the original get/put fsm monitor code on the 1.2-perf branch and found that pure load performance was ~10% higher, but it's only when you're really pushing the system that you notice it.

src/riak_kv_get_fsm.erl
lordnull, line 384: Using active read repairs was discussed, unfortunately it was brought up after freeze deadline, so it's been slated as a future improvement.  If there are more gets active than the hard cap, there is a 100% chance that a read repair will not happen. It is a trade-off, as one would rely much more heavily on AAE if the system is loaded enough to constantly skip read repair.

src/riak_kv_get_put_monitor.erl
lordnull, line 123: Unless I need to match in the function head to differ between clauses, I pull out what I need from on the line below.  This means that when I look at the function clauses, only what is needed for the clause to match is there, and the line is shorter.

eg: if I needed to test if the monitor list was a certain length, the match would be in the head. Since I don't, I went for a shorted function head declaration.

russelldb:
Ok. Thanks. +1

PR #447, opened 2012-12-07
Add basic health check to Riak KV

jtuple:
This commit adds a simple health check to riak_kv to support the new health check functionality added to riak_core. This health check monitors the process mailboxes of k/v vnodes. The check fails if any vnode on the node has a mailbox containing over a configurable limit of messages -- default 5000.

Depends on basho/riak_core#257 which in turn depends on basho/riak_core#240

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -77,7 +77,7 @@ start(_Type, _StartArgs) ->
     case code:ensure_loaded(StorageBackend) of
         {error,nofile} ->
             lager:critical("storage_backend ~p is non-loadable.",
-                                   [StorageBackend]),
+                           [StorageBackend]),
             throw({error, invalid_storage_backend});
         _ ->
             ok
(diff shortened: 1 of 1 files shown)


jtuple:
@jrwest The latest commit addresses issue we discussed in backchannels. Specifically, there is now a lower and upper threshold as well as output messages to detail when transitions occur (and when disabling, printing the vnode/queue lengths as well).

There are two items that should still be addressed at some point
1. We should provide a way to disable all health check logic on a node.
2. We should fix the stupid exit error messages that are based on the `riak_core_node_watcher` code use exits for signaling. I gave a pass on the exit approach when I reviewed that PR, but have become sufficiently annoyed with the messages that I think we should address them. Message example:

```
14:58:42.462 [error] CRASH REPORT Process <0.2530.0> with 0 neighbours exited with reason: false in riak_core_node_watcher:'-start_health_check/2-fun-0-'/4 line 723
```

Both of the above two issues have more to do with the base health check logic in `riak_code_node_watcher` and not Riak K/V or this basic K/V health check. So, I'd say we address those in a future PR and get the current stuff merged in.

jrwest:
+1

PR #448, opened 2012-12-10
use riak_capability:get for vnode_vclocks

joecaswell:


seancribbs:
This needs to merge to master as well.

jtuple:
This was merged into master for Riak 1.3 in commit 65be0808509b47d8cd1055f28c2586978274c127

As no known 1.2.x point releases are planned, I'm going to consider this resolved (fixed in 1.3, yay!) and close out this PR. Thanks for catching this bug, very easy to overlook given that the new PB API code and the capability system essentially merged into Riak the same week and I overlooked this use of the older app var.

Technically, these should probably be calls to `riak_core_capability:get/3` rather than `get/2` to ensure we never throw an exception if the call is made while Riak is still starting up the capability system, but it's unclear to me if the PB server would ever been up early enough for that to happen. Plus an exception would likely just kill a single request or so, and things would stabilize quickly once capability system was fully up.

PR #449, opened 2012-12-13
Update stats in process

russelldb:
When a node is giving good thoughput then riak_kv_stat cannot keep
up. Large message queues ensue, memory balloons, the node dies.
This commit only forces stat creation through the gen_server, any
other stat update is direct in process.

This does cause a degradation in throughput & latency as each get / put
process is waiting to acquire the locks on the folsom tables.

NOTE: this works a whole lot better with these `folsom` optimisations https://github.com/basho/folsom/tree/adt-faster-spiral

With the `folsom` improvements, put throughput is as good as 1.2.1 (though worse than master without stats.) Even with the `folsom` improvements, there is extra latency, most noticeable in the 95th%ile.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -39,16 +39,14 @@
 %% API
 -export([start_link/0, get_stats/0,
          update/1, register_stats/0, produce_stats/0,
-         leveldb_read_block_errors/0]).
-
+         leveldb_read_block_errors/0, stop/0]).
 -export([track_bucket/1, untrack_bucket/1]).
+-export([active_gets/0]).
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
-% TODO temporarily silence warnings
--export([get_put_monitor_stats/0]).
 -define(SERVER, ?MODULE).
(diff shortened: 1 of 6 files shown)


joedevivo:
./riak_test -t verify_riak_stats -> pass

engelsanchez:
The change is simple looks good as far as I can tell. :+1: Was that your new plus one @joedevivo ?

src/riak_kv_stat.erl
```diff
@@ -85,8 +99,9 @@ init([]) ->
...
     {ok, ok}.
 
-handle_call(_Req, _From, State) ->
-    {reply, ok, State}.
+handle_call({register, Name, Type}, _From, State) ->
+    Rep = do_register_stat(Name, Type),
+    {reply, Rep, State}.
 
 handle_cast({update, Arg}, State) ->
     do_update(Arg),
```
joedevivo: You should remove this handle_cast clause, since it's no longer used.

joedevivo:
@engelsanchez no, that was just saying the code passed the tests, still was awaiting manual review. I told @russelldb  I'd remove the extra handle_cast and run riak_test again.

joedevivo:
Ok, riak_test passed again after the clause was removed. +1

PR #451, opened 2012-12-14
Change AAE hashtree to buffer and batch write to LevelDB

jtuple:
Change AAE to batch write to LevelDB rather than write hash updates one at a time. While AAE writes are small and asynchronous, having 2x the writes (normal Riak write + AAE write) going into LevelDB has shown to have an impact on Riak throughput. This pull-request has hashtrees buffer writes in memory until there are 200 hash updates, and then send all writes to LevelDB as a single batch write. This has been benchmarked to show a noticeable performance improvement.

In addition to batch writing once every 200 writes, the hashtree will also always flush writes before updating a tree and performing a tree exchange. Thus, AAE semantics are identical between non-batched and batched approach.

If a `riak_kv_index_hashtree` crashes for some reason, and therefore crashes the `hashtree` processes, we may lose the last few writes buffered in RAM. This isn't a problem, as AAE is designed to be lossy anyway, and exchanges with other replicas and/or eventual re-building of the tree will sort things out.

This pull-request also makes the LevelDB options used by AAE configurable, as well as bumps the default write-buffer size from 1 MB to 4 MB.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -455,9 +455,11 @@ new_segment_store(Opts, State) ->
                   SegmentPath ->
                       SegmentPath
               end,
-    Options = [{create_if_missing, true},
-               {write_buffer_size, 1024*1024},
-               {max_open_files, 20}],
+    Config = app_helper:get_env(riak_kv,
+                                anti_entropy_leveldb_opts,
+                                [{write_buffer_size, 4*1024*1024},
+                                 {max_open_files, 20}]),
+    Options = [{create_if_missing, true} | Config],
     filelib:ensure_dir(DataDir),
     {ok, Ref} = eleveldb:open(DataDir, Options),
     State#state{ref=Ref, path=DataDir}.
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -171,6 +171,8 @@
                 ref            :: term(),
                 path           :: string(),
                 itr            :: term(),
+                write_buffer   :: [{binary(), binary()}],
+                write_buffer_count :: integer(),
                 dirty_segments :: array()
                }).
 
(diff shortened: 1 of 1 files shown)


src/hashtree.erl
```diff
@@ -253,14 +257,33 @@ insert(Key, ObjHash, State, Opts) ->
...
+            State3 = maybe_flush_buffer(State2),
+            %% Dirty = gb_sets:add_element(Segment, State3#state.dirty_segments),
+            Dirty = bitarray_set(Segment, State3#state.dirty_segments),
+            State3#state{dirty_segments=Dirty};
         false ->
             State
     end.
 
+maybe_flush_buffer(State=#state{write_buffer_count=WCount}) ->
+    Threshold = 200,
```
joedevivo, line 277: Is there a reason we'd ever want this to be configurable?

joedevivo:
update the @docs in hashtree.erl around line 82: "updated in real-time as writes occur"

src/hashtree.erl
```diff
@@ -290,14 +313,16 @@ should_insert(HKey, Opts, State) ->
...
+    SnapState = snapshot(State2),
+    State3 = SnapState#state{dirty_segments=bitarray_new(NumSegments)},
+    {SnapState, State3}.
 
 -spec update_tree(hashtree()) -> hashtree().
 update_tree(State) ->
-    State2 = snapshot(State),
-    update_perform(State2).
+    State2 = flush_buffer(State),
+    State3 = snapshot(State2),
```
joedevivo, line 329: What happens if an insert comes in between flush_buffer/1 and snapshot/1?

jtuple, line 329: There is no "in between". `insert` and `update_tree` are both top-level functions that operate on a hash tree. You put a hash tree in, get a hash tree out. You have two options:

``` erlang
B = hashtree:insert(Key, Hash, A),
C = hashtree:update_tree(B).
```

or 

``` erlang
B = hashtree:update_tree(A),
C = hashtree:inesrt(Key, Hash, B)
```

The behavior is clear in either case.

src/hashtree.erl
jtuple, line 277: I keep going back and forth on this one. My current opinion is hardcode it and QA things against a static value. How much you want to buffer or not has little to do with user workload, cluster size, etc. It's almost entirely just a "magic number" that we want to optimize against Riak / LevelDB.

joedevivo:
+1 to merge, once that documentation is fixed.

PR #454, opened 2012-12-18
Try to make KV shutdown cleaner

Vagabond:
Unregister PB services and Webmachine routes and wait for any active put
FSMs to complete.

The goal of this is to prevent writes from continuing to happen while KV is stopping. Once we mark the service down and unregister http/pb endpoints, the node can no longer service client writes or act as the coordinating node for puts.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -162,13 +162,18 @@ prep_stop(_State) ->
         %% no error message or logging about the failure otherwise.
 
         lager:info("Stopping application riak_kv - marked service down.\n", []),
-        riak_core_node_watcher:service_down(riak_kv)
+        riak_core_node_watcher:service_down(riak_kv),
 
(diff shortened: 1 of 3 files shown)


bookshelfdave:
Tested with https://github.com/basho/riak_test/pull/128, these changes seem fine.

src/riak_kv_app.erl
```diff
@@ -204,3 +208,12 @@ check_epoch() ->
                 "but your system says the epoch is ~p", [Epoch]),
             ok
     end.
+
+wait_for_put_fsms() ->
```
bookshelfdave: I know 1000 ms is plenty of time for the fsm's to complete, but should this give up after N tries?

bookshelfdave:
riak_kv tests passed (retry_test failed, but I believe Jesse has a fix for it).
riak_test replication2 test passes.

bookshelfdave:
+1 for Andrew's commit

PR #458, opened 2012-12-28
Make index_hashtree process exit when related vnode exits

jtuple:
When a riak_kv_vnode exits, the associated riak_kv_index_hashtree should also shutdown because a future riak_kv_vnode process will try to create a new riak_kv_index_hashtree. If the old index_hashtree is still alive, it retains a lock to the LevelDB instance, causing the newer index_hashtree to fail endlessly when trying to open the LevelDB instance.

This commit changes riak_kv_index_hashtree to monitor the associated riak_kv_vnode, shutting down if the riak_kv_vnode ever exits.

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -30,7 +30,7 @@
 -include_lib("riak_kv_vnode.hrl").
 
 %% API
--export([start/2, start_link/2]).
+-export([start/3, start_link/3]).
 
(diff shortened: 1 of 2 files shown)


engelsanchez:
Well, the code looks good to me and I just ran this branch against the partition_repair test with the massive slaughter of kv nodes without the hashtree problems that happened before virtually all of the time. So I would say :+1:

PR #459, opened 2012-12-28
Fix health check code to handle dead pids

jtuple:
If the health check code calls `process_info` on a dead pid, the result will be `undefined` rather than the expected list of information.  This causes the list comprehension to crash. This pull-request fixes the list comprehension to handle this case.


slfritchie:
+1 to merge

PR #460, opened 2012-12-31
Fix AAE exchange bug for the N=1 case

jtuple:
When `N=1`, AAE doesn't need to do any KV exchanges. However, entropy manager used pattern match that expected there would always be exchanges to be done. This pull-request fixes that assumption and prevents the `riak_kv_entropy_manager` from constantly crashing every tick when a Riak cluster only includes `N=1` buckets.


coderoshi:
+1 As advertised

PR #462, opened 2013-01-08
Expose FSM timeouts via the HTTP API

evanmcc:
adding the header 'X-Riak-Timeout=<some_int_in_ms>' will now cause the request to return a timeout error if it takes longer than the specified time interval.  This works for DELETEs, PUTs, and GETs, but in the case of a PUT or a get, a timeout failure does not guarantee anything about the state of the cluster when it happens.  The PUT or DELETE may or may not have succeeded; the correct course of action after the timeout is to retry the same action until a 200 is received.

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -142,7 +142,7 @@
               links,        %% [link()] - links of the object
               index_fields, %% [index_field()]
               method,       %% atom() - HTTP method for the request
-              timeout=60000 %% integer() - passed-in timeout value in ms
+              timeout       %% integer() - passed-in timeout value in ms
              }).
 %% @type link() = {{Bucket::binary(), Key::binary()}, Tag::binary()}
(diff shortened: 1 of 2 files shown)


evanmcc:
The associated riak_test is here:  https://github.com/basho/riak_test/pull/150

seancribbs:
Too late for 1.3, but I like the general idea.

evanmcc:
Couldn't make heads or tails of the travis output.

seancribbs:
@evanmcc Not to delay this PR, but we should also add these to the appropriate PB messages.

evanmcc:
Yup; it's been on my todo list since I submitted the first one, but it's a
bit more involved, so I am only halfway done.

As a related question: is the riak erlang client easily available in
riak_test?

chardan:
+1

seancribbs:
@evanmcc Yes, `rt:pbc(Node)` will grant you a new client.

PR #463, opened 2013-01-11
Reduce spamminess of hashtree creation.

seancribbs:
It shouldn't be necessary to log at the `info` level that new hash trees are being created.

/cc @rzezeski @jtuple


rzezeski:
I'm +1 on this but make sure Joe concurs before merging.

rzezeski:
Actually, re-reviewing this I'm -1 now.  These msgs should only occur when n-val is changed and I think it would be good to log them by default so that we can tell from logs that n-val has been changed.  It should only happen once per partition.

rzezeski:
I'm going to close this out.  @seancribbs if you get caremad about this I'm happy to discuss.

seancribbs:
It's fine, IIRC this was problematic early in the 1.3 cycle for some other reason, but I have not noticed it lately.

Sean Cribbs

PR #465, opened 2013-01-15
Modified HTTP interface user metadata handling

cdahlqvist:
Have modified the HTTP interface to strip off the 'x-riak-meta-' prefix before storing user metadata onto the record. It also appends the 'x-riak-meta-' prefix when adding HTTP response headers. This allows the setting of short user metadata parameter names like is possible through the PB clients.

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -663,7 +663,12 @@ extract_user_meta(RD) ->
                         string:to_lower(riak_kv_wm_utils:any_to_list(K)))
                 end,
                 mochiweb_headers:to_list(wrq:req_headers(RD))),
-    [{string:sub_string(A, 13),B} || {A,B} <- List].
+    case application:get_env(riak_kv, strip_http_user_metadata_prefix) of
+        {ok, true} ->
+            [{string:sub_string(A, 13),B} || {A,B} <- List];
+        _ ->
+            List
+    end.
 
 %% @spec multiple_choices(reqdata(), context()) ->
 %%          {boolean(), reqdata(), context()}
@@ -713,7 +718,12 @@ produce_doc_body(RD, Ctx) ->
             UserMetaRD = case dict:find(?MD_USERMETA, MD) of
                         {ok, UserMeta} ->
                             lists:foldl(fun({K,V},Acc) ->
-                                            wrq:merge_resp_headers([{?HEAD_USERMETA_PREFIX ++ K,V}],Acc)
+                                            case application:get_env(riak_kv, strip_http_user_metadata_prefix) of
+                                                {ok, true} ->
+                                                    wrq:merge_resp_headers([{?HEAD_USERMETA_PREFIX ++ K,V}],Acc);
+                                                _ ->
+                                                    wrq:merge_resp_headers([{K,V}],Acc)
+                                            end
                                         end,
                                         LinkRD, UserMeta);
                         error -> LinkRD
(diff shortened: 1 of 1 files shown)


cdahlqvist:
This fix addresses issue https://github.com/basho/riak/issues/234

cdahlqvist:
Made new HTTP user metadata behavior configurable through 'riak_kv' section app.config entry as follows: 

%% Enable stripping of prefix for user metadata over the HTTP interface
%{strip_http_user_metadata_prefix, true},

This will require user to deliberately enable it, which will ensure backwards compatibility.

jaredmorrow:
@cdahlqvist any status on this, has this made it on the core/kv cabal?

cdahlqvist:
@jaredmorrow I do not think it has.

slfritchie:
Pulling PRs out of the freezer ... @seancribbs or @broach any opinions?

src/riak_kv_wm_object.erl
```diff
@@ -657,12 +657,18 @@ extract_content_type(RD) ->
...
 extract_user_meta(RD) ->
-    lists:filter(fun({K,_V}) ->
+    List = lists:filter(fun({K,_V}) ->
                     lists:prefix(
                         ?HEAD_USERMETA_PREFIX,
                         string:to_lower(riak_kv_wm_utils:any_to_list(K)))
                 end,
-                mochiweb_headers:to_list(wrq:req_headers(RD))).
+                mochiweb_headers:to_list(wrq:req_headers(RD))),
+    case application:get_env(riak_kv, strip_http_user_metadata_prefix) of
```
seancribbs, line 666: @cdahlqvist Can you change this to use `app_helper:get_env/3` with the default of `true`?

src/riak_kv_wm_object.erl
```diff
@@ -712,7 +718,12 @@ produce_doc_body(RD, Ctx) ->
             UserMetaRD = case dict:find(?MD_USERMETA, MD) of
                         {ok, UserMeta} ->
                             lists:foldl(fun({K,V},Acc) ->
-                                            wrq:merge_resp_headers([{K,V}],Acc)
+                                            case application:get_env(riak_kv, strip_http_user_metadata_prefix) of
+                                                {ok, true} ->
+                                                    wrq:merge_resp_headers([{?HEAD_USERMETA_PREFIX ++ K,V}],Acc);
+                                                _ ->
+                                                    wrq:merge_resp_headers([{K,V}],Acc)
```
seancribbs, line 725: If the user has created metadata from the PB client without the prefix, won't this emit headers without the prefix? Clients expect the prefix, regardless of the internal representation.

src/riak_kv_wm_object.erl
```diff
@@ -712,7 +718,12 @@ produce_doc_body(RD, Ctx) ->
             UserMetaRD = case dict:find(?MD_USERMETA, MD) of
                         {ok, UserMeta} ->
                             lists:foldl(fun({K,V},Acc) ->
-                                            wrq:merge_resp_headers([{K,V}],Acc)
+                                            case application:get_env(riak_kv, strip_http_user_metadata_prefix) of
```
seancribbs, line 721: @cdahlqvist Same here with `app_helper` instead of `application`. Also, you should probably pull out this env value before entering the iteration.

nickelization:
I'm going to go ahead and close this. In a perfect world, it would be nice if the behavior were consistent for user metadata across HTTP and protobufs. However, in practice, we would have to leave this config flag turned off by default so that we don't break backward compatibility. As a result, this fix would only get used if a user specifically knew about the flag, and I'm not at all convinced it's worth cluttering up our tests, docs, and config files to support a flag to work around such a small, esoteric corner case.

I'll leave basho/riak#234 (RIAK-1893) open so that people know about this, but I suspect it will just have to remain a known limitation for the foreseeable future. If someone does want to address it, then we will at least have to address the review comments and update this code, and I doubt that will rise to the top of our priority list anytime soon.

PR #467, opened 2013-01-17
add stats for coverage query starts

evanmcc:
- index fsm: create, create_error, active
- keys and bucket fsms (as 'list'): create, create_error, active

This should give us a better picture as to how many index and list queries are being executed against each riak node.

cc @russelldb

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -45,10 +45,12 @@
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
-         terminate/2, code_change/3]).
+         terminate/2, code_change/3, monitor_loop/1]).
 
 -define(SERVER, ?MODULE).
 -define(APP, riak_kv).
+
+-record(state, {monitors}).
 
 start_link() ->
     gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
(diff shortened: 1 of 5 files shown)


src/riak_kv_stat.erl
```diff
@@ -102,12 +104,28 @@ stop() ->
 
 init([]) ->
     register_stats(),
-    {ok, ok}.
+    State = {state, [spawn(?MODULE, monitor_loop, [index]),
```
evanmcc: should these be spawn_link() instead?

src/riak_kv_stat_bc.erl
```diff
@@ -232,7 +232,13 @@ legacy_stat_map() ->
      {coord_redirs_total, {riak_kv,node,puts,coord_redirs}, counter},
      {executing_mappers, {riak_kv,mapper_count}, counter},
      {precommit_fail, {riak_kv, precommit_fail}, counter},
-     {postcommit_fail, {riak_kv, postcommit_fail}, counter}
+     {postcommit_fail, {riak_kv, postcommit_fail}, counter},
+     {index_fsm_create, {{riak_kv, index, fsm, create}, one}, spiral},
+     {index_fsm_create_error, {{riak_kv, index, fsm, create}, one}, spiral},
+     {index_fsm_active, {riak_kv, index, fsm, active}, counter},
+     {list_fsm_create, {{riak_kv, list, fsm, create}, one}, spiral},
+     {list_fsm_create_error, {{riak_kv, list, fsm, create}, one}, spiral},
```
russelldb: Should be `{riak_kv, list, fsm, create, error}`

src/riak_kv_stat_bc.erl
```diff
@@ -232,7 +232,13 @@ legacy_stat_map() ->
      {coord_redirs_total, {riak_kv,node,puts,coord_redirs}, counter},
      {executing_mappers, {riak_kv,mapper_count}, counter},
      {precommit_fail, {riak_kv, precommit_fail}, counter},
-     {postcommit_fail, {riak_kv, postcommit_fail}, counter}
+     {postcommit_fail, {riak_kv, postcommit_fail}, counter},
+     {index_fsm_create, {{riak_kv, index, fsm, create}, one}, spiral},
+     {index_fsm_create_error, {{riak_kv, index, fsm, create}, one}, spiral},
```
russelldb: Should be `{riak_kv, index, fsm, create, error}`

russelldb:
Running each of bucket list, key list of index query and then running stats shows incorrect results. The stats output shows that for each [index | list]_fsm_create there is an [index | list]_fsm_create_error. I've highlighted the lines above that are the cause of this (incorrect stat used in error output.) Probably just copypasta.

That aside, it looks good to me, and it works (checking folsom shows correct stats updated.)

src/riak_kv_stat_bc.erl
evanmcc: oops

evanmcc:
updated.

russelldb:
wfm +1 to merge. Many thanks!

PR #469, opened 2013-01-23
use tuple modules instead of parameterized modules

vinoski:
Erlang R16, coming soon, will do away with parameterized modules (see Issue
4 under http://www.erlang.org/news/35 for details). Change riak_client to
use tuple modules instead, since they will continue to be supported in R16
and beyond. These changes are backward compatible.

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_client: object used for access into the riak system
 %%
-%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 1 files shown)


engelsanchez:
Steve: any reason why you went with passing the client as the last parameter vs as the first? That would make it the same as riakc_pb_socket. Not that Erlang libraries are famous for consistency in that area, but...

vinoski:
The parameterized module feature always passed its module argument as the last argument. I therefore made the tuple module the last argument as well to keep it completely backwards compatible for existing callers.

engelsanchez:
So they passed it as the last argument behind the curtains, blergh. Too bad. We'll get a lot of flak for this client incompatibility in the future :(

engelsanchez:
For future readers of this thread: my comments above were due to my not knowing what "tuple modules" are. Total non-issue.

cmeiklejohn:
@engelsanchez any outstanding concerns here?  also, how far did you get with the testing of this so I know where to continue?

engelsanchez:
@cmeiklejohn This was a total drive by comment, sorry. Start from scratch

cmeiklejohn:
:+1: looks good, builds and passes eunit.

evanmcc:
have rebased this and pulled it into #602.

PR #472, opened 2013-01-28
Various minor AAE changes prompted by Riak 1.3 QA

jtuple:
This pull-request contains three changes to AAE based on on-going 1.3 release testing and LevelDB changes.

First, this pull-request makes calls between KV vnodes and `riak_kv_index_hashtree` synchronous, using calls rather than casts. This provides backpressure to the vnode if the hashtree writes cannot occur fast enough. Technically, this shouldn't happen. It happened during tests because LevelDB was unfairly throttling AAE writes, preventing AAE from making progress. This has since been changed. However, we want to play it safe and add backpressure anyway. Of course, this backpressure just shifts the problem. If hashtree writes can't happen fast enough, then we will have unbounded message queue growth somewhere. Without this pull-request, the growth would occur in a `riak_kv_index_hashtree` process, with this patch the growth occurs in a `riak_kv_vnode` process. However, because of the new KV health check code in Riak 1.3, Jon/I decided that having potential message queue growth in the KV vnode is preferred. If this scenario happens, the health check will notice, kick in, and disable the KV service on the node.

NOTE: It is intentional that I use 'catch' when generating the now synchronous calls. This is because the previous use of cast was assumed in the design, and there are scenarios where the `riak_kv_index_hashtree` process may be dead when a message is sent. For a cast, the message would be dropped. For a call, it will generate an exception. We catch the exception as to not crash the KV vnode. Easier to do this rather than re-writing/re-validating all of AAE to guarantee never passing in a potentially dead PID. Future plans for 1.4 already include a completely different approach to backpressure, so this is largely a temporary change/situation anyway. 

Second, this pull-request adds code to ensure that all hashtree iterators are closed before attempting to close the LevelDB instance backing a group of hashtrees. This change was originally necessary as new LevelDB code changed semantics requiring iterators to be closed before a database could be closed/cleaned up. Very recent LevelDB changes have once again returned to the original semantics (closing a database handle will close/cleanup iterators automatically), but Jon/I/Matthew decided that committing this code anyway can't hurt and is a good "just in case".

Third, this pull-request updates how `hashtree.erl` calls `eleveldb:iterator_move` to account for how LevelDB now generates error messages. In the past, passing in an already closed iterator would generate an error tuple, now it generates a `badarg` exception. The change to hashtree ensure that these exceptions are caught and ignored as desired, as hashtree was already coded with the assumption that `iterator_move` may be called on already closed iterators in some scenarios.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -241,12 +241,22 @@ new({Index,TreeId}, LinkedStore, Options) ->
 
 -spec close(hashtree()) -> hashtree().
 close(State) ->
-    eleveldb:close(State#state.ref),
-    State.
+    close_iterator(State#state.itr),
+    catch eleveldb:close(State#state.ref),
+    State#state{itr=undefined}.
+
+close_iterator(Itr) ->
+    try
+        eleveldb:iterator_close(Itr)
+    catch
+        _:_ ->
+            ok
+    end.
 
 -spec destroy(hashtree()) -> hashtree().
 destroy(State) ->
-    eleveldb:close(State#state.ref),
+    %% Assumption: close was already called on all hashtrees that
+    %%             use this LevelDB instance,
     ok = eleveldb:destroy(State#state.path, []),
     State.
 
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -633,7 +633,7 @@ multi_select_segment(#state{id=Id, itr=Itr}, Segments, F) ->
                _ ->
                    encode(Id, First, <<>>)
            end,
-    IS2 = iterate(eleveldb:iterator_move(Itr, Seek), IS1),
+    IS2 = iterate(iterator_move(Itr, Seek), IS1),
     #itr_state{current_segment=LastSegment,
                segment_acc=LastAcc,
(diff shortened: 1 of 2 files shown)


jtuple:
Concerning the call/casts, I really want to move away from using calls for backpressure in 1.4. The plan would be to return to using casts, but only allow X writes to to the AAE tree before stopping to wait on an ACK from the tree. Ie. the standard backpressure approach we typically use in Riak. The problem is that the current protocol between the KV vnode and the hash tree is stateless, so it's much easier to just add the calls for now to do the "minimal change necessary" to get something working so late in the 1.3 cycle.

No issue deleting the handle_casts for now. They'll just return in one form or another for 1.4.

The match and spec change sound reasonable. Might as well make our expected input/expected preconditions explicit. Will push those changes out tomorrow.

slfritchie:
Spiffy.  The {noreply, State} hackery to share the handle_cast code is an optional hack.  Otherwise, +1 from me with a too-late-brain-not-working-didn't-actually-execute-the-stuff caveat.

slfritchie:
+1, many thanks!

PR #473, opened 2013-01-29
Fix AAE to not attempt to repair non-divergent keys over and over

jtuple:
When AAE exchange determines keys are divergent, AAE triggers read repair to repair divergent keys. This repairs the divergent replicas as well as updates the respective AAE trees. However, if an AAE tree is divergent from the underlying KV data, it it is possible that AAE will try to repair keys that do not have divergent KV replicas. In that case, read repair is never triggered, and therefore the AAE trees are never updated. In this case, AAE attempts to repair the same non-divergent keys over and over.

This commit changes AAE such that any key repair also triggers an update of the respective AAE trees, regardless of if read repair was triggered or not. As such, repair for non-divergent keys will be triggered at most once.

### DIFF
--- a/src/riak_kv_exchange_fsm.erl
+++ b/src/riak_kv_exchange_fsm.erl
@@ -227,7 +227,7 @@ read_repair_keydiff(RC, LocalVN, RemoteVN, {_, KeyBin}) ->
     %%       redbug to trace read_repair_keydiff when needed. Of course,
     %%       users can't do that.
     %% lager:debug("Anti-entropy forced read repair: ~p/~p", [Bucket, Key]),
-    RC:get(Bucket, Key, [{r, all}]),
+    RC:get(Bucket, Key),
     %% Force vnodes to update AAE tree in case read repair wasn't triggered
     riak_kv_vnode:rehash([LocalVN, RemoteVN], Bucket, Key),
     ok.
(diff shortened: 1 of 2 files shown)


slfritchie:
Joe, what happens to read_repair_keydiff() when it's blocked by slow reponses to the Riak get op?  Prior to this patch, it could still be hosed/slowed by a quorum of slow vnodes, but with {r,all} you can be delayed significantly by a single vnode.  #lazyweb I haven't looked yet but will try to figure out what might get congested while waiting.

slfritchie:
Hrm, so the entropy manager starts the exchange FSM to run in a separate Erlang process to do the exchange between two vnodes.  So if that one is bogged down by a bad-by-being-slow actor, only a single exchange FSM proc is affected ... well, all procs doing exchanges are likely to be affected by such a bad actor for the same reason.

Is the change to {r,all} meant to avoid keeping extra Erlang processes from quasi-leaking while waiting for laggard responses?

jtuple:
Exchange FSMs are designed to be passive entities that impact normal Riak performance as little as possible. If the get is slow, that means we repair keys at a slower rate. Something that we're willing to live with in the design. The logic by the change to R=all was to ensure that we didn't rehash the object until after any potential read repair had already occurred. Of course, after thinking about things again, I'm not sure that matters. If we rehash before the read repair, then the subsequent read repair will cause the newer hash to be written in KV write path. So, no reason I can't change this back to the unspecified, R=quorum case. Any strong opinions either way?

slfritchie:
Your logic about the tree update via read repair seems sound: extra work but not really harmful.  I think {r,quorum}/unspecified should return.

slfritchie:
+1, many thanks!

PR #474, opened 2013-01-30
Fix AAE rehash to work when data is missing

jtuple:


jaredmorrow:
+1 to the code

jaredmorrow:
Proxy +1 from @argv0 who tested the fix

PR #334, opened 2013-01-31
Every read triggers a read-repair when Last-write-wins=true

engelsanchez:
When LWW=true on a bucket, every replica increments the vclock. On read a read-repair is triggered due to mismatched vclocks at each replica, which means a put, and an incremented vclock, so the next read triggers a read-repair (and so on ad infinitum.)

This does not occur with an index capable backend, it is a product of the fast path optimisation for LWW to avoid a read before write https://github.com/basho/riak_kv/blob/master/src/riak_kv_vnode.erl#L650

```
LWW = true and backend is not index capable.
```

Funnily, we get a write after every read rather than a read before every write.

Potential fix is only increment vclock on the co-ordinating vnode.

Thanks to Martin Sumner at the NHS for the catch.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -794,6 +794,7 @@ prepare_put(State=#state{vnodeid=VId,
                          modstate=ModState},
             PutArgs=#putargs{bkey={Bucket, _Key},
                              lww=LWW,
+                             coord=Coord,
                              robj=RObj,
                              starttime=StartTime}) ->
     %% Can we avoid reading the existing object? If this is not an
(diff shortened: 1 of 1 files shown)


engelsanchez:
The problem is real.  You can verify it by running the riak_test referenced above: verify_no_writes_on_read before and after the fix. Alternatively, simply do the steps by hand: 
- start a node with bitcask backend
- set bucket property last_write_wins to true
- write one object
- repeatedly read the same object

You should see the write and read repair stats go up. You can also redbug the bitcask backend put operation and you'll see those going by every time you re-read the object.

engelsanchez:
The riak_test to verify this is not in master yet. Use the eas-fix-lww-writes-on-read branch if you want to run it (verify_no_writes_on_read).

tisba:
@engelsanchez any idea when this will land in riak? 1.3.1 does not have this, correct?

slfritchie:
Am I reading the history correctly, that this PR was merged to master about a day after the 1.3.x branch was created?

engelsanchez:
@tisba @slfritchie this was my first bit of work for the 1.4 release, which was supposed to go into code freeze this week. The 1.3.1 bugfix release was all about a specific data corruption problem (2i bigints encoding) and only major but very simple fixes went along with that. Since now we are working on a 1.3.2 bugfix release which delays 1.4 and this fix has gotten some visibility, I will make sure it goes out with that.

tisba:
@engelsanchez this is great, thanks!

PR #475, opened 2013-02-01
Using riak_pipe for Map Reduce task may return inconsistent results when one of cluster node is down. 

dreyk:
Using riak_pipe for Map Reduce task may return inconsistent results(for example empty list,however all data are available).
1. We must generate constant hash for the output reduce pase taking account of which nodes are up.
2. We must return an error If can't send result of reduce(prereduce phase to the output)  because some node are failed during map reduce job and we can't send result to it.

See http://lists.basho.com/pipermail/riak-users_lists.basho.com/2013-January/010870.html

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -323,12 +323,12 @@ mr2pipe_phases([]) ->
 mr2pipe_phases(Query) ->
     %% now() is used as a random hash to choose which vnode to collect
     %% the reduce inputs
-    Now = now(),
+    ConstantHashCookie = const_hash_cookie(),
 
     %% first convert phase
     QueryT = list_to_tuple(Query),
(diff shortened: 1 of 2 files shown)


beerriot:
As the github history indicates, I cleaned up the first patch in this PR (for the constant reduce hash) for incorporation into Riak's 1.3 RC2. We'll review the second patch (for handling reduce phase errors) separately.

Thank you!

evanmcc:
@beerriot Did the second patch ever get merged?

beerriot:
@evanmcc Nope, it sure didn't. Want to submit a PR?

slfritchie:
Pulling PRs out of the freezer ... Hrm, looks like an impasse.  Does the 2nd patch exist someplace?

evanmcc:
I don't think so.  It'd be a process of looking at this PR, and Bryan's
part 1 PR and diffing them and cleaning up/updating.

yorkshire-steve:

PR #476, opened 2013-02-02
Take node liveness into account during contant hash choice for reduce phase

beerriot:
This PR builds on work submitted in #475 (I'd suggest squashing on merge). I'd like to have this reviewed for inclusion in the RC2 of Riak 1.3, since it fixes a bug in pipe's MR emulation that was handled by the now-removed luke system.

Before this PR, generating a constant hash for a reduce phase in an MR query did not take node-liveness into account. We use `nval=1` in reduce phase fittings to make sure that all inputs end up at one vnode, instead of accidentally being spread due to liveness flickering. But, this meant that a down vnode might be chosen to reduce, and this could cause surprising (incorrect) results in some cases.

The test submitted in basho/riak_test#195 is a reproducer. It should fail before this PR, and pass after it.

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -321,14 +321,12 @@ mr2pipe_phases([]) ->
                    module=riak_pipe_w_pass,
                    chashfun=follow}];
 mr2pipe_phases(Query) ->
-    %% now() is used as a random hash to choose which vnode to collect
-    %% the reduce inputs
-    ConstantHashCookie = const_hash_cookie(),
+    ConstantHash = random_constant_hash(),
 
     %% first convert phase
     QueryT = list_to_tuple(Query),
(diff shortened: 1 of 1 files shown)


Vagabond:
Eyeballed the code at @jaredmorrow 's request. +1 on a read-through, Jared ran the tests.

jaredmorrow:
#### Tested the following on OSX:

##### Before PR applied
- basho/riak_test#195 failed (repeated x 5)
- mapred_verify_rt test passed

##### With #476 applied
- basho/riak_test#195 passed (repeated x 5)
- mapred_verify_rt test passed 
- verify_link_walk_urls test passed

#### Tested the following on Centos 6 64:
- basho/riak_test#195 passed
- mapred_verify_rt test passed
- verify_link_walk_urls test passed
- client_java_verify test passed

+1

PR #478, opened 2013-02-04
Improve interaction between AAE and K/V deletion

jtuple:
This pull-request improves AAE's handling around object deletion.

First, the pull-request changes the rehash logic in `riak_kv_vnode` to delete AAE hashes if an object is found to be missing. Thus, if a K/V delete fails to update the relevant AAE hashtree, the hashtree will be fixed on the first repair. Without this change, AAE could continually attempt to repair keys that have been deleted without informing the relevant AAE hashtree. This is scenario is hard to trigger in practice, but it's better to be safe as the fix is easy and straightforward.

Second, this pull-request changes the AAE hashtrees to buffer deletes using the same logic used for buffering writes. Thus, both deletes and writes are buffered and submitted to LevelDB as a single batch operation. As previously shown, write-batching is necessary to make AAE have low impact on the cluster, thus this change is also important for users that frequently delete data.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -271,17 +271,21 @@ insert(Key, ObjHash, State, Opts) ->
     HKey = encode(State#state.id, Segment, Key),
     case should_insert(HKey, Opts, State) of
         true ->
-            WBuffer = [{HKey, ObjHash} | State#state.write_buffer],
-            WCount = State#state.write_buffer_count + 1,
-            State2 = State#state{write_buffer=WBuffer,
-                                 write_buffer_count=WCount},
-            State3 = maybe_flush_buffer(State2),
-            %% Dirty = gb_sets:add_element(Segment, State3#state.dirty_segments),
-            Dirty = bitarray_set(Segment, State3#state.dirty_segments),
-            State3#state{dirty_segments=Dirty};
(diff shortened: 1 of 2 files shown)


evanmcc:
code looks fine, any testing suggestions?

jrwest:
was about to ask same question re: testing. +1 to code, clearly fixes potential bug with deletions getting re-ordered/lost.

jtuple:
Easiest way to test is to change the code to occasionally drop things. Was planning on writing an intercept or meck to do this, but didn't get around to it. But, easy enough to just modify the code (that's how I tested things).

Take this line:
https://github.com/basho/riak_kv/blob/jdb-aae-deleted/src/riak_kv_vnode.erl#L781

Change to something like (randomly dropping 2/3 AAE deletes):

``` erlang
(random:uniform(3) == 1) orelse riak_kv_index_hashtree:delete(BKey, State#state.hashtrees),
```

Write a bunch of data. Then delete said data. Should see repairs start happening (some vnodes saw deletes, others did not). But, repairs should eventually stop after enough rehashing calls fix things up. If you do similar modification + test against master / 1.3.0, you'll see the repairs continue forever.

jrwest:
confirmed described behavior in local test. +1

PR #479, opened 2013-02-07
More Compact Riak Object Binary Format

jrwest:
This PR changes how `riak_object`s are serialized before storing them on disk, with the goal of reducing overhead. The current `term_to_binary` approach used results in a significant amount of unnecessary overhead due to the metadata dict, among other things. This is especially true for small objects. It also means we store the bucket and key twice for each object (once as the actual key in the backend, and once in the serialized object). The new format addresses these issues by building its own binary that more compactly stores the metadata dictionary in addition to not including the bucket and key.

To support the new format several other changes were made: `riak_kv_vnode` uses a new capability to determine when the new format is supported and encodes the object for writes using the version negotiated by capabilities. Several sub-systems, e.g. aae, backup, handoff were also changed to account for the new version. The commit notes have specific details on what was changed and why.

Additionally, in order for rolling downgrades to be possible after new objects have been written, `riak_kv_format` is included in this PR with the intention of it being run on a live node before starting the downgrade. A `riak_test` for this can be found [here](https://github.com/basho/riak_test/pull/197). `riak_kv_format` is general enough that it can be used to upgrade the node's data as well.

Finally, this PR includes some general cleanup of the `riak_object` module.

There are several possible improvements, such as some potential binary optimizations, that could be made before this PR is merged or in a separate PR. I figured those were best left to discussion with a reviewer.

EDIT: much thanks to @argv0 @buddhisthead and @seancribbs who also worked on this branch. The author history was lost as part of rebasing many commits from the portland meetup.

### DIFF
--- a/src/riak_kv_backup.erl
+++ b/src/riak_kv_backup.erl
@@ -93,16 +93,20 @@ backup_vnodes([Partition|T], Node) ->
     receive stop -> stop end,
     backup_vnodes(T, Node).
 
-backup_folder(_, V, Pid) ->
-    Pid ! {backup, V},
+backup_folder(K, V, Pid) ->
+    Pid ! {backup, {K, V}},
     Pid.
 
 result_collector(PPid) ->
     receive
         stop -> 
             PPid ! stop;
-        {backup, M} when is_binary(M) ->
-            disk_log:log(?TABLE, M),
+        {backup, {{B, K}, M}} when is_binary(M) ->
+            %% make sure binary is encoded using term_to_binary (v0)
+            %% not v1 format. restore does not have access to bucket/key
+            %% so must include them in encoded format, which v1 does not
+            ObjBin = riak_object:to_binary_version(v0, B, K, M),
+            disk_log:log(?TABLE, ObjBin),
             result_collector(PPid)
     end.
 
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -107,10 +107,11 @@ insert(_Id, _Key, _Hash, undefined, _Options) ->
 insert(Id, Key, Hash, Tree, Options) ->
     catch gen_server:call(Tree, {insert, Id, Key, Hash, Options}, infinity).
 
-%% @doc Add a term_to_binary encoded riak_object associated with a given
+%% @doc Add an encoded (binary) riak_object associated with a given
 %%      bucket/key to the appropriate hashtree managed by the provided
 %%      index_hashtree pid. The hash value is generated using
-%%      {@link hash_object/1}.
+%%      {@link hash_object/2}. Any encoding version is supported. The encoding
+%%      will be changed to the appropriate version before hashing the object.
 -spec insert_object({binary(), binary()}, riak_object_t2b(), pid()) -> ok.
 insert_object(_BKey, _RObj, undefined) ->
     ok;
@@ -240,7 +241,7 @@ handle_call({insert, Id, Key, Hash, Options}, _From, State) ->
 handle_call({insert_object, BKey, RObj}, _From, State) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     IndexN = riak_kv_util:get_index_n(BKey, Ring),
-    State2 = do_insert(IndexN, term_to_binary(BKey), hash_object(RObj), [], State),
+    State2 = do_insert(IndexN, term_to_binary(BKey), hash_object(BKey, RObj), [], State),
     {reply, ok, State2};
 handle_call({delete, BKey}, _From, State) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
@@ -300,7 +301,7 @@ handle_cast(stop, State) ->
 handle_cast({insert_object, BKey, RObj}, State) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     IndexN = riak_kv_util:get_index_n(BKey, Ring),
-    State2 = do_insert(IndexN, term_to_binary(BKey), hash_object(RObj), [], State),
+    State2 = do_insert(IndexN, term_to_binary(BKey), hash_object(BKey, RObj), [], State),
     {noreply, State2};
 
 handle_cast(build_failed, State) ->
@@ -381,13 +382,13 @@ load_built(#state{trees=Trees}) ->
     end.
 
 %% Generate a hash value for a binary-encoded `riak_object'
--spec hash_object(riak_object_t2b()) -> binary().
-hash_object(RObjBin) ->
+-spec hash_object({riak_object:bucket(), riak_object:key()}, riak_object_t2b()) -> binary().
+hash_object({Bucket, Key}, RObjBin) ->
     %% Normalize the `riak_object' vector clock before hashing
-    RObj = binary_to_term(RObjBin),
+    RObj = riak_object:from_binary(Bucket, Key, RObjBin),
     Vclock = riak_object:vclock(RObj),
     UpdObj = riak_object:set_vclock(RObj, lists:sort(Vclock)),
-    Hash = erlang:phash2(term_to_binary(UpdObj)),
+    Hash = riak_object:hash(UpdObj),
     term_to_binary(Hash).
 
 %% Fold over a given vnode's data, inserting each object into the appropriate
@@ -402,7 +403,7 @@ fold_keys(Partition, Tree) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     Req = ?FOLD_REQ{foldfun=fun(BKey={Bucket,Key}, RObj, _) ->
                                     IndexN = riak_kv_util:get_index_n({Bucket, Key}, Ring),
-                                    insert(IndexN, term_to_binary(BKey), hash_object(RObj),
+                                    insert(IndexN, term_to_binary(BKey), hash_object(BKey, RObj),
                                            Tree, [if_missing]),
                                     ok
                             end,
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -386,9 +386,7 @@ load_built(#state{trees=Trees}) ->
 hash_object({Bucket, Key}, RObjBin) ->
     %% Normalize the `riak_object' vector clock before hashing
     RObj = riak_object:from_binary(Bucket, Key, RObjBin),
-    Vclock = riak_object:vclock(RObj),
-    UpdObj = riak_object:set_vclock(RObj, lists:sort(Vclock)),
-    Hash = riak_object:hash(UpdObj),
+    Hash = riak_object:hash(RObj),
     term_to_binary(Hash).
 
 %% Fold over a given vnode's data, inserting each object into the appropriate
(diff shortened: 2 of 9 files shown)


buddhisthead:
+1 on the tests. I can't really be the main reviewer since I contributed to the early code changes.

src/riak_kv_backup.erl
```diff
@@ -93,16 +93,19 @@ backup_vnodes([Partition|T], Node) ->
...
     Pid.
 
 result_collector(PPid) ->
     receive
         stop -> 
             PPid ! stop;
-        {backup, M} when is_binary(M) ->
-            disk_log:log(?TABLE, M),
+        {backup, {{B, K}, M}} when is_binary(M) ->
+            %% make sure binary is encoded using term_to_binary (v0)
```
russelldb, line 105: One day someone will ask "why?", maybe that day is today.

src/riak_kv_index_hashtree.erl
```diff
@@ -102,10 +102,11 @@ insert(Id, Key, Hash, Tree) ->
 insert(Id, Key, Hash, Tree, Options) ->
     gen_server:cast(Tree, {insert, Id, Key, Hash, Options}).
 
-%% @doc Add a term_to_binary encoded riak_object associated with a given
+%% @doc Add a encoded (binary) riak_object associated with a given
```
russelldb: super_picky "Add an encoded…"

src/riak_kv_vnode.erl
```diff
@@ -876,6 +894,37 @@ perform_put({true, Obj},
...
+                               reqid=undefined,
+                               index_specs=IndexSpecs},
+            case perform_put({true, RObj}, State, PutArgs) of
+                {{fail, _, _}, UpdState}  ->
+                    Reply = {error, backend_error};
+                {_, UpdState} ->
+                    Reply = ok
+            end
+    end,
+    {Reply, UpdState}.
```
russelldb, line 964: Stylistic thing: how do you feel about exporting `Repy` and `UpdState` from case statements? I try and avoid it just 'cos my editor complains. Personally I think it requires more reading to track down the source of the bindings.

src/riak_kv_vnode.erl
```diff
@@ -876,6 +894,37 @@ perform_put({true, Obj},
...
+        {error, not_found, _UpdModState} ->
+            Reply = {error, not_found},
+            UpdState = State;
+        {ok, ObjBin, _UpdModState} ->
+            %% since it is assumed capabilities have been properly set
+            %% to the desired version, to reformat, all we need to do
+            %% is submit a new write
+            RObj = riak_object:from_binary(Bucket, Key, ObjBin),
+            {ok, Capabilities} = Mod:capabilities(Bucket, ModState),
+            IndexBackend = lists:member(indexes, Capabilities),
```
russelldb: Why do we need to re-write the indexes?

src/riak_object.erl
```diff
@@ -129,13 +138,29 @@ equal_contents([C1|R1],[C2|R2]) ->
         andalso (C1#r_content.value =:= C2#r_content.value)
         andalso equal_contents(R1,R2).
 
-%% @spec reconcile([riak_object()], boolean()) -> riak_object()
+
```
russelldb, line 141: I see that the PR description includes a mention of "general cleanup". I find it harder to review with these non-pertinent changes mixed in. IM(not so)HO a separate PR for all these `@spec` to `-spec` changes would have been easier to review.

src/riak_object.erl
```diff
@@ -541,6 +559,181 @@ syntactic_merge(CurrentObject, NewObject) ->
...
+    <<1, Bin/binary>>;
+encode_maybe_binary(Bin) ->
+    <<0, (term_to_binary(Bin))/binary>>.
+
+decode_maybe_binary(<<1, Bin/binary>>) ->
+    Bin;
+decode_maybe_binary(<<0, Bin/binary>>) ->
+    binary_to_term(Bin).
+
+contents_size(Vsn, Contents) ->
```
russelldb: Sorry if I am the worlds pickiest reviewer, but this would be so much easier to read if it was at L570+ (directly where it is used in calculating size)

src/riak_object.erl
```diff
@@ -541,6 +559,181 @@ syntactic_merge(CurrentObject, NewObject) ->
...
+bin_content(#r_content{metadata=Meta, value=Val}) ->
+    ValBin = encode_maybe_binary(Val),
+    ValLen = byte_size(ValBin),
+    MetaBin = meta_bin(Meta),
+    MetaLen = byte_size(MetaBin),
+    <<ValLen:32/integer, ValBin:ValLen/binary, MetaLen:32/integer, MetaBin:MetaLen/binary>>.
+
+bin_contents(Contents) ->
+    F = fun(Content, Acc) ->
+                <<Acc/binary, (bin_content(Content))/binary>>
```
russelldb, line 688: I've never seen this `(bin_content(Content))` before, is it to save setting an intermediate variable?

src/riak_object.erl
```diff
@@ -541,6 +559,181 @@ syntactic_merge(CurrentObject, NewObject) ->
...
+      Deleted:1/binary-unit:8, RestBin/binary>>.
+
+fold_meta_to_bin(?MD_VTAG, Value, {{_Vt,Del,Lm},RestBin}) ->
+    {{Value, Del, Lm}, RestBin};
+fold_meta_to_bin(?MD_LASTMOD, Value, {{Vt,Del,_Lm},RestBin}) ->
+     {{Vt, Del, Value}, RestBin};
+fold_meta_to_bin(?MD_DELETED, true, {{Vt,_Del,Lm},RestBin})->
+     {{Vt, <<1>>, Lm}, RestBin};
+fold_meta_to_bin(?MD_DELETED, "true", Acc) ->
+    fold_meta_to_bin(?MD_DELETED, true, Acc);
```
russelldb, line 715: Why not the same as the previous clause to save the call?

src/riak_kv_backup.erl
jrwest, line 105: Ah, "why?" is somewhere in the commit notes but probably best as a comment. will add that.

src/riak_kv_index_hashtree.erl
jrwest: proper use of a/an is always good :). will fix when i make changes to the kv backup comment

src/riak_object.erl
```diff
@@ -541,6 +559,181 @@ syntactic_merge(CurrentObject, NewObject) ->
...
+sibs_of_binary(0, <<>>, Result) -> lists:reverse(Result);
+sibs_of_binary(0, _NotEmpty, _Result) ->
+    {error, corrupt_contents};
+sibs_of_binary(Count, SibsBin, Result) ->
+    {Sib, SibsRest} = sib_of_binary(SibsBin),
+    sibs_of_binary(Count-1, SibsRest, [Sib | Result]).
+
+sib_of_binary(<<ValLen:32/integer, ValBin:ValLen/binary, MetaLen:32/integer, MetaBin:MetaLen/binary, Rest/binary>>) ->
+    <<LMMega:32/integer, LMSecs:32/integer, LMMicro:32/integer, VTagLen:8/integer, VTag:VTagLen/binary, Deleted:1/binary-unit:8, MetaRestBin/binary>> = MetaBin,
+    DeletedVal = case Deleted of <<1>> -> true; _False -> false end,
```
russelldb: I re-factored this section to

https://github.com/russelldb/riak_kv/compare/basho:jrw-pdx-binrobj-rebased2...russelldb:rdb-refctor-sib-of-binary

Dunno if you want me to push that, or if you prefer yours. Again it is just a stylistic thing, but there for discussion.

src/riak_kv_vnode.erl
jrwest, line 964: I actually prefer it too and get the same warning but most of our code I was looking at did it and I wasn't sure if it was a whitespace indentation thing. Up to change it if you prefer.

russelldb:
Pending answers to questions +1 from me. Code works. Tests pass.

Basho benches on my limited set up on bolth level and bitcask show no discernible performance degradation, maybe there was even a slight perf increase (especially for small object values.) Also, for a million 1kish objects, the level folder was nearly a gig smaller on each node.

src/riak_kv_vnode.erl
jrwest: I was just extracting them here because they are necessary to pass to `perform_put` which now that I think about it probably does rewrite the indexes, which is not ideal, although probably not harmful. The reason I didn't want to add yet another put function is we have so many already. Should be changed I think though.

russelldb: Just pass  the empty list and all will remain as before, (edit) Erm, I think. best to double check.

src/riak_object.erl
jrwest: wfm, reads much nicer

src/riak_object.erl
jrwest, line 715: stylistic thing i guess. Usually I prefer to have only one version of the function that returns when converting a value (like `"true" -> true`) because if the return value changes it only requires changing one function clause. Can be convinced otherwise if you have strong feelings the other way.

src/riak_object.erl
jrwest: I can move it up there. I always have a hard time separating private functions from public API ones for that reason. No objection to moving it up though

src/riak_object.erl
jrwest, line 688: yea, need the extra parens to make `erlc` happy

src/riak_kv_vnode.erl
jrwest: thanks for the tip. will verify and give it a shot

src/riak_object.erl
jrwest, line 141: ah my bad. you are correct. OCD got the best of me after staring at the file for so long. It should be easy enough with a rebase to back out those changes if you want...There is only one thing from the last commit that really should be in this PR (the sorting of vnode entries in `riak_object:hash/1`)

src/riak_kv_vnode.erl
russelldb, line 964: I'm easy either way…I agree it is good to have the code in keeping with the general conventions of the larger corpus. I prefer not to export from case statements. Really up to you. I guess leave for now.

src/riak_object.erl
russelldb, line 715: Ah, good reason. The 'optimisation' of _not_ doing is probably far less than the read/maintainability. I like your way. Let's leave it.

src/riak_kv_vnode.erl
jrwest: @rdb  so it looks like handling of the IndexSpecs is delegated to the kv backend on `Mod:put`. if level doesn't update the index we may be able to take advantage of that by passing `[]` but not sure that is safe to rely upon. Thoughts?

russelldb: @jrwest so, index specs are of the form `{add, name, value}` and `{remove, name, value}`, the backend uses these as instructions for what to do (write or delete indexes) in this case, isn't it that only `adds` will be generated? And aren't those `adds` for indexes that are already added (since your just reading and writing the object locally) So, I guess the question is 'is the empty list a valid index spec to pass for all backends?' I'm pretty sure the answer is 'yes'

jrwest:
force pushed an update to the branch w/ comments addressed. @russelldb I squashed your changes into the cleanup commit

src/riak_kv_vnode.erl
reiddraper, line 964: I'll throw my opinion in the ring and say exporting from case is A Bad Idea.

russelldb:
WFM, +1. Nice work.

PR #482, opened 2013-02-08
Randomize the LevelDB write_buffer_size used for AAE

jtuple:
LevelDB as used for normal Riak K/V data randomizes the write_buffer_size to protect against multiple buffers filling up simultaneously under heavy uniform load. This commit does the same for the LevelDB instances created by AAE.


chardan:
+1.

PR #483, opened 2013-02-08
Improve AAE backpressure on K/V vnode write-path

jtuple:
Original AAE design had `riak_kv_vnode` send async casts to the `riak_kv_index_hashtree` to decouple the two processes and ensure AAE had little impact on K/V performance. Testing uncovered issues where the K/V vnode could operate faster the AAE hashtree, leading to unbounded message queue growth. To fix this, the updates were changes to be synchronous calls to add backpressure.

However, doing a blocking call for every update impacts K/V performance.  This is especially noticeable when running a fast backend, such as Memory or Bitcask. This commit moves to a hybrid approach. A fixed number of updates are sent asynchronously after which the next update is performed synchronously. The value is currently hardcoded to 90 asynchronous requests. The worse-case message queue size is therefore 91 outstanding messages.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -96,6 +96,8 @@
 -type index() :: non_neg_integer().
 -type state() :: #state{}.
 
+-define(HASHTREE_TOKENS, 90).
+
 -record(putargs, {returnbody :: boolean(),
                   coord:: boolean(),
                   lww :: boolean(),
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -96,6 +96,8 @@
 -type index() :: non_neg_integer().
 -type state() :: #state{}.
 
+-define(HASHTREE_TOKENS, 90).
```
engelsanchez: Although this sounds reasonable, it might be a good idea to provide a dial on a change this late in the game and make this configurable? It would only be fetched every ~90 operations anyway, so I imagine performance of that should not be an issue (and if it was, it can always be cached in that stinky pdict anyway)

src/riak_kv_vnode.erl
```diff
@@ -1199,10 +1201,28 @@ do_diffobj_put({Bucket, Key}, DiffObj,
 
 -spec update_hashtree(binary(), binary(), binary(), state()) -> ok.
 update_hashtree(Bucket, Key, Val, #state{hashtrees=Trees}) ->
-    riak_kv_index_hashtree:insert_object({Bucket, Key}, Val, Trees).
+    case get_hashtree_token() of
+        true ->
+            riak_kv_index_hashtree:async_insert_object({Bucket, Key}, Val, Trees);
+        false ->
+            riak_kv_index_hashtree:insert_object({Bucket, Key}, Val, Trees),
+            put(hashtree_tokens, ?HASHTREE_TOKENS)
```
engelsanchez: The function no longer returns ok in this case. Not sure if that would be a problem, but...

jtuple:
Just pushed code that should address both comments.

engelsanchez:
Well, that's all I had :+1: :dancer:

PR #485, opened 2013-02-12
Fix PR/PW

Vagabond:
Originally, PR/PW simply checked, in the validate phase of the GET/PUT
FSM that the required number of primaries were online.

However, that was insufficient. Consider PW=2, PR=2, N=3 case; if 2 primaries
are online. You do the write, but one primary and one fallback receive
the write, the third write fails (but you've already returned a result
to the client, because 2 writes succeeded). Now, when you go to read
that key, 2 primaries are still online, but you read from the primary
that just replaced the fallback and the primary that failed the write
before. Congratulations, you just failed to read-your-own-writes.

It gets worse, even. Because the node_watcher can lag behind the actual
up/down state of a node, especially if the node kernel panics or has its
NIC unplugged, primaries can be reported up for some time after they've
stopped responding (until the distributed erlang connection closes).
This means that its even easier to fail PR/PW but still have the FSM
pass the validation stage. And since R and PR are unrelated, even PW=N
could spuriously validate (if R < PR).

The solution to all this is to simply use the preflist to check we get
the required number of responses from a primary before the FSM will
return. However, this required significant modification to the
Quickcheck tests. Additionally, some simple sanity check eunit tests
were added. Most of the actual changes in this branch are just
EQC/Eunit changes and refactoring.

This also means that PR/PW will now _block_ a response from returning
until they've been satisfied. Thusly, PR=2 now will wait for at least 2
replies, possibly all N if necessary, to satisfy the constraint, meaning
you don't need to specify PR _and_ R anymore. The same is true for PW,
except that PW waits for DW responses, not W responses, to satisfy the
constraint.

Additionally, the rules on how DW relates to W have been changed, DW is
always at least 1 (for the coordinating put), but it is no longer
demoted to match W.

### DIFF
--- a/src/riak_kv_get_core.erl
+++ b/src/riak_kv_get_core.erl
@@ -20,7 +20,7 @@
 %%
 %% -------------------------------------------------------------------
 -module(riak_kv_get_core).
--export([init/7, add_result/4, result_shortcode/1, enough/1, response/1,
+-export([init/8, add_result/3, result_shortcode/1, enough/1, response/1,
          has_all_results/1, final_action/1, info/1]).
 -export_type([getcore/0, result/0, reply/0, final_action/0]).
 
@@ -39,6 +39,7 @@
                         {read_repair, [{non_neg_integer() | repair_reason()}], riak_object:riak_object()} |
                         delete.
 -type idxresult() :: {non_neg_integer(), result()}.
+-type idx_type() :: [{non_neg_integer, 'primary' | 'fallback'}].
 
 -record(getcore, {n :: pos_integer(),
                   r :: pos_integer(),
@@ -54,7 +55,8 @@
                   num_pok = 0 :: non_neg_integer(),
                   num_notfound = 0 :: non_neg_integer(),
                   num_deleted = 0 :: non_neg_integer(),
-                  num_fail = 0 :: non_neg_integer()}).
+                  num_fail = 0 :: non_neg_integer(),
+                  idx_type :: idx_type()}).
 -opaque getcore() :: #getcore{}.
 
 %% ====================================================================
@@ -62,20 +64,23 @@
 %% ====================================================================
 
 %% Initialize a get and return an opaque get core context
--spec init(pos_integer(), pos_integer(), pos_integer(), pos_integer(), boolean(), boolean(),
-           boolean()) -> getcore().
-init(N, R, PR, FailThreshold, NotFoundOk, AllowMult, DeletedVClock) ->
+-spec init(N::pos_integer(), R::pos_integer(), PR::pos_integer(),
+           FailThreshold::pos_integer(), NotFoundOK::boolean(),
+           AllowMult::boolean(), DeletedVClock::boolean(),
+           IdxType::idx_type()) -> getcore().
+init(N, R, PR, FailThreshold, NotFoundOk, AllowMult, DeletedVClock, IdxType) ->
     #getcore{n = N,
              r = R,
              pr = PR,
              fail_threshold = FailThreshold,
              notfound_ok = NotFoundOk,
              allow_mult = AllowMult,
-             deletedvclock = DeletedVClock}.
+             deletedvclock = DeletedVClock,
+             idx_type = IdxType}.
 
 %% Add a result for a vnode index
--spec add_result(non_neg_integer(), result(), primary | fallback, getcore()) -> getcore().
-add_result(Idx, {ok, RObj} = Result, Status, GetCore) ->
+-spec add_result(non_neg_integer(), result(), getcore()) -> getcore().
+add_result(Idx, {ok, RObj} = Result, GetCore) ->
     Dels = case riak_kv_util:is_x_deleted(RObj) of
         true ->  1;
         false -> 0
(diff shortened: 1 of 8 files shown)


Vagabond:
cc @russelldb

Vagabond:
I just re-ran the scenario described here:

http://lists.basho.com/pipermail/riak-users_lists.basho.com/2012-January/007034.html

I'm unable to reproduce the bad behaviour that was reported. Although, note that test/foo on my ring hashed to a preflist that had 2 members on one node, test2/foo hashed better for me, so that is what I used.

I switched back to the master branch, and the bug came back.

src/riak_kv_get_core.erl
```diff
@@ -222,7 +249,159 @@ merge(Replies, AllowMult) ->
...
+                    ?assertMatch({{error, {pr_val_unsatisfied, 3, 1}}, _},
+                        response(#getcore{n= 3, r = 2, pr=3,
+                                fail_threshold = 1, num_ok = 1, num_pok = 1,
+                                allow_mult = false,
+                                num_notfound = 0, num_deleted = 0,
+                                num_fail = 2,
+                                results= [
+                                    {1, {ok, RObj}},
+                                    {2, {ok, RObj}},
+                                    {3, {ok, RObj}}]})),
```
beerriot: Small nit: this test setup lies a bit, doesn't it? There should be no way for `add_result/4` to produce a list of three `{ok, Obj}` results with `num_ok+num_pok` less than three, I think. Even more so with `num_fail > 0` as well, no? The same goes for the following two tests as well.

src/riak_kv_get_fsm.erl
```diff
@@ -67,6 +67,7 @@
                 startnow :: {non_neg_integer(), non_neg_integer(), non_neg_integer()},
                 get_usecs :: non_neg_integer(),
                 tracked_bucket=false :: boolean(), %% is per bucket stats enabled for this bucket
+                idx_type :: {non_neg_integer(), 'primary' | 'fallback'}, %% mapping of index to partition type (primary | fallback)
```
beerriot: Looks like this is actually a list of that tuple.

src/riak_kv_put_core.erl
```diff
@@ -47,7 +53,10 @@
                   final_obj :: undefined | riak_object:riak_object(),
                   num_w = 0 :: non_neg_integer(),
                   num_dw = 0 :: non_neg_integer(),
-                  num_fail = 0 :: non_neg_integer()}).
+                  num_pw = 0 :: non_neg_integer(),
+                  num_fail = 0 :: non_neg_integer(),
+                  idx_type :: {non_neg_integer(), 'primary' | 'fallback'} %% mapping of idx -> primary | fallback
```
beerriot: Why add `idx_type` to get_FSM, but put_CORE? Why not both FSMs, or both cores?

src/riak_kv_put_core.erl
```diff
@@ -56,14 +65,19 @@
...
-           pos_integer(), pos_integer(), boolean(), boolean()) -> putcore().
-init(N, W, DW, WFailThreshold, DWFailThreshold, AllowMult, ReturnBody) ->
-    #putcore{n = N, w = W, dw = DW,
+           non_neg_integer(), pos_integer(), pos_integer(),
+           pos_integer(), boolean(), boolean(),
+           orddict:orddict()) -> putcore().
+init(N, W, PW, DW, WFailThreshold, PWFailThreshold,
+     DWFailThreshold, AllowMult, ReturnBody, IdxType) ->
+    #putcore{n = N, w = W, pw = PW, dw = DW,
              w_fail_threshold = WFailThreshold,
```
beerriot: It looks like `w_fail_threshold` is no longer used. Get rid of it?

src/riak_kv_put_core.erl
```diff
@@ -153,3 +179,128 @@ maybe_return_body(PutCore = #putcore{returnbody = true}) ->
...
+
+
+
+-ifdef(TEST).
+%% simple sanity tests
+enough_test_() ->
+    [
+        {"Checking W",
+            fun() ->
+                    %% you can never fail W directly...
```
beerriot, line 204: There should be a test with `num_fail > 0`, if they are to prove that "you can never _fail_ W directly".

src/riak_kv_put_fsm.erl
```diff
@@ -274,11 +275,14 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
...
             %% The coord vnode is responsible for incrementing the vclock
-            DW = erlang:min(DW1, erlang:max(1, W))
+            DW = erlang:max(DW1, 1)
     end,
+
+    IdxType = [{Part, Type} || {{Part, _Node}, Type} <- Preflist2],
     NumPrimaries = length([x || {_,primary} <- Preflist2]),
     NumVnodes = length(Preflist2),
-    MinVnodes = erlang:max(1, erlang:max(W, DW)), % always need at least one vnode
+    MinVnodes = erlang:max(1, erlang:max(W, erlang:max(DW, PW))), % always need at least one vnode
```
beerriot: suggestion: `lists:max([1, W, DW, PW])`

src/riak_kv_put_core.erl
```diff
@@ -86,34 +100,46 @@ add_result(_Other, PutCore = #putcore{num_fail = NumFail}) ->
...
-            {{error, {w_val_unsatisfied, NumW, NumDW, W, DW}}, PutCore}
-    end.
+%% Perfect world - all quora met
+response(PutCore = #putcore{w = W, num_w = NumW, dw = DW, num_dw = NumDW, pw = PW, num_pw = NumPW}) when
+      NumW >= W, NumDW >= DW, NumPW >= PW ->
+    maybe_return_body(PutCore);
+%% Everything is ok, except we didn't meet PW
+response(PutCore = #putcore{w = W, num_w = NumW, dw = DW, num_dw = NumDW, pw = PW, num_pw = NumPW}) when
+      NumW >= W, NumDW >= DW, NumPW < PW ->
+    {{error, pw_val_unsatisfied, PW, NumPW}, PutCore};
```
beerriot: Why not `{error, {pw_val_unsatisfied, PW, NumPW}}` like the other error tuples? This would remove the need to modify `riak_kv_put_fsm:add_client_info`, I think.

src/riak_kv_get_core.erl
```diff
@@ -93,44 +110,54 @@ result_shortcode(_)                 -> -1.
...
+enough(#getcore{r = R, num_ok = NumOK, pr= PR, num_pok = NumPOK}) when
+      NumOK >= R andalso NumPOK >= PR ->
+    true;
+%% too many failures
+enough(#getcore{fail_threshold = FailThreshold, num_notfound = NumNotFound,
+        num_fail = NumFail}) when NumNotFound + NumFail >= FailThreshold ->
+    true;
+%% Got all N responses, but unable to satisfy quorum
+enough(#getcore{n = N, num_ok = NumOK, num_notfound = NumNotFound,
+        num_fail = NumFail}) when NumOK + NumNotFound + NumFail >= N ->
```
beerriot: This looks like a new case that wasn't checked before, but isn't specifically related to PR. Is it just a new short circuit? Does it have any other effects? Or, am I just missing how it was handled in old code?

src/riak_kv_put_core.erl
Vagabond, line 204: W failures don't increment num_fail, only DW failures do.

W results are always either 'good' or are not sent at all:

https://github.com/basho/riak_kv/blob/master/src/riak_kv_vnode.erl#L367

src/riak_kv_get_core.erl
Vagabond: No, we failed to meet PR here, we got N responses, but NumPOK was not high enough. Will update the comment.

Vagabond:
Riak test here: basho/riak_test#218

Vagabond:
Rebased and pushed a fix for webmachine error handling.

beerriot:
The given test makes sense, fails pre-patch, passes post-patch. I like it. +1!

PR #486, opened 2013-02-14
Make AAE default to 'off' when not configured

jtuple:
The intention of this commit is to make AAE be disabled by default
when performing an in-place upgrade of Riak, but be enabled by default
for new installs.

This change disables AAE when missing from the top-level app.config.
In Riak, the top-level app.config defaults to having AAE enabled, thus
overriding the default. However, users performing an in-place rolling
upgrade of Riak maintain their existing app.config which is missing
the override. Users who want to enable AAE after an upgrade will need
to add the appropriate settings to their app.config.

### DIFF
--- a/src/riak_kv.app.src
+++ b/src/riak_kv.app.src
@@ -36,8 +36,8 @@
          %% use the legacy routines for tracking kv stats
          {legacy_stats, true},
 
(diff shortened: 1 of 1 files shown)


jaredmorrow:
+1

PR #487, opened 2013-02-14
provide a Location header for the same api version on POST

beerriot:
If a POST comes in to /buckets/B/keys, the Location header generated
should be of the form /buckets/B/keys/K. If the post arrived at /riak/B,
the Location should be /riak/B/K.

This PR addresses @seancribbs concern in #432.

A test for this behavior is provided in basho/riak_test#205.


russelldb:
WFM, and test is good +1.

PR #491, opened 2013-02-20
Remove Link headers from bucket and key lists

seancribbs:
While webmachine/mochiweb only accepts headers up to 8KB in size, it has no problem producing huge headers. However, many clients cannot handle large headers. Since the same information is produced in the body of these requests anyway, we have no need to produce Link headers as well. If we want to really use HATEOAS, other things need to be reconsidered. Since we don't really have HATEOAS, let's not pretend anymore.

This also adds some ignores for EQC artifacts.

### DIFF
--- a/src/riak_kv_wm_keylist.erl
+++ b/src/riak_kv_wm_keylist.erl
@@ -128,9 +128,7 @@ encodings_provided(RD, Ctx) ->
 %%      is specified, keys will be streamed back to the client in JSON chunks
 %%      like so: {"keys":[Key1, Key2,...]}.
 %%      A Link header will also be added to the response by this function
-%%      if the keys are included in the JSON object.  The Link header
-%%      will include links to all keys in the bucket, with the property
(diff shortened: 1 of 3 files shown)


src/riak_kv_wm_keylist.erl
```diff
@@ -131,13 +131,10 @@ encodings_provided(RD, Ctx) ->
 %%      if the keys are included in the JSON object.  The Link header
```
russelldb: Shouldn't these comments be changed / expunged?

russelldb:
I never knew about this 'feature', madness.

Definitely +1 on removing the links. Technically changes the API…does that matter? We seem to do that anyway every release(I mean, we're not very semver). 

WFM +1 after that comments comment has been addressed.

src/riak_kv_wm_keylist.erl
seancribbs: Good catch.

PR #492, opened 2013-02-20
Make hashtree_eqc close trees before destroy.

seancribbs:
The hashtree API expects its backing leveldb instances to be closed before destroying them. Without this patch, the property would abort quickly with a badmatch error.


Vagabond:
+1, works great.

PR #495, opened 2013-02-25
Add encoding capability for handoff.

chardan:
Motivation:
Zlib forces serialization on an internal Erlang lock when opening new port drivers, which we are concerned may adversely affect scalability on larger systems. This feature provides a new capability for specifying the encoding method used for handoff, and offers a new option to use raw encoding so that users may choose a tuning most appropriate for their situation.

Note: Since capabilities are not re-negotiated, in some corner cases of rolling upgrade, it's possible for handoff to crash if a capability changes mid-up/downgrade; in this case, handoff will still finish, as it is designed to crash and will restart. Future work is expected to eliminate this case by encoding capabilities for the entire handoff rather than per-session.

See also:
https://github.com/basho/riak_test/pull/223
https://github.com/basho/riak_core/pull/276


src/riak_kv_vnode.erl
```diff
@@ -1351,6 +1352,23 @@ object_info({Bucket, _Key}=BKey) ->
     Hash = riak_core_util:chash_key(BKey),
     {Bucket, Hash}.
 
+%% @private
+%% Encoding and decoding selection:
+
+handoff_data_encoding_method() ->
+    riak_core_capability:get({riak_core, handoff_data_encoding}, encode_zlib).
+
+decode_binary_object(BinaryObject) ->
```
engelsanchez, line 1524: Would you consider fallbacks here? That is, if decoding fails, to fall back to the next encoding type? That way you don't have to worry about mid-flight changes and would make it more robust. Failure on decoding would be very rare anyway so an extra try shouldn't realistically affect performance.

chardan, line 1524: I think the plan right now is to address such cases in the future (see new change description).

engelsanchez:
:+1: :dancer: from me. You will have a conflict when merging this, though. I squashed this for local benchmarking and had to fix conflicts with the binary object code.

The benchmarking's results so far were funny: my test with 10k basho_bench generated binaries showed a huge (60%) improvement from the old code to the **code using Jordan's binary object format**, but only a 10% improvement from there to using raw encoding.

Also, the benchmark with zlib encoding was using a lot more CPU, of course (peaking at 90% total CPU usage) vs raw (peaking at 65% CPU usage) where I/O became the bottleneck.  I don't understand why ZLIB, which supposedly has this serialization problem, was able to use all my CPUs so well.

The test involved upping all configuration parameters to allow all handoffs in a 64 vnode node to happen in parallel.

jtuple:
@engelsanchez "I don't understand why ZLIB, which supposedly has this serialization problem, was able to use all my CPUs so well."

Erlang uses locks that spin-wait for a short duration before sleeping. The serialization concern with zlib was a lock around opening a new port (zlib is a port driver), which is a reasonably quick operation but one with high contention given how we were using zlib. It wouldn't surprise me if the common case was: try to acquire lock, spin wait, get lock, open port, unlock; rinse and repeat. Ie. 90% CPU that's spent mostly spin waiting.

Not saying this is what you observed, but certainly seems plausible.

For the curious, the lock we hit in `open_port` is the `ERTS_PROC_LOCK_LINK` lock which is acquired using `erts_smp_proc_lock`. Like all uses of `erts_smp_proc_lock`, contended locks end up in `erts_proc_lock_failed` which does the spin wait before sleep dance.

slfritchie:
Further comments on the VM problems regarding extremely rapid port allocation & deallocation patterns:
- Both `zlib:zip()` and `zlib:unzip()` allocated a port and deallocate that port for a single call.
- There's a single POSIX mutex for allocation & deallocation of all ports (as Joe notes above)
- The mutex contention only gets worse with more CPU cores.  We have support tickets documenting problems with 16-24 core machines.
- The R15B VM delays final deallocation of ports until after the owner Erlang process has been scheduled out.  That helps reduce contention on the global mutex for deallocation, but it also means that it's possible to hit `system_limit` errors when (in theory) the maximum number of ports is generously high.  (Also seen in support tickets.)

engelsanchez:
Thanks @jtuple  and @slfritchie for satisfying my curiosity. The busy waiting explanation sounds very plausible.

engelsanchez:
@chardan @jtuple Is there more work coming to this PR or can we merge?

src/riak_kv_vnode.erl
```diff
@@ -1512,6 +1513,56 @@ object_info({Bucket, _Key}=BKey) ->
...
+                                                   throw(invalid_handoff_encoding)
+                                end;
+
+        _                   ->  lager:error("Request to decode invalid handoff object"),
+                                throw(invalid_handoff_object)
+
+    %% An exception means we have a legacy handoff object:
+    catch
+        _:_                 -> do_zlib_decode(BinaryObject)
+    end.
```
evanmcc, line 1542: Not sure this is an issue, but how expensive is setting up the monitor environment for the exception in Erlang? Have we tested that this change doesn't make handoff in mixed clusters even slower?

src/riak_kv_vnode.erl
```diff
@@ -1512,6 +1513,56 @@ object_info({Bucket, _Key}=BKey) ->
...
+    PBObj = riak_core_pb:decode_riakobject_pb(DecodedObject),
+    BKey = {PBObj#riakobject_pb.bucket,PBObj#riakobject_pb.key},
+    {BKey, PBObj#riakobject_pb.val}.
+
+encode_binary_object(Bucket, Key, Value) ->
+    Method = handoff_data_encoding_method(),
+
+    case Method of
+        encode_raw  -> EncodedObject = term_to_binary({ Bucket, Key, iolist_to_binary(Value) }),
+                       return_encoded_binary_object(Method, EncodedObject);
```
evanmcc, line 1555: note that we call t2b here twice (one here and once in return_encoded).  wouldn't it make more sense just to t2b as {Method, {Bucket, Key, iolist_to_binary(Value) }} in the first place?

src/riak_kv_vnode.erl
chardan, line 1542: An excellent question. We have not yet measured this, but my expectation would indeed be that mixed handoff should be slower than previous versions: we've added more mechanism (for instance, the capability check). I don't know of a more efficient way to do this right now, as binary_to_term() either throws or succeeds.

engelsanchez, line 1542: Wouldn't this only happen briefly when the capability finally switches? That is, in a mixed cluster, the cap has not changed and all nodes will be talking the same protocol. During the switch, messages in flight might reach a destination with the new setting and hit this path. But that should be transient if I'm not mistaken. Will test.

chardan, line 1542: Let me know what you see. What I saw matches your expectations. Bear in mind that here we expect zlib to be the legacy method most of the time after upgrade. If efficiency here is a concern, perhaps there's a way to do this with BIF classification functions, I could look into that for another release if there's an issue. (I do not especially like using exceptions for flow control, but I'm not aware of another way to do this right now.)

src/riak_kv_vnode.erl
chardan, line 1555: Thanks for spotting this-- corrected.

src/riak_kv_app.erl
```diff
@@ -140,6 +140,10 @@ start(_Type, _StartArgs) ->
                                           [enabled_v1, disabled],
                                           disabled),
 
+            riak_core_capability:register({riak_kv, handoff_data_encoding},
+                                          [encode_raw, encode_zlib],
+                                          encode_raw),
```
engelsanchez: The default needs to be changed to encode_zlib so that it will work when upgrading from a previous version. Right now, adding a cluster with new code to a previous version node results in the new node default to using raw. But the old node only understands zlib encoded, and handoffs fail.  I tested manually and changing it here to encode_zlib allows the transfers to succeed.

engelsanchez:
Tested again. I found the problem above with the wrong default. After the fix, I've manually verified handoffs when adding a new node to an old cluster, upgrading a node in a mixed cluster and manually messing with capabilities to have mixed capabilities in the cluster. That is actually hard to reproduce because it would not normally happen on rolling upgrades, so I did not try to measure the performance during that mixed capability situation.

Things seem to work as advertised. Ima :+1: this sucker now.

PR #500, opened 2013-03-06
vclock capability

chardan:
Adds a new capability to support different encoding options for vector clocks. Moves existing zlib support to a capability, implements raw encoding. The purpose is to allow users to better tune for compression or performance.

### DIFF
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -309,11 +309,11 @@ erlify_rpbvc(undefined) ->
 erlify_rpbvc(<<>>) ->
     vclock:fresh();
 erlify_rpbvc(PbVc) ->
-    riak_kv_wm_utils:vclock_to_term(PbVc).
+    binary_to_term(riak_object:decode_vclock(PbVc)).
 
 %% Convert a vector clock to protocol buffers
 pbify_rpbvc(Vc) ->
-    riak_kv_wm_utils:term_to_vclock(Vc).
+    riak_object:encode_vclock(term_to_binary(Vc)).       
 
 %% ===================================================================
 %% Tests
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -313,7 +313,7 @@ erlify_rpbvc(PbVc) ->
 
 %% Convert a vector clock to protocol buffers
 pbify_rpbvc(Vc) ->
-    riak_object:encode_vclock(term_to_binary(Vc)).       
+    riak_object:encode_vclock(term_to_binary(Vc)).
 
 %% ===================================================================
 %% Tests
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -309,11 +309,11 @@ erlify_rpbvc(undefined) ->
 erlify_rpbvc(<<>>) ->
     vclock:fresh();
 erlify_rpbvc(PbVc) ->
-    binary_to_term(zlib:unzip(PbVc)).
+    riak_kv_wm_utils:vclock_to_term(PbVc).
 
 %% Convert a vector clock to protocol buffers
 pbify_rpbvc(Vc) ->
-    zlib:zip(term_to_binary(Vc)).
+    riak_kv_wm_utils:term_to_vclock(Vc).
 
 %% ===================================================================
 %% Tests
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -309,11 +309,11 @@ erlify_rpbvc(undefined) ->
 erlify_rpbvc(<<>>) ->
     vclock:fresh();
 erlify_rpbvc(PbVc) ->
-    binary_to_term(riak_object:decode_vclock(PbVc)).
+    riak_object:decode_vclock(PbVc).
 
 %% Convert a vector clock to protocol buffers
 pbify_rpbvc(Vc) ->
-    riak_object:encode_vclock(term_to_binary(Vc)).
+    riak_object:encode_vclock(Vc).
 
 %% ===================================================================
 %% Tests
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -831,19 +831,11 @@ select_doc(#ctx{doc={ok, Doc}, vtag=Vtag}) ->
 %% @spec encode_vclock_header(reqdata(), context()) -> reqdata()
 %% @doc Add the X-Riak-Vclock header to the response.
 encode_vclock_header(RD, #ctx{doc={ok, Doc}}) ->
-    {Head, Val} = vclock_header(Doc),
+    {Head, Val} = riak_object:vclock_header(Doc),
     wrq:set_resp_header(Head, Val, RD);
 encode_vclock_header(RD, #ctx{doc={error, {deleted, VClock}}}) ->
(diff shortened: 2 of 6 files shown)


chardan:
See-also accompanying tests in riak_test br jfw-vclock_capability:
https://github.com/basho/riak_test/pull/233

src/riak_object.erl
```diff
@@ -389,10 +390,17 @@ assemble_index_specs(Indexes, IndexOp) ->
 set_contents(Object=#r_object{}, MVs) when is_list(MVs) ->
     Object#r_object{contents=[#r_content{metadata=M,value=V} || {M, V} <- MVs]}.
 
+%% @spec vclock_header(riak_object()) -> {Name::string(), Value::string()}
+%% @doc Transform the Erlang representation of the document's vclock
+%%      into something suitable for an HTTP header
+vclock_header(Doc) ->
```
engelsanchez, line 415: Moving this method here doesn't seem that clean to me.  We now have a bit of web machine http knowledge embedded in the internal riak object representation.  I think that goes beyond the scope of this fix.

src/riak_object.erl
```diff
@@ -541,6 +549,31 @@ syntactic_merge(CurrentObject, NewObject) ->
...
+%%
+%% Helpers for managing vector clock encoding and related capability:
+%%
+
+%% Fetch the preferred vclock encoding method:
+vclock_encoding_method() ->
+    riak_core_capability:get({riak_kv, vclock_data_encoding}, encode_zlib).
+
+%% Encode a vclock in accordance with our capability setting:
+-spec encode_vclock(VClock :: term()) -> binary().
```
engelsanchez: So, it seems from above that this is really is taking a binary, not a term. Also, a plain string will work just due to the fact that zlib:zip and base64:encode both would take it (not that we need to advertise this internal detail in the spec).

src/riak_object.erl
```diff
@@ -541,6 +549,31 @@ syntactic_merge(CurrentObject, NewObject) ->
...
+encode_vclock(VClock) ->
+    case vclock_encoding_method() of
+        encode_zlib -> base64:encode(zlib:zip(VClock));
+        encode_raw  -> base64:encode(VClock)
+    end.
+
+%% Decode a vclock against our capability settings:
+-spec decode_vclock(VClock :: term()) -> vclock:vclock().
```
engelsanchez: Same here for the input as above (term() does not seem right), and we are returning a binary()

src/riak_object.erl
```diff
@@ -389,10 +390,17 @@ assemble_index_specs(Indexes, IndexOp) ->
...
+%% @doc Transform the Erlang representation of the document's vclock
+%%      into something suitable for an HTTP header
+vclock_header(Doc) ->
+    {?HEAD_VCLOCK,
+        binary_to_list(encode_vclock(term_to_binary(riak_object:vclock(Doc))))}.
+
 %% @spec to_json(riak_object()) -> {struct, list(any())}
 %% @doc Converts a riak_object into its JSON equivalent
 to_json(Obj=#r_object{}) ->
-    {_,Vclock} = riak_kv_wm_utils:vclock_header(Obj),
```
engelsanchez, line 412: I see now why the vclock_header method was moved here.  Still, we have the typical dilemma of the simplicity of having all of the encoding stuff built into the basic object and the cleanliness and flexibility of having anything to do with encoding moved to a different module.

src/riak_kv_wm_object.erl
```diff
@@ -815,18 +815,16 @@ encode_vclock_header(RD, #ctx{doc={ok, Doc}}) ->
...
 %% @spec vclock_header(riak_object()) -> {Name::string(), Value::string()}
 %% @doc Transform the Erlang representation of the document's vclock
 %%      into something suitable for an HTTP header
 vclock_header(Doc) ->
-    {?HEAD_VCLOCK,
-        encode_vclock(riak_object:vclock(Doc))}.
-
-encode_vclock(VClock) ->
-    binary_to_list(base64:encode(zlib:zip(term_to_binary(VClock)))).
+    %% JFW: I'm slightly suspicious about this because we're dropping the initial term_to_binary(Doc):
```
engelsanchez: The original vclock_header took a riak_object and produced the header tuple. This can't be right. Now Doc is expected to be a binary version of the vclock. Unless something else has changed upstream, this is definitely different.

engelsanchez:
Also, I would recommend merging from master. I'm doing that now locally to be able to test this. There are some conflicts.

engelsanchez:
@chardan were my comments on this PR addressed with the last changes? Can I continue reviewing this?  Don't forget to comment on a PR if you do address review comments so the reviewer gets notified and knows when to continue.

chardan:
I believe so, have a look!

src/riak_kv_pb_object.erl
```diff
@@ -293,11 +293,11 @@ erlify_rpbvc(undefined) ->
 erlify_rpbvc(<<>>) ->
     vclock:fresh();
 erlify_rpbvc(PbVc) ->
-    binary_to_term(zlib:unzip(PbVc)).
+    binary_to_term(riak_object:decode_vclock(PbVc)).
```
engelsanchez: Because of this change, the representation of a vclock through the PB interface changes from a binary to a base64 encoded binary (see riak_object:decode_vclock).  So, if a node is upgraded in a cluster, vclocks fetched from other nodes will be incompatible with this node. For instances where the vclock is sent to a different node than the original one, I believe this will fail.  This is in addition to requests failing when getting a vclock from a cluster right before the capabilities change to a new format, which we should probably handle too. Thoughts?

chardan: Very good observation, I'll address it. Thanks!

src/riak_object.erl
```diff
@@ -937,7 +1001,6 @@ new_with_ctype_test() ->
     O = riak_object:new(<<"b">>, <<"k">>, <<"{\"a\":1}">>, "application/json"),
     ?assertEqual("application/json", dict:fetch(?MD_CTYPE, riak_object:get_metadata(O))).
 
-new_with_md_test() ->
```
engelsanchez, line 940: @chardan, this is not compiling. I see the aftermath of a rebase here. There is no merge history to figure out what else broke with the rebase.  Please correct and re-check if you have lost anything else if you can.  Please be careful not to check in broken code man. Compile and run the tests after a merge/rebase.

chardan, line 940: Apologies for the rebase. I missed the conditional compilation, it's been corrected, along with bugfixes for the ancestor/sibling test.

src/riak_kv_pb_object.erl
engelsanchez: Unfortunately the way this was addressed does not fix the problem. See this commit for an updated riak_test that shows how it fails in a mixed cluster when going from a new node to an old node.  Ideally, vclocks through PB should be encoded exactly the same way in a mixed cluster than they were in the old cluster, even though that makes things more complicated because PB was encoding in a different way to the HTTP interface, which messes with the refactoring to clean it up.

engelsanchez:
I have attempted to fix all problems I reported with this commit https://github.com/basho/riak_kv/commit/6a05f3706e6be87f46d55ac5526567e2b2f7baa1

It also takes care of making sure we only binary to term/term to binary raw encoded values once.  Jesse will be on vacation for a bit, so we'll need someone else to review my fixes.

PR #502, opened 2013-03-11
Feature request, optionally return indexed values with keys

russelldb:
https://github.com/basho/platform_tasks/issues/34

Depends on https://github.com/basho/riak_pb/pull/34

Currently a 2i query's results are a list of keys. This change adds
an optional request parameter to pb and http that, when set, will
cause pairs of `key:index_value` to be returned as the result of
a 2i query.

Has no effect on MapReduce.

Uses capabilities for rolling upgrade
situations.

### DIFF
--- a/src/riak_index.erl
+++ b/src/riak_index.erl
@@ -32,7 +32,8 @@
          format_failure_reason/1,
          normalize_index_field/1,
          timestamp/0,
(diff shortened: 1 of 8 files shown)


russelldb:
Superseded by #540

PR #503, opened 2013-03-13
Fix eleveldb fold_objects to start at 1st object

engelsanchez:
Also fixed special bucket option so it starts at the beginning of the
given bucket. But that doesn't appear to be used anywhere...

/cc @jrwest


jrwest:
+1. consistently fixed failing case for me.

jrwest:
@engelsanchez going to merge this so I can get a PR out. It will get reviewed again as part of that.

Issue #499, opened 2013-03-04
Big integer 2i indexes sometimes sort incorrectly


reiddraper:
[As reported on the mailing list](http://lists.basho.com/pipermail/riak-users_lists.basho.com/2013-March/011276.html), sometimes bigger integers don't sort properly. This bug affects Riak 1.x. It only affects the leveldb backend, memory backend is _not_ affected. One manifestation of the issue can be reproduced like this:

```
curl -X PUT -H 'x-riak-index-foo_int:1362400142028' -d foo localhost:8098/buckets/new/keys/bar

# properly returns the 'bar' key
curl http://localhost:8098/buckets/new/index/foo_int/1100000000000/1362400142028

# does _not_ return the 'bar' key
curl http://localhost:8098/buckets/new/index/foo_int/1000000000000/1362400142028
```

We're pretty sure the issue has been tracked down to `sext`, specifically uwiger/sext#5.

reiddraper:
Confirmed @uwiger's fix in sext also fixes the 2i issue.

uwiger:
Are you happy with the migration support? The change in encoding may well make the problem worse while there are mixed encodings in the data set.

reiddraper:
> Are you happy with the migration support?

We're still working through what the migration story is going to be. I think `sext:legacy_encode_bignum/1` will prove helpful though.

uwiger:
Ok, let me know if you have any other ideas. Thanks for pinpointing the bug, and apologies for it being there in the first place.

reiddraper:
Reminder for later, the fix is on `uwiger/sext`, not `esl/sext` yet, so we may just change the rebar.config if it doesn't make it over.

uwiger:
It should be there now.

I haven't been pushing to esl in a while, but can start doing so again.

BR,
Ulf

2013/3/6 Reid Draper notifications@github.com

jrwest:
addressed by #504. closing

PR #504, opened 2013-03-14
Upgrade/Downgrade Path for 2i Bigint Bug

jrwest:
As reported in #499, 2i range queries involving bigints could return incorrect results. The issue was caused by https://github.com/uwiger/sext/issues/5. The fix introduces an incompatibility in sext's encoding of bigints. This PR addresses handling of both the fixed and legacy encoding in Riak and provides a path for 2i users to upgrade stored indices that may be using the legacy format. A path for downgrading is also provided. 

With this change, when first started, Riak will write new indicies (or update indicies) using the legacy format. When indicies are removed (when an object is deleted or updated), initially, the backend will attempt to remove both the legacy and fixed encoding of the index. After `riak_kv_util:fix_incorrect_indices/0,1` has been run the additional delete of the legacy index will no longer occur. The additional delete will also not occur in partitions that are empty on startup. 

Users wishing to downgrade can use `riak_kv_util:fix_incorrect_indices/2` passing the option `{downgrade, true}`. This performs a similar operation to upgrading, however, after starting, index writes will be encoded in the legacy format.

In the case `riak_kv_util:fix_incorrect_indices/0,1` crashes or returns errors it is expected that it will be re-run. Re-running will "continue from where it left off" in the sense that indices rewritten before the crash will not be rewritten again. Results are written to the log as `info` messages.

Additionally, this PR changes the dependency on `sext`. It bumps the version to 1.1, which is necessary to include the encoding fix and uses Ulf Wiger's repository instead of ESL's since Ulf's is now the official one.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -594,10 +594,9 @@ fold_objects_fun(FoldObjectsFun, FilterBucket) ->
 %% Augment the fold options list if a
 %% bucket is defined.
 fold_opts(undefined, FoldOpts) ->
-    FoldOpts;
+    [{first_key, to_first_key(undefined)} | FoldOpts];
 fold_opts(Bucket, FoldOpts) ->
-    BKey = sext:encode({Bucket, <<>>}),
-    [{first_key, BKey} | FoldOpts].
+    [{first_key, to_first_key({bucket, Bucket})} | FoldOpts].
 
 
 %% @private Given a scope limiter, use sext to encode an expression
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -54,7 +54,7 @@
 
 -define(API_VERSION, 1).
 -define(CAPABILITIES, [async_fold, indexes]).
--define(FIXED_INDEXES_KEY, sext:encode(fixed_indexes)).
+-define(FIXED_INDEXES_KEY, fixed_indexes).
 
 -record(state, {ref :: reference(),
                 data_root :: string(),
@@ -192,7 +192,7 @@ put(Bucket, PrimaryKey, IndexSpecs, Val, #state{ref=Ref,
     end.
 
 indexes_fixed(#state{ref=Ref,read_opts=ReadOpts}) ->
-    case eleveldb:get(Ref, ?FIXED_INDEXES_KEY, ReadOpts) of
+    case eleveldb:get(Ref, to_md_key(?FIXED_INDEXES_KEY), ReadOpts) of
         {ok, <<1>>} ->
             true;
         {ok, <<0>>} ->
@@ -241,7 +241,7 @@ mark_indexes_fixed(State=#state{ref=Ref, write_opts=WriteOpts}, ForUpgrade) ->
                 true -> <<1>>;
                 false -> <<0>>
             end,
-    Updates = [{put, ?FIXED_INDEXES_KEY, Value}],
+    Updates = [{put, to_md_key(?FIXED_INDEXES_KEY), Value}],
     case eleveldb:write(Ref, Updates, WriteOpts) of
         ok ->
             {ok, State#state{fixed_indexes=ForUpgrade}};
@@ -679,6 +679,11 @@ from_index_key(LKey) ->
         _ ->
             undefined
     end.
+
+%% @doc Encode a key to store partition meta-data attributes.
+to_md_key(Key) ->
+    sext:encode({md, Key}).
+
 
 %% ===================================================================
 %% EUnit tests
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -35,7 +35,7 @@
          drop/1,
          fix_index/4,
          mark_indexes_fixed/2,
-         set_legacy_indices/2,
+         set_legacy_indexes/2,
          fold_buckets/4,
          fold_keys/4,
          fold_objects/4,
@@ -63,8 +63,8 @@
                 read_opts = [],
                 write_opts = [],
                 fold_opts = [{fill_cache, false}],
-                fixed_indexes = false, %% true if legacy indices have be rewritten
-                legacy_indices = false %% true if new writes use legacy indices (downgrade)
+                fixed_indexes = false, %% true if legacy indexes have be rewritten
+                legacy_indexes = false %% true if new writes use legacy indexes (downgrade)
                }).
 
 
@@ -164,7 +164,7 @@ get(Bucket, Key, #state{read_opts=ReadOpts,
                  {error, term(), state()}.
 put(Bucket, PrimaryKey, IndexSpecs, Val, #state{ref=Ref,
                                                 write_opts=WriteOpts,
-                                                legacy_indices=WriteLegacy,
+                                                legacy_indexes=WriteLegacy,
                                                 fixed_indexes=FixedIndexes}=State) ->
     %% Create the KV update...
     StorageKey = to_object_key(Bucket, PrimaryKey),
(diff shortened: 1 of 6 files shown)


jrwest:
accompanying `riak_test` PR: https://github.com/basho/riak_test/pull/234

src/riak_kv_eleveldb_backend.erl
```diff
@@ -51,6 +54,7 @@
 
 -define(API_VERSION, 1).
 -define(CAPABILITIES, [async_fold, indexes]).
+-define(FIXED_INDEXES_KEY, sext:encode(fixed_indexes)).
```
engelsanchez: I've been thinking about this and wondering: we are going to live with these markers for all eternity. Should we maybe make something more compatible with existing keys like:
- Objects are {o,Bucket, Key}
- Index data is {i,Bucket,Field,Term,Key}
- Meta-data {md, Key}
  To accommodate similar markers in the future without making a mess that will need maintenance forever. What do you think?

src/riak_kv_eleveldb_backend.erl
```diff
@@ -159,13 +191,74 @@ put(Bucket, PrimaryKey, IndexSpecs, Val, #state{ref=Ref,
...
+        not_found ->
+            false;
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+index_deletes(FixedIndexes, Bucket, PrimaryKey, Field, Value) ->
+    KeyDelete = [{delete, to_index_key(Bucket, PrimaryKey, Field, Value)}],
+    LegacyDelete = [{delete, to_legacy_index_key(Bucket, PrimaryKey, Field, Value)}
+                    || FixedIndexes =:= false],
```
engelsanchez: How about avoiding adding LegacyDelete if the result of to_index_key and to_legacy_index_key is the same? One less possibility of a performance regression right there.

src/riak_kv_vnode.erl
```diff
@@ -511,7 +511,57 @@ handle_command(?KV_VNODE_STATUS_REQ{},
...
+        case Mod:fix_index(StorageKey, ForUpgrade, ModState) of
+            {ok, _UpModState} ->
+                ok;
+            {ignore, _UpModState} ->
+                ignore;
+            {error, Reason, _UpModState} ->
+                {error, Reason}
+        end,
+    {reply, Reply, State};
+handle_command({get_index_entries, ForUpgrade},
```
engelsanchez, line 537: Not that it matters for 1.3.1, but this names looks like a generic index scan instead of a scan returning entries to upgrade. What's in a name? :)

src/riak_kv_vnode.erl
```diff
@@ -511,7 +511,57 @@ handle_command(?KV_VNODE_STATUS_REQ{},
...
+               Sender,
+               State=#state{mod=Mod,
+                            modstate=ModState0}) ->
+    case Mod of
+        riak_kv_eleveldb_backend ->
```
engelsanchez: Oh shite, we have to support multi backend. When I wrote this I didn't intend for it to live for too long.  More work!

src/riak_kv_eleveldb_backend.erl
jrwest: i actually chose to not use a tuple for that exact reason but maybe that was my mistake. As my reasoning went, using an atom guarunteers `FIXED_INDEX_KEY` is always less than (in erlang term order, and thus sext) any tuple term.

EDIT: "using an atom"

src/riak_kv_eleveldb_backend.erl
jrwest: ah good call, so that way we dont issue the deletes for string indices and the like. will make that change

src/riak_kv_vnode.erl
jrwest: hmm multibackend will be interesting. IIRC the multibackend only supports indexes when all "sub-backends" support it [1]. 

[1] https://github.com/basho/riak_kv/blob/master/src/riak_kv_multi_backend.erl#L104

src/riak_kv_eleveldb_backend.erl
engelsanchez: This is a very small issue. The order doesn't really matter that much. Actually, a 2 tuple will always sort before a 3 tuple anyway when using sext. I'm just thinking that down the line we might want more markers and people will get creative with the names. So maybe adding the concept of keys that are global meta-data for the partition might make that less likely.  That way we could avoid even more special code by having some to_md_key function and such.

src/riak_kv_vnode.erl
jrwest, line 537: haha i changed it b/c now it fixes and "unfixes" indexes. naming is hard. realize i didn't continue that up the list of commands though. may just change it back while im in there to be consistent

src/riak_kv_eleveldb_backend.erl
jrwest: if you'd really like that cleanup to be made i have no objection. personally, i would make that change when we actually ran across the issue (its easy to still write a `to_md_key` that works on atoms and 2-tuples).

engelsanchez: I'm thinking of to_md_key(fixed_indexes), to_md_key(some_other_crazy_marker) that map to {md,Key}. Instead of then having sometimes ?FIXED_INDEXES_KEY + another generic path for to_md_key(some_other...). That is, creating a new namespace that can be used for all markers of this kind.  We could also just have atoms I suppose, making this key compatible and other markers in the future being plain atoms, but then someone comes along and adds a marker that's a tuple... I think wrapping it up in {md, <marker key>} would be more future proof.

src/riak_kv_eleveldb_backend.erl
```diff
@@ -99,11 +105,30 @@ start(Partition, Config) ->
...
 
+determine_fixed_index_status(State) ->
+    case indexes_fixed(State) of
+        {error, Reason} ->
+            {error, Reason};
+        true ->
+            {ok, State#state{fixed_indexes=true}};
+        false ->
+            case is_empty(State) of
+                true -> mark_indexes_fixed_on_start(State);
```
russelldb, line 121: I get it, a db with no indexes has no broken indexes, sometimes I'm pretty dumb, took me a few minutes to grok this one.

src/riak_kv_eleveldb_backend.erl
```diff
@@ -58,7 +62,9 @@
                 config :: config(),
                 read_opts = [],
                 write_opts = [],
-                fold_opts = [{fill_cache, false}]
+                fold_opts = [{fill_cache, false}],
+                fixed_indexes = false, %% true if legacy indices have be rewritten
+                legacy_indices = false %% true if new writes use legacy indices (downgrade)
```
russelldb: why is this `indices` and the previous `indexes`?

jrwest: bad judgement on my part. i noticed the same thing with `mark_indexes_fixed` and `set_legacy_indices`. will change (will hold off until all review comments tho).

src/riak_kv_eleveldb_backend.erl
```diff
@@ -159,13 +191,76 @@ put(Bucket, PrimaryKey, IndexSpecs, Val, #state{ref=Ref,
...
+            end,
+    Updates = [{put, to_md_key(?FIXED_INDEXES_KEY), Value}],
+    case eleveldb:write(Ref, Updates, WriteOpts) of
+        ok ->
+            {ok, State#state{fixed_indexes=ForUpgrade}};
+        {error, Reason} ->
+            {error, Reason, State}
+    end.
+
+set_legacy_indices(State, WriteLegacy) ->
```
russelldb: yeah, what you said about consistent naming

russelldb:
Tests pass. Poking, noodling and smoke testing on my local cluster all works fine. You've addressed the naming comments so +1 from me. Nice work.

engelsanchez:
The results of this branch have been squashed into #516 
That PR will be merged and this will be closed and left behind as a detailed trail of what went on here.

PR #506, opened 2013-03-14
Fix bug where stats endpoints were calculating _all_ riak_kv stats

russelldb:
Since adding many more stats, and most of the infrastructure for
ad hoc querying of stats, the stat calculation code for the
(not yet legacy) endpoints was calculating all stats for riak_kv.

As there are about (ring_size \* vnode stats) + (fsm stages \* fsm stats)
more stats now, this calculation, understandbly, took a long time.

This patch instead only calculates the minimum subset of stats
needed to support the (not yet legacy) stats endpoints.

### DIFF
--- a/src/riak_kv_stat_bc.erl
+++ b/src/riak_kv_stat_bc.erl
@@ -330,7 +330,7 @@ config_stats() ->
 %% @doc add the pipe stats to the blob in a style consistent
 %% with those stats already in the blob
 pipe_stats() ->
-    Stats = riak_core_stat_q:get_stats([riak_pipe]),
+    Stats = riak_pipe_stat:get_stats(),
     lists:flatten([bc_stat(Name, Val) || {Name, Val} <- Stats]).
 
 %% old style blob stats don't have the app name
--- a/src/riak_kv_stat_bc.erl
+++ b/src/riak_kv_stat_bc.erl
@@ -171,38 +171,47 @@ legacy_stats() ->
 
 %% @doc legacy stats uses multifield stats for multiple stats
 %% don't calculate the same stat many times
-get_stat(Name, Cache, CalcFun) ->
+get_stat(Name, Type, Cache) ->
+    get_stat(Name, Type, Cache, fun(S) -> S end).
+
+get_stat(Name, Type, Cache, ValFun) ->
     case proplists:get_value(Name, Cache) of
         undefined ->
-            S =  folsom_metrics:CalcFun(Name),
-            {S, [{Name, S} | Cache]};
-        Cached -> {Cached, Cache}
+            case riak_core_stat_q:calc_stat({Name, Type}) of
+                unavailable -> {unavailable, Cache};
+                Stat ->
+                    {ValFun(Stat), [{Name, Stat} | Cache]}
+            end;
+        Cached -> {ValFun(Cached), Cache}
     end.
 
 bc_stat({Old, {NewName, Field}, histogram}, Acc, Cache) ->
-    {Stat, Cache1} = get_stat(NewName, Cache, get_histogram_statistics),
-    Val = proplists:get_value(Field, Stat),
-    {[{Old, trunc(Val)} | Acc], Cache1};
+    ValFun = fun(Stat) -> trunc(proplists:get_value(Field, Stat)) end,
+    {Val, Cache1} = get_stat(NewName, histogram, Cache, ValFun),
+    {[{Old, Val} | Acc], Cache1};
 bc_stat({Old, {NewName, Field}, histogram_percentile}, Acc, Cache) ->
-    {Stat, Cache1} = get_stat(NewName, Cache, get_histogram_statistics),
-    Percentile = proplists:get_value(percentile, Stat),
-    Val = proplists:get_value(Field, Percentile),
-    {[{Old, trunc(Val)} | Acc], Cache1};
+    ValFun = fun(Stat) ->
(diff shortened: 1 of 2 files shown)


src/riak_kv_stat_bc.erl
```diff
@@ -163,26 +162,43 @@ produce_stats() ->
...
 %% with commas for underscores. Uses legacy_stat_map to generate
 %% legacys stats from the new list of stats.
-backwards_compat(Stats) ->
-    [bc_stat(Old, New, Type, Stats) || {Old, New, Type} <- legacy_stat_map()].
+legacy_stats() ->
+    {Legacy, _Calculated} = lists:foldl(fun({Old, New, Type}, {Acc, Cache}) ->
+                                                bc_stat({Old, New, Type}, Acc, Cache) end,
+                                        {[], []},
+                                        legacy_stat_map()),
+    lists:reverse(Legacy).
```
seancribbs, line 170: Why not `foldr` instead of `foldl` + `reverse`?

seancribbs:
@russelldb I don't know an adequate way to test the latency improvement, but I can see how computing less would be an improvement. I left only one nitpick for you, otherwise :+1:. You might want someone more capable than I to make a before/after comparison.

src/riak_kv_stat_bc.erl
russelldb, line 170: Why? The `reverse` was an add on when I was eyeballing the results. There is no API guarantee of order for the stats, but adding the reverse maintains the old order. I'll make it a `foldr` instead. Looks cleaner. Thank you.

russelldb:
I've pushed a change.

As regards testing…

This branch https://www.dropbox.com/s/km0v3086l89ts4q/kv505.png
1.3.0 https://www.dropbox.com/s/45l76mld8l3m0jv/1.3.0.png
1.2.1 https://www.dropbox.com/s/3lr17jasj036xu5/1.2.1.png

As you can see current branch is way better than 1.3.0, but worse than 1.2.1. I haven't been able to pinpoint that difference yet.

However, using this branch (https://github.com/basho/riak_core/tree/kv505-serve-from-cache) that only ever serves stats from the cache, and fetches them in the background gets this

https://www.dropbox.com/s/hcdjuzffskqma7b/kv505%2Bcache_bg.png

Which is pretty much ideal.

To reproduce results run the basho_bench configs against riak with the app.config found in this gist https://gist.github.com/russelldb/5301024b1b39ff4f35b0

seancribbs:
Thanks! You still have my :+1:

jrwest:
running against `verify_riak_stats` in riak test I was able to provoke an error (w/ or w/o basho/riak_core#286 ). The error does not appear against 1.3 w/o this PR (regardless of whether basho/riak_core#286 is included). not sure if this error is local to my environment:

```
webmachine error: path="/stats"
{error,{error,function_clause,[{proplists,delete,[disk,{error,{function_clause,[{riak_kv_stat_bc,'-pipe_stats/0-lc$^0/1-0-',[{riak_pipe,[{{riak_pipe,pipeline,active},0},{{riak_pipe,pipeline,create},[{count,0},{one,0}]},{{riak_pipe,pipeline,create,error},[{count,0},{one,0}]}]}],[{file,"src/riak_kv_stat_bc.erl"},{line,334}]},{riak_kv_stat_bc,pipe_stats,0,[{file,"src/riak_kv_stat_bc.erl"},{line,334}]},{riak_kv_stat_bc,produce_stats,0,[{file,"src/riak_kv_stat_bc.erl"},{line,149}]},{timer,tc,3,[{file,"timer.erl"},{line,194}]},{folsom_metrics,histogram_timed_update,4,[{file,"src/folsom_metrics.erl"},{line,177}]},{riak_core_stat_cache,'-do_get_stats/2-fun-0-',4,[{file,"src/riak_core_stat_cache.erl"},{line,182}]}]}}],[{file,"proplists.erl"},{line,362}]},{riak_kv_wm_stats,get_stats,0,[{file,"src/riak_kv_wm_stats.erl"},{line,88}]},{riak_kv_wm_stats,produce_body,2,[{file,"src/riak_kv_wm_stats.erl"},{line,77}]},{webmachine_resource,resource_call,3,[{file,"src/webmachine_resource.erl"},{line,183}]},{webmachine_resource,do,3,[{file,"src/webmachine_resource.erl"},{line,141}]},{webmachine_decision_core,resource_call,1,[{file,"src/webmachine_decision_core.erl"},{line,48}]},{webmachine_decision_core,decision,1,[{file,"src/webmachine_decision_core.erl"},{line,555}]},{webmachine_decision_core,handle_request,2,[{file,"src/webmachine_decision_core.erl"},{line,33}]}]}}
```

russelldb:
Yeah, problem is here https://github.com/basho/riak_kv/pull/506/files#L0L322 it's fixed in amongst https://github.com/basho/riak_kv/issues/508. I should have lumped them all together. It has been a messy journey.

I've pushed a change that fixes it on this branch.

jrwest:
confirmed that fix addressed my issue. no longer seeing that error

jrwest:
here are my perf numbers from running on r1s02.bos ( 1 x quad core 2.4Ghz E5620 w/ 32GB RAM and 24 x 1TB 7200 RPM drives in RAID 5 (22TB usable)):

https://gist.github.com/jrwest/bc63fcd33eaed326191d

russelldb:
Squashed and merged

Issue #507, opened 2013-03-15
multibackend invalid capabalities


jrwest:
The multibackend claims to have the capabalities that are the intersection of all "sub-backend" capabilities. However, the only thing the multi backend is capable of currenlty is `async_fold` because of the implementation of `riak_kv_multibackend:capabilities/1`. The function folds over all sub-backends and takes the intersection of the accumulator set and the sub-backends capabilities. However, Acc0 is the empty set and and set intersection w/ empty set is always the empty set. 

A simple fix would be to use `lists:map` + `ordsets:intersection/1` instead.

```
1> ordsets:intersection(ordsets:new(), ordsets:from_list([indexes])).
[]
```

https://github.com/basho/riak_kv/blob/master/src/riak_kv_multi_backend.erl#L104-L106

jtuple:
Resolved in #509 / #504

PR #509, opened 2013-03-18
make multibackend caps work as advertised

jrwest:
The multibackend performs set intersection to determine intersecting capabilities
of all sub-backends but it used the empty set as `Acc0`, which resulted in the empty set
always. This addresses this issue by using ordsets:intersection/1 instead.

~~Note: this has the side-effect of making the multi-backend index capable if all
sub-backends support indexing (e.g. some mix of memory and leveldb).~~ (the multibackend supported indexing due to the fact that `capabilities/2` returned the proper values based on the sub-backend's capabilities, this just fixes an inconsistency).

Addresses #507


jrwest:
closing this out. going to resubmit as part of #504 since its needed for fix

PR #512, opened 2013-03-19
Protocol Buffers interface allows the creation of records with an empty key

seancribbs:
Clients using protobufs can create records using an empty string as the key. This can be reproduced using an example from CorrugatedIron: https://github.com/DistributedNonsense/CorrugatedIron/issues/112 The HTTP interface falls down on this sort of behavior and attempts lists bucket props instead. 

Although both behaviors may be correct, the difference in client interfaces may create confusion for users - both interfaces should agree on basic CRUD operations.

### DIFF
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -98,6 +98,10 @@ process(#rpbsetclientidreq{client_id = ClientId}, State) ->
                end,
     {reply, rpbsetclientidresp, NewState};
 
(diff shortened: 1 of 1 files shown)


seancribbs:
See also basho/riak-ruby-client#86

evanmcc:
This no longer merges cleanly, but the conflicts are easy to resolve.

Once resolved, unit tests pass, and the patch behaves as expected with both zero length buckets and keys

:+1:

PR #513, opened 2013-03-19
When a broken stat is detected, register it.

russelldb:
Re-use the 'safe' stat calc funs in riak_core.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -132,9 +132,9 @@ handle_cast(stop, State) ->
 handle_cast(_Req, State) ->
     {noreply, State}.
 
-handle_info({'DOWN', MonRef, process, Pid, _Cause}, #state{repair_mon={Pid, MonRef}}) ->
+handle_info({'DOWN', MonRef, process, Pid, _Cause}, State=#state{repair_mon={Pid, MonRef}}) ->
     RepairMonitor = spawn_monitor(fun() -> stat_repair_loop() end),
-    {noreply, #state{repair_mon=RepairMonitor}};
+    {noreply, State#state{repair_mon=RepairMonitor}};
 handle_info(_Info, State) ->
     {noreply, State}.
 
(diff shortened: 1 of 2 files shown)


src/riak_kv_stat.erl
```diff
@@ -106,18 +109,32 @@ stop() ->
...
+    Pid ! {re_register_stat, Arg},
+    {noreply, State};
 handle_cast(stop, State) ->
     {stop, normal, State};
 handle_cast(_Req, State) ->
     {noreply, State}.
 
+handle_info({'DOWN', MonRef, process, Pid, _Cause}, #state{repair_mon={Pid, MonRef}}) ->
+    RepairMonitor = spawn_monitor(fun() -> stat_repair_loop() end),
+    {noreply, #state{repair_mon=RepairMonitor}};
```
engelsanchez: This is correct, but it might be more future proof to do State#state{repair_mon=RepairMonitor} just in case something's added later and this goes unnoticed.

russelldb: I kind of prefer this way. Why introduce a variable we don't need? This also makes it clear that the state that goes out is new, and not a "change" to the existing state.

engelsanchez: If later someone adds a new field to State and at that point a fresh new state is the wrong thing to do and that person does not notice this line so far away from the record field, it will wipe out that new field upon receiving the DOWN message. Imagine how hard that would be to track. An OTP gen_server is usually almost always just updating the state and not wiping it clean. I really think that

handle_info({'DOWN', MonRef, process, Pid, _Cause}, State=#state{repair_mon={Pid, MonRef}}) ->
    RepairMonitor = spawn_monitor(fun() -> stat_repair_loop() end),
    {noreply, State#state{repair_mon=RepairMonitor}};

would be more future proof.  If any new field is added to state, it will be preserved.  This could be you in 3 months! :)

russelldb: I understand that. I can't argue with it. I'll change it.

engelsanchez:
:+1: :dancer:

PR #520, opened 2013-03-26
Adds X-Riak-Deleted where missing

### LINKED ISSUE #518: HTTP API tombstone reporting
With the HTTP API regarding tombstones here's what we currently have:
1. If _only_ a tombstone exists when you do a GET for a key, you will
   receive a 404 but it will contain the `X-Riak-Vclock` header. A "normal" 404 (when there's no object) will not have this header.
2. If there is a set of siblings, and one of them is a tombstone:
   - Retrieving all the siblings at once by including `Accept:
     multipart/mixed` in the GET request will return all the siblings, and the
     tombstone will include the `X-Riak-Deleted` header
   - Retrieving each sibling manually by adding ?vtag=XXXXXX to the GET
     will (unfortunately) return a `200 OK` for the tombstone but it will
     have an empty body (`Content-Length: 0`).

It seems the easiest way to approach this since we're returning a 404 in `1.` is to add the `X-Riak-Deleted` header there for consistency with `2a.`  

In the case of `2b.` ... ? @seancribbs is lobbying that we remove the fetch each sibling via the `?vtag=XXXXX` method since there's no logical reason to do so. At the very least we should add the `X-Riak-Deleted` header (and IMHO make it return a 404).

broach:
In the case of a single object being returned which is a
tombstone and the case where a single sibling has been
requested via a vtag and it is a tombstone, the
X-Riak-Deleted header will now be added and
returned.

Ideally both these would return 204 rather than 404 and 200
respectively but doing so would have breaking consequences
for clients.

Fixes #518

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -704,7 +704,20 @@ multiple_choices(RD, Ctx=#ctx{vtag=undefined, doc={ok, Doc}}) ->
     end;
 multiple_choices(RD, Ctx) ->
     %% specific vtag was specified
-    {false, RD, Ctx}.
+    %% if it's a tombstone add the X-Riak-Deleted header
+    case select_doc(Ctx) of
+        {M, _} ->
+            case dict:find(?MD_DELETED, M) of
+                {ok, true} ->
+                    {false,
+                        wrq:set_resp_header(?HEAD_DELETED, "true", RD),
+                        Ctx};
+                error ->
+                    {false, RD, Ctx}
+            end;
+        multiple_choices ->
+            throw({unexpected_code_path, ?MODULE, multiple_choices, multiple_choices})
+    end.
 
 %% @spec produce_doc_body(reqdata(), context()) -> {binary(), reqdata(), context()}
 %% @doc Extract the value of the document, and place it in the
@@ -1038,9 +1051,10 @@ handle_common_error(Reason, RD, Ctx) ->
         {error, {deleted, _VClock}} ->
             {{halt, 404},
                 wrq:set_resp_header("Content-Type", "text/plain",
-                    wrq:append_to_response_body(
-                        io_lib:format("not found~n",[]),
-                        encode_vclock_header(RD, Ctx))),
+                    wrq:set_resp_header(?HEAD_DELETED, "true",
+                        wrq:append_to_response_body(
+                            io_lib:format("not found~n",[]),
+                            encode_vclock_header(RD, Ctx)))),
                 Ctx};
         {error, {n_val_violation, N}} ->
             Msg = io_lib:format("Specified w/dw/pw values invalid for bucket"
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -708,7 +708,7 @@ multiple_choices(RD, Ctx) ->
     case select_doc(Ctx) of
         {M, _} ->
             case dict:find(?MD_DELETED, M) of
-                {ok, true} ->
+                {ok, "true"} ->
                     {false,
                         wrq:set_resp_header(?HEAD_DELETED, "true", RD),
                         Ctx};


seancribbs:
Quick manual test case and failure:

``` bash
$ curl -X PUT localhost:8098/buckets/test/props -H "content-type: application/json" -d '{"props":{"allow_mult":true}}'

$ curl -X PUT localhost:8098/buckets/test/keys/1 -H "Content-Type: text/plain"  -d "hello, world"

$ curl localhost:8098/buckets/test/keys/1 -i
HTTP/1.1 200 OK
X-Riak-Vclock: a85hYGBgzGDKBVIcypz/fgbGLL6bwZTImMfK4KFw8RRfFgA=
Vary: Accept-Encoding
Server: MochiWeb/1.1 WebMachine/1.9.2 (someone had painted it blue)
Link: </buckets/test>; rel="up"
Last-Modified: Wed, 03 Apr 2013 21:51:04 GMT
ETag: "3SHUhQSMLhfjR0X7uQCHgf"
Date: Wed, 03 Apr 2013 21:51:17 GMT
Content-Type: text/plain
Content-Length: 11

hello world

$ curl -X PUT localhost:8098/buckets/test/keys/1 -H "Content-Type: text/plain"  -H "X-Riak-Vclock: a85hYGBgzGDKBVIcypz/fgbGLL6bwZTImMfK4KFw8RRfFgA=" -d "hello 2"

$ curl -X DELETE localhost:8098/buckets/test/keys/1 -H "X-Riak-Vclock: a85hYGBgzGDKBVIcypz/fgbGLL6bwZTImMfK4KFw8RRfFgA="

$ curl -i localhost:8098/buckets/test/keys/1
HTTP/1.1 300 Multiple Choices
X-Riak-Vclock: a85hYGBgzGDKBVIcypz/fgbGLL6bwZTInMfK0Ktw8RRfFgA=
Vary: Accept, Accept-Encoding
Server: MochiWeb/1.1 WebMachine/1.9.2 (someone had painted it blue)
Last-Modified: Wed, 03 Apr 2013 21:52:13 GMT
ETag: "ByOt5fgRS2Fba75z0GB1W"
Date: Wed, 03 Apr 2013 21:52:20 GMT
Content-Type: text/plain
Content-Length: 56

Siblings:
6k3PJsbhPOTH4La6QudyXJ
4MqWbAYPO8Y0sQbwfLjkAj

$ curl -i localhost:8098/buckets/test/keys/1 -H "Accept: multipart/mixed"
HTTP/1.1 500 Internal Server Error
Vary: Accept, Accept-Encoding
Server: MochiWeb/1.1 WebMachine/1.9.2 (someone had painted it blue)
Last-Modified: Wed, 03 Apr 2013 21:52:13 GMT
ETag: "ByOt5fgRS2Fba75z0GB1W"
Date: Wed, 03 Apr 2013 21:52:29 GMT
Content-Type: text/html
Content-Length: 1233

<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request:<br><pre>{error,
    {error,
        {case_clause,{ok,true}},
        [{riak_kv_wm_utils,multipart_encode_body,4,
             [{file,"src/riak_kv_wm_utils.erl"},{line,126}]},
         {riak_kv_wm_object,'-produce_multipart_body/2-lc$^0/1-0-',5,
             [{file,"src/riak_kv_wm_object.erl"},{line,810}]},
         {riak_kv_wm_object,produce_multipart_body,2,
             [{file,"src/riak_kv_wm_object.erl"},{line,809}]},
         {webmachine_resource,resource_call,3,
             [{file,"src/webmachine_resource.erl"},{line,183}]},
         {webmachine_resource,do,3,
             [{file,"src/webmachine_resource.erl"},{line,141}]},
         {webmachine_decision_core,resource_call,1,
             [{file,"src/webmachine_decision_core.erl"},{line,48}]},
         {webmachine_decision_core,decision,1,
             [{file,"src/webmachine_decision_core.erl"},{line,546}]},
         {webmachine_decision_core,handle_request,2,
             [{file,"src/webmachine_decision_core.erl"},{line,33}]}]}}</pre><P><HR><ADDRESS>mochiweb+webmachine web server</ADDRESS></body></html>
```

broach:
This is not caused by my PR. 

Building riak from master (which pulls riak_kv master), creating a sibling, then fetching via HTTP with `accept: multipart/mixed` causes this error currently. I just retested to verify. 

This does not occur in 1.3.1 which is pulling `{tag,"1.3.1"}` for riak_kv (also verified just now)

broach:
The issue is this in `src/riak_kv_wm_utils.erl`

`case dict:find(?MD_DELETED, MD) of
         {ok, "true"} ->`

If you remove the double quotes around `true` it fixes it which makes sense given the error message. However ... I have no idea what/why that changed. I'm guessing the client puts that value in the dict?

jrwest:
@broach is correct, this is caused by [1], the new riak object format inadvertently changed the type of deleted from string to atom. I think the short term fix is to just change the type back (one line change). Long term, we should probably stop leaking riak object in places like `kv_wm_utils` as @seancribbs has outlined before. 

[1] https://github.com/basho/riak_kv/blob/master/src/riak_object.erl#L640

seancribbs:
With #532, this works as expected. :+1:

PR #521, opened 2013-03-26
Changes needed to expose FSM timeouts to clients

evanmcc:
- change delete in riak_client to honor timeouts passed
  in Options
- change the various object interfaces to collect and
  pass on the timeout values

requires https://github.com/basho/riak_pb/issues/38

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -251,12 +251,7 @@ delete(Bucket,Key,Options,Timeout) when is_list(Options) ->
     riak_kv_delete_sup:start_delete(Node, [ReqId, Bucket, Key, Options, Timeout,
                                            Me, ClientId]),
     RTimeout = recv_timeout(Options),
-    lager:info("timeouts r ~p t ~p > ~p",
-               [RTimeout, Timeout, RTimeout > Timeout]),
-    case RTimeout > Timeout of 
-        true -> wait_for_reqid(ReqId, Timeout);
-        false -> wait_for_reqid(ReqId, RTimeout)
-    end;
+    wait_for_reqid(ReqId, erlang:min(Timeout, RTimeout));
 delete(Bucket,Key,RW,Timeout) ->
     delete(Bucket,Key,[{rw, RW}], Timeout).
 
@@ -302,10 +297,7 @@ delete_vclock(Bucket,Key,VClock,Options,Timeout) when is_list(Options) ->
     riak_kv_delete_sup:start_delete(Node, [ReqId, Bucket, Key, Options, Timeout,
                                            Me, ClientId, VClock]),
     RTimeout = recv_timeout(Options),
-    case RTimeout > Timeout of 
-        true -> wait_for_reqid(ReqId, Timeout);
-        false -> wait_for_reqid(ReqId, RTimeout)
-    end;
+    wait_for_reqid(ReqId, erlang:min(Timeout, RTimeout));
 delete_vclock(Bucket,Key,VClock,RW,Timeout) ->
     delete_vclock(Bucket,Key,VClock,[{rw, RW}],Timeout).
 
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -250,7 +250,13 @@ delete(Bucket,Key,Options,Timeout) when is_list(Options) ->
     ReqId = mk_reqid(),
     riak_kv_delete_sup:start_delete(Node, [ReqId, Bucket, Key, Options, Timeout,
                                            Me, ClientId]),
-    wait_for_reqid(ReqId, Timeout);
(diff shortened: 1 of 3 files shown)


evanmcc:
associated riak test: https://github.com/basho/riak_test/pull/243

src/riak_client.erl
```diff
@@ -250,7 +250,13 @@ delete(Bucket,Key,Options,Timeout) when is_list(Options) ->
     ReqId = mk_reqid(),
     riak_kv_delete_sup:start_delete(Node, [ReqId, Bucket, Key, Options, Timeout,
                                            Me, ClientId]),
-    wait_for_reqid(ReqId, Timeout);
+    RTimeout = recv_timeout(Options),
+    lager:info("timeouts r ~p t ~p > ~p",
```
seancribbs: This should be at debug level and probably have a more useful message.

src/riak_kv_wm_object.erl
```diff
@@ -848,19 +873,27 @@ ensure_doc(Ctx=#ctx{doc=undefined, key=undefined}) ->
...
 
 %% @spec delete_resource(reqdata(), context()) -> {true, reqdata(), context()}
 %% @doc Delete the document specified.
 delete_resource(RD, Ctx=#ctx{bucket=B, key=K, client=C, rw=RW, r=R, w=W,
-        pr=PR, pw=PW, dw=DW, timeout=Timeout}) ->
-    Options = lists:filter(fun({_, default}) -> false; (_) -> true end,
-        [{rw, RW}, {r, R}, {w, W}, {pr, PR}, {pw, PW}, {dw, DW}, 
-         {timeout, Timeout}]),
+        pr=PR, pw=PW, dw=DW}) ->
+    Options0 = lists:filter(fun({_, default}) -> false; (_) -> true end,
```
seancribbs: Suggested refactoring:

``` erlang
Opts0 = [{rw, RW}, {r, R}, {w, W}, {pr, PR}, {pw, PW}, {dw, DW}, {timeout, Timeout}],
Options = [ {Opt, Val} || {Opt, Val} <- Opts0, Val /= undefined, Val /= default ]
```

That pattern might also be useful in these other places where you conditionally include the timeout option, e.g.

``` erlang
Opts0 ++ [ {timeout, Timeout} || Timeout /= undefined ]
```

src/riak_client.erl
```diff
@@ -295,7 +301,11 @@ delete_vclock(Bucket,Key,VClock,Options,Timeout) when is_list(Options) ->
     ReqId = mk_reqid(),
     riak_kv_delete_sup:start_delete(Node, [ReqId, Bucket, Key, Options, Timeout,
                                            Me, ClientId, VClock]),
-    wait_for_reqid(ReqId, Timeout);
+    RTimeout = recv_timeout(Options),
+    case RTimeout > Timeout of
```
seancribbs: Why not:

``` erlang
wait_for_reqid(ReqId, erlang:max(Timeout, RTimeout))
```

src/riak_client.erl
evanmcc: meant to remove.

src/riak_client.erl
evanmcc: erlang:stdlib() > evan:brain()

src/riak_kv_wm_object.erl
evanmcc: it's done in three different ways in three different places.  Would be happy to refactor it to be uniform, but wasn't sure that this PR was the right place.

seancribbs: I'm ok with adding refactoring all three, or simply making a function to call for the same effect. Cleaning up debt is always good.

Sean Cribbs

src/riak_kv_wm_object.erl
```diff
@@ -245,29 +230,66 @@ malformed_request(RD, Ctx) when Ctx#ctx.method =:= 'POST'
...
+        TimeoutStr -> 
+            try
+                Timeout = list_to_integer(TimeoutStr),
+                {false, RD, Ctx#ctx{timeout=Timeout}}
+            catch
+                _:_ ->
+                    {true,
+                     wrq:append_to_resp_body(io_lib:format("Bad timeout "
+                                                           "value ~p, "
+                                                           "using ~p~n",
```
seancribbs: If you provide a bad timeout value, it doesn't actually use the default but returns a 400 response.

seancribbs:
I also wonder about the arbitrary 100ms being added to the timeout, which produces some strange results if you set the timeout very low. In a sense, you cannot set a timeout lower than 100ms.

seancribbs:
:+1: to merge.

PR #523, opened 2013-03-29
Fix perf problems and bug in 2i reformat

engelsanchez:
Add batch size parameter to control how many keys
are fixed at a time.
Add backpressure to query of bad idx entries to avoid
having the reformat process flooded with msgs.
Fixed problem with list keys and code triggering a second
scan during 2i reformat transition.

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -422,39 +422,26 @@ format_timestamp(_Now, undefined) ->
 format_timestamp(Now, TS) ->
     riak_core_format:human_time_fmt("~.1f", timer:now_diff(Now, TS)).
 
-reformat_indexes([]) ->
-    start_index_reformat([]),
-    io:format("index reformat started with default options~n"),
-    io:format("check console.log for status information~n"),
-    ok;
-reformat_indexes(["--downgrade"]) ->
-    start_index_reformat([{downgrade, true}]),
-    io:format("index reformat downgrade started with default options~n"),
-    io:format("check console.log for status information~n"),
-    ok;
-reformat_indexes([ConcurrencyStr | Rest]) ->
-    {Valid, Opts} = try list_to_integer(ConcurrencyStr) of
-                        Concurrency ->
-                            BaseOpts = [{concurrency, Concurrency}],
-                            Final = case lists:member("--downgrade", Rest) of
-                                        true -> [{downgrade, true} | BaseOpts];
-                                        false -> BaseOpts
-                                    end,
-                            {true, Final}
-                    catch
-                        _:_ ->
-                            {false, []}
-                    end,
-    case Valid of
+index_reformat_options([], Opts) ->
+    Opts;
+index_reformat_options(["--downgrade" | Rest], Opts) ->
+    index_reformat_options(Rest, [{downgrade, true} | Opts]);
+index_reformat_options([IntStr | Rest], Opts) ->
+    %% Letting it crash here if not an integer
+    IntVal = list_to_integer(IntStr),
+    case lists:keymember(concurrency, 1, Opts) of
         true ->
-            start_index_reformat(Opts),
-            io:format("index reformat started with options: ~p~n", [Opts]),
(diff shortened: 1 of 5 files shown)


src/riak_kv_console.erl
```diff
@@ -422,40 +422,27 @@ format_timestamp(_Now, undefined) ->
...
-                        _:_ ->
-                            {false, []}
-                    end,
-    case Valid of
+index_reformat_options([], Opts) ->
+    Opts;
+index_reformat_options(["--downgrade" | Rest], Opts) ->
+    index_reformat_options(Rest, [{downgrade, true} | Opts]);
+index_reformat_options([IntStr | Rest], Opts) ->
+    %% Letting it crash here if not an integer
```
jrwest: Other console commands seem to catch the bad format: https://github.com/basho/riak_core/blob/master/src/riak_core_console.erl#L546

src/riak_kv_vnode.erl
```diff
@@ -524,24 +524,28 @@ handle_command({fix_incorrect_index_entry, {done, ForUpgrade}}, _Sender,
...
             {reply, error, State}
     end;
-handle_command({fix_incorrect_index_entry, {Bucket, StorageKey}, ForUpgrade},
+handle_command({fix_incorrect_index_entry, Keys, ForUpgrade},
                _Sender,
                State=#state{mod=Mod,
                             modstate=ModState}) ->
     Reply =
-        case Mod:fix_index(Bucket, StorageKey, ForUpgrade, ModState) of
+        case Mod:fix_index(undefined, Keys, ForUpgrade, ModState) of
```
jrwest: passing `undefined` for the bucket here will break the multi backend I'm pretty sure. It uses it to select which "sub-backend" to pass on the `fix_index` call to: https://github.com/basho/riak_kv/blob/a03f2e96aedf38d462a378d3af4ee541016c9aa5/src/riak_kv_multi_backend.erl#L392

src/riak_kv_vnode.erl
```diff
@@ -550,12 +554,24 @@ handle_command({get_index_entries, ForUpgrade},
...
-                    FoldFun = fun(Bucket, StorageKey, _) ->
-                                      riak_core_vnode:reply(Sender, {Bucket, StorageKey})
-                              end,
+                    BufferMod = riak_kv_fold_buffer,
+                    ResultFun =
+                        fun(Results) ->
+                            BatchRef = make_ref(),
+                            riak_core_vnode:reply(Sender, {self(), BatchRef, Results}),
+                            Monitor = riak_core_vnode:monitor(Sender),
+                            receive
```
jrwest, line 565: comment about what this is doing (assuming backpressure?) would be a nice addition imo

engelsanchez, line 565: Will do

engelsanchez:
@jrwest The changes are ready. These should address the multi-backend and reformat command issues.

engelsanchez:
To give you an idea of the performance characteristics of the reformat with this fix: 
For this test about 31 million keys were updated. The leveldb data on disk was about 80G. On Ubuntu 12, 16 CPU box with striped disks I believe (BOS r1s02). The load came from 32 basho bench workers doing a mix of gets, puts, queries returning 1 results, queries returning 5 results (equal distribution among those 4).

concurrency=2, batch_size = 100   -> 1 hr              (no load)
concurrency=2, batch size  = 100   -> 1hr 17 mins  (under load)
concurrency=8, batch_size = 100   -> 17 mins        (no load)
concurrency=1, batch size  = 1      -> 1h 35 mins    (no load)

engelsanchez:
@jrwest Added printing of default values to reformat-indexes console command

jrwest:
code changes look good. tested against leveldb and multibackend (w/ and w/o sub-backends other than leveldb). clearly improves observability (although the logs are a bit "loud") and performance. Besides merging #525 into this first I'm +1. Should we get a second reviewer/tester @engelsanchez?

1.3.1 release notes also need updating to add info about batch size parameter

PR #527, opened 2013-04-04
List buckets timeout & streaming

seancribbs:
To ease exporting of cluster data for specific customer cases, we need to reduce the operational cost of the list-buckets operation so that it can succeed when there are large numbers of keys in the cluster.
- [x] Allow buckets FSM to stream results
- [x] Expose timeout option in `riak_client`
- [x] Add timeout option to `riak_kv_wm_buckets`
- [x] Add timeout option to `riak_kv_pb_buckets` and associated message in `riak_pb`
- [x] Add `riak_client:stream_list_buckets()`
- [x] Add streaming option to `riak_kv_wm_buckets` (preferably using multipart)
- [x] Add streaming option to `riak_kv_pb_buckets` and associated messages in `riak_pb`.

This is a WIP that needs to be taken up by someone else in order to land for 1.4, perhaps @evanmcc.


evanmcc:
updated and ready for review. riak_test PR referenced above.

src/riak_kv_pb_bucket.erl
```diff
@@ -71,8 +71,23 @@ encode(Message) ->
...
+            case C:list_buckets(T) of
+                {ok, Buckets} ->
+                    {reply, #rpblistbucketsresp{buckets = Buckets}, State};
+                {error, Reason} ->
+                    {error, {format, Reason}, State}
+            end
+    end;
+
+%% this should remain for backwards compatibility
+process(rpblistbucketsreq, #state{client=C} = State) ->
```
kellymclaughlin: Can this just become a call to the `process` with an `#rpblistbucketsreq` record that has `stream=false` and avoid the code duplication?

evanmcc: So I don't actually know enough about PB usage to answer that question.
In older versions of the client, there was no message, so we get an atom rather than a record, which are the two cases that we're taking care of.  So my thought was that for backwards compatibility we needed to handle both cases.  However, if clients and the server require version lockstep, then we can delete the older, redundant looking case.

kellymclaughlin: I'm just suggesting to replace the body of this function with a call to process with the `#rpblistbucketsreq` record with stream set to false which will end up executing the very same case statement.

evanmcc:
updated.

kellymclaughlin:
+1 to merge

PR #529, opened 2013-04-09
Count async MR results against the sink buffer size cap

beerriot:
While chasing another issue, I found that MR results delivered to an FSM sink via `gen_fsm:send_event/2` are not counted against the buffer size limit. This method is used to deliver 10 results before `gen_fsm:sync_send_event/3` is used once. Since only the results sent synchronously were counted against the buffer size limit, the limit actually imposed was effectively 10 times higher than expected.

The first commit in this PR modifies the eunit test to demonstrate the issue. The second commit fixes the problem.

### DIFF
--- a/src/riak_kv_mrc_sink.erl
+++ b/src/riak_kv_mrc_sink.erl
@@ -189,12 +189,14 @@ collect_output(next, State) ->
             end
     end;
 collect_output(#pipe_result{ref=Ref, from=PhaseId, result=Res},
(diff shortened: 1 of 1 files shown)


engelsanchez:
I studied the code and it looks like an obvious omission. Also verified the eunit test fails when data is sent asynchronously to the mrc sink, but passes again after the changes are applied. Other eunit tests are failing but seem unrelated.

:+1: :dancer: :boat: !

PR #530, opened 2013-04-17
Ring Resizing Support

jrwest:
support for https://github.com/basho/riak_core/pull/301

This PR includes the following changes:
- exports the already existing `riak_kv_vnode:object_info/1` function, which is used during resize transfers
- add `riak_kv_vnode:request_hash/1`. has the effect of only forwarding puts/deletes during ring resizing, all other requests are handled locally
- slight modifications to AAE so entropy_manager and vnodes don't crash when ring is expanding
- modifications to PR/PW in get/put_core (see updated comments in source)


jtuple:
+1 merge, nothing in the code that I have a complaint about. Tested as part of the larger basho/riak_core#301 review.

PR #532, opened 2013-04-19
kv_wm_utils expects ?MD_DELETED to be "true" not 'true'

jrwest:
minimal possible fix. long term these types of things should be encapsulated
in riak_object

found here (includes steps to recreate): https://github.com/basho/riak_kv/pull/520#issuecomment-16616515 (Note: to recreate issue you must create a _tombstone_ sibling, not just a sibling)

Java code that exposes issue: https://gist.github.com/broach/57ce9237ec32b47d8897

cc/ @broach @seancribbs


seancribbs:
:+1: This fixes the problem.

PR #540, opened 2013-05-01
2i pagination support

russelldb:
Adds a `max_results` option to PB and HTTP 2i endpoints for limiting
the number of results returned for a 2i query.  Will return an opaque
`continuation` if there are `max_results` results returned. A
subsequent query that includes the `continuation` as a parameter will
return the next `max_results` results.

The results will be returned sorted ascending. In the case of $key
and $bucket queries, and equals queries, the results will be sorted by
primary key In the case of range queries the results will be sorted by
index value, then primary key. Since the values must be returned from the backend for
sorting, there is a further option `return_terms` that will cause the
indexed terms to be returned to the user also, where relevant (range
queries on index fields.)

Sorting is performed by the module `sms.erl` originally written by Reid
Draper. It is used as a buffer by `riak_kv_index_fsm`.

A new vnode function `stop_fold` has been added so the fsm can tell
the backend to break out of the fold when enough results have been
gathered.

The option to stream 2i results has also been added. Adding the
`stream` option to either HTTP or PB queries will cause the API
endpoints to return chunks of results as they are received from the
fsm This can lead to much lower perceived latencies as the application
recieves the first results of a large set much sooner.

An example HTTP query: 

`curl "http://127.0.0.1:10018/buckets/2ibucket/index/field2_int/1/999?return_terms=true&max_results=100"`

Will return the first 100 {value, key} pairs for the `field2_int`
index, sorted by value, then primary key.  If there are 100 results returned, it will
also return an opaque `continuation` that may be sumbitted to continue
the query: 

`curl "http://127.0.0.1:10018/buckets/2ibucket/index/field2_int/1/999?return_terms=true&continuation=some_gibberish_looking_value&max_results=100"`

A further `continuation` maybe returned with the results.

---

Riak tests can be found at https://github.com/basho/riak_test/tree/pt29-2i-pagination
Riak-Erlang-Client support at https://github.com/basho/riak-erlang-client/tree/pt29-2i-pagination

NOTE: requires
https://github.com/basho/riak_core/tree/pt29-2i-pagination
https://github.com/basho/riak_pb/tree/pt29-2i-pagination


include/riak_kv_index.hrl
```diff
@@ -0,0 +1,13 @@
+%% Index query records
```
engelsanchez, line 23: Don't forget to stick a licence header or the lawyers will get you!

src/riak_index.erl
```diff
@@ -258,6 +271,152 @@ to_index_query(IndexField, Args) ->
...
+to_index_query(IndexField, Args, Continuation, ReturnBody, {Start, StartInc}, {End, EndInc}) ->
+    BaseQuery = ?KV_INDEX_Q{return_body=ReturnBody,
+                            start_key=Start, start_inclusive=StartInc,
+                            end_term=End, end_inclusive=EndInc},
+    to_index_query(IndexField, Args, Continuation, BaseQuery).
+
+%% @doc Create an index quey of the current highest version supported by
+%% cluster capability.
+%% `IndexField' is either a user supplied term or one of
+%% the inbuilt indexes (`<<"$key">>' and <<"$bucket">>').
```
engelsanchez: A little mismatched edoc quote here around bucket.

src/riak_kv_app.erl
```diff
@@ -144,6 +145,10 @@ start(_Type, _StartArgs) ->
                                           [v1, v0],
                                           v0),
 
+            riak_core_capability:register({riak_kv, '2i_version'},
```
engelsanchez: Consider using an atom name that does not require quotes.  It would be one more thing that requires Erlang knowledge when overriding this in app.config. You can see how that could be a pain in the neck for users. Also, not sure 2i is prominent as a name in the docs. secondary_index_version?

src/riak_index.erl
```diff
@@ -258,6 +271,152 @@ to_index_query(IndexField, Args) ->
...
+%% returned from a previous query, or `undefined'.
+%% @see make_continuation/1
+to_index_query(IndexField, Args, Continuation) ->
+    to_index_query(IndexField, Args, Continuation, ?KV_INDEX_Q{}).
+
+to_index_query(IndexField, Args, Continuation, BaseQuery) ->
+    Version = riak_core_capability:get({riak_kv, '2i_version'}, v1),
+    Query = to_index_query(IndexField, Args),
+    case {Version, Query} of
+        {_Any, {error, _Reson}=Error} -> Error;
```
engelsanchez: Little harmless typo: _Reson -> _Reason

src/riak_index.erl
```diff
@@ -258,6 +271,152 @@ to_index_query(IndexField, Args) ->
...
+%% from the given last result
+-spec make_continuation(list()) -> continuation().
+make_continuation([]) ->
+    undefined;
+make_continuation(L) ->
+    Last = lists:last(L),
+    base64:encode(term_to_binary(Last)).
+
+%% @doc decode a continuation received from the outside world.
+-spec decode_contintuation(continuation() | undefined) -> last_result() | undefined.
```
engelsanchez: Props for being consistent throughout with this typo! decode_contintuation -> decode_continuation

src/riak_index.erl
```diff
@@ -258,6 +271,152 @@ to_index_query(IndexField, Args) ->
...
+make_v2_query({eq, Field, Value}, Q) ->
+    Q?KV_INDEX_Q{filter_field=Field, start_term=Value, end_term=Value, return_terms=false};
+make_v2_query({range, ?KEYFIELD, Start, End}, Q) ->
+    Q?KV_INDEX_Q{filter_field=?KEYFIELD, start_term=Start, start_key=Start,
+                end_term=End, return_terms=false};
+make_v2_query({range, Field, Start, End}, Q) ->
+    Q?KV_INDEX_Q{filter_field=Field, start_term=Start,
+                 end_term=End};
+make_v2_query(V1Q, _) ->
+    {error, {invalid_v1_query, V1Q}}.
```
engelsanchez, line 324: The usual Erlang controversy: crash or return an error here.  Notice how this is directly passed to apply_continuation above, which would end up in a continuation error. Also, the result of this function is sometimes directly passed to some folding functions with side effects quickly getting out of control. Consider throwing or letting it crash instead.

src/riak_index.erl
russelldb: :blush: Sorry.

src/riak_index.erl
russelldb, line 324: This is just the existing behaviour, isn't it?

engelsanchez, line 324: make_v2_query is new. I'm talking about how it returns either a new query record or an error tuple, and it's used without matching in places like this: https://github.com/basho/riak_kv/pull/540/files#L2R306 or here through update_query https://github.com/basho/riak_kv/pull/540/files#L5R692
Versus returning something like {ok, NewVal} | {error, Reason} and matching on that and failing right at that point if something went wrong. Instead, the error would not crash immediately, but would be passed down and end up wrapped in some other error down the line.

russelldb, line 324: I see. `{ok, Q} | {error, Reason}` then `{ok, Q} = make_v2_query(V1Q, V2Q)` where it is used for bad match crashing?

engelsanchez, line 324: Yes, exactly. That would be more idiomatic and should fail with a message close to the original problem.

russelldb, line 324: Word! Will do.

src/riak_index.erl
```diff
@@ -258,6 +271,152 @@ to_index_query(IndexField, Args) ->
...
+%% @doc is an index key in range for a 2i query?
+index_key_in_range({Bucket, Key, Field, Term}=IK, Bucket,
+                   ?KV_INDEX_Q{filter_field=Field,
+                               start_key=StartKey,
+                               start_inclusive=StartInc,
+                               start_term=StartTerm,
+                               end_term=EndTerm})
+  when Term >= StartTerm,
+       Term =< EndTerm ->
+    in_range(gt(StartInc, Key, StartKey), true, IK);
```
engelsanchez: It might be worth doing some performance timing of this function and its used of in_range() and gt(). They sound like good candidates to be inlined or become macros instead, since they could be used quite a lot. I'll do some of that after reviewing the code, but I was wondering if you have already.

russelldb: I haven't, no. I did a little b_b on the whole PR, but no micro benching.

src/riak_kv_vnode.erl
```diff
@@ -677,14 +678,30 @@ handle_coverage_index(Bucket, ItemFilter, Query,
...
+handle_coverage_keyfold(Bucket, ItemFilter, Query,
+                      FilterVNodes, Sender, State,
+                      ResultFunFun) ->
+    handle_coverage_keyfold(fold_keys, Bucket, ItemFilter, Query,
+                            FilterVNodes, Sender, State, ResultFunFun).
+
+%% Until a bit of a refactor can occur to better abstract
+%% index operations, allow the ModFun for folding to be declared
+%% to support index operations that can return objects
+handle_coverage_keyfold(FoldType, Bucket, ItemFilter, ResultFun,
```
engelsanchez: We now have a bit of a misnomer here.  Consider changing to something like handle_coverage_fold instead now that we do both key only and key/value folds?

src/riak_kv_wm_index.erl
```diff
@@ -45,9 +45,13 @@
           client,       %% riak_client() - the store client
           riak,         %% local | {node(), atom()} - params for riak client
           bucket,       %% The bucket to query.
-          index_query   %% The query..
+          index_query,   %% The query..
+          max_results :: all | pos_integer(), %% maximum nunber of 2i results to return, the page size.
```
engelsanchez: Little typo here, nunber

src/riak_kv_wm_index.erl
```diff
@@ -89,24 +93,57 @@ malformed_request(RD, Ctx) ->
     IndexField = list_to_binary(riak_kv_wm_utils:maybe_decode_uri(RD, wrq:path_info(field, RD))),
     Args1 = wrq:path_tokens(RD),
     Args2 = [list_to_binary(riak_kv_wm_utils:maybe_decode_uri(RD, X)) || X <- Args1],
+    ReturnTerms0 = wrq:get_qs_value(?Q_2I_RETURNTERMS, "false", RD),
+    ReturnTerms = normalize_boolean(string:to_lower(ReturnTerms0)),
```
engelsanchez, line 97: Here normalize_boolean considers anything not in (0,no,false) to be true. This easily ends up frustrating clients due to confusion. Consider having a white list. That is, allowing only "1", "yes", "true" only as true values. I have not looked around to see what if any other boolean values we are using through HTTP here. Have you? It would be good to stay standard if we have any other parameters of the same kind.

src/riak_kv_wm_index.erl
```diff
@@ -127,19 +164,121 @@ encodings_provided(RD, Ctx) ->
 %% @spec produce_index_results(reqdata(), context()) -> {binary(), reqdata(), context()}
 %% @doc Produce the JSON response to an index lookup.
 produce_index_results(RD, Ctx) ->
-    %% Extract vars...
+    case wrq:get_qs_value("stream", "false", RD) of
+        "true" ->
```
engelsanchez, line 172: Whereas here stream has to be true to work, and anything else is considered false. We should try to stay consistent.

src/riak_kv_wm_index.erl
```diff
@@ -127,19 +164,121 @@ encodings_provided(RD, Ctx) ->
...
+                    LastResult1 = last_result(Results),
+                    Count1 = Count + length(Results),
+                    {iolist_to_binary(Body),
+                     index_stream_helper(ReqID, Boundary, ReturnTerms, MaxResults, LastResult1, Count1)};
+                {ReqID, Error} ->
+                    lager:error("Error in index wm: ~p", [Error]),
+                    Body = ["\r\n--", Boundary, "\r\n",
+                            "Content-Type: text/plain\r\n\r\n",
+                            "there was an error..."],
+                    {iolist_to_binary(Body), done}
```
engelsanchez, line 228: I'm no HTTP expert so I'm not sure what the best way to return an error while streaming index results is. Did you check with Mr. @seancribbs ?

src/riak_kv_wm_index.erl
```diff
@@ -127,19 +164,121 @@ encodings_provided(RD, Ctx) ->
...
+mochify_continuation(undefined) ->
+    [];
+mochify_continuation(Continuation) ->
+    [{?Q_2I_CONTINUATION, Continuation}].
+
+filter_values([]) ->
+    [];
+filter_values([{_, _} | _T]=Results) ->
+    [K || {_V, K} <- Results];
+filter_values(Results) ->
```
engelsanchez, line 272: What other type of results are you expecting here?

src/riak_kv_wm_index.erl
seancribbs, line 97: :+1: @engelsanchez Please use a single value there. Our client interfaces are leaky enough already.

engelsanchez, line 97: Looking around, it seems we use strict "true", "false" (and a third "default") in riak_kv_wm_object (see riak_kv_wm_object:malformed_boolean_param) for booleans. It would be better if we stick to that I think.

src/riak_kv_wm_index.erl
engelsanchez, line 228: Discussed with Sean Cribbs offline. The recommendation is to definitely return a JSON response here. Something reasonable would be:
`{ "error" : "Some reason" }`

seancribbs, line 228: loese

russelldb, line 228: behoove always wins

src/riak_kv_index_fsm.erl
```diff
@@ -77,14 +82,58 @@ req(Bucket, ItemFilter, Query) ->
 %% should cover, the service to use to check for available nodes,
 %% and the registered name to use to access the vnode master process.
 init(From={_, _, _}, [Bucket, ItemFilter, Query, Timeout]) ->
+    %% http://erlang.org/doc/reference_manual/expressions.html#id77404
+    %% atom() > number()
```
engelsanchez, line 86: I'd recommend not to leave this note here. I won't stop you from tattooing this on your wrist though.

russelldb, line 86: Previous reviewer recommended adding it, you guys have me in a spin!

engelsanchez, line 86: Ha! Bah, ignore me then. It doesn't matter.

src/riak_kv_index_fsm.erl
```diff
@@ -77,14 +82,58 @@ req(Bucket, ItemFilter, Query) ->
...
+    {ok, State#state{merge_sort_buffer=sms:new(CoverageVNodes)}}.
+
+process_results(_VNode, {error, Reason}, _State) ->
+    {error, Reason};
+process_results(_VNode, {From, _Bucket, _Results}, State=#state{max_results=X, results_sent=Y})  when Y >= X ->
+    riak_kv_vnode:stop_fold(From),
+    {done, State};
+process_results(VNode, {From, Bucket, Results}, State) ->
+    {ok, State2} = process_results(VNode, {Bucket, Results}, State),
+    riak_kv_vnode:ack_keys(From),
```
engelsanchez: I'm not sure if I got this right, but wouldn't it make sense to check for Count reaching max results here and calling stop_fold instead of ack_keys in that case here? It seems we are waiting for the next batch of results to check the count.

src/riak_kv_index_fsm.erl
```diff
@@ -77,14 +82,58 @@ req(Bucket, ItemFilter, Query) ->
...
+    {ok, State2};
+process_results(VNode, {_Bucket, Results}, State) ->
+    #state{merge_sort_buffer=MergeSortBuffer,
+           from={raw, ReqId, ClientPid}, results_sent=ResultsSent, max_results=MaxResults} = State,
+    %% add new results to buffer
+    BufferWithNewResults = sms:add_results(VNode, lists:reverse(Results), MergeSortBuffer),
```
engelsanchez: So the FSM reverses results as they are received, and the client reverses the entire list once it has received everything (https://github.com/basho/riak_kv/pull/540/files#L1R584). Without really thinking too much about the details, I wonder if we can optimize the code to minize that.

src/riak_kv_index_fsm.erl
```diff
@@ -77,14 +82,58 @@ req(Bucket, ItemFilter, Query) ->
...
+                                {BufferWithNewResults, 0};
+                            {ToSend, NewBuff} ->
+                                DownTheWire = case (ResultsSent + length(ToSend)) > MaxResults of
+                                                  true ->
+                                                      lists:sublist(ToSend, MaxResults - ResultsSent);
+                                                  false ->
+                                                      ToSend
+                                              end,
+                                ClientPid ! {ReqId, {results, DownTheWire}},
+                                {NewBuff, length(DownTheWire)}
```
engelsanchez: As a minor optimization you can avoid computing the O(n) length of the list here again (pass the length in the case statement above).

src/riak_kv_index_fsm.erl
russelldb: I'm open to suggestions. I admit it seems sort of wasteful and is apparently suboptimal…I haven't measured the cost.

engelsanchez: I think it's OK to put my little optimization notes in the backburner. Once we verify this as correct and are happy with the test coverage, we can try to squeeze the extra cycles in a different PR.

src/riak_kv_pb_csbucket.erl
```diff
@@ -0,0 +1,122 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_pb_index: Expose secondary index queries to Protocol Buffers
+%%
+%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
```
engelsanchez: Already outdated!

src/riak_kv_pb_csbucket.erl
```diff
@@ -0,0 +1,122 @@
...
+%% <p>This service produces the following responses:</p>
+%%
+%% <pre>
+%%  41 - RpbCSBucketResp
+%% </pre>
+%% @end
+
+-module(riak_kv_pb_csbucket).
+
+-include_lib("../../riak_pb/include/riak_kv_pb.hrl"). %% @TODO reset this
```
engelsanchez: You should be able to use `-include_lib("riak_pb/include/riak_kv_pb.hrl").` here, no?

russelldb: yes, see the TODO, I shoulda todone it

src/riak_kv_pb_csbucket.erl
```diff
@@ -0,0 +1,122 @@
...
+
+
+%% @doc process_stream/3 callback. Handle streamed responses
+process_stream({ReqId, done}, ReqId, State=#state{req_id=ReqId,
+                                                  continuation=Continuation,
+                                                  req=Req,
+                                                  result_count=Count}) ->
+    %% Only add the continuation if there may be more results to send
+    #rpbcsbucketreq{max_results=MaxResults} = Req,
+    Resp = case is_integer(MaxResults) andalso Count =:= MaxResults of
```
engelsanchez: I believe this is correct, but consider doing `Count >= MaxResults` for total future proofing paranoia. Same with riak_kv_pb_index actually.

src/riak_kv_pb_csbucket.erl
```diff
@@ -0,0 +1,122 @@
...
+    {ignore, State}.
+
+encode_result(B, {K, V}) ->
+    RObj = riak_object:from_binary(B, K, V),
+    Contents = riak_pb_kv_codec:encode_contents(riak_object:get_contents(RObj)),
+    VClock = pbify_rpbvc(riak_object:vclock(RObj)),
+    GetResp = #rpbgetresp{vclock=VClock, content=Contents},
+    #rpbindexobject{key=K, object=GetResp}.
+
+pbify_rpbvc(Vc) ->
```
engelsanchez, line 116: This will need revisiting once the Vclock encoding is merged! Add a TODO note somewhere. We're going to merge that one today.

engelsanchez:
There is one funny side effect of the way max results and continuations work: Currently we make no attempt to deduplicate results from a 2i query. So, if item 1 is indexed on field f1 for values 1,2,3, and you do a range query from 1 to 3, you will get the same item 3 times.  Now, if you specify max results as 2, you get the value twice, plus a continuation. If you query the continuation, you get nothing.  Since we are now sorting the results, deduplicating becomes very cheap.  I'm wondering if we should start doing it.  On the other hand, if users rely on this property and use the number of duplicates for something, continuations might break that. Thoughts?

engelsanchez:
More on the duplicates being dropped behaviour: if return_terms is true the duplicates that we are dropping may contain information useful to the client. Or at least it's easier to imagine they might care about that, since some terms will sometimes be retrieved, sometimes lost.

russelldb:
I don't understand. Can you provide a test or some minimal set of data and a query that reproduce the condition, please?

engelsanchez:
https://gist.github.com/engelsanchez/5630275
While I was playing with this I noticed the above: a single entry indexed by 3 values (1,2,3).  A range query from 1-3 will return 3 items. With return_terms, each item is tagged with one of (1,2,3). With max results = 2, you get two of those, but of course the continuation won't get that key again, so you don't get the third one.

I'm not saying this is bad or unintended. Just that it might interfere with what the users of return_terms might want to do with that.  Maybe ask them directly if it matters?

russelldb:
It is bad and unintended. I really hadn't noticed it or thought of it. I'm not sure what the answer it, except to expose the `start_key_incl` option to the end user, even then it is a bit of a guess on their part.

russelldb:
OK. I think  this is a straight up bug from considering the key and the index value in isolation. Let me work on it this morning, I have some ideas.

russelldb:
Pooshed a fix for the dropped-duplicate-key-after-continuation bug, (good catch BTW! Glad we didn't ship that.) Pooshed test for same to riak_test.

Would you say now is a good time to address your other comments (i.e. are you through with the majority of the review?)

src/riak_kv_pb_csbucket.erl
engelsanchez, line 116: OK, vclock encoding is on master now. Take a look at riak_pb_object.erl for the new way of doing this (basically use riak_object encode_vclock/decode_vclock).

engelsanchez:
I've verified the fix works on my example. Good work man! Simple, small and effective fix. Will keep banging on it to see what else I can find.

src/riak_kv_vnode.erl
russelldb: I'm not sure. I get your meaning…but…hmmm…I think really a deeper refactor is needed. The query should dictate the fold type, probably the backend should decide. I see this as a hack / halfway house. Changing the name might even add to the confusion? I totally agree with you, but think there is some more discussion to be had.

src/riak_kv_wm_index.erl
russelldb, line 272: Erm…asking the past…maybe just keys from a backend that isn't v2 yet?

src/riak_kv_index_fsm.erl
russelldb: I guess that would be an optimisation, yes. I'll do it!

russelldb: Better still would be to tell _all_ vnodes to stop_fold at once.

src/riak_kv_pb_csbucket.erl
russelldb, line 116: But that means merging with master to test. Sheesh! Or maybe…should I rebase this branch and push??

engelsanchez:
A couple more things: remember to look at my branch adding an EQC test for sms, which also contains a couple of proposed edge case fixes: https://github.com/basho/riak_kv/commits/eas-add-sms-eqc

And you will need to merge from master before the PR gets its green button again (there is a little conflict in riak_kv_app where capabilities are added). Please do that before the final :+1: so that I can go over the final version of the code.

engelsanchez:
I'm having trouble using max_results and return_terms together through the HTTP interface

```
curl http://localhost:10018/buckets/bucket/index/i1_int/1/4?return_terms=true&max_results=2
{"results":[{"1":"1"},{"2":"1"},{"2":"2"},{"3":"1"},{"3":"2"},{"4":"2"}]}
```

```
curl http://localhost:10018/buckets/bucket/index/i1_int/1/4?max_results=2&return_terms=true
{"keys":["1","1"],"continuation":"g2gCYQJtAAAAATE="}
```

Also, validation doesn't seem to kick in for the second parameter. Maybe I'm doing something stupid, but I haven't figured it out. I'll debug this more right after lunch.

russelldb:
This is why I hate changes on review. I tested this stuff in the past. Ok. I'll add some more tests to riak_test and see what is up. Thanks.

russelldb:
Huh. Works for me. I enquoted my url+query params thussly:

```
curl "http://localhost:8098/buckets/mybucket/index/field1_int/1/4?max_results=2&return_terms=true"
{"results":[{"1":"1"},{"2":"2"}],"continuation":"g2gCYQJtAAAAATI="}
```

and

```
curl "http://localhost:8098/buckets/mybucket/index/field1_int/1/4?return_terms=true&max_results=2"
{"results":[{"1":"1"},{"2":"2"}],"continuation":"g2gCYQJtAAAAATI="}
```

and

```
curl "http://localhost:8098/buckets/mybucket/index/field1_int/1/4?max_results=2&return_terms=bum"
Invalid "return_terms". "bum" is not a boolean
```

and

```
curl "http://localhost:8098/buckets/mybucket/index/field1_int/1/4?return_terms=false&max_results=hippo"
Invalid "max_results". "hippo" is not a positive integer
```

Are all your branches up-to-date? I'll push the latest, which merges master and your EQC test (thanks for that, great work!)

/me shrugs

engelsanchez:
Yep. I tried on the browser and it works fine too. Nothing to see here, move along, nevermind... Sorry

russelldb:
Heh, better safe, no worries.

engelsanchez:
Ok, I think we can call it a day. I believe all comments were addressed. New testing has not uncovered any new problems. Ship it!
:+1: :dancer: :boat:

slfritchie:
From the commit log:

@rdb Please check to see how much effort vs. value might be required
to address the TODO comment.  If it's too much work now, then please
create an issue for fixing & removing that TODO item so that the matter
doesn't get lost.

slfritchie:
Oh, and the bug fixed by commit  6934d65 is something that only pops up every few minutes, in my experience with the backend_eqc test.  The namespace (bucket + key + index) is likely too big to hit a case where a bucket + key + index is written and then sometime later a fold_keys is attempted with a query on the exact same index.

So, whenever a developer plays with areas that would affect the code that this test exercises, it would be extremely wise to run the backend_eqc test for a much longer period of time than the usual "make test" does.  See my comment on that commit for an example.

russelldb:
Awesome @slfritchie. Thanks for finding and fixing. I think I'll fix that test properly during this testing cycle (add all the new query types, make it use V2 queries etc.) Thanks again. It was on the list to update the test, but time is cruel.

PR #541, opened 2013-05-01
Deprecate KV health checks

engelsanchez:
We are deprecating KV health checks in their current form as there are problems with flapping with clusters under load. Better load protection mechanisms are being implemented and/or planned.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -144,15 +144,16 @@ start(_Type, _StartArgs) ->
                                           [v1, v0],
                                           v0),
 
+            HealthCheckOn = app_helper:get_env(riak_kv, enable_health_checks, false),
             %% Go ahead and mark the riak_kv service as up in the node watcher.
             %% The riak_core_ring_handler blocks until all vnodes have been started
             %% synchronously.
             riak_core:register(riak_kv, [
                 {vnode_module, riak_kv_vnode},
-                {health_check, {?MODULE, check_kv_health, []}},
                 {bucket_validator, riak_kv_bucket},
                 {stat_mod, riak_kv_stat}
-            ]),
+            ]
+            ++ [{health_check, {?MODULE, check_kv_health, []}} || HealthCheckOn]),
 
             ok = riak_api_pb_service:register(?SERVICES),
 


jrwest:
does this add another config param to `riak_kv`? If so, why not just reuse `vnode_mailbox_limit` from the actual health check [1] w/ a special token like `undefined`?

[1] https://github.com/basho/riak_kv/blob/master/src/riak_kv_app.erl#L224

engelsanchez:
Hmmm... That seems a bit obscure, and it would actually enable health checks for people who have tweaked it. We want to make sure that it's disabled for everybody unless they want it back, or so is my understanding of our intention. Also, vnode_mailbox_limit looks a great name for a config parameter that could be used in future versions of health checks. This config parameter is meant never to be surfaced, and silently dropped in the next release along with the kv health checking code probably.

jrwest:
im not sure the plan is to outright drop the kv health check code but perhaps expand on it. 

Personally, I dont find `{vnode_mailbox_limit, undefined}` to be too obscure since the documented meaning of that parameter is to control the mailbox thresholds used by the kv health check, if its undefined it means "has no threshold" which is what we are doing by disabling it. Also, overloading that meaning for another check is not a good idea imo (confusing on upgrade). 

Not sure I follow the it would "actually enable the health checks for people who have tweaked it part" but to clarify I was suggesting reading this config var where you are reading `health_checks_enabled` now and treating `undefined` like you are treating `false` now.

jrwest:
Also fwiw, there is already an `enable_health_checks` in core: https://github.com/basho/riak_core/blob/master/src/riak_core_node_watcher.erl#L206

engelsanchez:
What I mean is that if we use that parameter, and it is set in an app.config that people don't tweak with the upgrade, health checks will be enabled. Which is exactly why I didn't just use the riak_core enable_health_checks. That is, we want people who upgrade to have these health checks disabled without doing anything, because many people don't change their configs anyway. Unless they really, really want them, in which case they can enable this new config value. Re-using any existing parameters would require people to have to go and change them. If we were going to do that, we can already tell them to change riak_core enable_health_checks to false and we wouldn't need to do anything. That is my understanding.

jrwest:
ah ok, yea i forgot the upgrade path can be a pain. its shame to have to add another parameter but i see your case now. fwiw, i think this would be solved by some sort of `riak-admin give-me-the-new-shiny`

engelsanchez:
This PR was opened against 1.4, but we want this feature sooner for 1.3.2. #549 is the one that will be merged and this one will be orphaned, poor thing.

PR #543, opened 2013-05-03
url encode/decode 2i headers

broach:
This stems from the StackOverflow Q found here:
http://stackoverflow.com/questions/16335296/url-unsafe-secondary-index-names-in-riak-do-not-work

Currently we don't URL-decode the field-name or the value of x-riak-index-*
headers on a POST/PUT. The problem then arrises that on a GET for an index
you need to url-encode then url-escape everything because we do url-decode the
hierarchical part of the URL. It also then means that the same query
using protocol buffers has to be url-encoded which is ... silly.

By url-decoding the field-name and the value then storing the raw bytes we
avoid all this confusion. GET requests work as expected using HTTP and PB
queries do not have to be URL encoded.

The one outstanding consideration is ... what to do about people who
are currently URL-encoding / escaping, etc. When the decoding was added
for the GET (which is actually what broke this, ironically enough) there
was a environment setting and an optional header added to toggle the feature
on/off (see [riak_kv_wm_utils:maybe_decode_uri/2](https://github.com/basho/riak_kv/blob/x-riak-index-url-decode/src/riak_kv_wm_utils.erl#L46) - today the default is "on")

I can easily add the same type of thing which I'm thinking we may have to do?

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -426,14 +426,16 @@ extract_index_fields(RD) ->
                     true ->
                         %% Isolate the name of the index field.
                         IndexField = list_to_binary(element(2, lists:split(PrefixSize, KList))),
-
+                        %% url-decode the header field-name and value
+                        IndexField1 = mochiweb_util:unquote(IndexField),        
+                        V2 = mochiweb_util:unquote(V),
                         %% HACK ALERT: Split values on comma. The HTTP
                         %% spec allows for comma separated tokens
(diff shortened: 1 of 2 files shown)


slfritchie:
Pulling PRs out of the freezer ... what's the status of this PR?

jtuple:
It looks like we're out of time to get this into 2.0, so I'm assigning to 2.1.

yorkshire-steve:

PR #545, opened 2013-05-04
Changes to support vnode overload protection

jtuple:
This pull-request provides the `riak_kv` components for the new vnode overload protection work that is implemented in basho/riak_core#314


jtuple:
Accidentally opened this PR against `master` instead of `1.3` branch. Correct PR here: #547

PR #546, opened 2013-05-07
Expose the put_fsm 'asis' option to clients

seancribbs:
This is an obscure feature of `riak_kv_put_fsm` that was originally only used for replication, but has potential benefits for bulk-loading. By adding the `asis` flag, clients can indicate that this is not an update and so the vector clock should not be incremented.


evanmcc:
:+1:

PR #547, opened 2013-05-07
Changes to support vnode overload protection

jtuple:
This pull-request provides the riak_kv components for the new vnode overload protection work that is implemented in basho/riak_core#314

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -770,6 +770,14 @@ mk_reqid() ->
 %% @private
 wait_for_reqid(ReqId, Timeout) ->
     receive
+        {ReqId, {error, overload}=Response} ->
+            case app_helper:get_env(riak_kv, overload_backoff, undefined) of
+                Msecs when is_number(Msecs) ->
+                    timer:sleep(Msecs);
+                undefined ->
+                    ok
+            end,
+            Response;
         {ReqId, Response} -> Response
     after Timeout ->
             {error, timeout}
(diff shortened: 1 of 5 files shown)


src/riak_client.erl
```diff
@@ -539,6 +539,14 @@ mk_reqid() ->
 %% @private
 wait_for_reqid(ReqId, Timeout) ->
     receive
+        {ReqId, {error, overload}=Response} ->
+            case app_helper:get_env(riak_kv, overload_backoff, undefined) of
+                Msecs when is_number(Msecs) ->
+                    timer:sleep(Msecs);
```
engelsanchez, line 545: I'm worried about internal backoff code like this. Couldn't this potentially lead to an accumulation of processes under heavy load? Ideally external clients would be doing exponential backoff or similar instead so internal Riak resources are not involved. /cc @jonmeredith I think this is from you.

jonmeredith, line 545: I agree with you.  The desirable solution is to have that built into the app layer.

The reason I did this as an app config is to give us an option to turn when things are going south.  Picture if you will, a cluster in overloaded meltdown, it's the middle of the night and the app team is nowhere to be found.   You have a very tired syseng on the line with you who needs to get back operational.  With this setting you can slow up the request rate to riak.

You don't have to enable it.  And it only fires when things are bad.

engelsanchez:
Besides my concern for the use of the backoff code in riak_client (which I hope we'd never have to do), this code has been observed to prevent overload in manual tests. I've also run several riak_test suites (java client, python client and a couple of others) without finding any functional problems.  We should follow up with more performance testing to keep validating this, but I've seen graphs from Joe that look really good (no apparent performance penalty under load).
:+1: :dancer: :boat:

PR #549, opened 2013-05-10
Deprecate KV health checks

engelsanchez:
We are adding other load protection mechanisms in KV and consider this
implementation of health checks as troublesome since it can cause nasty
flappings when the cluster is under load, so it's being deprecated.

This leaves in place the riak_core health checks mechanism, but disables
its use by riak_kv unless {enable_health_checks, true} is added in the
riak_kv section.  That way, customers who have the previously shipped
config with {enable_health_checks,true} in the riak_core section will
also stop having health checks, but could re-add if really necessary.

This is a retargeting of #541 to merge against 1.3. See that one for original review comments, etc.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -137,15 +137,16 @@ start(_Type, _StartArgs) ->
                                           [enabled_v1, disabled],
                                           disabled),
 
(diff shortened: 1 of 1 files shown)


jrwest:
started riak w/ this change and verified kv health check was not registered with `riak_core_node_watcher`. +1.

PR #551, opened 2013-05-15
Backend capability to use Riak r_object records

slfritchie:
Add the optional riak_kv backend capability to pass Riak r_object records to/from the backend rather than binary blobs that are serialized/unserialized by `riak_kv_vnode`.

Also, rearrange the QuickCheck tests to permit easier testing of the fs2 backend.


slfritchie:
Derp, this is based on the wrong base, closing, will open another.

PR #553, opened 2013-05-15
Add sloppy_quorum and n_val options

slfritchie:
Add experimental `sloppy_quorum` and `n_val` options.  "Experimental" because there's only one user for them at the moment, Riak CS; any other app that uses them should be aware that they may change or disappear without warning.

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -362,11 +362,12 @@ finalize(StateData=#state{get_core = GetCore}) ->
 %% Maybe issue deletes if all primary nodes are available.
 %% Get core will only requestion deletion if all vnodes
 %% replies with the same value.
-maybe_delete(_StateData=#state{n = N, preflist2=Sent,
-                               req_id=ReqId, bkey=BKey}) ->
+maybe_delete(StateData=#state{n = N, preflist2=Sent,
+                              req_id=ReqId, bkey=BKey}) ->
     %% Check sent to a perfect preflist and we can delete
     IdealNodes = [{I, Node} || {{I, Node}, primary} <- Sent],
-    case length(IdealNodes) == N of
+    NotCustomN = not using_custom_n_val(StateData),
+    case NotCustomN andalso (length(IdealNodes) == N) of
         true ->
             ?DTRACE(?C_GET_FSM_MAYBE_DELETE, [1],
                     ["maybe_delete", "triggered"]),
@@ -375,6 +376,14 @@ maybe_delete(_StateData=#state{n = N, preflist2=Sent,
             ?DTRACE(?C_GET_FSM_MAYBE_DELETE, [0],
                     ["maybe_delete", "nop"]),
             nop
+    end.
+
+using_custom_n_val(#state{n=N, bucket_props=BucketProps}) ->
+    case proplists:get_value(n_val, BucketProps) of
+        N ->
+            false;
+        _ ->
+            true
     end.
 
 %% based on what the get_put_monitor stats say, and a random roll, potentially
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -46,7 +46,7 @@
                   {details, true} |
                   details |
                   {sloppy_quorum, boolean()} | %% default = true
-                  {n_val, boolean()}.          %% default = bucket props
+                  {n_val, pos_integer()}.      %% default = bucket props
 
 -type options() :: [option()].
 -type req_id() :: non_neg_integer().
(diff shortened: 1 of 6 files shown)


slfritchie:
NOTE: Branches are apparently hard.  This should be merged to _master_.  Ignore the diffs from Brian's two commits.

src/riak_kv_get_fsm.erl
```diff
@@ -44,7 +44,9 @@
                   {timeout, pos_integer() | infinity} | %% Timeout for vnode responses
                   {details, details()} |       %% Return extra details as a 3rd element
                   {details, true} |
-                  details.
+                  details |
+                  {sloppy_quorum, boolean()} | %% default = true
+                  {n_val, boolean()}.          %% default = bucket props
```
russelldb: should this be `pos_integer()`?

slfritchie: No, it shouldn't.  Commit 2bbee66 takes care of that.  Odd that Dialyzer output isn't altered with the correction.

Which reminds me, there's a cut-and-paste error in a comment in the put FSM code.  I'll go fiddle with that now.

jtuple:
In cases where the user provides an `n_val` we don't check that it's less than the bucket N-value. This is experimental and all, so don't need to address now, but probably should note that problem in the code somewhere as a TODO to fix if this sticks around longterm.

jtuple:
In the get/put FSMs, I suggest using `get_primary_apl` rather than filtering in the FSMs. Makes it easier to change things in the future.

Something like:

``` erlang
Preflist2 = case proplists:get_value(sloppy_quorum, Options, true) of
                true ->
                    riak_core_apl:get_apl_ann(DocIdx, N, Ring, UpNodes);
                false ->
                    riak_core_apl:get_primary_apl(DocIdx, N, Ring, UpNodes)
            end,
```

jtuple:
General question. Why are we providing ability to specific N-value for puts? Thought the plan was only to do so for gets. There's many, many areas in Riak that assume bucket-N-value replicas exist. Obviously, experimental and all; but I would tread very lightly on less than configured N-value writes given how pervasive that implicit assumption is.

slfritchie:
About the general question: Riak CS use case.  If the block (which is large'ish by Riak recommended guidelines)  is immutable, we don't need to pester N replicas for the exact same large'ish block.  If we get a notfound, fine, ask again with the default n_val and let read-repair do its thing.

jtuple:
Right, I understand that for gets. I was asking about puts. Is Riak CS planning to do puts that are smaller than N-val?

slfritchie:
If it does, it will do so knowing full well that things that use n_val, e.g. AAE and manual vnode repair, won't know how to honor the n_val.  So, mostly likely not right now.

But given the problem that the only other ways to configure n_val is default bucket props or per-bucket props in the ring ... any app like Riak CS that manufactures zillions of Riak buckets, the existing methods are insufficient.

slfritchie:
Joe, I think I've addressed both review bits with a new commit.  Anything else?

jtuple:
Semantics around tombstone harvesting are undesirable when performing `n_val` gets.

Basic issue: issue normal delete but have tombstone reach < N replicas. Read the object using `{n_val, 1}`. Object is deleted entirely on the replica that replies. This is different than normal Riak that will only reap tombstones if all primaries reply with a tombstone. Straightforward to extend this scenario to something like "delete, get, partition/node heals, get" sequence that resurrects deleted data.

I can't imagine any scenario where we **want** tombstones to be harvested with less than bucket-N-value primaries responding. So, I'm going to assume this is just a corner case we want to fix.

Seems like the easiest approach is to modify [maybe_delete/1](https://github.com/basho/riak_kv/blob/84bee8a27cc5ed58dff7b2761c9451d86923dc61/src/riak_kv_get_fsm.erl#L362-L378) to either check against actual bucket-N-value or (even simpler) just default to false if the new `n_val` option was provided in the get request.

slfritchie:
From internal chat: In the Riak CS use case, there is no get of those objects that needs to trigger a delete by itself.  Those n_val=1 gets are only for fetching S3 object contents.  If the object's blocks are tombstoned, the get to reap them will be a standard get as done by riak_kv_delete.erl [SLF: without specifying n_val].

jtuple:
Understood. So seems like the second approach I suggested would work best for Riak CS then? Ie. disable tombstone harvesting anytime we specify `n_val` in the get request. I think that's also the simplest change to make.

slfritchie:
@jtuple I think a8831a6 implements what you had in mind?

slfritchie:
@jtuple I took your gist from earlier today and committed it into 8826a92.

jtuple:
Nothing else on my list. +1 merge.

PR #555, opened 2013-05-16
Add init/final to AAE remote interface

buddhisthead:
Init and final message handlers allow us to integrate the compare function with more complex applications.


rzezeski:
+1 to merge

Chris and I already discussed this but I just wanted to state for the record that I think the init/final stages are a bit outside of the scope of the `Remote` fun.  It's original intention is so that a hashtree has a generic way to make calls to a remote hashtree.  These stages, at least in the case of repl, are being used outside the context of the hashtree.  However, the reason for doing so is because repl needs the init/final stages to run on the compare process and this was the easiest way to do that.  We could refactor AAE to make init/final explicit but it would be slightly more complicated.  I think this is a good pragmatic solution.  However, in the future if we factor AAE into a common framework in core we may want to revisit the init/final stages and make them an explicit part of AAE exchange.

PR #556, opened 2013-05-21
Enable stat updates to use sidejob workers

jtuple:
This pull-request enables Riak K/V stat updates to be sent to `sidejob` workers to avoid the impact of updating stats directly in get/put FSM processes. To keep the overload low, this change uses `sidejob:unbounded_cast` to dispatch the requests. Of course, this prevents `sidejob` from being able to enforce a capacity limit. In practice, message queues are kept reasonable given the use of multiple workers processes to handle the requests.

This pull-request makes the use of `sidejob` for stat updates the default behavior. However, adding `{direct_stats, true}` to the `riak_kv` section of `app.config` can override this behavior, returning to the direct, in-process approach.

Requires related pull-request: basho/sidejob#3

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -11,7 +11,7 @@
 
 {deps, [
         {riak_core, "1.3.2", {git, "git://github.com/basho/riak_core", {tag, "1.3.2"}}},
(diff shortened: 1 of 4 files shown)


russelldb:
+1. Ran perf tests and looks good. Nice that there is an app.config option so we can switch if needed.

PR #557, opened 2013-05-21
Add fsm active and error stats to the blob

russelldb:
Somehow these went missing from the stat blob, and it made people sad when trying to figure out how many fsm processes were active.


jtuple:
This change already made it into master (and thus 1.4) when the same change was made to 1.3 and 1.3 was merged to master. 

Relevant commit: e8a8fa8a3057ed9112a540fb653fc1ef4e712f5b

These stats were further changed in a later commit: 543692063489186e88b85806bac15f9b13ce75a8

Closing out pull-request.

PR #558, opened 2013-05-21
Add fsm active and error stats to the blob

russelldb:
Somehow these went missing from the stat blob, and it made people sad when trying to figure out how many fsm processes were active.


jrwest:
+1. fetched stats w/ and w/o this change. w/ change the added fields are included as expected.

PR #560, opened 2013-05-22
Fix regression in 2i reformat status flag & add extra status function

jrwest:
This PR fixes a regression introduced by  https://github.com/basho/riak_kv/commit/b50d1df73aee20c4260e72d73fdfcf6493603636 (see commit/patch for more details) and adds `riak_kv_status:fixed_index_status` which is meant to be used by riaknostic


engelsanchez:
I've reviewed the code changes and verified them using the riak_test. I've discovered a different issue with the reformat, which I'll open a different issue for (it does not handle a worker crash: it hangs forever in that case).

:+1: :dancer: :boat:

PR #561, opened 2013-05-23
Handle worker errors and timeout on 2i reformat

engelsanchez:
Before this fix, I noticed the reformat basically hang forever with a
crashed worker error once. This fix handles the error message from the
vnode worker and adds a big 2 minute timeout to make sure we don't get
stuck. In reality, it's mostly harmless, but annoying to leave a process
hanging forever like that.

/cc @jrwest


evanmcc:
:+1:, test passes with small manual change due to change in riak-erlang-clent's handling of index results.

PR #562, opened 2013-05-23
use old object format by default on upgrade

jrwest:
also provide convenient config var instead of having to use capability overrides.

as the TODO in the patch mentions, this is the quick and easy but not ideal solution for future upgrades.

related riak_test change: https://github.com/basho/riak_test/pull/284


evanmcc:
+1

PR #563, opened 2013-05-24
Add binary format for counters

russelldb:
May not be optimal, but adds Tag and Vsn fields to make upgrade / downgrade simpler in future.

### DIFF
--- a/src/riak_kv_counter.erl
+++ b/src/riak_kv_counter.erl
@@ -36,9 +36,17 @@
 
 -module(riak_kv_counter).
 
--export([update/3, merge/1, value/1]).
+-export([update/3, merge/1, value/1, new/2]).
 
 -include("riak_kv_wm_raw.hrl").
+-include_lib("riak_kv_types.hrl").
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-define(TAG, 69).
+-define(V1_VERS, 1).
 
 %% @doc Update `Actor's entry by `Amt' and store it in `RObj'.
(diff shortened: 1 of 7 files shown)


src/riak_kv_counter.erl
```diff
@@ -145,3 +157,33 @@ later(TS1, TS2) ->
...
+-ifdef(TEST).
+
+roundtrip_bin_test() ->
+    PN = riak_kv_pncounter:new(),
+    PN1 = riak_kv_pncounter:update({increment, 2}, <<"a1">>, PN),
+    PN2 = riak_kv_pncounter:update({decrement, 1000000000000000000000000}, douglas_Actor, PN1),
+    PN3 = riak_kv_pncounter:update(increment, [{very, ["Complex"], <<"actor">>}, honest], PN2),
+    PN4 = riak_kv_pncounter:update(decrement, "another_acotr", PN3),
+    Bin = to_binary(PN4),
+    ?debugFmt("Bin ~p t2b ~p", [byte_size(Bin),
```
jrwest: missed a debug print. although i found it kind of useful to quickly verify the size difference.

jrwest:
riak_test/eqc/eunit tests still pass, booted it up and used the http api a bit, all looks good. A debug print (which I commented on) shows a few bytes of savings and having versioned binaries is always better. +1!

src/riak_kv_counter.erl
russelldb: oops, actually, that test does nothing, except print that size diff

Wait, it tests it too, ignore me, it's late for me.

PR #565, opened 2013-05-25
Fix put_fsm_eqc after local_put_failed change

engelsanchez:
As part of the overload protection, the put fsm code was changed from
returning local_put_failed on failure to returning the request id, which
could sometimes be instead the overload token (hacky hack hack)
See commit https://github.com/basho/riak_kv/commit/29720ed4cc06663116caaffdfa497cbd6804e10e
Also, add sidejob to hard coded list of deps for test
I am not sure why we need it, but the riak_kv application will not start
during the test without it. Sigh...

### DIFF
--- a/src/riak_kv_test_util.erl
+++ b/src/riak_kv_test_util.erl
@@ -245,7 +245,7 @@ dep_apps(Test, Extra) ->
         end,
 
     [sasl, Silencer, crypto, public_key, ssl, riak_sysmon, os_mon,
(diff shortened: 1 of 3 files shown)


engelsanchez:
Added Scott's fix to ec_eqc from master

jrwest:
all tests previously failed on 1.3 (or wouldn't even run) and now pass. :+1: good digging @engelsanchez

PR #566, opened 2013-05-25
Eas 1.3.2 to master

engelsanchez:
Staged merge from the 1.3 branch.  This passed all eunit tests in riak_kv (through make test) so I think it finally is good enough to merge.

/cc @jrwest @slfritchie @jtuple @russelldb please take a peek when you can so we can merge this sucker.

Later today I'll prepare the staged merges of the other repos (those are ready, but got no time right now to put them up)


russelldb:
@engelsanchez Looks legit to me. Thanks for doing this.

PR #567, opened 2013-05-26
Get rid of parametrized modules

zinid:
Mostly the same as #469, but without tuple modules, e.g. C:function(...) becomes riak_client:function(C, ...). Eunit passes without any errors.
NOTE: absolutely backward _incompatible_.
This is _not_ a concurrent pull-request. This is just an option if you don't like tuple modules.


evanmcc:
Missed this one earlier.  Also closing b/c we went with the original patch or something similar.

PR #568, opened 2013-05-27
Hscic 2i applyregex

martinsumner:
Follow discussions with Russell Brown et al, this is the regular expression feature for 2i performance improvements written as a branch to extend the other 2i changes (streaming, pagination etc).  This has been written to support the requirements of the NHS Spine project in England.

The idea is that a regular expression can be passed as a parameter in a 2i query (using the HTTP protocol) - with the key 'apply_regexp' used with the value being a regular expression. If the regular expresion is invalid (which means it won't compile when erlang attempts to re:compile/1 it with no options), then a a bad request response is returned with details of the compilation failure.  The regular expression will be applied to all index terms which are returned in a 2i range query (and will be ignored if it is anything but a range query) - and only passing terms will be added to the buffer of results.  This allows for index terms to be overloaded with concatenated pieces of information, and results filtered by this information when the ordering of the information in the term prevents the filtering being applied to the range.  As this reduces the information being sent between vnodes, and horizontally distributes the process of the regex across the cluster, it is expected that this may often be 50% faster than using return_terms and filtering terms in the application. 

The feature should work independently of return_terms - return_terms can be either true or false, and the expression will still be applied.  With max_results set, only those results within the range and passing the regular expression will be returned and count towards the maximum results.  It works with streaming (although you may get repeated sequences of empty result sets) if there is a low pass rate on the regular expression.

Implementation has been done by extending the v2 query record to support a term_regexp field.  A skip_result atom has been added as a potential alternative Val to be passed to the accumulator/buffer, when there is nomatch for the regular expression on the term - and that accumulator/buffer then doesn't add the result to the set.

I've run a set of local tests to prove counts match between sending the regexp to the cluster, applying it in the client after using return_terms.  Also basic functional tests that behaviour with different other parameter combinations is as expected.  I've not forked riak_test, and added any tests here, due to understanding issues on my part.

I've only considered the eleveldb backend, as I believe this is the only one to support v2 queries.

Martin


include/riak_kv_index.hrl
```diff
@@ -7,7 +7,8 @@
           return_terms=true :: boolean(), %% Note, should be false for an equals query
           start_inclusive=true :: boolean(),
           end_inclusive=true :: boolean(),
-          return_body=false ::boolean() %% Note, only for riak cs bucket folds
+          return_body=false ::boolean(), %% Note, only for riak cs bucket folds
+          term_regexp %% This should carry a compiled regular expression or the atom false
```
russelldb, line 11: I think it is ok for it to be a compiled regexp or the atom `undefined`. That is quite idiomatic for erlang.

src/riak_index.erl
```diff
@@ -308,18 +309,17 @@ to_index_query(IndexField, Args, Continuation, BaseQuery) ->
 
 %% @doc upgrade a V1 Query to a v2 Query
 make_v2_query({eq, ?BUCKETFIELD, _Bucket}, Q) ->
-    Q?KV_INDEX_Q{filter_field=?BUCKETFIELD, return_terms=false};
+    Q?KV_INDEX_Q{filter_field=?BUCKETFIELD, return_terms=false, term_regexp=false};
```
russelldb, line 312: As mentioned in the comment in `riak_index.hrl` leaving this field alone to its default `undefined` is ok. Same for lines 315, 317, 320

src/riak_kv_app.erl.orig
```diff
@@ -0,0 +1,284 @@
+%% -------------------------------------------------------------------
```
russelldb: Not sure why this file is in the PR. Can you remove it, please?

src/riak_kv_eleveldb_backend.erl
```diff
@@ -640,16 +663,33 @@ fold_keys_fun(FoldKeysFun, {index, FilterBucket, Q=?KV_INDEX_Q{filter_field=Filt
...
-                              true -> Key
-                          end,
+                    Val = 
+                        case {Terms, TermRegExp} of
+                            {false, false} -> 
+                                Key;
+                            {true, false} ->
+                                {Term, Key};
+                            {false, _} ->
+                                case re:run(Term, TermRegExp) of
```
russelldb: Maybe move this logic into `riak_index:index_key_in_range/3`? There is already, in that code, the facility to skip an item without aborting the fold, and saves adding skipping to the fold buffer.

src/riak_kv_fold_buffer.erl
```diff
@@ -64,6 +64,8 @@ new(MaxSize, Fun) ->
 %% the accumlated buffer items and
 %% then the buffer is emptied.
 -spec add(any(), buffer()) -> buffer().
+add(skip_result, Buffer) ->
```
russelldb: see comment about using `riak_index:index_key_in_range/3` on `riak_kv_eleveldb_backend`

russelldb:
I've made some initial comments.

2 other things
1. It will be easier to read this PR when its parent branch is merged to master. I think that will happen today so it is not worth re-pointing it at pt29-2i-pagination
2. It would be great if the PB endpoint also supported this feature, but I understand if you leave that for now.

We're having a bit of a discussion internally about alternative ways to provide the regexp from the client, but in general we like this feature and we're very grateful for the PR.

Happy to discuss the comments via email or on here as you prefer. Cheers!

src/riak_index.erl
martinsumner, line 312: I think I was confused here about what make_v2 query was attempting to do.  I'm forcing this to false at this stage to make sure that a passed term_regexp is ignored when it doesn't make sense in context (e.g. if it is a bucket/key query or an exact term match).  This is what I thought was going on between riak_index:return_terms/2 and make_v2_query, and so I was trying to do something similar between riak_index:apply_regex/2 and make_v2_query.  

If I leave these as undefined, then I may pass a regular expression through when it made no sense, as apply_regex wouldn't detect this.  

I suspect I've just got the wrong end of the stick here so some further advice would be helpful.

src/riak_kv_app.erl.orig
martinsumner: Oops, will do.

src/riak_kv_eleveldb_backend.erl
martinsumner: Yes, will do

src/riak_index.erl
russelldb, line 312: Ah, I see. Ok. So we're dealing with a sort of tri-state thing here: `undefined | false | regexp()`, where `undefined` is the initial state (and should never be seen by the backend), `false` means `unsupported` or `not_applicable`, and tells the backend there is no regexp filter, and anything else _is_ a compiled regexp to apply. I get it now. I guess I'm ok with `false`, I was thrown 'cos we're not dealing with a boolean.

slfritchie:
Pulling PRs out of the freezer ... given that it's a crazy part of our work cycle plus end-of-year, but I'll ask anyway: should this ticket be kept open?

russelldb:
This has been taken over by @engelsanchez and merged into the 1.4 branch for imminent release. This can be closed.

engelsanchez:
Yep. This has now been implemented in 1.4 and develop for 2.0

PR #573, opened 2013-06-06
Rework riak-admin test to support local and cluster options

bsparrow435:
Add `riak:client_test(cluster)` which creates a list of bucket/keys which maps to all preflists in the ring.  Each keys does a get/put/get cycle with R=quorum and W|DW=N through a client connected to the local node. 

The list of `{Preflist,{Bucket,Key}}` is saved to `Bucket: <<"__riak_client_test__">> Key: <<"list">>` for future use. This is ok because of consistent hashing as long as the n_val is not changed on the <<"**riak_client_test**">> bucket or the ring_size is not changed. I plan on adding logic to check if the ring_size has been changed(something like check the length of the stored list versus the length of `riak_core_ring:all_preflists(Ring, NValue).`)

Legacy functionality is not changed. The local node test still uses a static key that may or may not map to the local node. I am planning on changing this soon.

### DIFF
--- a/src/riak.erl
+++ b/src/riak.erl
@@ -155,8 +155,10 @@ code_hash() ->
 
 client_test(cluster) ->
     {ok, Client} = riak:local_client(),
-    BKeyList = get_test_key_list([node()|nodes()]),
-    client_test_phase_loop(BKeyList,Client);
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    NValue = proplists:get_value(n_val,riak_core_bucket:get_bucket(?CLIENT_TEST_BUCKET, Ring)),
+    BKeyList = get_test_key_list(Ring, Client),
+    client_test_phase_loop(BKeyList, Client, NValue);
     %io:format("Keylist: ~p~n", [L]);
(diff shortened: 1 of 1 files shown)


bsparrow435:
basho/riak#307

slfritchie:
Pulling PRs out of the freezer ... Brian, any chance that you could hunt down a reviewer for this PR?

jrwest:
added to kill bill so maybe it has a better chance of getting a reviewer (maybe not?)

jtuple:
Anytime to do this for 2.0 or should we punt to 2.1?

reiddraper:
Is this blocked on something? Should we be moving to 2.1?

bsparrow435:
Nope, just needs a review. I haven't touched it in a while. Will need a
rebase before merge.

reiddraper:
This has languished for quite some time. I will be moving to 2.1 if I don't hear any objections before the end-of-day.

reiddraper:
@bsparrow435 going once before moving to 2.1. That ok?

reiddraper:
Moved to 2.1.

yorkshire-steve:

PR #575, opened 2013-06-07
When backend has uses_r_object capability, encode correctly (e.g. handoff)

slfritchie:
Required bugfix for 1.4.0 release, sorry I missed it earlier.

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -607,8 +607,8 @@ to_binary_version(v1, _, _, <<?MAGIC:8/integer, 1:8/integer, _/binary>>=Bin) ->
     Bin;
 to_binary_version(Vsn, B, K, Bin) when is_binary(Bin) ->
     to_binary(Vsn, from_binary(B, K, Bin));
-to_binary_version(Vsn, _B, _K, O) when is_record(O, r_object) ->
-    to_binary(Vsn, O).
+to_binary_version(Vsn, _B, _K, Obj = #r_object{}) ->
+    to_binary(Vsn, Obj).
 
 %% @doc return the binary version the riak object binary is encoded in
 -spec binary_version(binary()) -> binary_version().
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -605,8 +605,11 @@ to_binary_version(v0, _, _, <<131,_/binary>>=Bin) ->
     Bin;
 to_binary_version(v1, _, _, <<?MAGIC:8/integer, 1:8/integer, _/binary>>=Bin) ->
     Bin;
-to_binary_version(Vsn, B, K, Bin) ->
-    to_binary(Vsn, from_binary(B, K, Bin)).
+to_binary_version(Vsn, B, K, Bin) when is_binary(Bin) ->
+    to_binary(Vsn, from_binary(B, K, Bin));
+%% r_object record definition isn't available, so hack the guard.
+to_binary_version(Vsn, _B, _K, O) when is_tuple(O), element(1, O) == r_object ->
+    to_binary(Vsn, O).
 
 %% @doc return the binary version the riak object binary is encoded in
 -spec binary_version(binary()) -> binary_version().
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -607,8 +607,7 @@ to_binary_version(v1, _, _, <<?MAGIC:8/integer, 1:8/integer, _/binary>>=Bin) ->
     Bin;
 to_binary_version(Vsn, B, K, Bin) when is_binary(Bin) ->
     to_binary(Vsn, from_binary(B, K, Bin));
-%% r_object record definition isn't available, so hack the guard.
-to_binary_version(Vsn, _B, _K, O) when is_tuple(O), element(1, O) == r_object ->
+to_binary_version(Vsn, _B, _K, O) when is_record(O, r_object) ->
     to_binary(Vsn, O).
 
 %% @doc return the binary version the riak object binary is encoded in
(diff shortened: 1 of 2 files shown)


src/riak_object.erl
```diff
@@ -605,8 +605,11 @@ to_binary_version(v0, _, _, <<131,_/binary>>=Bin) ->
     Bin;
 to_binary_version(v1, _, _, <<?MAGIC:8/integer, 1:8/integer, _/binary>>=Bin) ->
     Bin;
-to_binary_version(Vsn, B, K, Bin) ->
-    to_binary(Vsn, from_binary(B, K, Bin)).
+to_binary_version(Vsn, B, K, Bin) when is_binary(Bin) ->
+    to_binary(Vsn, from_binary(B, K, Bin));
+%% r_object record definition isn't available, so hack the guard.
```
jrwest: not sure i follow, why can't this just match on the r_object defined in this module? im probably missing something...

slfritchie: Heh, cut-and-paste error from another module, sorry, good catch.  Will push a fix in a moment.

src/riak_object.erl
```diff
@@ -605,8 +605,10 @@ to_binary_version(v0, _, _, <<131,_/binary>>=Bin) ->
     Bin;
 to_binary_version(v1, _, _, <<?MAGIC:8/integer, 1:8/integer, _/binary>>=Bin) ->
     Bin;
-to_binary_version(Vsn, B, K, Bin) ->
-    to_binary(Vsn, from_binary(B, K, Bin)).
+to_binary_version(Vsn, B, K, Bin) when is_binary(Bin) ->
+    to_binary(Vsn, from_binary(B, K, Bin));
+to_binary_version(Vsn, _B, _K, O) when is_record(O, r_object) ->
```
jrwest: sorry a little more nitpicking. doesn't most of the riak code use the `#r_object{}` style [1]

[1] https://github.com/basho/riak_kv/blob/master/src/riak_object.erl#L114

jrwest:
@slfritchie to make sure I understand correctly: this change is necessary because backends using `uses_r_object` pass an `r_object` instead of the binary to the fold which would crash handoff when that `r_object` was passed into `to_binary_version`. Do you have a quick way I can test/reproduce?

In hindsight i wonder if `to_binary_version` is even necessary or if I should have just made it part of `to_binary`. Meh, too much of a change this late in the game.

slfritchie:
Use a file system that has the new capability, `uses_r_object`.  (Sorry, the originaly commit tried using backticks for GH markdown, but my terminal shell replaced them for me, oops.)  At the moment, there's only one: fs2.  Otherwise, hack on the yessir backend to support `uses_r_object`.

Evidence in the (lab) wild:

```
2013-06-07 16:03:43.962 [error] <0.25310.7> gen_server <0.25310.7> terminated with reason: no function clause matching riak_object:from_binary(<<48,98,58,228,222,232,92,94,66,158,249,217,95,172,49,136,69,118,125>>, <<23,163,204,214,164,168,79,24,166,171,164,136,221,38,197,117,0,0,0,39>>, {r_object,<<48,98,58,228,222,232,92,94,66,158,249,217,95,172,49,136,69,118,125>>,<<23,163,204,214,...>>,...}) line 618
```

slfritchie:
ping.  :-)

jrwest:
@slfritchie meant to ping it this morning my bad. Went searching for new fs backend to test with but couldn't find a branch. Do you have it available? Otherwise this change is pretty straight forward. I'll run a few riak tests when I get home and give it the official +1

jrwest:
checked `verify_riak_object_format` still passes (note the test itself needs to be updated for 1.4 but can be made to pass easily for review). Also, setup a devrel and made sure normal handoff operations are not affected. Did not have a chance to test w/ the updated fs2 backend since I dont have a copy but I think its sufficient that this issue affects only backends using the new capability and @slfritchie reports the issue is gone w/ this PR. ~~+1.~~ (removed +1 since more changes came)

jrwest:
@slfritchie imo the last commit should be a seperate PR. The change seems to be related to a different bug introduced by different patch.

slfritchie:
Don't you know that the code freeze is happening any second now?

jrwest:
well technically it happened already....lets call these bug fixes found during testing of 1.4?

jrwest:
Also, I'm not sure a better fix would be to update `riak_core:get_primary_apl/3` to take either the service or the up nodes. The chashbin changes should have made it easy to no longer pass in the ring with minimal interface change (e.g. call the 3-arity version w/o the ring). This is why this should be in a seperate PR, so it can be discussed w/o ruining the discussion history of the original changes in this one.

jrwest:
tested again w/ the updated fs2 backend [1]. Confirmed that there were no errors when doing handoff w/ data. However, I was unable to reproduce the `riak_object:from_binary` error in the comment above in this PR by removing the change. I was able to produce a badmatch on `riak_object:to_binary_version` during handoff by removing that change. @slfritchie any details on how I can reproduce the `from_binary` error (I thought handoff would do it)?

NOTE: all the above was tested w/o f17cd6e

My config in `riak_kv` section of app.config for fs2 backend:

```
{add_paths, ["/home/jordan/sandbox/basho/review/slf-backend-robj-cap/riak_cs/ebin"]},
{storage_backend, riak_kv_fs2_backend},
{data_root, "./data/fs2"}, %% fs2 data root
{block_size, 1024}, %% fs2 block size, picked a random power of 2
```

[1] https://github.com/basho/riak_cs/blob/gh555-fs2-stability/src/riak_kv_fs2_backend.erl

slfritchie:
Jordan, I saw handoff do it on a Riak cluster with fs2 underneath and Riak CS the sole app talking to Riak. {shrug}

jrwest:
@slfritchie ok. i'll have to take your word for it since I am still unable to reproduce w/ just riak + fs2. looking at the other change in this PR re: n_val now

buddhisthead:
+1 based on https://github.com/basho/riak_test/pull/304, which used to exhibit the error: insufficient_vnodes before this fix. With it, the test passes.

jrwest:
with @buddhisthead's verification + my testing I'm +1 on this as well.

slfritchie:
https://github.com/basho/riak_core/issues/337 created for lingering review comment.

PR #576, opened 2013-06-12
Add a capability for counters

russelldb:
Please also see (run?) https://github.com/basho/riak_test/pull/305

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -176,9 +176,9 @@ start(_Type, _StartArgs) ->
                                           [encode_zlib, encode_raw],
                                           encode_zlib),
 
-            riak_core_capability:register({riak_kv, counters},
-                                          [true, false],
-                                          false),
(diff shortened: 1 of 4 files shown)


src/riak_kv_app.erl
```diff
@@ -176,6 +176,10 @@ start(_Type, _StartArgs) ->
                                           [encode_zlib, encode_raw],
                                           encode_zlib),
 
+            riak_core_capability:register({riak_kv, counters},
```
jrwest: unfortunately I don't know what the future holds but assuming we will add support for other CRDTs in the future what do you think about making the capability something like `{riak_kv, data_structures}` (needs better name) where the values are a list of supported CRDTs on that node (empty list meaning what `false` does right now)?

jrwest:
+1. tested http by hand, `verify_counter_converge` still passes and basho/riak_test#305 passes for me as well (w/ @russelldb's recent change to riak-erlang-http-client)

Issue #537, opened 2013-04-26
riak_kv_stat:re_register_stat/1 has case_clause exception


slfritchie:
Today's 'master' branch has `case_clause` exceptions being thrown ... example errors from MvM:

```
2013-04-26 12:09:27.979 [error] emulator Error in process <0.29138.49> on node 'riak@127.0.0.1' with exit value: {{case_clause,{add_pid,<0.29109.49>}},[{riak_kv_stat,re_register_stat,1,[{file,"src/riak_kv_stat.erl"},{line,485}]},{riak_kv_stat,stat_repair_loop,0,[{file,"src/riak_kv_stat.erl"},{line,478}]}]}
```

My first guess is that the clauses of `do_update/1` that have `add_monitor(MumbleSomething, Pid)` are the cause: `add_monitor` is not returning `ok`.

russelldb:
Can I get some more? There will be a warning like `ErrorClass:Error updating stat StatName` in the logs. Looks like a merge error, maybe, can't see any clause of `do_update` that handles an argument `{add_pid, pid()}`

slfritchie:
This is all that I have from @matthewvon, he's the generator & owner of the logs.

```
2013-04-26 12:09:27.973 [error] <0.29006.49> gen_server riak_kv_stat terminated with reason: no match of right hand value [] in riak_kv_stat:handle_call/3 line 126
2013-04-26 12:09:27.973 [error] <0.29006.49> CRASH REPORT Process riak_kv_stat with 2 neighbours exited with reason: no match of right hand value [] in riak_kv_stat:handle_call/3 line 126 in gen_server:terminate/6 line 747
2013-04-26 12:09:27.973 [error] <0.28981.49> Supervisor riak_core_stats_sup had child riak_kv_stat started with riak_kv_stat:start_link() at <0.29006.49> exit with reason no match of right hand value [] in riak_kv_stat:handle_call/3 line 126 in context child_terminated
2013-04-26 12:09:27.979 [error] emulator Error in process <0.29138.49> on node 'riak@127.0.0.1' with exit value: {{case_clause,{add_pid,<0.29109.49>}},[{riak_kv_stat,re_register_stat,1,[{file,"src/riak_kv_stat.erl"},{line,485}]},{riak_kv_stat,stat_repair_loop,0,[{file,"src/riak_kv_stat.erl"},{line,478}]}]}
```

russelldb:
Ah! Got it, thank you.

[...]

PR #577, opened 2013-06-13
Fix 1.3 -> master merge error

russelldb:
add_monitor does not return `ok`. `re_register_stat` expects all
stat updates to return `ok`.

Also, the `list` and `index` fsm active stat was not recreated.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -230,13 +230,15 @@ do_update({fsm_error, Type}) when Type =:= gets; Type =:= puts ->
 do_update({index_create, Pid}) ->
     folsom_metrics:notify_existing_metric({?APP, index, fsm, create}, 1, spiral),
     folsom_metrics:notify_existing_metric({?APP, index, fsm, active}, {inc, 1}, counter),
-    add_monitor(index, Pid);
+    add_monitor(index, Pid),
+    ok;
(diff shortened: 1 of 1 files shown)


evanmcc:
I can't seem to force this to happen.  The code looks good, but can't figure out how to test it.  Suggestions?

russelldb:
Something like

in erl console:

```
{spiral, Tid, _Pid} = folsom_metrics_spiral:get_value({riak_kv,list,fsm,create}).
ets:delete(Tid).
folsom_metrics:get_metric_value({riak_kv,list,fsm,create}). %% You will see an error 'cos the stat is gone
```

From cli

```
curl localhost:8098/riak/bob?keys=true
```

then back in the console

```
folsom_metrics:get_metric_value({riak_kv,list,fsm,create}).
```

Ta da, it exists again.

Do all that before and after. (You can use the stats endpoint if you want rather than calling folsom from the console.) Feel free to repeat with `index` replacing `list` and running a 2i query.

In the 'before' test there will be errors like those in #537, after, there should not be any.

evanmcc:
+1.  Still having trouble reproducing, suspect that the error might be masked by other code changes since.  That said, the code change is so minor and clearly correct from the diff that there isn't a lot of reason to spend a huge amount of time on reproduction.

PR #579, opened 2013-06-14
Skip start {val, key} pair if start_inclusive is false

russelldb:
Memory backend was not updated with the bug fix for skipping only
the start {val, key} pair.


russelldb:
ayo @engelsanchez, thanks for reporting this, sorry I missed it before.

engelsanchez:
Sweet: verified that this fixes the verify_2i_limit riak test. Looks good :+1:  /cc @javajolt

PR #581, opened 2013-06-19
Wire up sidejob stats to /stats endpoint

jtuple:
Wire a subset of the stats tracked by sidejob for the get/put FSM resources to the `/stats` endpoint. Also ensure that we don't include stats formerly provided by the get/put monitor when using sidejob, as those stats will always be 0 when sidejob is managing FSMs.


evanmcc:
:+1:, should also be merged to master.  @russelldb might also want to take a look, he had some questions, IIRC.

PR #582, opened 2013-06-20
Refactor handle_common_error/3 to remove case statement

lenary:
I found it quite odd that the first argument of `riak_kv_wm_object:handle_common_error/3` and `riak_kv_wm_counter:handle_common_error/3` had a case statement immediately, so I refactored it to use erlang's function argument pattern matching.

It also becomes very clear that there is: 
a) lots of duplication between these two functions (in different modules sadly)
b) lots of duplication between the clauses
c) slight changes between the error handling (ie some set Content-type to "text/plain", others don't)

I could do further refactoring, but I just wanted to start with a simple and obvious change first.

### DIFF
--- a/src/riak_kv_wm_counter.erl
+++ b/src/riak_kv_wm_counter.erl
@@ -295,72 +295,70 @@ ensure_doc(Ctx=#ctx{doc=undefined, bucket=B, key=K, client=C, r=R,
                 {basic_quorum, Quorum}, {notfound_ok, NotFoundOK}])};
 ensure_doc(Ctx) -> Ctx.
 
-handle_common_error(Reason, RD, Ctx) ->
-    case {error, Reason} of
-        {error, too_many_fails} ->
-            {{halt, 503}, wrq:append_to_response_body("Too Many write failures"
-                    " to satisfy W/DW\n", RD), Ctx};
-        {error, timeout} ->
-            {{halt, 503},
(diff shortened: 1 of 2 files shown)


src/riak_kv_wm_counter.erl
```diff
@@ -295,72 +295,69 @@ ensure_doc(Ctx=#ctx{doc=undefined, bucket=B, key=K, client=C, r=R,
...
-                Ctx}
-    end.
+handle_common_error(allow_mult_false, RD, Ctx) ->
+    Msg = "Counters require bucket property 'allow_mult=true'",
+    {{halt, 409}, wrq:append_to_response_body(Msg, RD), Ctx};
+handle_common_error(too_many_fails, RD, Ctx) ->
+    {{halt, 503}, wrq:append_to_response_body("Too Many write failures"
+                      " to satisfy W/DW\n", RD), Ctx};
+handle_common_error(timeout, RD, Ctx) ->
+    {{halt, 503},
```
evanmcc, line 305: I suspect that this pattern could also be factored out into a function.  e.g error_response(ErrCode, Msg, Ctx) -> ...

russelldb:
I already have a branch that refactors out the common functions (not just the error fun.) It didn't get completed / reviewed in time for code freeze. This is a duplication of that effort. Why?

PR #587, opened 2013-06-25
Fix incorrect arg in call to get_primary_apl/3 by put FSM

jtuple:
Fix incorrect arg in call to `riak_core_apl:get_primary_apl/3` by the put FSM.

Additionally, optimize the get and put FSMs to only compute a list of up nodes when using sloppy quorums. The strict quorum code path uses `get_primary_apl/3` which already computes the list of up nodes internally.

Without this change, puts that attempt to use strict quorums (ie. disable sloppy quorums) always fail:

``` erlang
> {ok, C} = riak:local_client().
> C:put(riak_object:new(<<"abc">>, <<"xyz">>, <<"123">>),
        [{sloppy_quorum, false}]).
{error,all_nodes_down}
```

With this patch, things work as desired.

``` erlang
> {ok, C} = riak:local_client().
> C:put(riak_object:new(<<"abc">>, <<"xyz">>, <<"123">>),
        [{sloppy_quorum, false}]).
ok
```

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -193,13 +193,11 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
             {stop, normal, StateData2};
         _ ->
             StatTracked = proplists:get_value(stat_tracked, BucketProps, false),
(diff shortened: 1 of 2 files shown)


jrwest:
+1. code changes look good. was able to reproduce and verify fix with example above.

PR #588, opened 2013-06-25
Multi backend was missing data_size function

engelsanchez:
It could report size as a capability if all backends had it (bitcask and
level both do) but there was no data_size method! So it would crash on
handoff.

/cc @jrwest

You can verify this with the verify_build_cluster riak_test with and without the multi backend

### DIFF
--- a/src/riak_kv_multi_backend.erl
+++ b/src/riak_kv_multi_backend.erl
@@ -37,6 +37,7 @@
          fold_keys/4,
          fold_objects/4,
          is_empty/1,
+         data_size/1,
          status/1,
          callback/3,
          fix_index/3,
@@ -313,6 +314,13 @@ is_empty(#state{backends=Backends}) ->
                   Module:is_empty(SubState)
           end,
     lists:all(Fun, Backends).
+
+%% @doc Not currently supporting data size
+%% @todo Come up with a way to reflect mixed backend data sizes,
+%% as level reports in bytes, bitcask in # of keys.
+-spec data_size(state()) -> undefined.
+data_size(_) ->
+    undefined.
 
 %% @doc Get the status information for this backend
 -spec status(state()) -> [{atom(), term()}].


engelsanchez:
You will need a recent riak_test after this was merged a few minutes ago: https://github.com/basho/riak_test/pull/259

jrwest:
was trying to catch this before it got opened but apparently I am too slow this morning. This is not the correct solution to this issue. The multi backend DOES NOT support progress tracking and it should have worked properly since it doesn't not advertise the capability. However, `riak_kv_multi_backend`  does [this stupid thing](https://github.com/basho/riak_kv/blob/master/src/riak_kv_multi_backend.erl#L101) which causes it to roll-up capabilities of all sub-backends, which is where we get bit. The correct fix is to not advertise the capability. This is a fine temporary soln if we want to go w/ it but imo this is a problem w/ the multi-backend (blindly advertising capabilities it may not have).

jrwest:
here is the original PR for reference, note the multi-backend was clearly not supported: https://github.com/basho/riak_kv/pull/526

src/riak_kv_multi_backend.erl
```diff
@@ -314,6 +315,13 @@ is_empty(#state{backends=Backends}) ->
           end,
     lists:all(Fun, Backends).
 
+%% @doc Not currently supporting data size
+%% @todo Come up with a way to reflect mixed backend data sizes,
```
jrwest, line 319: this `@todo` should be removed

jrwest, line 319: an additional comment to the affect of "this function is only here because the capability can accidentally be supported". Just want it to be clear that multi backend is not getting progress tracking any time soon (its too inaccurate most likely).

jrwest:
Test crashed w/o this branch and passed w/ it as expected (using the new multi-backend support). As mentioned above there are a few comments i think need changing and although this is not the most ideal fix its probably the one w/ most minimal impact. +1.

PR #589, opened 2013-06-25
Expose Preflist via HTTP Header

lenary:
This adds a `X-Riak-Preflist` header to all non-error HTTP responses, which contains the http interface's ip/port for all the current Bucket/Key's primary nodes that the current node can reach.

In the [Dynamo paper](http://www.allthingsdistributed.com/files/amazon-dynamo-sosp2007.pdf), results are shown about allowing Clients to coordinate writes, with a significant reduction in both regular and tail latencies (§ 6.4 & Table 2).

Though we cannot allow clients to actually coordinate a read [citation needed], there should [citation needed] be a performance increase by hinting to clients which servers to aim their reads and writes at. 

I'm going to work on adding support for this header to the Erlang HTTP Client, so that I can start running benchmarks to see if there is actually any performance increase. The idea is that this information is cached for a short amount of time, to speed up tight GET-PUT cycles.

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -787,7 +787,14 @@ produce_doc_body(RD, Ctx) ->
                                           UserMetaRD, IndexMeta);
                           error -> UserMetaRD
                       end,
-            {riak_kv_wm_utils:encode_value(Doc), encode_vclock_header(IndexRD, Ctx), Ctx};
+
+            % Add Preflist Header to response
+            PreflistRD = preflist_header(IndexRD, Ctx),
+            
+            % Add Vclock Header
+            VclockRD = encode_vclock_header(PreflistRD, Ctx),
+
+            {riak_kv_wm_utils:encode_value(Doc), VclockRD, Ctx};
         multiple_choices ->
             throw({unexpected_code_path, ?MODULE, produce_doc_body, multiple_choices})
     end.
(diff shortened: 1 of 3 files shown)


seancribbs:
Some things:
1. I think it would better for the FSM to reply with the preflist that it used/is using than to recompute it in the WM resource.
2. We have decided to focus support of our official clients on PB, I'd much rather see this added there. 
3. This is not incredibly useful without some mapping of node to client interface, a feature which has not been built yet, i.e. "discovery" or "auto-configuration".

lenary:
@seancribbs Right:
1. Yeah, I was thinking this, but need to work out how to do it. I'm not sure it would be _that_ trivial.
2. Not yet looked at PB code, I will do so. point 1 makes a lot more sense if it's used for both PB and HTTP. I assume benchmarks are done for both.
3. It did feel a bit hacky to just get the info from riak_core. Please can you explain the discovery stuff to me more?

seancribbs:
@lenary One other thing I noticed is that if the interface is bound to the catch-all address (i.e. "0.0.0.0"), this will not return useful information to the client. One would actually have to figure out what addresses "make sense" to the client, which is part of point 3.

lenary:
So 3 is definitely nontrivial to do properly, so that we don't surprise any users.

However, 1 is also looking increasingly complex, due to how other state is passed between the webmachine object, riak_client and the FSM. 

It's probably best at this stage just to shelve it, and later we might be able to look at this stuff again.

seancribbs:
@lenary Sounds like a good plan. Maybe you could familiarize yourself with the FSMs, then revisit?

PR #591, opened 2013-06-28
Added functions to riak_kv_console to support managing index repairs through riak-admin

cdahlqvist:


cdahlqvist:
Also added support for handling search index repairs.

slfritchie:
Pulling PRs out of the freezer ... isn't there some ongoing work with 2I repair, @engelsanchez and @jtuple?

jrwest:
This PR is for the "old" style repair of 2i indexes via primary partitions not the 2i aae stuff engel is working on. Last we discussed this there was still some debate about whether or not we wanted to expose this version of repair from `riak-admin`. Now that we have 2i aae I wonder even more if this is not necessary... @engelsanchez @jtuple ?

EDIT: i meant not necessary....

jtuple:
In my opinion, we need to decide on this issue for 2.0. Or close it otherwise. Punting to 2.1 doesn't make a lot of sense, as that's not really punting the work but instead punting the decision to do this or not. Let's just make a decision. Assigning to 2.0-beta so we hopefully take notice and make a call. FYI, not prepared to make a decision just yet, but will think about it this week. Anyone else with opinions should chime in.

rzezeski:
@jrwest Actually repair was written for Riak Search and not 2i. There is also KV repair but it's not really for 2i (although obviously it will repair 2i since it's just metadata). This particular form of repair is useful in the case where the disk array on a node is destroyed/lost (or entire node is lost). You can use adjacent partitions from other nodes to repair partition on recovered one. HOWEVER, this only works if the partitions on the other nodes are fully intact. If you repair a partition and its adjacent partitions are missing data then the repaired partition will be missing data. I never added this repair command to riak-admin because I wasn't sure this very important nuance is clear to users. E.g. if a user has multiple nodes with merge_index corruption then repair will not help (in fact it might even just cause a bunch of additional errors in the log).

Given that legacy riak_search has been deprecated the only repair type to survive will be riak_kv. It can still be useful in certain situations but I don't know if it should be exposed to users. Maybe it's better as an internal tool for other repair mechanisms (e.g. AAE could use it if it noticed lots of missing data) or for the CSEs when they know it applies.

jrwest:
thanks for the clarification @rzezeski. fwiw, I had thought I had seen it used in cases for 2i repair most recently. Anyways, I agree we should close as wontfix and leave it as an internal/advanced (it is documented) tool.

jtuple:
Looks like we reached consensus on keeping repair feature somewhat hidden (ie. we have docs, but aren't going to put it in `riak-admin`). Closing issue.

PR #594, opened 2013-07-04
Additions to Counters for Yokozuna

lenary:
This is associated with the work in basho/yokozuna#135 - and hence will have to wait until yokozuna supports 1.4.

Firstly, the Yokozuna extractor extracts on a sibling-by-sibling basis, not the entire riak object. Hence it's useful to have a method in `riak_kv_counter` that can give the counter value from a single sibling.

Secondly, we require that Yokozuna merges all CRDTs before doing extraction, so that the most up to date value of the counter is merged. `riak_kv_counter:merge/1` is "safe" to call on a riak_object without counter siblings (ie it won't affect anything), but is quite a heavyweight thing to do. We include a simple check to see if any siblings could be a counter, before merging. This probably needs (micro-)benchmarking to see if it really does increase performance.

### DIFF
--- a/src/riak_kv_counter.erl
+++ b/src/riak_kv_counter.erl
@@ -65,8 +65,21 @@ update(RObj, Actor, Amt) ->
 -spec merge(riak_object:riak_object()) ->
                    riak_object:riak_object().
 merge(RObj) ->
-    {Meta, Counter, NonCounterSiblings} = merge_object(RObj),
-    update_object(RObj, Meta, Counter, NonCounterSiblings).
+    case might_have_counter(RObj) of
+        true -> {Meta, Counter, NonCounterSiblings} = merge_object(RObj),
+                update_object(RObj, Meta, Counter, NonCounterSiblings);
+        _    -> RObj
+    end.
+    
+%% @doc returns true if any of the siblings of the riak_object
+%% could be a counter (using the content type or the binary tag and version)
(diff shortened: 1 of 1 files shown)


src/riak_kv_counter.erl
```diff
@@ -80,6 +97,17 @@ value(RObj) ->
             riak_kv_pncounter:value(Counter)
     end.
 
+%% @doc Turn a single sibling into a counter value. Used by the YZ Extractor.
+-spec sibling_value(binary()) -> {ok, integer()} | {error, term()}.
+sibling_value(Sibling) ->
+    try
+        PNCounter = riak_kv_counter:from_binary(Sibling),
+        Value = riak_pn_counter:value(PNCounter),
```
coderoshi: It looks like this is called `riak_kv_pncounter` rather than `riak_pn_counter`

lenary: Yes, Yes it is. I must have got confused between the names here and the names in riak_dt. Just a second while I update it in this branch.

slfritchie:
Pulling PRs out of the freezer ... Sam, any chance that you could bribe a reviewer?  Was @coderoshi doing a full review or just a drive-by comment?

russelldb:
@seancribbs is working on this in a different approach. Can wait for @lenary but I vote to close this.

seancribbs:
@russelldb @slfritchie Yes, closing. See basho/yokozuna#250.

PR #595, opened 2013-07-05
Sibling index: On merge conflict, add index if asked to do so.

eriksoe:
This is for providing an easy, cheap and reliable way to query for conflict-burdened keys.

The mechanism is: on merge conflicts (i.e. whenever the result of a merge has >1 contents), if user-metadata "X-Riak-Meta-X-Index-Siblings" has a value of "true" (in any of the contents), then add to the resulting object an index item "x_riak_siblings_int: <number of siblings>" (and remove any old such indexing).

The two special names are rather arbitrary; please change them if you find better ones.

(This patch was mentioned on riak-users: subject "Indexing conflicts (feature)")

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -226,12 +226,45 @@ compare_content_dates(C1,C2) ->
 -spec merge(riak_object(), riak_object()) -> riak_object().
 merge(OldObject, NewObject) ->
     NewObj1 = apply_updates(NewObject),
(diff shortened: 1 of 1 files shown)


slfritchie:
Pulling PRs out of the freezer ... we definitely missed the last Riak major release with this PR.  @engelsanchez or @jtuple any opinions now?

yorkshire-steve:

PR #596, opened 2013-07-08
Generic AAE status

rzezeski:
This pull request makes AAE entropy info parameterized on type.  This allows both KV and Yokozuna to share almost all the same status code without copy pasta.
- Add a `Type` to entropy info.  It represents the type of AAE being performed.  Currently `riak_kv` or `yokozuna` but could be other things in the future.
- Export the `get_build_time` function as it can be reused by Yokozuna.
- Parameterize the module and function name used to compute all exchanges.  KV has inter-partition exchange whereas Yokozuna is purely intra.  I.e. Yokozuna exchange is always between the same partition, different services.  By parameterizing the "all exchanges" calculation can let different services have different exchange patterns.

### DIFF
--- a/src/riak_kv_entropy_info.erl
+++ b/src/riak_kv_entropy_info.erl
@@ -24,7 +24,9 @@
          create_table/0,
          dump/0,
          compute_exchange_info/1,
-         compute_tree_info/1]).
+         compute_exchange_info/2,
+         compute_tree_info/1,
+         all_exchanges/2]).
 
 -define(ETS, ets_riak_kv_entropy).
 
@@ -89,11 +91,14 @@ dump() ->
                                    [{index(), Last :: t_now(), All :: t_now(),
                                      repair_stats()}].
 compute_exchange_info(Type) ->
+    compute_exchange_info(Type, {?MODULE, all_exchanges}).
+
+compute_exchange_info(Type, {M,F}) ->
     filter_index_info(),
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     Indices = riak_core_ring:my_indices(Ring),
     Defaults = [{Index, undefined, undefined, undefined} || Index <- Indices],
-    KnownInfo = [compute_exchange_info(Ring, Index, Info)
+    KnownInfo = [compute_exchange_info({M,F}, Ring, Index, Info)
                  || {{Type2, Index}, Info} <- all_index_info(), Type2 == Type],
     merge_to_first(KnownInfo, Defaults).
 
@@ -166,9 +171,9 @@ all_exchanges(Ring, Index) ->
     L2 = [{RemoteIdx, IndexN} || {_, RemoteIdx, IndexN} <- L1],
     {Index, L2}.
 
-compute_exchange_info(Ring, Index, #index_info{exchanges=Exchanges,
-                                               repaired=Repaired}) ->
-    {_, AllExchanges} = all_exchanges(Ring, Index),
+compute_exchange_info({M,F}, Ring, Index, #index_info{exchanges=Exchanges,
+                                                      repaired=Repaired}) ->
+    {_, AllExchanges} = M:F(Ring, Index),
     Defaults = [{Exchange, undefined} || Exchange <- AllExchanges],
     KnownTime = [{Exchange, EI#exchange_info.time} || {Exchange, EI} <- Exchanges],
     AllTime = merge_to_first(KnownTime, Defaults),
--- a/src/riak_kv_entropy_info.erl
+++ b/src/riak_kv_entropy_info.erl
@@ -19,12 +19,12 @@
 %% -------------------------------------------------------------------
 -module(riak_kv_entropy_info).
 
--export([tree_built/2,
-         exchange_complete/4,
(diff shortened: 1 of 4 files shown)


src/riak_kv_entropy_info.erl
```diff
@@ -135,7 +144,7 @@ filter_index_info() ->
     Indices = ets:select(?ETS, [{{{index, '$1'}, '$2'}, [], ['$1']}]),
```
jrwest, line 144: with the changes to the key to support types I think this code is now a bit buggy. The `$1` matches the entire key so `Indicies` is a list of `{type(), index()}` instead of just `index()`. When used below, the `Others` set is now wrong. Adding some quick `io:formats` (results below) you can see how we always get `Indicies =:= Others`. Its easy to change the match but I wonder if it wouldn't be useful to keep the type information around because the `match_delete` afterwards is not deleting anything since `Idx` is `{type(), index()}`. Leaving it as a `delete`, in addition, avoids the full table scan. 

Also, and this is a more a comment for the original code but since we are changing it, we shouldn't assume ordsets are a list. There should be some `to_list` and/or `from_list` calls in here imo.

```
INDICIES: [{riak_kv,159851741583067506678528028578343455274867621888},
           {riak_kv,137015778499772148581595453067151533092743675904}]
PRIMARIES: [0,22835963083295358096932575511191922182123945984,
            45671926166590716193865151022383844364247891968,
            68507889249886074290797726533575766546371837952,
            ...
OTHERS: [{riak_kv,137015778499772148581595453067151533092743675904},
         {riak_kv,159851741583067506678528028578343455274867621888}]
```

src/riak_kv_exchange_fsm.erl
```diff
@@ -284,6 +284,7 @@ next_state_with_timeout(StateName, State, Timeout) ->
     {next_state, StateName, State, Timeout}.
 
 exchange_complete({LocalIdx, _}, {RemoteIdx, RemoteNode}, IndexN, Repaired) ->
-    riak_kv_entropy_info:exchange_complete(LocalIdx, RemoteIdx, IndexN, Repaired),
+    riak_kv_entropy_info:exchange_complete(riak_kv, LocalIdx, RemoteIdx,
+                                           IndexN, Repaired),
     rpc:call(RemoteNode, riak_kv_entropy_info, exchange_complete,
-             [RemoteIdx, LocalIdx, IndexN, Repaired]).
+             [riak_kv, RemoteIdx, LocalIdx, IndexN, Repaired]).
```
jrwest, line 290: Instead of changing all the call sites it would probably be safer to make the original arity functions pass in `riak_kv` by default and then export the new arity functions, in addition. Wouldn't this call crash the exchange if its with an older node?

src/riak_kv_entropy_info.erl
```diff
@@ -84,24 +87,30 @@ dump() ->
...
 %% exchange; time since the index completed exchanges with all sibling indices;
 %% as well as statistics about repairs triggered by different exchanges.
--spec compute_exchange_info()  -> [{index(), Last :: t_now(), All :: t_now(),
-                                    repair_stats()}].
-compute_exchange_info() ->
+-spec compute_exchange_info(atom())  ->
+                                   [{index(), Last :: t_now(), All :: t_now(),
+                                     repair_stats()}].
+compute_exchange_info(Type) ->
+    compute_exchange_info(Type, {?MODULE, all_exchanges}).
```
jrwest, line 94: does it make sense to use `riak_kv_entropy_info:all_exchanges/2` with types other than `riak_kv`? Should this remain zero arity and pass the type directly (either way I think we should have the zero arity version and not modify the call site as mentioned elsewhere).

jrwest:
rebased branch against develop to start review (it rebases cleanly). eunit tests pass on both develop and this branch for me. Found a few issues looking over the code. Planning to write an eunit for issue described [here](https://github.com/basho/riak_kv/pull/596/files#r6200996) but wanted to get comments out for the sake of time.

EDIT: i wasn't planning to test w/ yokozuna, just verify kv, but if there are some instructions I'd be glad to take them for a spin :)

rzezeski:
I created a new PR based on develop under #654.  I'm going to keep this open until I address the comments here.

src/riak_kv_entropy_info.erl
rzezeski, line 144: Nice catch.

I fixed this here: https://github.com/basho/riak_kv/pull/654#commits-pushed-74efbd7

I decided to simply fix the select and leave everything else the same.  I'm not too worried about run time given what this code is for.  But if you have a strong objection let me know and we can create a separate issue to make things faster.  I just want to get this in so Yokozuna can work.

src/riak_kv_exchange_fsm.erl
rzezeski, line 290: https://github.com/basho/riak_kv/pull/654#commits-pushed-bfe7f40

src/riak_kv_entropy_info.erl
rzezeski, line 94: https://github.com/basho/riak_kv/pull/654#commits-pushed-bfe7f40

PR #600, opened 2013-07-10
Avoid t2b/b2t between vnode and get FSM

engelsanchez:
This is a port of Krestenkrab's PR :
https://github.com/basho/riak_kv/pull/83
Avoids doing a binary to term on an object that is sent from a vnode to
the get FSM requesting it. This means that the FSM will have to do it on
that side, but it avoids the implicit term_to_binary/binary_to_term when
the object is sent through disterl.  Some tests using a memory backend
on OSX show a decent improvement (~5% in throughput and more in
latency).
The get_fsm fake vnode had to be updated to support the get binary
request.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -114,71 +114,7 @@ start(_Type, _StartArgs) ->
     %% Spin up supervisor
     case riak_kv_sup:start_link() of
         {ok, Pid} ->
-            %% Register capabilities
-            riak_core_capability:register({riak_kv, vnode_vclocks},
-                                          [true, false],
-                                          false,
-                                          {riak_kv,
-                                           vnode_vclocks,
-                                           [{true, true}, {false, false}]}),
-
-            riak_core_capability:register({riak_kv, legacy_keylisting},
-                                          [false],
-                                          false,
-                                          {riak_kv,
-                                           legacy_keylisting,
-                                           [{false, false}]}),
-
-            riak_core_capability:register({riak_kv, listkeys_backpressure},
-                                          [true, false],
-                                          false,
-                                          {riak_kv,
-                                           listkeys_backpressure,
-                                           [{true, true}, {false, false}]}),
-
-            riak_core_capability:register({riak_kv, index_backpressure},
-                                          [true, false],
-                                          false),
-
-            %% mapred_system should remain until no nodes still exist
-            %% that would propose 'legacy' as the default choice
-            riak_core_capability:register({riak_kv, mapred_system},
-                                          [pipe],
-                                          pipe,
-                                          {riak_kv,
-                                           mapred_system,
-                                           [{pipe, pipe}]}),
-
-            riak_core_capability:register({riak_kv, mapred_2i_pipe},
-                                          [true, false],
-                                          false,
-                                          {riak_kv,
-                                           mapred_2i_pipe,
-                                           [{true, true}, {false, false}]}),
-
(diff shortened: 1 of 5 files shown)


engelsanchez:
@russelldb reminds me to look into what this means for mixed clusters that may have nodes temporarily using different versions of the binary format. I may need to add a capability to control this protocol between the FSM and the vnode.

src/riak_kv_get_fsm.erl
```diff
@@ -292,6 +297,17 @@ preflist_for_tracing(Preflist) ->
      end || {Idx, Nd} <- lists:sublist(Preflist, 4)].
 
 %% @private
+
+waiting_vnode_r({r_bin, {ok, BinObj}, Idx, _ReqId},
+                StateData = #state{bkey={B, K}}) ->
+    case riak_object:from_binary(B, K, BinObj) of
+        {error, R} ->
```
evanmcc, line 304: Could this lager message be brought in line with the other serialization error stuff (that's been added since, admittedly)?

Also, I don't think that we can just log and drop it like this, I think that we need to fake up a `notfound` and recall `waiting_vnode_r` with that, otherwise I think that the FSM is going to crash every time we see a bad message.

src/riak_kv_get_fsm.erl
```diff
@@ -314,6 +330,16 @@ waiting_vnode_r(request_timeout, StateData) ->
     finalize(S2).
 
 %% @private
+waiting_read_repair({r_bin, {ok, BinObj}, Idx, _ReqId},
+                StateData = #state{bkey={B, K}}) ->
+    case riak_object:from_binary(B, K, BinObj) of
+        {error, R} ->
+            lager:error("Invalid object ~p/~p : ~p", [B, K, R]);
```
evanmcc, line 337: same as the above.

src/riak_kv_app.erl
```diff
@@ -201,6 +137,79 @@ start(_Type, _StartArgs) ->
             {error, Reason}
     end.
 
+
```
evanmcc, line 140: when/if you're updating stuff, this has mergerotted a little.

evanmcc:
Overall this is a good patch, but a thought:

Maybe it'd be better, if we're adding a new request, to add get_req_v2, which is a get request record with an options list.  Then we could add new options when/if we need them.  I feel like this would duplicate fewer code paths, and have more long-run flexibility.

evanmcc:
I guess also this needs to be merged into develop instead of master.

src/riak_kv_app.erl
engelsanchez, line 140: A lot has changed since this. And the PR is against master. I'll have to redo against develop. I think we've since added some incompatibilities to this approach. Darn robject/binary dance!

evanmcc, line 140: I merged to it latest develop, it was only a few minutes of work.  Other than the deserialzation thing in the FSM, everything merges cleanly and works fine.

evanmcc:
Moving some discussion over from chat: 

The mutator code path has broken this optimization in a way that's hard to fix without totally redoing the object format, so I am closing this for now.

slfritchie:
Closing per Evan's comment.

evanmcc:
Was considering re-opening this now that mutators are gone, but since it's pointed at master, I'll open a new one.

evanmcc:
totally failed to port this in time.  moving to 2.1 so we won't forget, hopefully.

yorkshire-steve:

PR #601, opened 2013-07-11
Move json encoding out of riak_object

engelsanchez:
Revisiting https://github.com/basho/riak_kv/pull/88
But using riak_object_json encode/decode and leaving old methods in
riak_object.erl with a deprecation warning.

### DIFF
--- a/src/riak_object_json.erl
+++ b/src/riak_object_json.erl
@@ -29,7 +29,6 @@
 
 -export([encode/1,decode/1]).
 
-%% @spec to_json(riak_object()) -> {struct, list(any())}
 %% @doc Converts a riak_object into its JSON equivalent
 -spec encode(riak_object:riak_object()) -> {struct, list(any())}.
 encode(Obj) ->
@@ -66,8 +65,8 @@ jsonify_metadata(MD) ->
               %% When the user metadata is empty, it should still be a struct
               ({?MD_USERMETA, []}) ->
                    {?MD_USERMETA, {struct, []}};
-              ({<<"Links">>, Links}) ->
-                   {<<"Links">>, [ [B, K, T] || {{B, K}, T} <- Links ]};
+              ({?MD_LINKS, Links}) ->
+                   {?MD_LINKS, [ [B, K, T] || {{B, K}, T} <- Links ]};
               ({Name, List=[_|_]}) ->
                    {Name, jsonify_metadata_list(List)};
               ({Name, Value}) ->
@@ -81,7 +80,7 @@ jsonify_metadata_list(List) ->
     Classifier = fun({Key,_}, Type) when (is_binary(Key) orelse is_list(Key)),
                                          Type /= array, Type /= string ->
                          struct;
-                    (C, Type) when is_integer(C), C >= 0, C =< 256,
+                    (C, Type) when is_integer(C), C >= 0, C < 256,
                                    Type /= array, Type /= struct ->
                          string;
                     (_, _) ->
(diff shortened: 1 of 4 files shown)


rzezeski:
- [x] verify compile
- [x] verify unit test/EQC
- [x] verify copy pasta
- [x] verify specs
- [x] while I'm here, check for any dumb code ( that has survived, not that you introduced Engel ;) )
- [x] verify riak test
- [ ] merge
- [x] close #88

src/riak_object_json.erl
```diff
@@ -0,0 +1,183 @@
...
+
+-module(riak_object_json).
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+-include("riak_kv_wm_raw.hrl").
+
+-export([encode/1,decode/1]).
+
+%% @spec to_json(riak_object()) -> {struct, list(any())}
```
rzezeski: Kill this doc spec.

src/riak_object_json.erl
```diff
@@ -0,0 +1,183 @@
...
+                         struct;
+                    (C, Type) when is_integer(C), C >= 0, C =< 256,
+                                   Type /= array, Type /= struct ->
+                         string;
+                    (_, _) ->
+                         array
+                 end,
+    case lists:foldl(Classifier, undefined, List) of
+        struct -> {struct, jsonify_proplist(List)};
+        string -> list_to_binary(List);
```
rzezeski, line 91: I need to look at this more closely but right off the bat wonder if we have potential utf-8 issues here?

Granted, it is the same as current behavior.  But making mental note to check.

src/riak_object_json.erl
```diff
@@ -0,0 +1,183 @@
...
+        array -> List
+    end.
+
+%% @doc converts a proplist with potentially multiple values for the
+%%    same key into a JSON object with single or multi-valued keys.
+jsonify_proplist([]) -> [];
+jsonify_proplist(List) ->
+    dict:to_list(lists:foldl(fun({Key, Value}, Dict) ->
+                                     JSONKey = if is_list(Key) -> list_to_binary(Key);
```
rzezeski, line 100: Once again, wonder about potential utf-8 issues.

src/riak_object_json.erl
```diff
@@ -0,0 +1,183 @@
...
+                                     end
+                             end, dict:new(), List)).
+
+dejsonify_values([], Accum) ->
+    lists:reverse(Accum);
+dejsonify_values([{<<"metadata">>, {struct, MD0}},
+                  {<<"data">>, D}|T], Accum) ->
+    Converter = fun({Key, Val}) ->
+                        case Key of
+                            <<"Links">> ->
```
rzezeski: `<<"Links">>` and `<<"X-Riak-Last-Modified">>` could be replaced with their macros in `riak_kv_wm_raw.hrl`.

src/riak_object_json.erl
engelsanchez, line 91: I don't see how we could have issues in this layer: UTF-8 or ASCII strings should look identical to this code (integer values between 0-255) and any errors would be in other layers interpreting this data. The funky thing I noticed now is that we are allowing 256!!! This code will probably never see one of those, but I think the correct thing is < 256 or =< 255

engelsanchez:
@rzezeski killed the spec, used the header constants and tweaked the string detection code. I don't _think_ I see potential UTF-8 issues, but let me know if you think there's something there.

seancribbs:
Couldn't those type guard BIFs be moved out of guards and replaced with unicode:c2l or c2b? Then you could at least verify that they are valid UTF-8.

Sean Cribbs

engelsanchez:
@rzezeski @seancribbs I'm not opposed to improve UTF-8 problem detection here, but could we open another issue for that? It's only taken 2 years to merge this simple refactor :)

seancribbs:
@engelsanchez Yeah, you're right. I think longer-term the better way would be to constrain the names/types of metadata to avoid the problem, but that's too much for this PR.

rzezeski:
+1 to merge

PR #602, opened 2013-07-11
R16B01 compatibility changes.

evanmcc:
- pull in sbv's param mods drop with some additional porting make it compatible with 1.4 features.
- export some opaque types
- change some crypto hash functions to remove warnings


Vagabond:
This looks fine, aside from the mess around pmods, but there's not much to be done about that.

+1

PR #392, opened 2012-09-10
remove -author attributes from source

argv0:

PR #603, opened 2013-07-12
Remove -author attributes

rzezeski:
They aren't kept current and the same information can be ascertained
from git.


rzezeski:
This supplants #392.

evanmcc:
+1

PR #606, opened 2013-07-22
Make the memory backend config behave as documented

Vagabond:
Also, fix yessir backend, which has the same potential issue.

See docs at:

http://docs.basho.com/riak/latest/ops/advanced/backends/memory/

cc @evanmcc


evanmcc:
Prior to merge only the multi-backend docs are correct.

Post merge, both of the docs are correct and buckets with ttl set expire as expected.

:+1:

PR #607, opened 2013-07-22
Don't return the timestamp information when folding objects

Vagabond:
The memory backend, with TTL enabled, stores the values as {Timestamp,
Value}, and it was leaking that information during fold_objects, which
would cause handoff to fail.

For a riak_test, see basho/riak_test#338


evanmcc:
:+1:

PR #608, opened 2013-07-25
fix riak_kv_backend standard tests

vinoski:
The riak_kv_backend standard tests, which are invoked by riak backend
tests, were not actually testing anything. The standard_test/2 function was
coded as an eunit test generator but it was returning functions that
themselves were returning lists of test generators. Unfortunately eunit
won't work through that many levels of test generator indirection so the
tests weren't actually running.

Change the standard tests to use regular assertMatch rather than the test
generator form _assertMatch so that the tests actually run. Due to
bitcask's use of the process dictionary, add the local test directive to
the top test spec to execute the tests in the same process where the
backend initialization function runs. Fix one of the fold_keys tests to
expect the correct result. Change all the delete/3 calls to delete/4 calls,
as that's what backends now implement. Modify the memory backend test setup
to ensure its ets tables are created appropriately for testing.

### DIFF
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_kv_backend: Riak backend behaviour
 %%
-%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
 %%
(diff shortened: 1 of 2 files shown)


vinoski:
Note that the Travis CI build failed because of existing `riak_kv_pb_object` unit test failures, not because of any changes in this PR.

rzezeski:
+1 to merge

PR #609, opened 2013-07-26
Passing to mochiweb:encode/1 a valid UTF-8 string (not latin1).

arcusfelis:
The old code sometimes calls `mochijson2:encode/1` with a wrong argument.
For example, here is a minimized version `mochijson2:encode({struct,[{input,<<"òB">>}]}).`.
This code fails with the reason `{ucs,{bad_utf8_character_code}}`, because `<<"òB">>` is not a valid UTF-8 string.
This PR fixes this error.

### DIFF
--- a/src/riak_kv_mapred_json.erl
+++ b/src/riak_kv_mapred_json.erl
@@ -491,7 +491,10 @@ pipe_phase_index(I)          -> I.
 -spec trunc_print(term()) -> binary().
 trunc_print(Term) ->
     {Msg, _Len} = lager_trunc_io:print(Term, 500),
-    iolist_to_binary(Msg).
+    %% `lager_trunc_io:print/2' returns an IO-list (in latin1).
+    %% `mochijson2:encode/1' expects, that binaries are encoded in UTF-8.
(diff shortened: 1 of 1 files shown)


slfritchie:
Pulling PRs out of the freezer ... @seancribbs or @javajolt opinions?

src/riak_kv_mapred_json.erl
```diff
@@ -491,7 +491,10 @@ pipe_phase_index(I)          -> I.
 -spec trunc_print(term()) -> binary().
 trunc_print(Term) ->
     {Msg, _Len} = lager_trunc_io:print(Term, 500),
-    iolist_to_binary(Msg).
+    %% `lager_trunc_io:print/2' returns an IO-list (in latin1).
+    %% `mochijson2:encode/1' expects, that binaries are encoded in UTF-8.
+    %% Do not use `iolist_to_binary/1' here.
+    unicode:characters_to_binary(Msg).
```
seancribbs, line 497: Shouldn't you qualify this with the `latin1` encoding (outputting `utf8`)?

yorkshire-steve:

PR #613, opened 2013-07-27
Fix counter protobuf message codes

titanous:
According to [this line](https://github.com/basho/riak_kv/blob/319e96717d5f4ee8afe6c47f65355f79035311f9/src/riak_kv_app.erl#L35) and my testing, these codes are the correct ones.

Is there a reason why they are not in the docs and `riak_pb` README?

### DIFF
--- a/src/riak_kv_pb_counter.erl
+++ b/src/riak_kv_pb_counter.erl
@@ -24,15 +24,15 @@
 %% following request messages:</p>
 %%
 %% <pre>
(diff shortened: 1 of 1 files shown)


seancribbs:
The initial version of the feature was written before other features, but merged later. The message codes clashed, and so had to be changed in the code, but this bit seems to have slipped through the cracks. Thanks!

PR #615, opened 2013-07-29
Fix gh611 - do not set the start_term to the last seen key for eq

### LINKED ISSUE #611: Pagination in 2i is broken for equals queries
When the continuation is decoded the last seen key is copied into the start_term, rather than preserving the index term in the query.

russelldb:
The clause for equals query pagination was missing.

Fixes #611
Test at basho/riak_test#340


engelsanchez:
Code looks good to me. I've tested the issue manually and using the accompanying riak test (plus a couple of additions). Ship it :+1: :dancer: :boat:

PR #616, opened 2013-07-29
Add millisecond timemout parameter to API 2i endpoints

russelldb:
Accepts non_neg_integer milliseconds (zero for inifinty)
Add support for an app.config setting for 2i timeouts

Depends on https://github.com/basho/riak_pb/pull/50

### DIFF
--- a/src/riak_index.erl
+++ b/src/riak_index.erl
@@ -40,7 +40,8 @@
          return_body/1,
          upgrade_query/1,
          object_key_in_range/3,
-         index_key_in_range/3
+         index_key_in_range/3,
+         add_timeout_opt/2
         ]).
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -51,6 +52,10 @@
 -define(TIMEOUT, 30000).
 -define(BUCKETFIELD, <<"$bucket">>).
 -define(KEYFIELD, <<"$key">>).
+
+%% See GH610, this default is for backwards compat, so 2i behaves as
+%% it did before the FSM timeout bug was "fixed"
+-define(DEFAULT_TIMEOUT, infinity).
 
 %% @type data_type_defs()  :: [data_type_def()].
 %% @type data_type_def()   :: {MatchFunction::function(), ParseFunction::function()}.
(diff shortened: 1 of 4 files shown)


src/riak_index.erl
```diff
@@ -419,6 +424,22 @@ decode_continuation(undefined) ->
...
+%% @doc add the `timeout' option tuple to the
+%% `Opts' proplist. if `Timeout' is undefined
+%% then the `app.config' property
+%% `{riak_kv, seconady_index timeout}' is used
+%% If that config property is defined, then
+%% a default of `infinity'  is used.
+%% We use `infinity' as the default to
+%% match the behavior pre 1.4
+add_timeout_opt(undefined, Opts) ->
+    Timeout = app_helper:get_env(riak_kv, secondary_index_timeout, ?DEFAULT_TIMEOUT),
```
rzezeski, line 436: IIRC app env uses ETS.  Should we worry about read contention on ETS table in hot-path here?

src/riak_index.erl
```diff
@@ -52,6 +53,10 @@
 -define(BUCKETFIELD, <<"$bucket">>).
 -define(KEYFIELD, <<"$key">>).
 
+%% See GH610, this default is for backwards compat, so 2i behaves as
+%% it did before the FSM timeout bug was "fixed"
+-define(DEFAULT_TIMEOUT, infinity).
```
rzezeski, line 58: I think default should be 60s like rest of system to be consistent.  But @jonmeredith and @russelldb seem to agree on infinity here.  My argument against is a) consistency and b) not tying up resource for an indefinite amount of time without the user explicitly saying to do so.  Hell, I would argue the 60s timeout is too large.

src/riak_index.erl
russelldb, line 436: Hope that people aren't doing that many 2i queries. Do you have a suggestion for an alternative?

src/riak_index.erl
russelldb, line 58: So the reasoning is this: there was never a user specified timeout for 2i, and there was a bug in the coverage FSM that meant the timeout (of 60 seconds) was reset every time a results were received from a vnode. When the fsm was fixed, some customers ran into the issue that their 2i queries were failing and they were unable to set the timeout (hence this fix.) The infinity default here means that 2i behaves as much like the broken 2i of old as possible.

I'm not wedded to it, but I'd hate to cause our users trouble over something like having consistent defaults.
We can change it to 60 seconds at 2.0, maybe?

russelldb, line 58: On second thoughts, I can get with your thinking here, especially if we keep the app env check. Users bitten by the featurebug can then update their app.config (rather than their client code) to get the infinite timeout.

src/riak_index.erl
rzezeski, line 436: After discussing in back-channel with @evanmcc and @russelldb decided to just leave it as is and wait to prove it is a problem before optimizing.

src/riak_index.erl
rzezeski, line 58: Exactly, you already said yourself in back-channel that due to other changes the reset 60s for every reply has already been changed.  That means that it no longer behaves as before.  Although that brings up the more thorny issue that to get a truly infinite timeout multiple timeout values need to be set in tandem.  Le sigh.

evanmcc, line 58: Have a gist that outlines a plan for reworking all the timeouts we can take
a look at during the meet up if you're caremad about the thousands of
parallel timeouts.

rzezeski, line 58: My last comment was based on incorrect assumptions.  It turns out that the 2i timeout is fed into the coverage FSM.  I.e. the 2i timeout is used to set the coverage timeout and that is what determines the time a 2i query has to finish.

That said, I would like to heard more next week @evanmcc.

rzezeski:
- [x] review code
- [x] verify compile/test/eunit
- [x] verify riak_test
- [x] merge to 1.4
- [ ] merge to master

rzezeski:
+1 to merge

russelldb:
I'm not 100% on the `infinity` default. Any final thoughts @jonmeredith @rzezeski ?

PR #618, opened 2013-07-29
Strip index term from result before passing to MR

russelldb:
MR expects only a Key, 2i ranges now return {index_term, key}

### DIFF
--- a/src/riak_kv_pipe_index.erl
+++ b/src/riak_kv_pipe_index.erl
@@ -104,12 +104,19 @@ keysend(_Bucket, [], _Partition, _FittingDetails) ->
     ok;
 keysend(Bucket, [Key | Keys], Partition, FittingDetails) ->
     case riak_pipe_vnode_worker:send_output(
-           {Bucket, Key}, Partition, FittingDetails) of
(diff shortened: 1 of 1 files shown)


src/riak_kv_pipe_index.erl
```diff
@@ -104,13 +104,20 @@ keysend(_Bucket, [], _Partition, _FittingDetails) ->
     ok;
 keysend(Bucket, [Key | Keys], Partition, FittingDetails) ->
     case riak_pipe_vnode_worker:send_output(
-           {Bucket, Key}, Partition, FittingDetails) of
+           {Bucket, strip_index(Key)}, Partition, FittingDetails) of
```
beerriot, line 107: It would be much cooler to transform `{Bucket, {IndexTerm, Key}}` into `{{Bucket, Key}, IndexTerm}`, because then the map phase function would get the `IndexTerm` as its `KeyData` argument. Depending on representation, that might break Javascript phases, though.

russelldb:
Add these tests, run without this patch you'll see failures https://github.com/basho/mapred_verify/pull/4

beerriot:
This fix makes sense to me, and makes the new mapred_verify tests pass. +1 to merge

PR #622, opened 2013-07-31
Fix gh611 - do not set the start_term to the last seen key for eq

russelldb:
The clause for equals query pagination was missing.


russelldb:
Actually no, will be merged from 1.4 to master after release.

PR #627, opened 2013-08-07
Add ack_execute to the put FSM's execute state for forwarded coordinator

slfritchie:
When a put operation has been forwarded to a remote coordinator node,
When the remote put FSM enters the execute state, it sends an ack_execute
message back to the put FSM on the source node.  In the event that the
source node does not receive the ack_execute message, it will choose a
new coordinator node.  The hardcoded timeout is 3 seconds.

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -34,6 +34,8 @@
 -define(DEFAULT_OPTS, [{returnbody, false}, {update_last_modified, true}]).
 -export([start/3, start/6,start/7]).
 -export([start_link/3,start_link/6,start_link/7]).
+-export([set_put_coordinator_failure_timeout/1,
+         get_put_coordinator_failure_timeout/0]).
 -ifdef(TEST).
 -export([test_link/4]).
 -endif.
@@ -105,7 +107,8 @@
                                           non_neg_integer()}}],
                 reply, % reply sent to client,
                 tracked_bucket=false :: boolean(), %% track per bucket stats
-                bad_coordinators = [] :: [atom()]
+                bad_coordinators = [] :: [atom()],
+                coordinator_timeout :: integer()
                }).
 
 -include("riak_kv_dtrace.hrl").
@@ -153,6 +156,18 @@ start_link(From, Object, PutOptions) ->
             end
     end.
 
+set_put_coordinator_failure_timeout(MS) when is_integer(MS), MS >= 0 ->
+    %% mochiglobal:put(put_coordinator_failure_timeout, MS);
+    application:set_env(riak_kv, put_coordinator_failure_timeout, MS);
+set_put_coordinator_failure_timeout(Bad) ->
+    lager:error("~s:set_put_coordinator_failure_timeout(~p) invalid",
+                [?MODULE, Bad]),
+    set_put_coordinator_failure_timeout(3000).
+
+get_put_coordinator_failure_timeout() ->
+    %% mochiglobal:get(put_coordinator_failure_timeout).
+    app_helper:get_env(riak_kv, put_coordinator_failure_timeout, 3000).
+
 make_ack_options(Options) ->
     case riak_core_capability:get({riak_kv, put_fsm_ack_execute}, disabled) of
         disabled ->
@@ -181,7 +196,7 @@ monitor_remote_coordinator(true = _UseAckP, MiddleMan, CoordNode, StateData) ->
     receive
         {ack, CoordNode, now_executing} ->
             {stop, normal, StateData}
-    after 3000 ->
+    after StateData#state.coordinator_timeout ->
             exit(MiddleMan, kill),
             Bad = StateData#state.bad_coordinators,
             prepare(timeout, StateData#state{bad_coordinators=[CoordNode|Bad]})
@@ -213,10 +228,12 @@ test_link(From, Object, PutOptions, StateProps) ->
 %% @private
 init([From, RObj, Options, Monitor]) ->
     BKey = {Bucket, Key} = {riak_object:bucket(RObj), riak_object:key(RObj)},
+    CoordTimeout = get_put_coordinator_failure_timeout(),
     StateData = add_timing(prepare, #state{from = From,
                                            robj = RObj,
                                            bkey = BKey,
-                                           options = Options}),
+                                           options = Options,
+                                           coordinator_timeout=CoordTimeout}),
     (Monitor =:= true) andalso riak_kv_get_put_monitor:put_fsm_spawned(self()),
     riak_core_dtrace:put_tag(io_lib:format("~p,~p", [Bucket, Key])),
     case riak_kv_util:is_x_deleted(RObj) of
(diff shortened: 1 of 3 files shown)


jrwest:
Corner case that I came across looking at the code and verified with a simple modification:

If the `ack_execute` message is dropped for some reason but the sending coordinator is allowed to keep working (does not crash) then the original put_fsm (the one waiting for `ack_execute`) can spawn another coordinator resulting in coordination for the same write on two nodes concurrently. This ends up w/ siblings.

I changed https://github.com/basho/riak_kv/pull/627/files#L0L390 to not send `ack_execute` 50% of the time and made a request to a key w/ no data stored (w/ the bucket set to `allow_mult: true`:

```
curl -v localhost:10018/riak/bucket/key   
```

Fetching this key after this particular run:

```
$ curl -v localhost:10018/riak/bucket/key
Siblings:
2cmcScReQKqjk2E7wA7QpG
5KaD1HCugysqsRSVl3P97T
7GcC3Jg7KvP5ZoNFO0BFH
7budz05j1phiU4W347mtTv
```

Some logs (added by me) that illustrate the issue:

```
==> dev/dev4/log/console.log <==
2013-09-17 19:58:26.727 [info] <0.2521.0>@riak_kv_put_fsm:execute:428 dropping ack_execute but proceeding
2013-09-17 19:58:27.034 [info] <0.2521.0>@riak_kv_put_fsm:client_reply:917 sending client reply
2013-09-17 19:58:27.034 [info] <0.2521.0>@riak_kv_put_fsm:finish:572 stopping in finish

==> dev/dev1/log/console.log <==
2013-09-17 19:58:29.715 [info] <0.3449.0>@riak_kv_put_fsm:monitor_remote_coordinator:166 Killing middle man on 'dev4@127.0.0.1': <0.3450.0>

==> dev/dev3/log/console.log <==
2013-09-17 19:58:29.737 [info] <0.2556.0>@riak_kv_put_fsm:execute:428 dropping ack_execute but proceeding
2013-09-17 19:58:29.738 [info] <0.2556.0>@riak_kv_put_fsm:client_reply:917 sending client reply
2013-09-17 19:58:29.738 [info] <0.2556.0>@riak_kv_put_fsm:finish:572 stopping in finish

==> dev/dev1/log/console.log <==
2013-09-17 19:58:32.716 [info] <0.3449.0>@riak_kv_put_fsm:monitor_remote_coordinator:166 Killing middle man on 'dev5@127.0.0.1': <0.3468.0>
2013-09-17 19:58:32.993 [info] <0.3449.0>@riak_kv_put_fsm:client_reply:917 sending client reply
2013-09-17 19:58:32.993 [info] <0.3449.0>@riak_kv_put_fsm:finish:572 stopping in finish

==> dev/dev5/log/console.log <==
2013-09-17 19:58:32.727 [info] <0.2520.0>@riak_kv_put_fsm:monitor_remote_coordinator:166 Killing middle man on 'dev3@127.0.0.1': <0.2521.0>
2013-09-17 19:58:32.735 [info] <0.2520.0>@riak_kv_put_fsm:execute:428 dropping ack_execute but proceeding
2013-09-17 19:58:32.992 [info] <0.2520.0>@riak_kv_put_fsm:client_reply:917 sending client reply
2013-09-17 19:58:32.992 [info] <0.2520.0>@riak_kv_put_fsm:finish:572 stopping in finish
```

slfritchie:
Naive question: what's worse: vclock siblings (but with identical values), or not having the write happen at all _plus_ a client-side timeout?

The Erlang answer to the dropped message problem is a monitor.  I'll look to see how easy it might be to add one.

jrwest:
We discussed this a bit more off-github. Since there doesn't seem to be an immediately known sol'n to the problem but timeouts are also bad, my vote is to keep this patch but adjust it to make it triggered by a request option (i would vote the current behavior be the default but major vsn numbers let you change things too). Also, for CRDTs since I think the affect would be more visible I think we shouldn't expose the option for the data types API or again make it not the default.

I have to make another pass through the code but I'll wait for any changes.

EDIT: @russelldb may or may not have some opinions re: CRDTs

slfritchie:
If we're going to make this an option that the client can pass in with the request, then it's my vote that the default is auto-retry.  If the client doesn't like that, then it can turn it off ... but absorbing the writes (even with the 3 sec/whatever delay) is better than waiting the default 60 second eternity.

slfritchie:
@jrwest @russelldb I've added a couple of commits, whaddya think?
1. Damn rolling upgrade.
2. Add `riak_kv_put_fsm` option, `{retry_put_coordinator_failure, boolean()}`, to control retry. Default = true.

src/riak_kv_put_fsm.erl
```diff
@@ -227,13 +268,16 @@ prepare(timeout, StateData0 = #state{from = From, robj = RObj,
...
+            Preflist1 = case proplists:get_value(sloppy_quorum, Options, true) of
                             true ->
                                 UpNodes = riak_core_node_watcher:nodes(riak_kv),
-                                riak_core_apl:get_apl_ann(DocIdx, N, UpNodes);
+                                riak_core_apl:get_apl_ann(
+                                  DocIdx, N, UpNodes -- BadCoordinators);
                             false ->
                                 riak_core_apl:get_primary_apl(DocIdx, N, riak_kv)
                         end,
+            Preflist2 = [X || X = {{_Index, Node}, _Type} <- Preflist1,
```
jrwest: is this necessary? doesn't `riak_core_apl` return a preflist that only contains owners from `UpNodes -- BadCoordinators` in this case?

src/riak_kv_put_fsm.erl
```diff
@@ -252,17 +296,22 @@ prepare(timeout, StateData0 = #state{from = From, robj = RObj,
...
-                        {stop, normal, StateData0}
+                        monitor_remote_coordinator(UseAckP, MiddleMan,
+                                                   CoordNode, StateData0)
                     catch
                         _:Reason ->
                             ?DTRACE(?C_PUT_FSM_PREPARE, [-2],
                                     ["prepare", dtrace_errstr(Reason)]),
-                            lager:error("Unable to forward put for ~p to ~p - ~p\n",
-                                        [BKey, CoordNode, Reason]),
+                            lager:error("Unable to forward put for ~p to ~p - ~p @ ~p\n",
```
jrwest, line 329: was this just for debugging? Adding the stacktrace seems helpful for us in the future but makes the log slightly messier. Just curious, I'm fine leaving it, if it was intentional.

jrwest:
re: config, the only thing missing is a way for the user to specify. I think it makes sense to be able to specify on a per request basis but doing it in app.config or bucket properties may be easier, since per-request means messing w/ PB and HTTP. Also, I think the timeout on ack_execute (currently at 3s) should be configurable in some way.

src/riak_kv_put_fsm.erl
slfritchie: It's necessary for filtering from sloppy_quorum=false. I suppose I should move it to that clause.

src/riak_kv_put_fsm.erl
slfritchie, line 329: It's intentional: messy-but-helpful vs. post-mortem-WTF.

src/riak_kv_put_fsm.erl
jrwest: ah yea that makes sense. moving it would be nice.

jrwest:
+1 to merge. 

imo https://github.com/basho/riak_kv/commit/6b6bef4adb6d49e7cc9d0d94c4b984013f3df695 could do w/o the extra commented out lines and the put_fsm record could have done w/o a change but ultimately minor.

slfritchie:
Merged into develop with commit 9517585608e53f46fa, closing.

PR #632, opened 2013-08-12
Store monitor state in gen_server state

russelldb:
Don't call into the monitor process to get Type


russelldb:
@evanmcc @reiddraper have at it.

reiddraper:
+1

engelsanchez:
After checking with Reid, I'm merging this request. This should fix another current customer issue, plus other fixes I'm preparing and I'd like to bundle them.

PR #633, opened 2013-08-12
Transform deserialization errors and CRC check failures into not_founds

evanmcc:
Backends should be lossy.  Since we typically have other replicas on the system, if we can't read something off the disk, or it seems to have been corrupted, we should drop it and count on read repair, AAE, or a subsequent write to put things right.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1083,7 +1083,10 @@ perform_put({true, Obj},
     {Reply, State#state{modstate=UpdModState}}.
 
 do_reformat({Bucket, Key}=BKey, State=#state{mod=Mod, modstate=ModState}) ->
-    case do_get_object(Bucket, Key, Mod, ModState) of
+    case (catch do_get_object(Bucket, Key, Mod, ModState)) of
+        {'EXIT', {badarg, _}} -> 
+            Reply = {error, not_found},
+            UpdState = State;
         {error, not_found, _UpdModState} ->
             Reply = {error, not_found},
             UpdState = State;
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1075,7 +1075,7 @@ perform_put({true, Obj},
     {Reply, State#state{modstate=UpdModState}}.
 
 do_reformat({Bucket, Key}=BKey, State=#state{mod=Mod, modstate=ModState}) ->
-    case catch do_get_object(Bucket, Key, Mod, ModState) of
+    case do_get_object(Bucket, Key, Mod, ModState) of
         {error, not_found, _UpdModState} ->
             Reply = {error, not_found},
             UpdState = State;
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -962,8 +962,7 @@ prepare_put(State=#state{vnodeid=VId,
         false ->
             prepare_put(State, PutArgs, IndexBackend)
     end.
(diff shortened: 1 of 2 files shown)


engelsanchez:
This change allows us to clean up this other mess I reluctantly did a while ago: https://github.com/basho/riak_kv/blob/master/src/riak_kv_vnode.erl#L982

Love it.

src/riak_kv_vnode.erl
```diff
@@ -1083,7 +1083,10 @@ perform_put({true, Obj},
     {Reply, State#state{modstate=UpdModState}}.
 
 do_reformat({Bucket, Key}=BKey, State=#state{mod=Mod, modstate=ModState}) ->
-    case do_get_object(Bucket, Key, Mod, ModState) of
+    case (catch do_get_object(Bucket, Key, Mod, ModState)) of
+        {'EXIT', {badarg, _}} ->
```
engelsanchez: Could you explain why we match against badarg here, and not just anything?

reiddraper:
Do we want some way to get at the corrupted data? Say the other replicas _aren't_ available for some reason, or they also have bad CRCs, it might still be helpful to be able to inspect the corrupted data.

src/riak_kv_vnode.erl
evanmcc: I suppose I was trying to be erlang-y and non-defensive (i.e. defending against the smallest possible error space), but it works either way from my perspective.  I don't know if other errors there might possibly be interesting.

evanmcc:
@reiddraper if you really need to see the data, you can go directly to the backend, assuming some heavy intervention.  If we change it so that we want to get it automatically, we can change this code as well at that time.

engelsanchez:
On Reid's question: I added a lager message when the bad_crc used to percolate all the way to vnode. I think it would be worth to report an error or warning in the logs when we hit this instead of the "ok" not found path. What do you think?

evanmcc:
Logging a warning sounds good to me, although the volume of them is
potentially very high.

engelsanchez:
I think it actually would not be in the common case: not found is received from a node, other replicas respond and a read repair would mask the old value.  If it is common even with that, perhaps we should know :)

reiddraper:
+1 to warning. But should we print the bucket/key in the warning?

shino:
Tiny nitpick: `bac_crc branch` at `prepare_put` is not necessary anymore?
https://github.com/basho/riak_kv/blob/70271e3e27b007787359569703d887b15607b68e/src/riak_kv_vnode.erl#L982

engelsanchez:
Yes @shino, that's what I meant in my comment above: that can be cleaned up now. Only bitcask returned bad_crc, and this is a more proper fix that makes backend responses uniform.  It'd be nice to remove with this PR.

shino:
@engelsanchez Oops, I just saw file diffs and have missed your comment.

One customer ran into this issue (not in production). Hope this will be merged soon :)

evanmcc:
@shino Working on this for another customer issue, plan right now is master, backport to 1.3 and 1.4

evanmcc:
Updated after a hilariously large # of git mistakes.  Should be good to go.  Note that I mistargeted this, it should be against 1.4, but the diff should be more or less identical. This should also go into 1.3, but the diff will be slightly different.

evanmcc:
Closing this in favor of the 1.4 version, all comments should be addressed on that branch: https://github.com/basho/riak_kv/pull/638

PR #635, opened 2013-08-13
Handle 2i timeout

russelldb:
Fix 60sec hardcoded timeout on http stream.
Fix mochiweb crash error on http stream timeout firing

Tests are here https://github.com/basho/riak_test/pull/350

### DIFF
--- a/src/riak_kv_wm_index.erl
+++ b/src/riak_kv_wm_index.erl
@@ -246,17 +246,20 @@ index_stream_helper(ReqID, Boundary, ReturnTerms, MaxResults, Timeout, LastResul
                     {iolist_to_binary(Body),
                      index_stream_helper(ReqID, Boundary, ReturnTerms, MaxResults, Timeout, LastResult1, Count1)};
                 {ReqID, Error} ->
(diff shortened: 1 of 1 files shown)


beerriot:
> Fix 60sec hardcoded timeout on http stream.

It looks like there were two fixes made, and I just want to make sure they were both intended:
- If a `timeout` query parameter is included, that value is used in the stream helper.
- If no `timeout` query parameter is included, the value of `DEFAULT_TIMEOUT` in `riak_index.erl` is used. This means the default timeout is now `infinity` instead of `60000`.

russelldb:
Yes, both intended. Not entirely happy with the `infinity` timeout, but it is consistent with the non-stream default timeout (which itself is somewhat contentious.)

beerriot:
Code makes sense. The change to `verify_2i_timeout` causes it to fail without this patch, and to pass with it. The `verify_2i_stream`, `verify_2i_returnterms`, `verify_2i_limit`, and `secondary_index_tests` tests all continue to pass with this patch in place.

The infinite default timeout has been discussed elsewhere (see #616), and is easily overridden in app.config.

+1. Merge away.

PR #636, opened 2013-08-15
Eas fix orphan stat repair

engelsanchez:
A customer had a good number of riak_kv_stat restarts due to problems in
folsom as of yet undetermined. The number of processes was observed to
balloon in size. The stat repair process was created every time a
riak_kv_stat process was restarted. riak_kv_stat monitored this process,
but this process did not monitor riak_kv_stat, so when that went away,
this process stayed around forever blocked waiting on its msg box.
This change ensures the process monitors riak_kv_stat so it goes down
with it.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -120,9 +120,10 @@ stop() ->
 
 init([]) ->
     register_stats(),
+    Me = self(),
     State = #state{monitors = [{index, spawn_link(?MODULE, monitor_loop, [index])},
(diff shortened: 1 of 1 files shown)


evanmcc:
This looks good.  Without the patch I can get processes to grown in an unbounded way using the code below, with it, I cannot.

```
[begin timer:sleep(100), exit(whereis(riak_kv_stat), kill) end || _ <- lists:seq(1,200)].
```

:+1:

PR #637, opened 2013-08-15
Transform deserialization errors and CRC check failures into not_founds for 1.3

evanmcc:
1.3 version of #633

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -143,6 +143,8 @@ get(Bucket, Key, #state{ref=Ref}=State) ->
             {ok, Value, State};
         not_found  ->
             {error, not_found, State};
+        {error, bad_crc}  ->
+            {error, not_found, State};
         {error, nofile}  ->
             {error, not_found, State};
         {error, Reason} ->
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -144,6 +144,7 @@ get(Bucket, Key, #state{ref=Ref}=State) ->
         not_found  ->
             {error, not_found, State};
         {error, bad_crc}  ->
+            lager:warning("Unreadable object discarded"),
             {error, not_found, State};
         {error, nofile}  ->
             {error, not_found, State};
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -144,7 +144,8 @@ get(Bucket, Key, #state{ref=Ref}=State) ->
         not_found  ->
             {error, not_found, State};
         {error, bad_crc}  ->
-            lager:warning("Unreadable object discarded"),
+            lager:warning("Unreadable object ~p/~p discarded",
+                          [Bucket, Key]),
             {error, not_found, State};
         {error, nofile}  ->
             {error, not_found, State};
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -749,9 +749,9 @@ handle_handoff_data(BinObj, State) ->
                 {reply, {error, Err}, State}
         end
     catch %% never let a bad object break a handoff
-        _ ->  
+        _:_ ->  
             lager:warning("Unreadable object discarded"),
-            {reply, {error, bad_object}, State}
+            {reply, ok, State}
     end.
 
 encode_handoff_item({B, K}, V) ->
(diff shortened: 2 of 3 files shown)


src/riak_kv_vnode.erl
```diff
@@ -737,16 +737,21 @@ handoff_finished(_TargetNode, State) ->
...
+                {reply, ok, State#state{modstate=UpdModState}};
+            {error, Reason, UpdModState} ->
+                {reply, {error, Reason}, State#state{modstate=UpdModState}};
+            Err ->
+                {reply, {error, Err}, State}
+        end
+    catch %% never let a bad object break a handoff
+        _ -> 
+            {reply, {error, bad_object}, State}
+    end
```
engelsanchez: Syntax error!

src/riak_kv_vnode.erl
```diff
@@ -923,7 +938,6 @@ prepare_put(#state{vnodeid=VId,
                          end,
             {{true, ObjToStore}, PutArgs#putargs{index_specs=IndexSpecs, is_index=IndexBackend}};
         {ok, Val, _UpdModState} ->
-            OldObj = binary_to_term(Val),
             case put_merge(Coord, LWW, OldObj, RObj, VId, StartTime) of
```
engelsanchez: OldObj referenced here, but removed above. Breaks compilation :(

src/riak_kv_bitcask_backend.erl
```diff
@@ -143,6 +143,8 @@ get(Bucket, Key, #state{ref=Ref}=State) ->
             {ok, Value, State};
         not_found  ->
             {error, not_found, State};
+        {error, bad_crc}  ->
```
engelsanchez, line 146: Consider logging the bucket/key here. It's great for testing. Otherwise the fact that there is corruption on disk is completely hidden. As a user I'd love to know that stuff.

src/riak_kv_vnode.erl
```diff
@@ -737,15 +737,21 @@ handoff_finished(_TargetNode, State) ->
...
+            {error, Reason, UpdModState} ->
+                {reply, {error, Reason}, State#state{modstate=UpdModState}};
+            Err ->
+                {reply, {error, Err}, State}
+        end
+    catch %% never let a bad object break a handoff
+        _:_ ->  
+            lager:warning("Unreadable object discarded"),
+            {reply, ok, State}
     end.
```
engelsanchez, line 756: Hmmm, Evan, you have force pushed over my change yesterday to fix this syntax error :(.

engelsanchez, line 756: Wait, nevermind. Github was showing me stale data somehow

src/riak_kv_vnode.erl
```diff
@@ -907,7 +913,18 @@ prepare_put(#state{vnodeid=VId,
...
                              prunetime=PruneTime},
             IndexBackend) ->
-    case Mod:get(Bucket, Key, ModState) of
+    Result = 
+        case Mod:get(Bucket, Key, ModState) of
+            {ok, OldObjBin, UpdModState} ->
+                try
+                    {ok, binary_to_term(OldObjBin), UpdModState}
+                catch _:_ ->
+                        lager:warning("Unreadable object discarded"),
```
engelsanchez: This and the other warnings are not very useful without the bucket/key. We should add them. Also, although we are recovering from it, I would consider this an error and something to trigger some kind of alarm, not a warning.  Not that I feel too strongly about.

engelsanchez:
This change looks good to me. I have tested by manually corrupting data and executing get/puts operations and performing handoffs. It seems to all work as advertised. :+1: :dancer: :boat:

src/riak_kv_vnode.erl
```diff
@@ -737,15 +737,21 @@ handoff_finished(_TargetNode, State) ->
...
+        BKey = {PBObj#riakobject_pb.bucket,PBObj#riakobject_pb.key},
+        case do_diffobj_put(BKey, binary_to_term(PBObj#riakobject_pb.val), State) of
+            {ok, UpdModState} ->
+                {reply, ok, State#state{modstate=UpdModState}};
+            {error, Reason, UpdModState} ->
+                {reply, {error, Reason}, State#state{modstate=UpdModState}};
+            Err ->
+                {reply, {error, Err}, State}
+        end
+    catch %% never let a bad object break a handoff
```
engelsanchez, line 751: I +1'd this already, but while looking at this code in the 1.4 I had a thought about logging here. This is just a reminder in case we choose to change this. https://github.com/basho/riak_kv/pull/638/files

PR #640, opened 2013-08-21
protect against corrupted indexes

evanmcc:
When sext is asked to decode something that it doesn't understand, it throws an exception, which we currently don't catch.  This can nuke a get or a put or a fold.  This patch tries to keep things working when possible.


engelsanchez:
Code looks good to me: all paths seemed covered. I manually tested and verified that 2i queries, list bucket, list keys as well as the index reformat process can all survive massive amounts of sext decoding failures.

:+1: :dancer: :boat:

PR #641, opened 2013-08-22
leveldb fold hardening for 1.4

evanmcc:
moved to 1.4 and adapted for changes there.  Requires the same tests.  Adding filtering to the index seemed like the minimal change to the code, but I can see the argument for handling it in the backend for separation-of-concerns reasons.  If it bothers you, please let me know.


engelsanchez:
Code looks good to me: all paths related to decoding seem to be covered, including new riak_index functions introduced with 2i pagination. Manually tested that 2i operations, bucket listing, key listing and the 1.3.1 index reformat can all finish in the presence of decoding errors (which become warnings in the log)
:+1: :dancer: :boat:

PR #642, opened 2013-08-27
Add bucket types

Vagabond:
see also, basho/riak#362

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -33,9 +33,9 @@
 -export([stream_list_keys/2,stream_list_keys/3,stream_list_keys/4]).
 -export([filter_buckets/2]).
 -export([filter_keys/3,filter_keys/4]).
--export([list_buckets/1,list_buckets/2,list_buckets/3]).
+-export([list_buckets/1,list_buckets/2,list_buckets/3, list_buckets/4]).
 -export([stream_list_buckets/1,stream_list_buckets/2,
-         stream_list_buckets/3,stream_list_buckets/4]).
+         stream_list_buckets/3,stream_list_buckets/4, stream_list_buckets/5]).
 -export([get_index/4,get_index/3]).
 -export([stream_get_index/4,stream_get_index/3]).
 -export([set_bucket/3,get_bucket/2,reset_bucket/2]).
@@ -375,8 +375,11 @@ stream_list_keys(Bucket, Timeout, {?MODULE, [_Node, _ClientId]}=THIS) ->
 %%      keys in Bucket on any single vnode.
 stream_list_keys(Input, Timeout, Client, {?MODULE, [Node, _ClientId]}) when is_pid(Client) ->
     ReqId = mk_reqid(),
+    lager:info("input is ~p", [Input]),
     case Input of
-        {Bucket, FilterInput} ->
+        %% buckets with bucket types are also a 2-tuple, so be careful not to
+        %% treat the bucket type like a filter
+        {Bucket, FilterInput} when not is_binary(FilterInput) ->
             case riak_kv_mapred_filters:build_filter(FilterInput) of
                 {error, _Error} ->
                     {error, _Error};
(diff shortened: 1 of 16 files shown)


src/riak_client.erl
```diff
@@ -375,8 +375,11 @@ stream_list_keys(Bucket, Timeout, {?MODULE, [_Node, _ClientId]}=THIS) ->
 %%      keys in Bucket on any single vnode.
 stream_list_keys(Input, Timeout, Client, {?MODULE, [Node, _ClientId]}) when is_pid(Client) ->
     ReqId = mk_reqid(),
+    lager:info("input is ~p", [Input]),
```
engelsanchez: Don't forget to remove this!

engelsanchez:
I don't think I've seen any typos, but consider using a macro instead of peppering <<"default">> around just in case.

src/riak_kv_pb_bucket.erl
```diff
@@ -91,9 +91,10 @@ process(rpblistbucketsreq, State) ->
     process(#rpblistbucketsreq{}, State);
 
 %% Start streaming in list keys
-process(#rpblistkeysreq{bucket=B,timeout=T}=Req, #state{client=C} = State) ->
+process(#rpblistkeysreq{type = Type, bucket=B,timeout=T}=Req, #state{client=C} = State) ->
     %% stream_list_keys results will be processed by process_stream/3
-    {ok, ReqId} = C:stream_list_keys(B, T),
+    Bucket = maybe_create_bucket_type(Type, B),
```
engelsanchez, line 96: Not a real issue, but I got confused for a second when I read this the first time thinking this could try to create a bucket type. Consider a name change maybe.

src/riak_object.erl
```diff
@@ -263,6 +273,17 @@ apply_updates(Object=#r_object{}) ->
...
+%% @doc Return the containing bucket for this riak_object without any type
+%% information, if present.
+-spec bucket_only(riak_object()) -> binary().
+bucket_only(#r_object{bucket={_Type, Bucket}}) -> Bucket;
+bucket_only(#r_object{bucket=Bucket}) -> Bucket.
+
+%% @doc Return the containing bucket-type for this riak_object.
+-spec type(riak_object()) -> binary() | undefined.
+type(#r_object{bucket={Type, _Bucket}}) -> Type;
+type(#r_object{bucket=_Bucket}) -> undefined.
```
engelsanchez, line 285: How about using bucket_type instead of type here? It could get confusing. It sounds like object type.

src/riak_object_json.erl
```diff
@@ -33,7 +33,8 @@
 -spec encode(riak_object:riak_object()) -> {struct, list(any())}.
 encode(Obj) ->
     {_,Vclock} = riak_object:vclock_header(Obj),
-    {struct, [{<<"bucket">>, riak_object:bucket(Obj)},
+    {struct, [{<<"type">>, riak_object:type(Obj)},
```
engelsanchez, line 36: Same here. I think it would be less confusing to use bucket_type (or btype or w/e) instead of plain type at the object level.

src/riak_kv_pipe_index.erl
```diff
@@ -105,7 +105,7 @@ keysend(_Bucket, [], _Partition, _FittingDetails) ->
     ok;
 keysend(Bucket, [Key | Keys], Partition, FittingDetails) ->
     case riak_pipe_vnode_worker:send_output(
-           {Bucket, strip_index(Key)}, Partition, FittingDetails) of
+           {{Bucket, strip_index(Key)}, undefined}, Partition, FittingDetails) of
```
engelsanchez, line 108: This makes the mapred_basic_compat riak test fail. Do we need to change this?

engelsanchez:
@Vagabond please see my note above about the MR riak test failing because it's expecting {Bucket, Key} and it's now getting {{Bucket, Key}, undefined}. I'm not following the need for that change. Could you explain?

PR #643, opened 2013-08-27
Add option for smaller bitcask keys

evanmcc:
The kv bucket/key to bitcask key conversion uses term_to_binary.  This results in 13 bytes of overhead per key.

This patch introduces a new key format:

```
|--|---bucket---|---key--|
 ^---- bucket_len
```

The first two bytes specify the length of the bucket, allowing kv to translate the compact binary back to the bucket/key pair that it needs.  In order to read old-format values off of disk, this code allows the insertion of a translation function, so that keys in the keydir are in a consistent format.  This saves 11 bytes per entry in the keydir.

When a large percentage of your keys on disk are in the old format, there is a small but noticable slowdown in startup time and fold time.  This should go down over time as keys are replaced.  If the slowdown is intolerable, adding `{small_keys, false}` to the bitcask section of your app.config will revert to the old format.  This feature is currently on by default; whether this should be the case is up for debate.

This PR depends on a change in bitcask adding support for key transformation functions: 
https://github.com/basho/bitcask/pull/103

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -56,11 +56,18 @@
 -define(API_VERSION, 1).
 -define(CAPABILITIES, [async_fold,size]).
 
+%% must not be 131, otherwise will match t2b in error
+%% yes, I know that this is horrible.
+-define(VERSION_1, 1).
+-define(VERSION_BYTE, ?VERSION_1).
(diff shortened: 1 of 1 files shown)


Vagabond:
This mostly looks ok, although it needs my refactor to support bucket types over in #663.

evanmcc:
Merged, your PR, is this ready to merge once the bitcask comments have been addressed?

Vagabond:
Yes.

PR #646, opened 2013-09-03
Sort AAE differences before acting upon them via read-repair

slfritchie:
```
        %% Sort the keys.  For vnodes that use backends that preserve
        %% lexicographic sort order for BKeys, this is a big
        %% improvement.  For backends that do not, e.g. Bitcask, sorting
        %% by BKey is unlikely to be any worse.  For Riak CS's use
        %% pattern, sorting may have some benefit since block N is
        %% likely to be adjacent on disk to block N+1.
```

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -117,6 +117,7 @@
          local_compare/2,
          compare/2,
          compare/3,
+         compare/4,
          top_hash/1,
          get_bucket/3,
          key_hashes/2,
@@ -401,7 +402,10 @@ compare(Tree, Remote) ->
 
 -spec compare(hashtree(), remote_fun(), acc_fun(X)) -> X.
 compare(Tree, Remote, AccFun) ->
-    compare(1, 0, Tree, Remote, AccFun, []).
+    compare(Tree, Remote, AccFun, []).
+
+compare(Tree, Remote, AccFun, Acc) ->
+    compare(1, 0, Tree, Remote, AccFun, Acc).
 
 -spec levels(hashtree()) -> pos_integer().
 levels(#state{levels=L}) ->
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -90,14 +90,14 @@
 %% to LevelDB using a single batch write. Writes are flushed whenever the
 %% buffer becomes full, as well as before updating the hashtree.
 %%
-%% Tree exchange is provided by the ``compare/2'' and ``compare/3'' functions.
-%% The behavior of these functions is determined through a provided function
+%% Tree exchange is provided by the ``compare/4'' function.
+%% The behavior of this function is determined through a provided function
 %% that implements logic to get buckets and segments for a given remote tree,
 %% as well as a callback invoked as key differences are determined. This
 %% generic interface allows for tree exchange to be implemented in a variety
 %% of ways, including directly against to local hash tree instances, over
 %% distributed Erlang, or over a custom protocol over a TCP socket. See
-%% ``local_compare/2'' and ``do_remote/1'' for examples.
+%% ``local_compare/2'' and ``do_remote/1'' for examples (-ifdef(TEST) only).
 
 -module(hashtree).
 
@@ -114,9 +114,6 @@
          destroy/1,
          read_meta/2,
          write_meta/3,
-         local_compare/2,
-         compare/2,
-         compare/3,
          compare/4,
          top_hash/1,
          get_bucket/3,
@@ -126,6 +123,7 @@
          width/1,
          mem_levels/1]).
 
+-ifdef(TEST).
 -export([run_local/0,
          run_local/1,
          run_concurrent_build/0,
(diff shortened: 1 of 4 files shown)


src/hashtree.erl
```diff
@@ -117,6 +117,7 @@
          local_compare/2,
          compare/2,
          compare/3,
+         compare/4,
```
engelsanchez, line 117: I would be in favor of reducing the number of convenience variations of this function, but that can be done later. Notice that in the documentation at the beginning of the file compare/2 and compare/3 are explicitly named.

src/riak_kv_index_hashtree.erl
```diff
@@ -600,12 +608,8 @@ do_compare(Id, Remote, AccFun, From, State) ->
         {ok, Tree} ->
             spawn_link(fun() ->
                                Remote(init, self()),
-                               Result = case AccFun of
-                                            undefined ->
```
engelsanchez, line 604: Notice how compare/2 passes undefined to compare/3 as the AccFun. That would make it use hashtree:compare/2, which defaults to an accumulator function that concatenates lists. Also notice how compare/2 is not used anywhere in the Riak tree. If something were to happen to compare/2, I would not tell anybody.  If someone was even more caremad and trimmed all these different "convenient" versions of compare through these modules, I wouldn't say anything either.

src/riak_kv_exchange_fsm.erl
```diff
@@ -186,27 +206,53 @@ key_exchange(timeout, State=#state{local=LocalVN,
...
-                                         case Acc2 of
-                                             [] ->
-                                                 [1];
-                                             [Count] ->
-                                                 [Count+1]
-                                         end
+                     lists:foldl(fun({DiffReason, BKeyBin}, Count) ->
+                                         {B, K} = binary_to_term(BKeyBin),
+                                         T = {B, K, DiffReason},
+                                         if Count rem 5000 == 0 ->
```
engelsanchez: It would be nice to have this batching parameter as a macro a the top of the file

src/riak_kv_exchange_fsm.erl
```diff
@@ -171,11 +171,31 @@ key_exchange(timeout, State=#state{local=LocalVN,
...
-    Remote = fun(get_bucket, {L, B}) ->
+    TmpDir = app_helper:get_env(riak_core, platform_data_dir, "/tmp"),
+    {NA, NB, NC} = Now = WriteLog = now(),
+    LogFile1 = lists:flatten(io_lib:format("~s/~s/in.~p.~p.~p",
+                                           [TmpDir, ?MODULE, NA, NB, NC])),
+    ok = filelib:ensure_dir(LogFile1),
+    LogFile2 = lists:flatten(io_lib:format("~s/~s/out.~p.~p.~p",
+                                           [TmpDir, ?MODULE, NA, NB, NC])),
+    ok = filelib:ensure_dir(LogFile2),
+    Remote = fun(get_bucket, {L, B}=_X) ->
```
engelsanchez: _X here doesn't really do anything.

src/hashtree.erl
jrwest, line 117: ooh, i just realized. This file is not in `riak_kv` any more. It has been in core since https://github.com/basho/riak_kv/pull/652

jrwest, line 117: another note, more pertinent to the original comment. Adding `compare/4`  would be useful for simplifying some of the code in `hashtree_tree.erl`.

engelsanchez, line 117: BTW I like compare/4. I would keep it and eye the others for cleanup. Again, not pressing now.

engelsanchez:
I'm done going over the code now. The comments I had were very minor, with the only little issue being the inconsistency of riak_kv_index_hashtree:compare/2 and the handling of the undefined AccFun (just getting rid of compare/2 would do).  I'll be testing this tomorrow.

engelsanchez:
And of course, the other "minor" issue of a file having completely moved out of the repo that Jordan mentioned :)

slfritchie:
I think I've addressed all of the review comments.  Dialyzer appears happy enough: the total number of warnings has dropped, and no new warnings have been added.

slfritchie:
Oh, oops.  So I patched a hashtree.erl that's now a ghost?

engelsanchez:
@slfritchie yes, we should do that simple hashtree patch in a separate PR in its new home in riak_core and update this. After that and a bit of testing on the final code, assuming it's all good, I should be able to give you a quick +1

slfritchie:
Engel: https://github.com/basho/riak_core/pull/411

engelsanchez:
I believe the comments have been addressed. I've verified sorted AAE repairs manually. EUnit tests have only failures that I've observed lately in the develop branch (get_put_monitor_eqc, crdts) but that don't seem to happen in the builder machines. It looks good. :+1: :dancer: 

I'm going to push a manual merge that I used for testing that takes care of the moved hashtree files.

PR #648, opened 2013-09-04
Add dialyzer targets.

cmeiklejohn:
@seancribbs Add a target to make it easier to build and keep the PLT up to date.


seancribbs:
:+1: MOAR DIALYZE

btw it doesn't take a 1/2 hour to rebuild anymore, this is 2013, dialyzer is parallel

PR #649, opened 2013-09-05
Fix keys fsm EQC test

engelsanchez:
The keys FSM EQC test was failing with lists of actual keys having many
duplicates for each key and failing to match.
The test relies on the memory backend tables to be named to be able to
wipe them out in between generated runs, which may and will use
different bucket n_val values, placing replicas in different sets of
vnodes each time. This was changed recently, so
adding named_table option at ETS creation time.

Make memory table options configurable to allow the keys_fsm_eqc test,
that relies on tables being named so it can clear them all between
runs and some memory_backend tests that will
break if they are named, as they create multiple ones with the same
partition number. This should allow them both to succeed.

### DIFF
--- a/src/riak_kv_memory_backend.erl
+++ b/src/riak_kv_memory_backend.erl
@@ -118,11 +118,16 @@ start(Partition, Config) ->
     TTL = riak_kv_util:get_backend_config(ttl, Config, memory_backend),
     MemoryMB = riak_kv_util:get_backend_config(max_memory, Config, memory_backend),
     %% leave this one alone, it is only for testing
+    NormalTableOpts = [ordered_set],
+    TestTableOpts = app_helper:get_prop_or_env(test_table_opts,
+                                               Config,
+                                               memory_backend,
+                                               [named_table, public]),
     TableOpts = case app_helper:get_prop_or_env(test, Config, memory_backend) of
                     true ->
-                        [ordered_set, public];
+                        NormalTableOpts ++ TestTableOpts;
                     _ ->
-                        [ordered_set]
+                        NormalTableOpts
                 end,
     case MemoryMB of
         undefined ->
@@ -632,8 +637,8 @@ object_size(Object) ->
 -ifdef(TEST).
 
 simple_test_() ->
-    application:set_env(memory_backend, test, true),
-    riak_kv_backend:standard_test(?MODULE, []).
+    Config = [{test, true}, {test_table_opts, [public]}],
+    riak_kv_backend:standard_test(?MODULE, Config).
 
 ttl_test_() ->
     Config = [{ttl, 15}],


kellymclaughlin:
How about just deleting this test? It is really an integration test and not a unit test and the setup and teardown is error prone. I don't think gives us much assurance we don't already get from the `verify_listkeys` riak_test. Just a thought.

engelsanchez:
I'm all for removing if people agree that's OK. I just wish it had been mentioned while I spent hours making the test work :(

kellymclaughlin:
I'm originally wrote this test and I'll put my :+1: towards a quorum for removal.

engelsanchez:
Jon agrees that if the same functionality is covered, it can go. I don't see anything in the test that's beyond what the riak_test has. I'm closing this and preparing to remove it.

engelsanchez:
Actually, had a change of heart: I'd like to postpone removing this until the listkeys riak_test can be enhanced to try different n_vals and use filters.  I opened this issue to remind me of that: https://github.com/basho/riak_test/issues/372. I actually have a version of the riak_test listkeys using EQC that I'd like to take mainstream soon, hopefully with this work.

rzezeski:
+1 to merge, keys_fsm failed before and works now.

reiddraper:
Broken again for me now on develop. Anyone else?

```
module 'keys_fsm_eqc'
  keys_fsm_eqc:49: eqc_test_...Failed! Reason:
{'EXIT',{{badmatch,{error,precommit_fail}},
         [{keys_fsm_eqc,'-prop_basic_listkeys/0-lc$^1/1-1-',3,
                        [{file,"test/keys_fsm_eqc.erl"},{line,76}]},
          {keys_fsm_eqc,'-prop_basic_listkeys/0-fun-3-',6,
                        [{file,"test/keys_fsm_eqc.erl"},{line,76}]}]}}
After 1 tests.
{3963010637,<<"=">>,none,3,1083,30136}
Shrinking....(4 times)
Reason:
{'EXIT',{{badmatch,{error,precommit_fail}},
         [{keys_fsm_eqc,'-prop_basic_listkeys/0-lc$^1/1-1-',3,
                        [{file,"test/keys_fsm_eqc.erl"},{line,76}]},
          {keys_fsm_eqc,'-prop_basic_listkeys/0-fun-3-',6,
                        [{file,"test/keys_fsm_eqc.erl"},{line,76}]}]}}
{3963010637,<<" ">>,none,1,10,10000}
*failed*
in function keys_fsm_eqc:'-eqc_test_/0-fun-3-'/1 (test/keys_fsm_eqc.erl, line 49)
**error:{assertEqual_failed,[{module,keys_fsm_eqc},
                     {line,49},
                     {expression,"quickcheck ( numtests ( ? TEST_ITERATIONS , ? QC_OUT ( prop_basic_listkeys ( ) ) ) )"},
                     {expected,true},
                     {value,false}]}
```

PR #650, opened 2013-09-05
Add CRDTs to Riak KV

russelldb:
Adds  Maps and Sets to Riak.

Depends on the `develop` branch of riak_dt.

What to say: generalises the 1.4 `riak_kv_*_counter` code to allow Maps and Sets to be stored in Riak.

What's not done? GC, MDC, backwards compatibility with 1. counters, bucket type support (we fake it.)

This is very much a pre1 PR to start to integrate and investigate.

Still needs a couple more tests (eqc if there is time) for `riak_kv_crdt` for update and update with preconditions.
Also, needs riak_tests writing (will start on that now, using REC branch here (https://github.com/basho/riak-erlang-client/compare/feature/sdc-datatypes))


src/riak_kv_pb_crdt.erl
```diff
@@ -0,0 +1,210 @@
...
+                           make_option(basic_quorum, BQ) ++
+                           make_option(sloppy_quorum, SloppyQ) ++
+                           make_option(n_val, NVal)) of
+                {ok, O} ->
+                    {Ctx0, Value} = riak_kv_crdt:value(O, Mod),
+                    Ctx = get_context(Ctx0, InclCtx),
+                    Resp = riak_pb_dt_codec:encode_fetch_response(Type, Value, Ctx, ?MOD_MAP),
+                    {reply, Resp, State};
+                {error, notfound} ->
+                    {reply, #dtfetchresp{}, State};
```
seancribbs: This results in a crash. Even when notfound, the DtFetchResp message requires a data-type.

`#dtfetchresp{type=Type}` should fix it, or `encode_fetch_response` with value and context of `undefined`.

seancribbs:
Be sure to add `riak_dt` as a dependency in the .app file. When I built the rel the first time, it couldn't find any of the type modules. Alternately, add it to `reltool.config` in the top-level repo.

russelldb:
Ah, I added it in riak. Ok.

src/riak_kv_pb_crdt.erl
russelldb: How do we communicate 'not_found' then? Or is bottom (hehehhehehe) the same as not found? I think it isn't 'cos it suggests the key exists, but at the same time, if a user then updates the datastructure it _will_ exist.

seancribbs: Well, to the user then the bottom-value and notfound are equivalent, right? The only rub on the client-side is that an empty fetch response will be decoded with no value in the case of counters, and empty values in the case of sets and maps (repeated fields and such).

seancribbs: Re-read what you said... maybe the fetch response should have a not found field? I think it's still important to communicate the expected type in case the client wants to create an empty data structure for the user.

src/riak_kv_wm_raw.hrl
```diff
@@ -35,6 +35,7 @@
 -define(HEAD_INDEX_PREFIX,    "x-riak-index-").
 -define(HEAD_DELETED,         "X-Riak-Deleted").
 -define(HEAD_TIMEOUT,         "X-Riak-Timeout").
+-define(HEAD_CRDT_CONTEXT,    "X-Riak-CRDT-Ctx").
```
seancribbs, line 38: What if the context gets too large for a header field? Most clients will bork on headers > several KB.

evanmcc:
Does riak_dt need to be reviewed along with this?

src/riak_kv_crdt.erl
```diff
@@ -0,0 +1,440 @@
...
+-ifdef(TEST).
+-ifdef(EQC).
+-export([test_split/0, test_split/1]).
+
+-include_lib("eqc/include/eqc.hrl").
+-endif.
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-define(TAG, 69).
```
evanmcc, line 41: really?

src/riak_kv_crdt.erl
```diff
@@ -0,0 +1,440 @@
...
+-define(V1_VERS, 1).
+
+-type crdts() :: [{DT_MOD::module(), crdt()}].
+-type ro_content() :: {Meta::dict(), Value::binary()}.
+-type ro_contents() :: [ro_content()].
+-type precondition_error() :: {error, {precondition, {not_present, term()}}}.
+
+%% @doc applies the gievn `Operation' to the merged value.  first
```
evanmcc: gievn -> given

src/riak_kv_crdt.erl
```diff
@@ -0,0 +1,440 @@
...
+                {orddict:new(), []},
+               Contents).
+
+%% @private worker for `merge_contents/1' if the content is a CRDT,
+%% de-binary it, merge it and store the most merged value in the
+%% accumulartor dictionary.
+-spec merge_value(ro_content(), {crdts(), ro_contents()}) ->
+                         {crdts(), ro_contents()}.
+merge_value({MD, <<?TAG:8/integer, ?V1_VERS:8/integer, CRDTBin/binary>>},
+            {Dict, NonCounterSiblings}) ->
```
evanmcc: Should this be NonCRDTSiblings?

src/riak_kv_crdt.erl
```diff
@@ -0,0 +1,440 @@
...
+-spec merge_contents(ro_contents()) ->
+                            {crdts(), ro_contents()}.
+merge_contents(Contents) ->
+    lists:foldl(fun merge_value/2,
+                {orddict:new(), []},
+               Contents).
+
+%% @private worker for `merge_contents/1' if the content is a CRDT,
+%% de-binary it, merge it and store the most merged value in the
+%% accumulartor dictionary.
```
evanmcc: accumulator

src/riak_kv_crdt.erl
```diff
@@ -0,0 +1,440 @@
...
+%% @see split_ops/1 for more explanation
+-spec update_crdt(orddict:orddict(), riak_dt:actor(), crdt_op()) -> orddict:ordddict() | precondition_error().
+update_crdt(Dict, Actor, ?CRDT_OP{mod=Mod, op=Op, ctx=undefined}) ->
+    {Meta, Record, Value} = fetch_with_default(Mod, Dict),
+    case Mod:update(Op, Actor, Value) of
+        {ok, NewVal} ->
+            orddict:store(Mod, {Meta, Record?CRDT{value=NewVal}}, Dict);
+        {error, _}=E -> E
+    end;
+update_crdt(Dict, Actor, ?CRDT_OP{mod=Mod, op=Ops, ctx=OpCtx}) when Mod==?MAP_TYPE;
```
evanmcc, line 164: do counters ever have context?  otherwise this seems like it'd fail w/ function clause

hazen:
Sorry. Wrong PR.

src/riak_kv_wm_crdt.erl
```diff
@@ -147,11 +150,16 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
             end;
         false  ->
             {false,
-             wrq:set_resp_body("Counters are not supported.",
+             wrq:set_resp_body(io_lib:format("~p are not supported.~n", [Type]),
```
evanmcc, line 153: may want to add an s after the ~p, otherwise this will say 'map are not supported'.

src/riak_kv_vnode.erl
```diff
@@ -1047,34 +1054,40 @@ prepare_put(#state{vnodeid=VId,
                                                                              PruneTime,
                                                                              BProps))
                     end,
-                    ObjToStore2 = handle_counter(Coord, CounterOp, VId, ObjToStore),
-                    {{true, ObjToStore2},
-                     PutArgs#putargs{index_specs=IndexSpecs, is_index=IndexBackend}}
+                    case handle_crdt(Coord, CRDTOp, VId, ObjToStore) of
+                        {error, _}=E ->
+                            {{fail, Idx, E}, PutArgs};%% NOTE subverts ReqId to Error
```
evanmcc, line 1059: extend perform_put with a replying error case rather than working around, maybe?

evanmcc, line 1059: Hm, I suppose that you already do.  Nevermind.

evanmcc:
Other than the comments, this is looking good, and Sean's test and variations on it pass.  Would like to see the comments addressed and the eunit tests fixed; they currently don't seem to run on their own, and other eunit failures have kept me from running it as part of the bigger suite.

src/riak_kv_vnode.erl
russelldb, line 1059: Yeah. This has precedence in the 'overload' case elsewhere in the vnode. Seems to be the way.

src/riak_kv_crdt.erl
russelldb, line 41: Sadly that slipped out the door in 1.4 with counters. It was pretty juvenile I guess, but I was writing the code, didn't want to break my stride thinking of a number, that was the first number that came into my head, I forgot about it / assumed it would get changed before release, and it never did. We can always claim it was just the next number in some internal sequence…which it sort of is (the 50s are used for riak_object binary format.) Anyway, there you go…to remove it would involve upgrade code for counters…your call. It is just a Bill And Ted reference anyway, try not to be to puritanical about it.

src/riak_kv_crdt.erl
russelldb, line 164: Counters never ever ever have a context. Ever.

src/riak_kv_wm_crdt.erl
russelldb, line 153: Not doing HTTP for pre1 or pre2, please to be ignoring riak_kv_wm_crdt

russelldb:
OK, think I addressed them all. Got the old counter eqc tests running, too.

We need to create an issue to make sure we test 1.4 counters, and mixed 1.4 / 2.0 counters.

russelldb:
And an issue for more | actually deterministic riak_tests…

src/riak_kv_pb_crdt.erl
```diff
@@ -0,0 +1,219 @@
...
+         process_stream/3]).
+
+-import(riak_pb_kv_codec, [decode_quorum/1]).
+
+-record(state, {client}).
+
+-define(DEFAULT_TIMEOUT, 60000).
+
+%% The empty counter that is the body of all new counter objects
+-define(NEW_COUNTER, {riak_kv_pncounter, riak_kv_pncounter:new()}).
```
engelsanchez: This is unused it seems.

russelldb: Oh yes, throwback from the 1.4 era

engelsanchez:
I haven't found any blocking issues that should prevent this from making it into pre2. The riak_test is working a lot better now and demonstrates the basic functionality. The code in riak_kv_vnode doesn't seem like it would affect other features as far as I can tell. There are some caveats:
- Work on the HTTP interface is expected to make it to pre3
- Compatibility issues with 1.4 counters are also expected to be addressed then. As a note, I was not able to run the counter convergence riak test, which uses the HTTP client I believe

With that said :+1: MOAR CRDTS!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

rebar.config
```diff
@@ -21,5 +21,6 @@
         {sext, ".*", {git, "git://github.com/uwiger/sext", {tag, "1.1-3-g3af5478"}}},
         {riak_pipe, ".*", {git, "git://github.com/basho/riak_pipe.git",
                                 "develop"}},
-        {riak_api, ".*", {git, "git://github.com/basho/riak_api.git", "develop"}}
+        {riak_api, ".*", {git, "git://github.com/basho/riak_api.git", "develop"}},
+	{riak_dt, ".*", {git, "git://github.com/basho/riak_dt.git", "develop"}}
```
lenary, line 25: TOOOOONCES. Yes, I'm going to highlight the tiny details.

src/riak_kv_crdt.erl
```diff
@@ -0,0 +1,501 @@
...
+value(RObj, Type) ->
+    {CRDTs, _NonCRDTSiblings} = merge_object(RObj),
+    crdt_value(Type, orddict:find(Type, CRDTs)).
+
+%% @TODO in riak_dt change value to query allow query to take an
+%% argument, (so as to query subfields of map, or set membership etc)
+-spec crdt_value(module(), error | {ok, {dict(), crdt()}}) ->
+                        {binary(), riak_dt:value()}.
+crdt_value(Type, error) ->
+    {<<>>, Type:value(Type:new())};
```
lenary, line 91: in some places you've used undefined for where context isn't supplied. Might it be an idea to do it here, and go from there instead of returning an empty binary.

I get this is going the other direction (crdt -> user), but it might still be an idea.

src/riak_kv_wm_raw.hrl
lenary, line 38: It doesn't matter, as we aren't doing HTTP for pre1 or pre2, so we can get to this later.

lenary:
This code looks good to me, and the eunits pass on my machine. :shipit: 

I have a few tiny niggles that can be worked out after a merge, some I commented above, some are a bit more involved, I'll send @russelldb and @seancribbs an email about them.

seancribbs:
Thanks @engelsanchez and @lenary for the extra review. Sam, we look forward to your niggley email.

seancribbs:
Manually resolved/merged in 79518bc291fc06aab4493142c88d5bb54e4b41bb

seancribbs:
Sorry, borked merge from stale version on my machine. Remerged with missing commits in 08c22ec80d0c5b994942c23f9a9bd1734ddc7c35

PR #653, opened 2013-09-06
Add Yokozuna index hook

rzezeski:
Add a low-level hook in the vnode for the purposes of indexing objects
in Yokozuna.  This hook provides a notification to Yokozuna _any time_
an object's bytes are modified on disk.


coderoshi:
+1

rzezeski:
This addresses basho/yokozuna#156.

PR #654, opened 2013-09-06
Generic AAE Status

rzezeski:
This is replacing #596.  It's been rebased against develop.

This pull request makes AAE entropy info parameterized on type.  This allows both KV and Yokozuna to share almost all the same status code without copy pasta.
- Add a `Type` to entropy info.  It represents the type of AAE being performed.  Currently `riak_kv` or `yokozuna` but could be other things in the future.
- Export the `get_build_time` function as it can be reused by Yokozuna.
- Parameterize the module and function name used to compute all exchanges.  KV has inter-partition exchange whereas Yokozuna is purely intra.  I.e. Yokozuna exchange is always between the same partition, different services.  By parameterizing the "all exchanges" calculation can let different services have different exchange patterns.

### DIFF
--- a/src/riak_kv_entropy_info.erl
+++ b/src/riak_kv_entropy_info.erl
@@ -24,7 +24,9 @@
          create_table/0,
          dump/0,
          compute_exchange_info/1,
-         compute_tree_info/1]).
+         compute_exchange_info/2,
+         compute_tree_info/1,
+         all_exchanges/2]).
 
 -define(ETS, ets_riak_kv_entropy).
 
@@ -89,11 +91,14 @@ dump() ->
                                    [{index(), Last :: t_now(), All :: t_now(),
                                      repair_stats()}].
 compute_exchange_info(Type) ->
+    compute_exchange_info(Type, {?MODULE, all_exchanges}).
+
+compute_exchange_info(Type, {M,F}) ->
     filter_index_info(),
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     Indices = riak_core_ring:my_indices(Ring),
     Defaults = [{Index, undefined, undefined, undefined} || Index <- Indices],
-    KnownInfo = [compute_exchange_info(Ring, Index, Info)
+    KnownInfo = [compute_exchange_info({M,F}, Ring, Index, Info)
                  || {{Type2, Index}, Info} <- all_index_info(), Type2 == Type],
     merge_to_first(KnownInfo, Defaults).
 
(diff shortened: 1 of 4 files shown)


src/riak_kv_entropy_info.erl
```diff
@@ -132,10 +141,10 @@ all_index_info() ->
 filter_index_info() ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     Primaries = riak_core_ring:my_indices(Ring),
-    Indices = ets:select(?ETS, [{{{index, '$1'}, '$2'}, [], ['$1']}]),
+    Indices = ets:select(?ETS, [{{{index, {'_', '$1'}}, '_'}, [], ['$1']}]),
     Others = ordsets:subtract(ordsets:from_list(Indices),
                               ordsets:from_list(Primaries)),
-    [ets:delete(?ETS, {index, Idx}) || Idx <- Others],
+    [ets:match_delete(?ETS, {{index, {'_', Idx}}, '_'}) || Idx <- Others],
```
jrwest, line 166: moving comments from https://github.com/basho/riak_kv/pull/596#discussion_r6211971 here:

I'm fine w/ the quick fix for now. I think we should open an issue as well since the table size does seem related to ring size (number of partitions owned by a single node). May not be that much of a hit vs. calling multiple deletes but since we did the large ring work we should be careful.

jrwest:
Changes look good. As mentioned in chat I'm +1 on this with the caveat that I think issues should be opened to export `tree_built/2` and to change to using `ets:delete` over `match_delete` (as pre2 issues). I didn't get a chance to run it with the changes but I can do that later today if we want to hold off a little longer on the merge.

src/riak_kv_exchange_fsm.erl
```diff
@@ -284,6 +284,7 @@ next_state_with_timeout(StateName, State, Timeout) ->
     {next_state, StateName, State, Timeout}.
 
 exchange_complete({LocalIdx, _}, {RemoteIdx, RemoteNode}, IndexN, Repaired) ->
-    riak_kv_entropy_info:exchange_complete(LocalIdx, RemoteIdx, IndexN, Repaired),
+    riak_kv_entropy_info:exchange_complete(LocalIdx, RemoteIdx,
```
jrwest: minor nitpick but do we even need this file in the patch now?

coderoshi: :fish:

src/riak_kv_entropy_info.erl
rzezeski, line 166: https://github.com/basho/riak_kv/issues/657

PR #656, opened 2013-09-06
use yokozuna or riak_search for {search, ...} MR inputs

beerriot:
_Note_: this should probably be considered non-merge-able until bucket-types land, since that has the potential for significant conflict, and redefinition of bucket-index mapping for yokozuna. I'm opening the PR now for comment and backup purposes.

There are two ways to use riak_search inputs for MapReduce:
- specify a {modfun, ...} input ({"module":...,"function":...} in json)
- specify a {search, ...} input ({"bucket":...,"query":...} in json)

This patch allows yokozuna to power the second type instead of
riak_search. The value of the "bucket" field is used as the name of the
yokozuna index.

This patch does not convert the modfun version, to keep from surprising
users, and to allow an escape hatch for people that need to use both.

If only one of riak_search or yokozuna is enabled, that system is chosen
to power search MR inputs.

If both are enabled, then the riak_kv's appenv 'mapred_search' must also
be defined as either 'riak_search' or 'yokozuna'. If it is not defined,
or it is defined as something else, and error will be returned from the MR.

If neither is enabled, an error is returned from the MR.

The automatic nature of the switch is intended to keep configuration as
minimal as possible: the extra riak_kv setting is only needed if _both_
riak_search and yokozuna are to be used, _and_ the user wants to use the
non-modfun forms. If only one search system is enabled, or only modfun
input forms are used, the correct thing just happens.

A test for this behavior is defined by `mapred_search_switch` in basho/riak_test#373.

### DIFF
--- a/src/riak_kv_mrc_pipe.erl
+++ b/src/riak_kv_mrc_pipe.erl
@@ -618,11 +618,21 @@ send_inputs(Pipe, {index, Bucket, Index, StartKey, EndKey}, Timeout) ->
             send_inputs(Pipe, NewInput, Timeout)
     end;
 send_inputs(Pipe, {search, Bucket, Query}, Timeout) ->
-    NewInput = {modfun, riak_search, mapred_search, [Bucket, Query, []]},
-    send_inputs(Pipe, NewInput, Timeout);
+    case search_module() of
+        {ok, Mod} ->
+            NewInput = {modfun, Mod, mapred_search, [Bucket, Query, []]},
+            send_inputs(Pipe, NewInput, Timeout);
+        {error, _}=Error ->
+            Error
+    end;
 send_inputs(Pipe, {search, Bucket, Query, Filter}, Timeout) ->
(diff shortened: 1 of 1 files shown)


beerriot:
After discussion today, this PR should be amended to have yokozuna mode expect "index" instead of bucket in the input spec.

seancribbs:
Not sure it was decided: does it make sense to allow the old "bucket" inputs, and have it imply "use the bucket's index, plus filter the 'bucket' field"?

beerriot:
@seancribbs We can definitely allow the old "bucket" input and look up the index to use from that. We'll have to be careful about filtering the results that produces, though, since it might behave strangely if sorting and limiting are in play.

seancribbs:
@beerriot Sure, that makes sense. I was thinking of the filter as part of the YZ query, but like Donnie I'm out of my element.

beerriot:
Closing in favor of #664, which takes into account the discussion above.

PR #658, opened 2013-09-11
Implement a maximum size for bucket identifiers

lenary:
After a discussion with @jtuple, he pointed out that there's a limit on key
length but not on bucket length. We realised that someday, someone will abuse
this possibility, so I decided to code up a fix.

I have no idea how many of our current customers will be affected. Hopefully not very many.

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -56,6 +56,7 @@
 -type binary_version() :: v0 | v1.
 
 -define(MAX_KEY_SIZE, 65536).
+-define(MAX_BUCKET_SIZE, 65536).
 
 -define(LASTMOD_LEN, 29). %% static length of rfc1123_date() type. Hard-coded in Erlang.
 -define(V1_VERS, 1).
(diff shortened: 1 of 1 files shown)


evanmcc:
Note that the current combined allowed length in bitcask is 16k.

lenary:
I... have no idea what to do about that. Is there a combined allowed length for leveldb too?

lenary:
Wait, isn't the combined max keysize for bitcask 64K too? https://github.com/basho/bitcask/blob/develop/include/bitcask.hrl#L37

evanmcc:
that line is trumped, sadly, by this one:

https://github.com/basho/bitcask/blob/develop/include/bitcask.hrl#L32

and this one:

https://github.com/basho/bitcask/blob/develop/c_src/bitcask_nifs.c#L73

Also your combined max is 128k, no?

I am not sure what the right thing to do is.  I am not sure that bitcask
even checks.  It's reasonable, since bitcask's keydir is all in-memory,
that it implement a tighter restriction than riak proper.  I have no actual
objection to this change, it just makes it a bit more complicated to do it
right.

It might be better to have it expressed by individual backends as a
capability, rather than building it into riak object, but that puts it very
far away from the frontend, where it might be a good idea to implement some
sort maximum message length anyway.

I dunno.

lenary:
My reaction is that we should have a limit in riak proper (rather than the backend), but a backend can impose stricter limits. 

I'll take a look at what bitcask does when I try to put in a huge bucket or key name.

evanmcc:
I think that it's just going to corrupt the entry.  Bleah, more stuff to
fix.

slfritchie:
Pulling PRs out of the freezer ... sounds like this one is still simmering?

lenary:
I'll take another look at this. I never really got very far with it.

evanmcc:
too late for 2.0 in any case.  this and other changes should be part of a
bigger push to harden riak against accidental and malicious misuse, I feel.

PR #659, opened 2013-09-11
Move riak_core.proto from riak_core.

seancribbs:
This protobuff definition is only used by KV. Whether we should
continue to use this encoding can be debated separately, but putting
the file in KV removes yet another dependency from Core. Third parties
who use Core can pull in protobuffs via riak_api or their own
applications. For KV, it is pulled as a dependency of riak_api.

/cc @Vagabond @evanmcc @jrwest

Pair of basho/riak_core#382 and should be considered together.

### DIFF
--- a/src/riak_core.proto
+++ b/src/riak_core.proto
@@ -1,0 +1,6 @@
+message RiakObject_PB {
+        required bytes bucket = 1;
+        required bytes key = 2;
+        required bytes val = 3;
+}
+


seancribbs:
Ping? This is a simple change and removes protobuffs as a dep of core.

Vagabond:
I'm +1 on the idea, and the changeset, although I have not tested it.

evanmcc:
should the module not be added to riak_kv_app.src?

evanmcc:
nevermind the previous comment, I understand now.

evanmcc:
This looks reasonable, seems to function the same before and after the patch.  I consistently get an error in protobuffs encoding in the bucket streaming, but that happens on `develop` as well.  I assume that is has something to do with bucket types.
:+1:, unless you think that issue might be related.  will dig into it more tomorrow.

seancribbs:
Thanks @evanmcc. Yes, that problem would most likely be in riak_pb or riak_kv_pb_bucket. I can look into it next week earliest, but if it is in KV, the fix is likely simple.

PR #661, opened 2013-09-16
Added Mutators to support changing object on read/write

lordnull:
This is to support https://github.com/basho/riak_repl/issues/223 .

This implements a 'mutator' interface. A mutator callback module can alter an object to be stored and after it has been retrieved from a storage backend.

Note this has been recently rebased and force-pushed (sept 26) to account for changes from develop.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -187,6 +187,9 @@ start(_Type, _StartArgs) ->
                                           [enabled, disabled],
                                           disabled),
 
+            % get/put mutator is supported, but not all nodes may have it
+            riak_core_capability:register({riak_kv, mutators}, [true, false], false),
+
             HealthCheckOn = app_helper:get_env(riak_kv, enable_health_checks, false),
             %% Go ahead and mark the riak_kv service as up in the node watcher.
             %% The riak_core_ring_handler blocks until all vnodes have been started
--- a/src/riak_kv_mutator.erl
+++ b/src/riak_kv_mutator.erl
@@ -90,8 +90,12 @@ mutate_get(Object) ->
 mutate_get(Object, []) ->
     Object;
 mutate_get(Object, [Mutator | Tail]) ->
-    Object2 = Mutator:mutate_get(Object),
-    mutate_get(Object2, Tail).
+    case Mutator:mutate_get(Object) of
+        notfound ->
+            notfound;
+        Object2 ->
+            mutate_get(Object2, Tail)
+    end.
 
 mutate_put(Object, BucketProps) ->
     Contents = riak_object:get_contents(Object),
--- a/src/riak_kv_mutator.erl
+++ b/src/riak_kv_mutator.erl
@@ -32,9 +32,14 @@
 
 -module(riak_kv_mutator).
 
+-include_lib("eunit/include/eunit.hrl").
+
 -export([register/1, register/2, unregister/1]).
 -export([get/0]).
 -export([mutate_put/2, mutate_get/1]).
+
+-callback mutate_put(Meta :: dict(), Value :: any(), ExposedMeta :: dict(), FullObject :: riak_object:riak_object(), BucketProps :: orddict:orddict()) -> {dict(), any(), dict()}.
+-callback mutate_get(Meta :: dict(), Value :: any(), FullObject :: riak_object:riak_object()) -> {dict(), any()}.
 
 -define(DEFAULT_PRIORITY, 0).
 
@@ -73,30 +78,44 @@ get() ->
     {ok, Modules}.
 
 mutate_get(Object) ->
-    Meta = riak_object:get_metadata(Object),
-    {AppliedMutators, Meta2} = case dict:find(mutators_applied, Meta) of
+    Contents = riak_object:get_contents(Object),
+    Contents2 = lists:map(fun({InMeta, InValue}) ->
+        maybe_get_mutate(InMeta, InValue, Object)
+    end, Contents),
+    riak_object:set_contents(Object, Contents2).
+
+maybe_get_mutate(Meta, Value, Object) ->
+    case dict:find(mutators_applied, Meta) of
         error ->
-            {[], Meta};
+            ?debugMsg("no mutators applied"),
+            {Meta, Value};
         {ok, Applied} ->
-            {Applied, dict:erase(mutators_applied, Meta)}
-    end,
-    Object2 = riak_object:update_metadata(Object, Meta2),
-    Object3 = riak_object:apply_updates(Object2),
-    FoldFun = fun(Module, Obj) ->
-        Module:mutate_get(Obj)
-    end,
-    lists:foldl(FoldFun, Object3, lists:reverse(AppliedMutators)).
+            ?debugFmt("muttors applied in put order: ~p", [Applied]),
+            lists:foldr(fun(Mutator, {InMeta, InValue}) ->
+                Mutator:mutate_get(InMeta, InValue, Object)
+            end, {Meta, Value}, Applied)
+    end.
 
 mutate_put(Object, BucketProps) ->
-    FoldFun = fun(Module, Obj) ->
-        Module:mutate_put(Obj, BucketProps)
-    end,
+    Contents = riak_object:get_contents(Object),
     {ok, Modules} = ?MODULE:get(),
-    Object2 = lists:foldl(FoldFun, Object, Modules),
-    Meta = riak_object:get_metadata(Object2),
-    Meta2 = dict:store(mutators_applied, Modules, Meta),
-    Object3 = riak_object:update_metadata(Object2, Meta2),
-    riak_object:apply_updates(Object3).
+    MetasValuesRevealeds = lists:map(fun({InMeta, InValue}) ->
+        {InMeta1, InValue1, InRevealed} = lists:foldl(fun(Module, {InInMeta, InInValue, InInRevealed}) ->
+            Module:mutate_put(InInMeta, InInValue, InInRevealed, Object, BucketProps)
+        end, {InMeta, InValue, dict:new()}, Modules),
+        InMeta2 = dict:store(mutators_applied, Modules, InMeta1),
+        {InMeta2, InValue1, InRevealed}
+    end, Contents),
+    Contents2 = [{M,V} || {M,V,_R} <- MetasValuesRevealeds],
+    Mutated = riak_object:set_contents(Object, Contents2),
+    FakedContents = lists:map(fun({InMeta, InContent, InRevealed}) ->
+        FixedMeta = dict:merge(fun(_Key, _NotMutated, MutatedVal) ->
+            MutatedVal
+        end, InMeta, InRevealed),
+        {FixedMeta, InContent}
+    end, MetasValuesRevealeds),
+    Faked = riak_object:set_contents(Object, FakedContents),
+    {Mutated, Faked}.
 
 merge_values([]) ->
     [];
--- a/src/riak_kv_mutator.erl
+++ b/src/riak_kv_mutator.erl
@@ -29,6 +29,50 @@
 %% can be run.
 %%
 %% This doubles as a behavior defining module for the mutators.
+%%
+%% == Callbacks ==
+%%
+%% A mutator callback must implement 2 function: mutate_put/5 and mutate_get/1.
+%%
+%% <code><b>mutate_put(MetaData, Value, ExposedMeta,
+%% FullObject, BucketProperties) -> Result</b></code>
+%%
+%% Types:
+%% ```
+%%     MetaData = dict()
+%%     Value = term()
+%%     ExposedMeta = dict()
+%%     FullObject = riak_object:riak_object()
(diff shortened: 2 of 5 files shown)


src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,210 @@
...
+%% This doubles as a behavior defining module for the mutators.
+%%
+%% == Callbacks ==
+%%
+%% A mutator callback must implement 2 function: mutate_put/5 and mutate_get/1.
+%%
+%% <code><b>mutate_put(MetaData, Value, ExposedMeta,
+%% FullObject, BucketProperties) -> Result</b></code>
+%%
+%% Types:
```
Vagabond, line 47: I think edoc can generate all this from dialyzer -type and -callback specs now.

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,210 @@
...
+%% `mutate_put' function. Also note how the entire object is given as opposed to
+%% simply a metadata/value pair. Care must be taken not to corrupt the object.
+%%
+%% A return of ``'notfound''' stops the mutator chain and returns immediately. This
+%% provides an escape hatch of sorts; if the mutator cannot reverse the mutation
+%% effectively, return ``'notfound'''.
+
+-module(riak_kv_mutator).
+
+-include_lib("eunit/include/eunit.hrl").
```
Vagabond: Shouldn't this be -ifsef(TEST)?

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,210 @@
...
+-spec register(Module :: atom()) -> 'ok'.
+register(Module) ->
+    ?MODULE:register(Module, ?DEFAULT_PRIORITY).
+
+%% @doc Register a module as a mutator with the given priority. Modules with
+%% equal priority are done in sort sort (alphabetical) order. A module
+%% can only be registered once.
+-spec register(Module :: atom(), Priority :: term()) -> 'ok'.
+register(Module, Priority) ->
+    Modifier = fun
```
Vagabond, line 110: It'd be nice to comment what resolution strategy this is trying to implement.

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,210 @@
...
+            orddict:erase(Module, Values2)
+    end,
+    riak_core_metadata:put({riak_kv, mutators}, list, Modifier, []).
+
+%% @doc Retrieve the list of mutators in the order to apply them when doing a
+%% a put mutation. To get the order when doing a get mutation, reverse the list.
+-spec get() -> [atom()].
+get() ->
+    Resolver = fun
+        ('$deleted', '$delete') ->
```
Vagabond: Is the second atom here a typo?

src/riak_kv_mutator.erl
lordnull, line 47: Looking into this (and trying it as well), edoc lists the functions but doesn't give type information.

Vagabond, line 47: I think if you did like -type metadata() :: dict(). and then used metadata() it'd show them better?

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,209 @@
...
+            mutate_get(Object2, Tail)
+    end.
+
+%% @doc Mutate an object in preparation to storage, returning a tuple of the
+%% object to store and the object to return to the client.
+-spec mutate_put(Object :: riak_object:riak_object(), BucketProps :: orddict:orddict()) -> {riak_object:riak_object(), riak_object:riak_object()}.
+mutate_put(Object, BucketProps) ->
+    Contents = riak_object:get_contents(Object),
+    {ok, Modules} = ?MODULE:get(),
+    MetasValuesRevealeds = lists:map(fun({InMeta, InValue}) ->
```
Vagabond, line 198: So for each sibling in the object, we apply all the mutators, then replace the contents of the object with the result?

Vagabond, line 198: Really, more comments here would be nice.

lordnull, line 198: Hopefully 5934e3b will address this particular issue.

src/riak_kv_mutator.erl
lordnull, line 47: I think we need to define better. I'd love to be able to have it automatically use the correct types, but that's a ways out. Adding links to types is a next step, but would it be better inline (in the code element) or in the expanded area?

Vagabond:
The tests fail for me, presumably because of changes to cluster metadata:

```
=ERROR REPORT==== 24-Sep-2013::15:38:43 ===
** Generic server riak_core_metadata_manager terminating
** Last message in was {put,{{riak_kv,mutators},list},
                            [],#Fun<riak_kv_mutator.0.123063518>}
** When Server state == {state,nonode@nohost,"test_data",876565}
** Reason for termination ==
** {noproc,{gen_server,call,
                       [riak_core_metadata_hashtree,
                        {insert,{{riak_kv,mutators},list},
                                <<0,170,139,16,5,28,164,71,218,221,64,20,238,
                                  83,39,191,48,160,195,137>>,
                                false},
                        infinity]}}
```

Vagabond:
I'm mostly OK with this change, although I think @jtuple was also going to look at it.

lordnull:
Eunit tests should pass now.

Also, there is now a riak_test pull request that enhances the tests further; it attempts to stress the interaction between aae and the replication mutator: https://github.com/basho/riak_test/pull/392

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,222 @@
...
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc There are circumstances where the object stored on disk is not
+%% the object to return; and there are times the object written to the
+%% data storage backend is not meant to be the object given. An
+%% example would be storing only meta data for an object on a remote
+%% cluster. This module is an interface to register mutators that will
+%% can be run.
```
engelsanchez: I can haz this sentence fixed next time you're in the file?

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,222 @@
...
+-define(DEFAULT_PRIORITY, 0).
+
+%% @doc Register the given module as a mutator with the default priority of 0.
+%% @see register/2
+-spec register(Module :: atom()) -> 'ok'.
+register(Module) ->
+    ?MODULE:register(Module, ?DEFAULT_PRIORITY).
+
+%% @doc Register a module as a mutator with the given priority. Modules with
+%% equal priority are done in sort sort (alphabetical) order. A module
```
engelsanchez: sort sort == sort ^ 2 ?

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,222 @@
...
+    end.
+
+mutate_get(Object, []) ->
+    Object;
+mutate_get(Object, [Mutator | Tail]) ->
+    % so event though the mutate_put callback has to return
+    % {Meta, Value, Exposed} values, the get callback gets away with just
+    % giving the object? This is to avoid complicated interaction with
+    % notfound return.
+    case Mutator:mutate_get(Object) of
```
engelsanchez, line 179: Notice that this makes it possible to invalidate all the data in the cluster if a mutator stored in object metadata is removed from the code. Users of this feature will have to be careful about some downgrade scenarios where data is written with a mutator in the list than then disappears on downgrade.

src/riak_kv_app.erl
```diff
@@ -187,6 +187,9 @@ start(_Type, _StartArgs) ->
                                           [enabled, disabled],
                                           disabled),
 
+            % get/put mutator is supported, but not all nodes may have it
+            riak_core_capability:register({riak_kv, mutators}, [true, false], false),
```
engelsanchez, line 191: This capability is not being queried anywhere that I can see.

src/riak_kv_mutator.erl
lordnull, line 179: Do we just leave this to documentation, or should there be some insurance put in the code here?

engelsanchez, line 179: A note in the documentation would be enough for now.  It's good to flag the worse case scenarios.

src/riak_kv_app.erl
lordnull, line 191: It is used in repl before attempting to install the reduced repl mutator: https://github.com/basho/riak_repl/blob/develop/src/riak_repl_sup.erl#L20

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,240 @@
...
+        ('$deleted', Values) ->
+            Values;
+        (Values, '$deleted') ->
+            Values;
+        (Values1, Values2) ->
+            merge_values([Values1, Values2])
+    end,
+    ModulesAndPriors = riak_core_metadata:get({riak_kv, mutators}, list, [{default, []}, {resolver, Resolver}]),
+    Flipped = [{P, M} || {M, P} <- ModulesAndPriors],
+    Sorted = lists:sort(Flipped),
```
engelsanchez: A sort in the critical path could become an issue with multiple mutators. I know we are not planning to add those soon, but it seems that the resolution and merge functions could take care of this directly and store values in sorted order, no?

engelsanchez:
While writing my own test mutators I was surprised by how they can be called twice on the same object (once in the local put, and the result of that is passed to the other vnodes in the preflist).  Unless the application of the mutator chain is completely idempotent, it will be very hard to reason about.  I had tried to do a simple "make value a list of two values" and a "make a list a tuple" mutator and put them in a chain, but the results end up just crazy.  The first one went from a value of 1 to {1, 1}, whereas the replicas ended up with {{1,1}, {1,1}}.  I'm leaving this here as a note of the headache that it could be to use this feature in a truly generic fashion.

engelsanchez:
I'm going to +1 this PR with the following caveats:
- We need to make sure we revisit and review performance before release. Part of my concerns will probably be addressed in the cluster metadata code by removing gen_server calls. The sort on get should be targeted for removal if resolution can create sorted lists without issue.
- I have my doubts that the code will survive as is when a second use case arrives. Of course, it's impossible to foresee what that may be.

The code as is should be enough to support the Repl use case.

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,247 @@
...
+    Acc;
+
+merge_values([{Priority, Module} = Mutator | Tail], Acc) ->
+    Acc2 = case lists:keyfind(Module, 2, Acc) of
+        false ->
+            ordsets:add_element(Mutator, Acc);
+        {P1, _} when P1 < Priority ->
+            Acc;
+        Else ->
+            ordsets:all_element(Mutator, ordsets:del_element(Else))
```
engelsanchez: I haven't even run it, but you surely meant add_element here, not all_element.

lordnull: D'oh!

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,247 @@
...
+            lists:keydelete(Module, 2, Values2)
+    end,
+    riak_core_metadata:put({riak_kv, mutators}, list, Modifier, []).
+
+%% @doc Retrieve the list of mutators in the order to apply them when doing a
+%% a put mutation. To get the order when doing a get mutation, reverse the list.
+-spec get() -> [atom()].
+get() ->
+    Resolver = fun
+        ('$deleted', '$deleted') ->
```
engelsanchez, line 141: A note for @jrwest: it seems that we should have the metadata tombstone in a public header or is there a way for resolvers not to use this atom directly?

engelsanchez:
I've looked over the sort change and tested manually. After the one issue I found was resolved, it looks good. Re +1 on this.

src/riak_kv_mutator.erl
jrwest, line 141: hrm thats a good point. we certainly want to expose tombstones to resolvers (or at least have the option to) but leaking this is a bit nasty. a macro would probably be quickest, but i wonder if there is a better way...

jrwest, line 141: for this code specifically, does it even call `riak_core_metadata:delete`? If not, don't really need to have these here.

engelsanchez, line 141: Well no, but I think it's worth handling anyway for our convenience at the very least. I used delete while testing it when the format was changed, for example.

src/riak_kv_mutator.erl
```diff
@@ -0,0 +1,247 @@
...
+    {ok, Modules}.
+
+%% @doc Unmutate an object after retrieval from storage. When an object is
+%% mutated, the mutators applied are put into the object's metadata.
+%% If the mutator does not exist anymore or has changed in a backwards
+%% incompatible manner, at best there will be corrupt objects, at worst
+%% a crash.
+-spec mutate_get(Object :: riak_object:riak_object()) -> riak_object:riak_object().
+mutate_get(Object) ->
+    [Meta | _] = riak_object:get_metadatas(Object),
```
russelldb, line 163: Is this right? What about siblings with no mutators in their metadata being higher up the list?

PR #662, opened 2013-09-16
Add security to Riak

Vagabond:
This PR uses the new security APIs in riak_core to do permissions checking on various riak API endpoints.

It currently does not secure ALL of the APIs, specifically stats, link walking and the 'CS bucket' APIs are not covered.

See also basho/riak#355

### DIFF
--- a/src/riak_kv_mapred_term.erl
+++ b/src/riak_kv_mapred_term.erl
@@ -24,7 +24,7 @@
 
 -module(riak_kv_mapred_term).
 
--export([parse_request/1]).
+-export([parse_request/1, get_required_permissions/2]).
 
 -define(DEFAULT_TIMEOUT, 60000).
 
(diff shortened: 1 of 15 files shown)


src/riak_kv_mapred_term.erl
```diff
@@ -119,3 +119,69 @@ parse_query(Query) when is_list(Query) ->
     {ok, Query};
 parse_query(Invalid) ->
     {error, {"Query takes a list of step tuples", Invalid}}.
+
+%% Look at a parsed MR Input and Query and return a list of required
+%% permissions.
+get_required_permissions(Inputs, _Query) ->
+    InputPermissions = case Inputs of
+        Bucket when is_binary(Bucket) ->
+            [{"riak_kv.list_keys", {<<"default">>, Bucket}},
```
engelsanchez, line 128: The sad thing about having all these default bucket name literals spread out through the code is that search and replace to change them into a macro will only miss the ones with typos, the only ones we need to fix :(

jaredmorrow:
From a usage standpoint and the understanding there is a lot of testing that's going to happen off of pre2, I'm +1 for this to go in.

coderoshi:
+1 here too. mapreduce is clear, and boilerplate code looks fine.

PR #664, opened 2013-09-19
Use yokozuna or riak_search for "search" MapReduce inputs

beerriot:
This PR replaces #656.

There are two ways to use riak_search inputs for MapReduce:
- specify a {modfun, ...} input ({"module":...,"function":...} in json)
- specify a {search, ...} input ({"bucket":...,"query":...} in json)

This PR allows yokozuna to power the second type instead of riak_search.

This patch does not convert the modfun version, to keep from surprising
users, and to allow an escape hatch for people that need to use both.

If only one of riak_search or yokozuna is enabled, that system is chosen
to power search MR inputs.

If both are enabled, then the riak_kv's appenv 'mapred_search' must also
be defined as either 'riak_search' or 'yokozuna'. If it is not defined,
or it is defined as something else, and error will be returned from the MR.

If neither is enabled, an error is returned from the MR.

The automatic nature of the switch is intended to keep configuration as
minimal as possible: the extra riak_kv setting is only needed if both
riak_search and yokozuna are to be used, and the user wants to use the
non-modfun forms. If only one search system is enabled, or only modfun
input forms are used, the correct thing just happens.

When YZ is chosen, it will translate the "bucket" parameter to the YZ index that has been associated with that bucket. Results from that index are filtered, such that only results matching the bucket specified are passed to MR.

This PR also allows the user to specify a YZ index directly. (as of basho/yokozuna#174) To do so, use `{index, Index}` where `Bucket` would normally go in the Erlang form, or use `"index":...` instead of `"bucket":...` in the JSON form.
- `{search, {index, Index}, Query, Filter}`
- `{"index":Index, "query":Query}`

This makes the JSON form easy to confuse with the JSON form of 2i inputs, which also uses "index" in its specification. The difference between the two is that search uses "query", while 2i uses "key" or "start" and "end".

If riak_search is chosen instead of YZ, it passes the index name straight through, since buckets map 1:1 with indexes in riak_search. (as of basho/riak_search#149)

A test for these new behaviors is defined by mapred_search_switch in basho/riak_test#387.


beerriot:
I expect the thing most likely to be wrong with this PR is around the filter query to single out one bucket in an index. It probably needs URL escaping, and may also be incompatible with bucket types.

src/riak_kv_mrc_pipe.erl
```diff
@@ -636,6 +646,34 @@ send_inputs(Pipe, {modfun, Mod, Fun, Arg} = Modfun, Timeout) ->
...
+search_module() ->
+    case {enabled(yokozuna), enabled(riak_search)} of
+        {true, true} ->
+            case application:get_env(riak_kv, mapred_search) of
+                %% being explicit here to help find typo errors earlier
+                {ok, yokozuna} ->
+                    {ok, yokozuna};
+                {ok, riak_search} ->
+                    {ok, riak_search};
+                undefined ->
```
rzezeski, line 660: If both systems are up and no `mapred_search` system is defined I would like to default to Riak Search.  This will be the case during Riak Search migration.

beerriot, line 660: WFM. ~~Willdo.~~ Done in 3ab3e01

rzezeski:
+1 to merge

Issue #666, opened 2013-09-20
You can upset your cluster by submitting a non-numeric n_val


russelldb:
And I imagine similar damage with other bucket properties. We have a simple bucket validation "Framework" in riak now, add some validators for n_val and other number properties so riak cuts fewer hands that try to hold it.

russelldb:
See https://github.com/basho/riak_kv/pull/667 and https://github.com/basho/riak_test/pull/389

jrwest:
addressed by merge of https://github.com/basho/riak_kv/pull/770

PR #667, opened 2013-09-22
Add validation for n_val and other riak_kv bucket props

russelldb:
Addresses issue #666 test at https://github.com/basho/riak_test/pull/389


slfritchie:
This seems quite sane, despite needing to check atom() and binary() versions of the same option, oh well.  +1.

src/riak_kv_bucket.erl
```diff
@@ -52,9 +52,46 @@ validate([{BoolProp, MaybeBool}|T], ValidProps, Errors) when is_atom(BoolProp),
...
+            validate(T, ValidProps, [{QProp, not_valid_quorum} | Errors])
+    end;
 validate([Prop|T], ValidProps, Errors) ->
     validate(T, [Prop|ValidProps], Errors).
 
+
+-spec is_quorum(term()) -> true | false.
+is_quorum(Q) when is_integer(Q), Q > 0 ->
+    true;
+is_quorum(Q) when is_atom(Q), Q =:= quorum
```
engelsanchez: There is no need for the is_atom here since you are matching exactly (=:=) against atoms anyway. The same with binaries below.

russelldb: oh yeah, doh!

russelldb:
Addressed comments.

src/riak_kv_bucket.erl
```diff
@@ -52,9 +52,45 @@ validate([{BoolProp, MaybeBool}|T], ValidProps, Errors) when is_atom(BoolProp),
...
+            validate(T, [Prop | ValidProps], Errors);
+        false ->
+            validate(T, ValidProps, [{QProp, not_valid_quorum} | Errors])
+    end;
 validate([Prop|T], ValidProps, Errors) ->
     validate(T, [Prop|ValidProps], Errors).
 
+
+-spec is_quorum(term()) -> true | false.
+is_quorum(Q) when is_integer(Q), Q > 0 ->
```
engelsanchez, line 91: The default for pr and pw is zero, which doesn't pass this validation. So, for example, getting the default properties and setting them again, maybe with a modification, will fail. I'm not sure why it's zero. The code seems to make sure that zero is never used I believe.

Maybe @Vagabond or @jrwest know the answer?

src/riak_kv_bucket.erl
```diff
@@ -52,9 +52,45 @@ validate([{BoolProp, MaybeBool}|T], ValidProps, Errors) when is_atom(BoolProp),
...
+                                                                  orelse IntProp =:= n_val
+                                                                  orelse IntProp =:= old_vclock
+                                                                  orelse IntProp =:= small_vclock ->
+case is_integer(MaybeInt) of
+    true when MaybeInt > 0 ->
+        validate(T, [Prop | ValidProps], Errors);
+    _ ->
+        validate(T, ValidProps, [{IntProp, not_integer} | Errors])
+end;
+validate([{QProp, MaybeQ}=Prop | T], ValidProps, Errors) when is_atom(QProp), QProp =:= dw
```
engelsanchez: And if you're back in the file, the comment about is_atom being redundant applies to the validate function too.

src/riak_kv_bucket.erl
jrwest, line 91: yea it can be zero https://github.com/basho/riak_kv/blob/develop/src/riak_kv_app.erl#L83-L97

russelldb, line 91: Do you think other quora can be zero? I mean can you zero R? W? DW? RW?

russelldb:
Ping @engelsanchez addressed comments

russelldb:
Bump?

russelldb:
re-re-bump ping bueller, bueller

jrwest:
Oh, I assumed this no longer merges cleanly but it looks like it does :). I'm doing some separate `n_val` validation from strong consistency, so I wouldn't mind double checking that it won't change anything here (I should know tomorrow). From what I can tell so far it shouldn't. If it does, I can merge this into that PR.

slfritchie:
Pulling PRs out of the freezer ... re-re-re-bump?

russelldb:
@slfritchie see riak_kv#747… @jrwest has said he'll merge this with that.

jrwest:
yep and I'm just looking for the final +1 from @jtuple on #747 before I do so

jrwest:
closing for rebased/merged version in #770

PR #669, opened 2013-09-24
Add permissions to crdt requests.

seancribbs:
@Vagabond @russelldb

### DIFF
--- a/src/riak_kv_pb_crdt.erl
+++ b/src/riak_kv_pb_crdt.erl
@@ -75,7 +75,8 @@ init() ->
 
 %% @doc decode/2 callback. Decodes an incoming message.
 decode(Code, Bin) ->
-    {ok, riak_pb_codec:decode(Code, Bin)}.
+    Msg = riak_pb_codec:decode(Code, Bin),
+    {ok, Msg, permission_for(Msg)}.
 
 %% @doc encode/1 callback. Encodes an outgoing response message.
 encode(Message) ->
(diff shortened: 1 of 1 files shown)


src/riak_kv_pb_crdt.erl
```diff
@@ -214,3 +215,8 @@ get_context(Ctx, true) ->
 mods_match(BucketMod, OpType) ->
     OpMod = riak_kv_crdt:to_mod(OpType),
     OpMod == BucketMod.
+
+permission_for(#dtupdatereq{bucket=B, type=T}) ->
```
Vagabond, line 348: Will type always be a binary here? Never undefined?

seancribbs, line 348: `type` is `required bytes` in the protobuffs message.

russelldb:
Looks good to me. And wow. Is that it? Bravo @Vagabond for making it so neat.

Vagabond:
Yes, this is all you need. +1 from me, although a test would be nice (maybe added to the pb_security test in the security riak_test PR:  basho/riak_test#385 )

seancribbs:
@Vagabond I'll add to basho/riak_test#385.

@russelldb We'll have to add them to HTTP when we get around to solidifying that API.

seancribbs:
I've rebased this so it applies cleanly. Still need a test in pb_security, I suppose.

seancribbs:
^^ @lenary pb_security and http_security need additions if you have time. If not, I'll add them next week.

russelldb:
test passes now, are you going to get this merged too?

seancribbs:
yup yup

PR #672, opened 2013-09-30
Bump rebar; fix makefile.

cmeiklejohn:
- Use a version of rebar which correctly builds the protobuffs
  dependencies with +debug_info.
- Compile needs to depend on deps; build dialyzer with all deps once
  rebar is upgraded.


seancribbs:
:+1: no more protobuffs warnings, but we have lots to fix in KV! :smile:

cmeiklejohn:
:beers:

PR #673, opened 2013-09-30
update bucket validator for changes in core

jrwest:
changes necessary for those made in basho/riak_core#402. 

NOTE: this PR does not actually implement the enforcement of any "create-only" types. It only makes the changes necessary for the code to run.


Vagabond:
Looks fine. +1

PR #677, opened 2013-10-01
bucket type console functions

jrwest:
these live in kv because they rely on mochijson2 which is no longer in riak_core

This PR relies on https://github.com/basho/riak_core/pull/402 and https://github.com/basho/riak_kv/pull/673

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -37,7 +37,13 @@
          aae_status/1,
          reformat_indexes/1,
          reformat_objects/1,
-         reload_code/1]).
+         reload_code/1,
+         bucket_type_status/1,
+         bucket_type_activate/1,
+         bucket_type_create/1,
+         bucket_type_update/1,
+         bucket_type_reset/1,
+         bucket_type_list/1]).
 
 join([NodeStr]) ->
     join(NodeStr, fun riak_core:join/1,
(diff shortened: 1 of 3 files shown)


Vagabond:
```
./rel/riak/bin/riak-admin bucket-type update foo '{"props": {"n_val": 6}}'

./rel/riak/bin/riak-admin bucket-type status foo
....
n_val: 3
postcommit: []
precommit: []
<<"n_val">>: 6
...
```

This seems bad?

jrwest:
@Vagabond hrm. how did you create the type? I think `n_val` should be an atom and looking at the examples in https://github.com/basho/riak/pull/395 that doesn't seem to be the case. I'll take a look, I thought I lifted the json parsing directly from the webmachine resources. I suppose this is one of the other annoyances w/ json....

jrwest:
ok, i see what I missed [1]. I'm going to move `erlify_bucket_props` somewhere common and use it here as well. Will have the changes up shortly.

[1]  https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_props.erl#L337-L369.

jrwest:
@Vagabond https://github.com/basho/riak_kv/commit/731a7167127f247a60860cd4644e6628884c815e should address the issue you found.

Vagabond:
This blows up if you try to set a bucket prop that isn't an existing atom in the VM. Is that OK?

jrwest:
meh, i guess thats maybe not ideal but all the atoms should be existing (they are used or validated somewhere i imagine?). If the blow up is really ugly user facing maybe we should catch it?

Vagabond:
It is good enough for now. +1

PR #678, opened 2013-10-02
Error reason for put is not propagated back to client

engelsanchez:
On a put error, the code in riak_kv_vnode will ignore the error and reply with a request id. We don't think this request id is necessary at all. We need to send back the error reason.  The overload code is already replacing that request id with the overload token when it triggers.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -391,7 +391,7 @@ init([Index]) ->
 
 
 handle_overload_command(?KV_PUT_REQ{}, Sender, Idx) ->
-    riak_core_vnode:reply(Sender, {fail, Idx, overload}); % subvert ReqId
+    riak_core_vnode:reply(Sender, {fail, Idx, overload});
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1055,8 +1055,8 @@ prepare_put(#state{vnodeid=VId,
                                                                              BProps))
                     end,
                     case handle_crdt(Coord, CRDTOp, VId, ObjToStore) of
-                        {error, _}=E ->
-                            {{fail, Idx, E}, PutArgs};%% NOTE subverts ReqId to Error
+                        {error, E}=E ->
```
russelldb: What gets sent to the fsm, can you overload `E` like this in one statement?

engelsanchez: That is not what I saw in my brain when I edited this. I forgot to delete the final =E

russelldb:
:+1: works and everything :D

PR #680, opened 2013-10-02
Add warning/max object limits

engelsanchez:
CSEs have wanted more visibility into sibling and object size explosions for a while. A warning threshold would make sure that the logs can be used to pinpoint or event prevent size and sibling number explosions.

We have also discussed putting hard limits on size and sibling number. Writes would fail for very large objects or objects with too many siblings. This should only be used as a last resort to prevent a cluster from experiencing severe performance problems that can result in cascaded deaths even

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1234,10 +1234,20 @@ do_get_binary(Bucket, Key, Mod, ModState) ->
 do_get_object(Bucket, Key, Mod, ModState) ->
     case uses_r_object(Mod, ModState, Bucket) of
         true ->
+            %% Non binary returns do not trigger size warnings
             Mod:get_object(Bucket, Key, false, ModState);
         false ->
             case do_get_binary(Bucket, Key, Mod, ModState) of
                 {ok, ObjBin, _UpdModState} ->
+                    BinSize = size(ObjBin),
+                    WarnSize = app_helper:get_env(riak_kv, warn_object_size),
+                    case BinSize > WarnSize of
+                        true ->
+                            lager:warning("Read large object ~p/~p (~p bytes)",
+                                          [Bucket, Key, BinSize]);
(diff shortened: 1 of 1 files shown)


engelsanchez:
Configuration schema changes for this are in basho/riak#397 and tests in basho/riak_test#398

src/riak_kv_vnode.erl
```diff
@@ -1701,13 +1711,55 @@ return_encoded_binary_object(Method, EncodedObject) ->
            {{error, Reason::term(), UpdModState::term()}, EncodedObj::binary()}.
 
 encode_and_put(Obj, Mod, Bucket, Key, IndexSpecs, ModState) ->
+    NumSiblings = riak_object:value_count(Obj),
+    case NumSiblings > app_helper:get_env(riak_kv, max_siblings) of
```
evanmcc, line 1715: should we handle the case that someone has set the number to some non_integer?

engelsanchez, line 1715: Any non integer term is larger than any integer in Erlang. So it handles itself: it would behave as if there wasn't a threshold

evanmcc:
A potential optimization here would be to stuff all of these values into a 4-tuple and have a function to unpack them.  As yet I have no idea if all of the extra ets calls are going to be an issue.

Code looks good, on to some testing.

engelsanchez:
You would be surprised by how efficient these ETS lookups are when measured with DTrace in the general case. Still, I added them as is because I have some plans to optimize configuration that I will tell you about elsewhere. It would be a post-ricon thing and would benefit from keeping plain app_helper:get_env calls instead of trying to optimize by hand. I have a todo item on that.

evanmcc:
No noticeable performance impacts in its current form, code seems sensible and does what is described.
:+1:  Additional comments on the default settings have been left on the appropriate PR.

PR #686, opened 2013-10-07
When binary encoding timestamps, 24 bits is enough for secs, micros

russelldb:
Their maximum value is 999999.

Also, I bet we could get away with fewer bytes for Mega Seconds
(16 bytes gives us another 272 years, 24 gives us ~0.5million years)


russelldb:
Actually, realised this bumps the version, not worth the trouble, is it?

jrwest:
Yea I agree its probably a pain to bump the version for just this change but I think its a good thing to keep on a list for `v2`. Unfortunately, I missed this in v1 :(. Perhaps open an issue so we can keep track of it? Or leave the PR open for now?

russelldb:
Added an issue to track this and other wishes for v2 binary format, @jrwest I reviewed the binary format, should have caught this then, but it is only ~2bytes, what's that between friends?

jrwest:
ha, not so much between friends, but a bit between millions of friends ;)

Issue #691, opened 2013-10-08
Call to missing function in `riak_kv_mutator`


rzezeski:
I noticed this while building PLT for Yokozuna.

```
riak_kv_mutator.erl:240: Call to missing or unexported function ordsets:del_element/1
```

engelsanchez:
Thanks for catching that @rzezeski.

engelsanchez:
This should be fixed now by #692

PR #692, opened 2013-10-08
Fixed call to undefined function in kv_mutator

lordnull:


engelsanchez:
This is a fix for https://github.com/basho/riak_kv/issues/691

engelsanchez:
Added a small unit test in riak_kv_mutator.erl to exercise the paths of the private merge_values function where this problem was found.  It all looks good to me now

:+1: :dancer: :boat: 

And I'm merging it now.

Issue #683, opened 2013-10-03
Better client timeout handling for get/put operations


engelsanchez:
If the protocol buffer client times out waiting for a get or put operation, the FSM remains alive and may later send messages to the PB process, which results in errors like:

2013-10-03 16:00:45.228 [error] <0.28542.0>@riak_api_pb_server:handle_info:141 Unrecognized message {86078617,{error,timeout}}
2013-10-03 16:00:45.228 [error] <0.28438.0>@riak_api_pb_server:handle_info:141 Unrecognized message {56705038,{error,timeout}}

This has been addressed in the 2i path by killing the FSM and draining the mailbox of incoming messages when the client side times out. Something similar should be done for the other operations.  This is probably a problem in the HTTP interface too.

evanmcc:
I suspect that I'll be making changes to this path for perf work in this dev cycle, so I can fix this when I am in there.  I think the right thing to do is just to make the fsm die silently.  This doesn't help the case where the reply is delayed and gets to the client after it has timed out, but that's rarer.

Tempted, along the lines of things that @rzezeski has suggested to make the FSM timeout 10s and the client timeout 12s to help obviate that.  I am not sure that 60s timeouts really work for anyone, and a major version bump is likely a good place to make that change.

reiddraper:
#693, which references this is on the 2.1 milestone. @engelsanchez @evanmcc, can this issue be moved to 2.1 as well?

evanmcc:
This will be handled by #693, yes.  I've changed the milestone.

reiddraper:
Thanks @evanmcc.

PR #693, opened 2013-10-08
Optionally remove use of BIF timers from get and put FSMs.

evanmcc:
The BIF timer lock is quite expensive to take, and as the number of
cores in your system increases, so does contention on it for commonly
spawned processes.

This patch has very little performance impact on systems below 20 cores,
but is necessary to achieve higher performance down the road.

It transfers responsibility for killing the FSMs to the client (although we
retain the older code path in case the new one is a problem), backstopping
the client by using cheaper state transition timers.  Logic is added to
reconnect the client with remotely spawned put coordinators.

Also addresses #683

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -80,19 +80,20 @@ get(Bucket, Key, {?MODULE, [_Node, _ClientId]}=THIS) ->
 %%       {error, Err :: term()}
 %% @doc Fetch the object at Bucket/Key.  Return a value as soon as R-value for the nodes
 %%      have responded with a value or error.
-get(Bucket, Key, Options, {?MODULE, [Node, _ClientId]}) when is_list(Options) ->
+get(Bucket, Key, Options0, {?MODULE, [Node, _ClientId]}) when is_list(Options0) ->
     Me = self(),
     ReqId = mk_reqid(),
-    case node() of
+    {Timeout, Options} = extract_timeout(Options0),
+    Pid = 
+        case node() of
         Node ->
-            riak_kv_get_fsm:start_link({raw, ReqId, Me}, Bucket, Key, Options);
-        _ ->
-            proc_lib:spawn_link(Node, riak_kv_get_fsm, start_link,
-                                [{raw, ReqId, Me}, Bucket, Key, Options])
-    end,
+                riak_kv_get_fsm:start_link({raw, ReqId, Me}, Bucket, Key, Options);
+            _ ->
+                proc_lib:spawn_link(Node, riak_kv_get_fsm, start_link,
+                                    [{raw, ReqId, Me}, Bucket, Key, Options])
+        end,
     %% TODO: Investigate adding a monitor here and eliminating the timeout.
-    Timeout = recv_timeout(Options),
-    wait_for_reqid(ReqId, Timeout);
+    wait_for_reqid(Pid, ReqId, Timeout);
 
 %% @spec get(riak_object:bucket(), riak_object:key(), R :: integer(), riak_client()) ->
 %%       {ok, riak_object:riak_object()} |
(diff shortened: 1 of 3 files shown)


src/riak_client.erl
```diff
@@ -643,11 +645,28 @@ wait_for_reqid(ReqId, Timeout) ->
...
+            wait_for_reqid(CoordPid, ReqId, Timeout1);            
+        {ReqId, Response} -> Response;
+        {'EXIT', _} -> 
+            %% if the client is part of a long-lived process, we need to 
+            %% be careful about selective receive problems, as can get 
+            %% put monitor messages.
+            Timeout2 = recalc_timeout(Timeout, StartTime),
+            wait_for_reqid(Pid, ReqId, Timeout2)
     after Timeout ->
+            exit(Pid, kill),
```
slfritchie, line 660: We're linked to this process aren't we?  If so, then we need to unlink first ... otherwise this is going to kill us transitively.

evanmcc, line 660: So just unlink, or should we change to spawn_monitor?

src/riak_client.erl
```diff
@@ -643,11 +645,28 @@ wait_for_reqid(ReqId, Timeout) ->
...
             {error, timeout}
     end.
 
+%% @private 
+recalc_timeout(Timeout, StartTime) ->
+    Now = os:timestamp(),
+    Elapsed = timer:now_diff(Now, StartTime),
+    Timeout - (Elapsed div 1000).
```
slfritchie, line 668: Do we need an erlang:max(0, ...) here to avoid a negative result?

src/riak_client.erl
slfritchie, line 660: Hrm, unlink seems alright, I think.

slfritchie:
I hope to be able to test & poke at this tomorrow (Friday JST).

slfritchie:
Aside from the small questions added earlier this week, everything looks good.  riak_client.erl appears to have a few less Dialyzer warnings, yay.

slfritchie:
Using this basho_bench config to apply a light load:

```
{mode, {rate,2}}.
{duration, 10}.
{concurrent, 88}.
{report_interval, 1}.

{driver, basho_bench_driver_riakc_pb}.

{key_generator, {int_to_bin_bigendian, {pareto_int, 250}}}.
{value_generator, {fixed_bin, 1}}.

{riakc_pb_ips, [
                {"localhost", 8087}
               ]}.

{riakc_pb_replies, quorum}.
{pb_timeout_general, 5000}.
{pb_timeout_read, 5000}.
{operations, [{get, 4},{update, 1}]}.
```

And then used the following to put some vnodes to sleep:

```
Vs = [Pid || {idxrec,{_,riak_kv_vnode},_,_,Pid,_} <- riak_core_vnode_manager:get_tab()].
[sys:suspend(X) || X <- lists:sublist(Vs, 8)].
```

After waiting for a while, I saw these in the Riak logs.  They're all the same.

```
2013-12-13 18:54:01.003 [error] <0.4186.0> gen_fsm <0.4186.0> in state connected terminated with reason: {error,badarg,[{erlang,exit,[{ok,<0.7719.0>},kill],[]},{riak_client,wait_for_reqid,3,[{file,"src/riak_client.erl"},{line,660}]},{riak_kv_pb_object,process,2,[{file,"src/riak_kv_pb_object.erl"},{line,240}]},{riak_api_pb_server,process_message,4,[{file,"src/riak_api_pb_server.erl"},{line,386}]},{riak_api_pb_server,connected,2,[{file,"src/riak_api_pb_server.erl"},{line,228}]},{riak_api_pb_server,decode_buffer,2,[{file,"src/riak_api_pb_server.erl"},{line,362}]},{gen_fsm,handle_msg,7,[{file,"gen..."},...]},...]}
```

slfritchie:
After I patched that error, it was a lot easier to see these.  (Silly me, I didn't use my "egrep -v" trick to weed out common log events, never forget!) There were many fewer of them.  I didn't pay attention to how long I'd waited, but b_b hadn't run for more than 285 seconds.

```
2013-12-13 19:07:07.156 [error] <0.10547.1> gen_fsm <0.10547.1> in state waiting_vnode_r terminated with reason: no function clause matching riak_kv_get_fsm:update_stats(timeout, {state,{raw,114377594,<0.631.0>},[{r,quorum}],3,[{{182687704666362864775460604089535377456991567872,...},...},...],...}) line 511
```

The hack patch, not sure if it's really doing the right thing by assuming that the return value is {ok, Pidx}.

```
diff --git a/src/riak_client.erl b/src/riak_client.erl
index ca7848b..40ba015 100644
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -87,7 +87,8 @@ get(Bucket, Key, Options0, {?MODULE, [Node, _ClientId]}) when is_list(Options0)
     Pid = 
         case node() of
         Node ->
-                riak_kv_get_fsm:start_link({raw, ReqId, Me}, Bucket, Key, Options);
+                {ok, Pidx} = riak_kv_get_fsm:start_link({raw, ReqId, Me}, Bucket, Key, Options),
+                Pidx;
             _ ->
                 proc_lib:spawn_link(Node, riak_kv_get_fsm, start_link,
                                     [{raw, ReqId, Me}, Bucket, Key, Options])
@@ -155,7 +156,8 @@ put(RObj, Options0, {?MODULE, [Node, ClientId]}) when is_list(Options0) ->
               undefined ->
                   case node() of
                       Node ->
-                          riak_kv_put_fsm:start_link({raw, ReqId, Me}, RObj, Options);
+                          {ok, Pidx} = riak_kv_put_fsm:start_link({raw, ReqId, Me}, RObj, Options),
+                          Pidx;
                       _ ->
                           proc_lib:spawn_link(Node, riak_kv_put_fsm, start_link,
                                               [{raw, ReqId, Me}, RObj, Options])
@@ -164,7 +166,8 @@ put(RObj, Options0, {?MODULE, [Node, ClientId]}) when is_list(Options0) ->
                   UpdObj = riak_object:increment_vclock(RObj, ClientId),
                   case node() of
                       Node ->
-                          riak_kv_put_fsm:start_link({raw, ReqId, Me}, UpdObj, [asis|Options]);
+                          {ok, Pidx} = riak_kv_put_fsm:start_link({raw, ReqId, Me}, UpdObj, [asis|Options]),
+                          Pidx;
                       _ ->
                           proc_lib:spawn_link(Node, riak_kv_put_fsm, start_link,
                                               [{raw, ReqId, Me}, RObj, [asis|Options]])
```

src/riak_client.erl
jonmeredith, line 660: Brutally killing request process prevents actions like read repair taking place for any partial results and reporting any stats on timeouts (not that it does a great job right now, but it should).

If there were mass timeouts around anything that uses a non-sidejob supervisor (for example deletion) could enough kills cause that supervisor to die and cause the application supervisor above to be unhappy enough to exit the app and take the node down?

Sending a message would be nicer (a Phil Collins "I don't care any more"), but that only works for requests on the local node, anything overdisterl would have issues.

evanmcc, line 660: Haven't had a chance to get back to this.  I am not sure how critical it is
until later on in our optimization efforts (or naturally over time as there
are more cores).  The kill is just a best effort anyway, given disterl.  I
am OK with sending a dont_care message, although we get bit again by the
gen_fsm 'have I covered all the states' thing.

Will try to rework for next week.

seancribbs:
Assigning to 2.1 milestone so we can remove "Next +1" milestone

yorkshire-steve:

PR #694, opened 2013-10-09
Add bucket types support to HTTP.

seancribbs:
This supports the proposed API from basho/riak#362.

Notes:
- Linkwalking is not supported over keys stored in bucket-types. Similarly, links can only be used between keys in the default type, as described in the proposal.
- JSON MapReduce was adjusted to support `[Type, Bucket]` as inputs for full-bucket jobs, and `[ [Type, Bucket, Key, KeyData], ...]` for specific keys. Note that the `KeyData` is required in this form, even if it is null/undefined.
- Bucket-type properties resources were added so that there is parity with the existing PB feature. Whether clients _should_ be able to modify those properties is still under debate. /cc @jrwest @Vagabond 

A riak_test and support in the Erlang HTTP client cover these new features.

### DIFF
--- a/src/riak_kv_web.erl
+++ b/src/riak_kv_web.erl
@@ -77,6 +77,9 @@ raw_dispatch(Name) ->
       riak_kv_wm_link_walker, Props1}
     ] ++
 
+   [ {["types", bucket_type, "props"], riak_kv_wm_bucket_type,
+      [{api_version, 3}|raw_props(Name)]} ] ++
+
    lists:flatten([
     [
      %% NEW API
--- a/src/riak_kv_web.erl
+++ b/src/riak_kv_web.erl
@@ -53,9 +53,8 @@ raw_dispatch() ->
 
 raw_dispatch(Name) ->
     Props1 = [{bucket_type, <<"default">>}, {api_version, 1}|raw_props(Name)],
-    Props2 = [{api_version, 2}|raw_props(Name)],
-    Props3 = [ {["types", bucket_type], Props2},
-               {[], [{bucket_type, <<"default">>}|Props2]}],
+    Props2 = [ {["types", bucket_type], [{api_version, 3}|raw_props(Name)]},
+               {[], [{bucket_type, <<"default">>}, {api_version, 2}|raw_props(Name)]}],
 
     [
      %% OLD API
@@ -105,7 +104,7 @@ raw_dispatch(Name) ->
      %% crdts
      {Prefix ++ ["buckets", bucket, crdt, key], {riak_kv_wm_crdt, known_type},
       riak_kv_wm_crdt, Props}
-    ] || {Prefix, Props} <- Props3 ]).
+    ] || {Prefix, Props} <- Props2 ]).
 
 is_post(Req) ->
     wrq:method(Req) == 'POST'.
--- a/src/riak_kv_web.erl
+++ b/src/riak_kv_web.erl
@@ -52,8 +52,11 @@ raw_dispatch() ->
     end.
 
 raw_dispatch(Name) ->
-    Props1 = [{api_version, 1}|raw_props(Name)],
+    Props1 = [{bucket_type, <<"default">>}, {api_version, 1}|raw_props(Name)],
     Props2 = [{api_version, 2}|raw_props(Name)],
+    Props3 = [ {["types", bucket_type], Props2},
+               {[], [{bucket_type, <<"default">>}|Props2]}],
+
     [
      %% OLD API
      {[Name],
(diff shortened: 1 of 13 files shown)


engelsanchez:
For quick lookup, related PRs are: 
- Erlang HTTP client PR https://github.com/basho/riak-erlang-http-client/pull/24
- Riak test https://github.com/basho/riak_test/issues/410

src/riak_kv_web.erl
```diff
@@ -52,8 +52,10 @@ raw_dispatch() ->
     end.
 
 raw_dispatch(Name) ->
-    Props1 = [{api_version, 1}|raw_props(Name)],
-    Props2 = [{api_version, 2}|raw_props(Name)],
+    Props1 = [{bucket_type, <<"default">>}, {api_version, 1}|raw_props(Name)],
+    Props2 = [ {["types", bucket_type], [{api_version, 3}|raw_props(Name)]},
+               {[], [{bucket_type, <<"default">>}, {api_version, 2}|raw_props(Name)]}],
+
```
engelsanchez, line 58: This not incorrect, but I noticed that we pass the modules the <<"default">> bucket_type from here, and then later riak_kv_wm_utils:ensure_bucket_type() will set it to <<"default">> if missing. It's just all a little redundant.

seancribbs, line 58: Yeah, call it paranoia? Which way is preferable, do you think?

src/riak_kv_wm_object.erl
```diff
@@ -1139,6 +1158,12 @@ handle_common_error(Reason, RD, Ctx) ->
                         io_lib:format("not found~n",[]),
                         RD)),
                 Ctx};
+        {error, bucket_type_unknown} ->
+            {{halt, 404},
+             wrq:set_resp_body(
+               io_lib:format("Unknown bucket type: ~p", [Ctx#ctx.bucket_type]),
```
engelsanchez: This is very minor, but perhaps using ~s here instead of ~p will print a less erlangy message? That is, it won't show the binary delimiters (<<"default">>  -> default).  That would require bucket types to always be strings (to avoid crashing the resource), which I think is the case.

src/riak_kv_wm_object.erl
```diff
@@ -685,10 +698,10 @@ process_post(RD, Ctx) -> accept_doc_body(RD, Ctx).
 %% @doc Store the data the client is PUTing in the document.
 %%      This function translates the headers and body of the HTTP request
 %%      into their final riak_object() form, and executes the Riak put.
-accept_doc_body(RD, Ctx=#ctx{bucket=B, key=K, client=C, links=L, index_fields=IF}) ->
-    Doc0 = case Ctx#ctx.doc of
+accept_doc_body(RD, Ctx=#ctx{bucket_type=T, bucket=B, key=K, client=C, links=L, index_fields=IF}) ->
+    Doc0 = case {Ctx#ctx.doc, T} of
```
engelsanchez: I haven't unraveled all the logic here, but certainly going from matching Ctx#ctx.doc to {ok, D} to matching {Ctx#ctx.doc, T} to {ok, D} has got to be wrong, right?

src/riak_kv_wm_object.erl
jrwest: there are no restrictions on the names of bucket types besides that they be represented as binaries fwiw

src/riak_kv_wm_utils.erl
```diff
@@ -180,6 +184,17 @@ format_uri(Bucket, Key, Prefix, 1) ->
...
 
+
+%% @doc Format the URI for a type/bucket/key correctly for the api version
+%% used. (APIVersion is the final parameter.)
+format_uri(_Type, Bucket, Key, Prefix, 1) ->
+    format_uri(Bucket, Key, Prefix, 1);
+format_uri(_Type, Bucket, Key, Prefix, 2) ->
+    format_uri(Bucket, Key, Prefix, 2);
+format_uri(Type, Bucket, Key, _Prefix, 3) ->
+    io_lib:format("/types/~s/buckets/~s/keys/~s", [Type, Bucket, Key]).
```
engelsanchez, line 195: Notice that we are using ~s on the bucket type (as well as bucket and key here). So using it in the error message would be consistent or both could go wrong in the same way.

src/riak_kv_wm_object.erl
seancribbs: You're right, that must have been an artifact of a previous pass.

seancribbs: Addressed in 1d6e639

src/riak_kv_wm_object.erl
seancribbs: One potential hole I just realized is that bucket types from the URL will most likely be strings. I agree about the `~s` though.

seancribbs: n/m re: binaries, `riak_kv_wm_utils:ensure_bucket_type` handles that.

seancribbs: Addressed in 2dec0b1

engelsanchez:
A general note: the WM resource modules have a lot of good documentation at the top. It might be worth keeping it up to date and add a note about the use of /types/<bucket type> where it applies.

src/riak_kv_wm_bucket_type.erl
```diff
@@ -0,0 +1,273 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_wm_bucket_type: Webmachine resource for bucket type properties
+%%
+%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
```
engelsanchez: Well this module is obsolete already! :)

src/riak_kv_wm_bucket_type.erl
```diff
@@ -0,0 +1,273 @@
...
+%%
+%% GET /types/Type/props
+%%   Modify bucket-type properties.
+%%   Content-type must be application/json, and the body must have
+%%   the form:
+%%     {"props":{Prop:Val}}
+%%   Where the "props" object takes the same form as returned from
+%%   a GET of the same resource.
+%%
+%% DELETE /types/Bucket/props
```
engelsanchez: Bucket -> Type here, right?

src/riak_kv_wm_bucket_type.erl
```diff
@@ -0,0 +1,273 @@
...
+%%     {"props":{Prop1:Val1,Prop2:Val2,...}}
+%%   Each bucket-type property will be included in the "props" object.
+%%   "linkfun" and "chash_keyfun" properties will be encoded as
+%%   JSON objects of the form:
+%%     {"mod":ModuleName,
+%%      "fun":FunctionName}
+%%   Where ModuleName and FunctionName are each strings representing
+%%   a module and function.
+%%
+%% GET /types/Type/props
```
engelsanchez: GET -> PUT/POST here?

src/riak_kv_wm_bucket_type.erl
```diff
@@ -0,0 +1,273 @@
...
+        {error, Details} ->
+            JSON = mochijson2:encode(Details),
+            RD2 = wrq:append_to_resp_body(JSON, RD),
+            {{halt, 400}, RD2, Ctx}
+    end.
+
+%% @spec delete_resource(reqdata(), context()) -> {boolean, reqdata(), context()}
+%% @doc Reset the bucket properties back to the default values
+delete_resource(RD, Ctx=#ctx{bucket_type=T}) ->
+    ok = riak_core_bucket_type:reset(T),
```
engelsanchez: It seems possible to get an error here by following the reset code, so maybe handle it and report it?

seancribbs: If `delete_resource` returns `false`, you get a 500 anyway. What sort of error should we report?

engelsanchez: So crashing here on a bad pattern match wouldn't kill the process and the connection, or are you saying that returning false would do the same so it doesn't matter?

seancribbs: It will not kill the connection process, Webmachine will catch it. However, a more useful error message would help, so I'll go ahead and do that.

src/riak_kv_web.erl
engelsanchez, line 58: It's fine, let's go with paranoia for now.

engelsanchez:
I'm done reviewing this for the moment.  I've ran the test and checked coverage. We can do better and I will personally be involved in increasing our coverage when the candidate build is out, so this is good for now. If you can address the documentation problems from my last few comments, I think we are basically +1.

src/riak_kv_wm_bucket_type.erl
seancribbs: Changed in cb83715 to print a more useful error.

engelsanchez:
Re :+1:

seancribbs:
:dancer: woot

PR #695, opened 2013-10-09
refactored in riak_kv bits from riak.schema with tests

joedevivo:
Here you'll find all the kv bits of the schema, and a unit test that validates them.


priv/riak_kv.schema
```diff
@@ -0,0 +1,440 @@
...
+}.
+
+%% {mapping, "buckets.default.basic_quorum", "riak_kv.default_bucket_props.basic_quorum", false},
+%% {mapping, "buckets.default.notfound_ok", "riak_kv.default_bucket_props.notfound_ok", true}
+
+%% @doc whether or not siblings are allowed.
+%% Note: See Vector Clocks for a discussion of sibling resolution.
+{mapping, "buckets.default.siblings", "riak_kv.default_bucket_props.allow_mult", [
+  {datatype, {enum, [on, off]}},
+  {default, off}
```
jrwest, line 140: this won't be THAT PR? ;). are we not changing the default or did you not want to shave that yak just yet?

jrwest:
I believe this is the case but want to confirm:

With the multi-backend schema, if a user has their own backend (e.g. riak_cs_multi_backend) it can still be set in the app.config/advanced.config but not in riak.conf?

priv/riak_kv.schema
joedevivo, line 140: ha! I give you the option to set the default. to change the default is on someone else :)

joedevivo:
Yes, that's correct. But, I do plan on adding the CS backend in the near future.

jrwest:
Speaking of the multi-backend. I mentioned in https://github.com/basho/eleveldb/pull/73 that I'm not a huge fan of having parts of the multi-backend schema in eleveldb (and bitcask). It doesn't have to be this PR, but given the tests and parts of the schema are already in `riak_kv` and and multi-backend is a riak_kv thing"" I think we should discuss moving those parts back here.

joedevivo:
I am open to that discussion. Over :beers: at ricon?

priv/riak_kv.schema
```diff
@@ -0,0 +1,440 @@
...
+%% are decoded, but buckets and keys are not. The compatibility
+%% mode will be removed in a future release.
+{mapping, "http_url_encoding", "riak_kv.http_url_encoding", [
+  {default, on},
+  {datatype, {enum, [on, off]}}
+]}.
+
+%% @doc Switch to vnode-based vclocks rather than client ids.  This
+%% significantly reduces the number of vclock entries.
+%% Only set on if *all* nodes in the cluster are upgraded to 1.0
```
jrwest, line 362: don't think this comment needs to be here. 2.0 clusters won't support legacy gossip so you can't join a pre-1.0 node to it. please correct me if I am wrong cc/ @rzezeski @jonmeredith (specifically leaving mr. tuple off the mention list since he asked not be bothered today).

priv/riak_kv.schema
```diff
@@ -0,0 +1,440 @@
...
+        end
+    end
+  end
+}.
+
+%% {mapping, "buckets.default.basic_quorum", "riak_kv.default_bucket_props.basic_quorum", false},
+%% {mapping, "buckets.default.notfound_ok", "riak_kv.default_bucket_props.notfound_ok", true}
+
+%% @doc whether or not siblings are allowed.
+%% Note: See Vector Clocks for a discussion of sibling resolution.
```
jrwest, line 137: should there be a link here?

priv/riak_kv.schema
```diff
@@ -0,0 +1,440 @@
...
+      yessir -> riak_kv_yessir_backend;
+      multi -> riak_kv_multi_backend;
+      _Default -> riak_kv_bitcask_backend
+    end
+  end}.
+
+%% @doc raw_name is the first part of all URLS used by the Riak raw HTTP
+%% interface.  See riak_web.erl and raw_http_resource.erl for
+%% details.
+{mapping, "raw_name", "riak_kv.raw_name", [
```
jrwest, line 210: I wonder if it wouldn't be good to group the names logically in the config, would cuttlefish support something like:

```
names.raw = ...
names.mapred = ...
```

with the same end result as its written now?

EDIT: modified my example, didn't mean to have the redundant `_name`

joedevivo, line 210: I remember something about tuning these being frowned upon. maybe @seancribbs has something to say here?

priv/riak_kv.schema
```diff
@@ -0,0 +1,440 @@
...
+      off -> false;
+      _Default -> true
+    end
+  end}.
+%% @doc Each of the following entries control how many Javascript
+%% virtual machines are available for executing map, reduce,
+%% pre- and post-commit hook functions.
+%% @datatype integer
+%% @mapping 
+{mapping, "javascript_vm.map_js_vm_count", "riak_kv.map_js_vm_count", [
```
jrwest: could this be shorter. "javascript_vm.map_count" or ".map_vm_count"? Same for similar javascript_vm settings.

joedevivo: yes, I like "javascript_vm.map_count". I'll push the change

jrwest:
minor comments here and there but everything looks good. tests pass w/ the addition of the `multi_backend.schema` file in https://github.com/basho/riak_kv/commit/983ad2df6696e44cb83952080eae9eda68a3b776. +1

jrwest:
my brain threw an async exception and realized after I gave the +1 that default bucket props are a riak_core thing [1]

[1] https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_props.erl#L57

priv/riak_kv.schema
rzezeski, line 362: I don't see what this has to do with legacy gossip.  Given that `riak.conf` is a 2.0 thing I see no sense in having a comment about all nodes being upgraded to 1.0.  This seems like a comment strictly dedicated to someone upgrade to 1.0.  It should have no place in 2.0.

priv/riak_kv.schema
seancribbs, line 210: Kill them with extreme prejudice, knobs that should never have existed.

Sean Cribbs

PR #696, opened 2013-10-09
Add 2i term regex filter

engelsanchez:
Adds a regular expression filter for secondary index terms. With this
you can scan a range for a binary index from, say A through Z, but only
use items that match terms containing "AA".
This filter does not work with integer queries, but it does work with
$key queries.

Regex compilation: currently it's done in the vnode, as different nodes
might have different versions of the regular expression library.
The PB and HTTP will compile the regular expression to validate it and
will return an error if it is invalid.

This PR requires:
- https://github.com/basho/riak_pb/pull/58 which adds the regex field
- https://github.com/basho/riak_test/pull/411 has some basic tests
- https://github.com/basho/riak-erlang-client/pull/122

### DIFF
--- a/include/riak_kv_index.hrl
+++ b/include/riak_kv_index.hrl
@@ -32,4 +32,16 @@
           return_body=false ::boolean() %% Note, only for riak cs bucket folds
          }).
 
--define(KV_INDEX_Q, #riak_kv_index_v2).
+-record(riak_kv_index_v3, {
+          start_key= <<>> :: binary(),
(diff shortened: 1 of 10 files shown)


slfritchie:
Pulling PRs out of the freezer ... the last 2 prereq PRs appear to be reviewed but not merged yet?  Is anything else missing?

engelsanchez:
Development of this feature was released in the 1.4 series, then ported back to 2.0 here: https://github.com/basho/riak_kv/pull/776

PR #697, opened 2013-10-10
Re-introduce the 1.4 counters API

russelldb:
Re-add the 1.4 API.

Support legacy counters at the FSM, Vnode, kv_crdt level.

I really think you don't need to review `riak_kv_gcounter` and `riak_kv_pncounter` as they are practically unchanged from 1.4 (The 2.0 API requires `update` returns {ok, Val} is the only change.)

The upgrade support is untested as upgrade seems to be broken. But running legacy clients against 2.0 has been tested (just run the riak_test for counters.)


include/riak_kv_types.hrl
```diff
@@ -5,6 +5,8 @@
 -define(CRDT_OP, #crdt_op).
 
 %% Top Level Key->Type Types
+-define(LEGACY_COUNTER_TYPE, riak_kv_pncounter).
+-define(LEGACY_COUNTER_TYPE(Val), #crdt{mod=?LEGACY_COUNTER_TYPE, ctype="application/riak_counter", value=Val}).
```
lenary: I'd prefer "V1_COUNTER_TYPE" just in case another type comes along.... but I'm relatively easy about it.

test/kv_crdt_eqc.erl
```diff
@@ -324,7 +324,7 @@ pncounterds() ->
 
 pncounter() ->
     ?LET(PNCounter, pncounterds(),
-         riak_kv_crdt:to_binary(?CRDT{mod=riak_dt_pncounter, value=PNCounter}, ?V1_VERS)).
+         riak_kv_crdt:to_binary(?CRDT{mod=riak_kv_pncounter, value=PNCounter}, ?V1_VERS)).
```
lenary: I think this should be `?LEGACY_COUNTER_TYPE`

src/riak_kv_web.erl
```diff
@@ -96,9 +96,9 @@ raw_dispatch(Name) ->
      {["buckets", bucket, "index", field, '*'],
       riak_kv_wm_index, Props2},
 
-     %% crdts
-     {["buckets", bucket, crdt, key], {riak_kv_wm_crdt, known_type},
-      riak_kv_wm_crdt, Props2}
+     %% counters
+     {["buckets", bucket, counters, key],
```
seancribbs: `counters` should be a string unless you intend to bind the value of that path segment.

src/riak_kv_gcounter.erl
```diff
@@ -0,0 +1,258 @@
...
+        false ->
+            merge(Rest, Clock2, [AC1|Acc])
+    end.
+
+%% @doc Are two `gcounter()'s structurally equal? This is not `value/1' equality.
+%% Two counters might represent the total `42', and not be `equal/2'. Equality here is
+%% that both counters contain the same actors and those actors have the same count.
+-spec equal(gcounter(), gcounter()) -> boolean().
+equal(VA,VB) ->
+    lists:sort(VA) =:= lists:sort(VB).
```
engelsanchez, line 109: It would be nice to add this method in the eunit test

engelsanchez, line 109: Ignore my previous comment. I was looking at riak_test coverage. Ignore. Ignore.

engelsanchez:
I noticed that the riak_test only exercises counters through the HTTP interface. Is there a test that uses the PB interface? If not, could we tweak the riak test to use it sometimes?

russelldb:
The riak_test verify_counter_capability uses the PB interface as does verify_dt_capability, but the latter fails due to upgrading to 2.0 failing.

russelldb:
verify_dt_converge uses the new PB interface. I guess you want a counter test that runs against 2.0 that uses the legacy 1.4 message? We'd have that if upgrade worked.

src/riak_kv_web.erl
engelsanchez: I suppose there will be some overlap with the URL configuration for this and upcoming HTTP API for CRDTs that @seancribbs and @russelldb have discussed? So for now going back to "counters" (instead of counters atom) should be enough?

src/riak_kv_crdt.erl
```diff
@@ -98,16 +98,22 @@ crdt_value(Type, {ok, {_Meta, ?CRDT{mod=Type, value=Value}}}) ->
...
+log_errors(_, _, []) ->
     ok;
-log_errors(RObj, Errors) ->
-    Bucket = riak_object:bucket(RObj),
-    Key = riak_object:key(RObj),
+log_errors(Bucket, Key, Errors) ->
     lager:info("Error(s) deserializing CRDT at ~p ~p: ~p~n", [Bucket, Key, Errors]).
 
+maybe_log_sibling_crdts(Bucket, Key, CRDTs) when length(CRDTs) > 1 ->
+    lager:info("Sibling CRDTs at ~p ~p: ~p~n", [Bucket, Key, orddict:fetch_keys(CRDTs)]);
```
engelsanchez: Should this be an error or warning? I think so. Also the above would work better as one too. I missed that before.

src/riak_kv_crdt.erl
```diff
@@ -148,12 +154,23 @@ deserialize_crdt(?V2_VERS, CRDTBin) ->
 deserialize_crdt(V, _Bin) ->
     {error, {invalid_version, V}}.
 
+counter_op(N) when N < 0 ->
+    {decrement, N*-1};
```
engelsanchez: Kind of a long winded way of writing -N yo

src/riak_kv_pb_counter.erl
```diff
@@ -39,8 +39,9 @@
 
 -module(riak_kv_pb_counter).
 
--include_lib("riak_pb/include/riak_kv_pb.hrl").
--include_lib("riak_pb/include/riak_pb_kv_codec.hrl").
+-include_lib("../../riak_pb/include/riak_kv_pb.hrl").
+-include_lib("../../riak_pb/include/riak_pb_kv_codec.hrl").
```
engelsanchez: Forgot to remove the ../../ bit?

russelldb: Again, bollocks.

src/riak_kv_pb_counter.erl
```diff
@@ -71,11 +69,11 @@ decode(Code, Bin) ->
     Msg = riak_pb_codec:decode(Code, Bin),
     %% no special permissions for counters, just get/put
     case Msg of
-        #rpbcountergetreq{type=T, bucket=B} ->
-            Bucket = bucket_type(T, B),
+        #rpbcountergetreq{bucket=B} ->
+            Bucket = bucket_type(B),
```
engelsanchez, line 73: Is there a reason to translate to {<<"default">>, Bucket} here? I thought all the code downstream from here could handle the old style, but I could be wrong.

engelsanchez:
I don't have anything else to add for the moment. Ping me when you address the existing comments.

src/riak_kv_pb_counter.erl
russelldb, line 73: @Vagabond  added it, I don't know the answer.

src/riak_kv_crdt.erl
russelldb: :blush:

src/riak_kv_pb_counter.erl
Vagabond, line 73: Either should be valid... I think... The 2-tuple form might be safer.

src/riak_kv_crdt.erl
russelldb: The second one is probably an error, as it means some kind of failure in our application of bucket types to CRDTs. The first one is tough to decide…it will happen if CRDT is serialized incorrectly or corrupt, but it will also happen if a users data as a binary _happens_ to start with the series of bytes that a CRDT does. It would be a big  coincidence I guess, but since EQC found a few…I dunno…pick one, you're the reviewer :smile:

russelldb:
When you "existing comments" you want the riak_tests writing too? If not, ping @engelsanchez

russelldb:
Sorry for the hefty refactor on riak_kv_pb_crdt, but adding another error (bucket type not found) just made the case nesting on the two `process` funs another level deep.

src/riak_kv_crdt.erl
engelsanchez: I get to pick? oh, oh. Well I think it's better to log it as an error then. If someone is using non-CRDT data as CRDT and they hit the serialization magic numbers, that is errorish enough right there.

engelsanchez:
@russelldb would you mind merging from develop? There are non-trivial merge conflicts in riak_kv_web.erl and riak_kv_pb_crdt.erl (around the section you refactored heavily).

engelsanchez:
for the merge: in riak_kv_web.erl the counters descriptor has to be reworked now that Sean has added descriptors for URLs including bucket types (which counters don't need). And in riak_kv_pb_crdt.erl, in develop they were added the retry_put_coordinator_failure request option set to false.

russelldb:
OK, did the merges.

I know you said maybe remove the roundtrip 1.4->2.0->1.4 stuff for counters, but I think this poosh answers the questions. If not then I will just cut it out for now. Going to work on a roundtrip test in the meantime.

russelldb:
Also, related https://github.com/basho/riak_test/pull/418 (verify_crdt_capability)

src/riak_kv_pb_crdt.erl
```diff
@@ -164,6 +108,112 @@ process_stream(_,_,State) ->
...
+make_options(#dtupdatereq{w=W0, dw=DW0, pw=PW0,
+                          timeout=Timeout, sloppy_quorum=SloppyQ,
+                          n_val=NVal, return_body=RetVal}) ->
+    W = decode_quorum(W0),
+    DW = decode_quorum(DW0),
+    PW = decode_quorum(PW0),
+    make_option(w, W) ++
+        make_option(dw, DW) ++
+        make_option(pw, PW) ++
+        make_option(timeout, timeout(Timeout)) ++
```
engelsanchez, line 213: @russelldb the merge missed adding the retry_put_coordinator_failure option present in develop, added by 0449970

russelldb, line 213: it is on line 168, it is an exceptional option, and not something passed in (we mandate false for crdts, since counters are not idempotent we want users to choose about retrying.) However, good point if you mention that the 1.4 adapter stuff doesn't do it…I mean if 2.0 counters opt out of fsm retry, 1.4 counters should too (I consider this something Scott missed but I should add it to this PR)

src/riak_kv_web.erl
```diff
@@ -104,6 +105,10 @@ raw_dispatch(Name) ->
      {Prefix ++ ["buckets", bucket, "index", field, '*'],
       riak_kv_wm_index, Props},
 
+     %% counters
+     {["buckets", bucket, "counters", key],
+      riak_kv_wm_counter, Props},
+
```
engelsanchez, line 115: Notice that this now adds the entry for counters twice, as this section is inside a list comprehension for two entries: one with types prefix and one without. I don't know if that has any side effects. It might be OK. But ideally this should be outside of it, adding only one entry.

russelldb, line 115: it is? Where is the other entry?

engelsanchez, line 115: Notice that this section is now within the list comprehension that ends here: https://github.com/basho/riak_kv/blob/ebf28d94cf7de8d78e3214cad5a5dc528efc22e3/src/riak_kv_web.erl#L115

russelldb, line 115: Oh I see. Ugly.

russelldb, line 115: I hope adding it twice is fine, I don't know though though, more and more I'm realising functional programming and no side effects aren't really the same thing. What can I say, what test we have pass.

russelldb, line 115: So riak_kv_web is basically wrong with regards to counters. There is no v3 or v1 api for counters, only v2. I guess it only passed review because I removed counters for ages to make 2.0 dev easier. More fool me.

engelsanchez, line 115: These are just properties passed to the module in the webmachine route. So, if the counters module doesn't care about these api or other properties, they are just ignored. Meh.

src/riak_kv_pb_crdt.erl
```diff
@@ -215,3 +272,79 @@ get_context(Ctx, true) ->
...
+        {true, {increment, Amt}} ->
+            Amt
+    end.
+
+%% delegate the legacy counter requests to the legacy pb service
+%% module
+process_legacy_counter({error, Reason}, State) ->
+    {error, {format, Reason}, State};
+process_legacy_counter(Req, State) ->
+    LegacyState = riak_kv_pb_counter:init(),
```
engelsanchez, line 330: This works fine here. It does create a riak_client on every request, which maybe we can optimize later.

engelsanchez:
Ok, I finally got a setup with all relevant tests passing (with 1.3.2 as legacy and 1.4.2 as previous). Sorry for the delay but I managed to completely botch my riak_test setup while preparing for that.  I think at this point the only remaining issue is the FSM retry flag for counters which you wanted to address here.

src/riak_kv_web.erl
russelldb, line 115: Looking at https://github.com/basho/riak_kv/blob/ebf28d94cf7de8d78e3214cad5a5dc528efc22e3/src/riak_kv_web.erl#L113 is it wrong? I thought CRDTs cannot be stored in the empty bucket type, that there is no default path for them @seancribbs?

russelldb:
Addressed comments cc @engelsanchez. I've pinged @seancribbs about that prefix + prefixless path for CRDTs.

seancribbs:
The way you have it is fine. We should remove the other resource's dispatch line until it's ready, as such it has not been tested adequately and we haven't thoroughly designed the API.

russelldb:
Removed!

engelsanchez:
Latest changes look good to me, relevant riak tests still pass and coverage is good.

:+1: :boat: :dancer:

Vagabond:
I think this PR forgot to add security checks to the WM interface.

russelldb:
Bum. Will start a new branch / PR for that. Thanks @Vagabond

PR #698, opened 2013-10-10
Coerce the datatype property into an atom.

seancribbs:
When creating a bucket-type from the command-line, the `datatype`
property might not be an atom, so we should try to convert it to one
before passing it to the CRDT code.


jrwest:
shouldn't the datatype coercion be handled in [`erlify_bucket_props`](https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_utils.erl#L343)? This is used by both wm and the command line bucket type stuff.

seancribbs:
@jrwest Addressed in d5b2326

jrwest:
+1. didn't actually run the code that uses the datatype property, but the change is straightforward enough

seancribbs:
Thanks! /cc @broach

PR #704, opened 2013-10-17
Adjust guard to tolerate bucket types

Vagabond:


engelsanchez:
Notice how the version this forwards to has no guards for bucket/key. In fact none of the other versions have them, and there is no other function clause. It might be better to simply remove them. We are forwarding to other methods that are more of the authority on what buckets should be.

engelsanchez:
@Vagabond I'm OK with merging as is. Tested that bucket types are allowed now where before they were not in this function. I would prefer no guards here, but that's mostly a matter of style.

:+1: to merge

PR #705, opened 2013-10-17
Fix include path.

cmeiklejohn:
This resolves compilation errors on develop/riak_kv.

@engelsanchez @seancribbs


jaredmorrow:
Thanks, worksforme +1

PR #707, opened 2013-10-22
Re-add security to the wm_counters endpoint

Vagabond:
cc @russelldb @engelsanchez I think this got lost as part of the refactor over in #697


russelldb:
Looks good, thanks for re-adding it, sorry it got lost. :+1:

PR #708, opened 2013-10-22
Add validator support for data types 

russelldb:
Only allow `datatype' bucket property to be set on`create`.
Only allow`allow_mult=true`for`datatype`buckets.
Only allow`datatype` bucket to be a supported type.

What do we do if we _retire_ a type in a future version? I dunno.

riak datatype support requires a bucket type of `datatype`
and `allow_mult` set to `true`. This PR enforces those two properties. We take the presence of a `datatype' property
as indication that a bucket type is a special type, somewhere to
store CRDTs. I realise this somewhat undermines the reason for
bucket types (no magic names) but there has to be some way to
indicate intent, and that way is the "special" property name
`datatype`.

Since we don't ever want sibling CRDT types (though we can handle
them see `riak_kv_crdt`), `datatype` is an immutable property. Once
you create a bucket with a certain datatype you can't change
it. The `update` bucket type path enforces this. It doesn't
validate the type, since it assumes that was done at creation.

I've hand tested. I'm adding riak_tests. Hopefully a review can start here while I finish that.

ping @jrwest

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -225,8 +225,8 @@ validate_create_dt_props(Unvalidated0, Valid, Invalid) ->
     Unvalidated = lists:keydelete(allow_mult, 1, Unvalidated0),
     case allow_mult(Unvalidated0) of
         true ->
-            {Unvalidated0, [{allow_mult, true} | Valid], Invalid};
-        false ->
+            {Unvalidated, [{allow_mult, true} | Valid], Invalid};
+        _ ->
             Err = io_lib:format("Data Type buckets must be allow_mult=true", []),
             {Unvalidated, Valid, [{allow_mult, Err} | Invalid]}
     end.
@@ -258,8 +258,8 @@ validate_update_dt_props(New, Valid, Invalid) ->
             {Unvalidated, Valid, Invalid};
         true ->
             {Unvalidated, [{allow_mult, true} | Valid], Invalid};
-        false ->
-            {Unvalidated, Valid, [{allow_mult, "Cannot change datatype bucket to allow_mult=false"} | Invalid]}
+        _ ->
+            {Unvalidated, Valid, [{allow_mult, "Cannot change datatype bucket from allow_mult=true"} | Invalid]}
 end.
 
 %% @private just grab the allow_mult value if it exists
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -381,12 +381,16 @@ prop_merges() ->
             begin
                 Result={Good, _Bad} = validate(update, Bucket, Existing, New),
 
-                Expected = case {default_bucket(Bucket), has_allow_mult(New), has_datatype(Existing)} of
-                               {true, _, _} ->
+                Expected = case {default_bucket(Bucket), has_allow_mult(New), allow_mult(New), has_datatype(Existing)} of
+                               {true, _, Mult,  _} when Mult /= error ->
                                    merge(New, Existing);
-                               {false, true, false} ->
+                               {true, true, error, _} ->
+                                   merge([proplists:lookup(datatype, New)], lists:keydelete(allow_mult, 1, Existing));
+                               {false, true, Mult,  false} when Mult /= error ->
                                    merge([proplists:lookup(allow_mult, New)], Existing);
-                               {false, true, true} ->
+                               {false,true,error,false} ->
+                                   Existing;
+                               {false, true, _Mult, true} ->
                                    Existing
                            end,
 
@@ -449,7 +453,7 @@ gen_new(create) ->
          Defaults ++ Mult ++ Datatype).
 
 gen_allow_mult() ->
-    ?LET(Mult, bool(), [{allow_mult, Mult}]).
+    ?LET(Mult, frequency([{9, bool()}, {1, binary()}]), [{allow_mult, Mult}]).
 
 gen_datatype_property() ->
     ?LET(Datattype, oneof([gen_datatype(), notadatatype]), [{datatype, Datattype}]).
@@ -484,6 +488,10 @@ immutable(update, New, Existing, {_Good, Bad}) ->
         {undefined, false, _Meh} ->
             has_allow_mult(Bad);
         {_Datatype, false, _Datatype2} ->
+            has_allow_mult(Bad) andalso has_datatype(Bad);
+        {_Datatype, _, _Datatype} ->
+            has_allow_mult(Bad);
+        {_, _, _} ->
             has_allow_mult(Bad) andalso has_datatype(Bad)
     end.
 
@@ -498,7 +506,7 @@ only_create_if_valid({Good, Bad}, New) ->
                 false ->
                     has_datatype(Bad) andalso has_allow_mult(Good)
             end;
-        {Datatype, false} ->
+        {Datatype, _} ->
             case lists:member(riak_kv_crdt:to_mod(Datatype), ?V2_TOP_LEVEL_TYPES) of
                 true ->
                     has_allow_mult(Bad) andalso has_datatype(Good);
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -25,27 +25,119 @@
 
 -export([validate/4]).
 
--include_lib("eunit/include/eunit.hrl").
 -include("riak_kv_types.hrl").
+
+-ifdef(TEST).
+-ifdef(EQC).
+-compile([export_all]).
+-include_lib("eqc/include/eqc.hrl").
+-endif.
+-include_lib("eunit/include/eunit.hrl").
+-endif.
 
 -type prop() :: {PropName::atom(), PropValue::any()}.
 -type error() :: {PropName::atom(), ErrorReason::atom()}.
 -type props() :: [prop()].
 -type errors() :: [error()].
 
+%% @doc called by riak_core in a few places to ensure bucket
+%%  properties are sane. The arguments combinations have the following
+%%  meanings:-
+%%
+%% The first argument is the `Phase' of the bucket/bucket type
+%% mutation and can be either `create' or `update'.
+%%
+%% `create' always means that we are creating a new bucket type or
+%% updating an inactive bucket type. In the first case `Existing' is
+%% the atom `undefined', in the second it is a list of the valid
+%% properties returned from the first invocation of `validate/4'. The
+%% value of `Bucket' will only ever be a two-tuple of `{binary(),
+%% undefined}' for create, as it is only used on bucket types. The
+%% final argument `BucketProps' is a list of the properties the user
+%% provided for type creation merged with the default properties
+%% defined in `riak_core_bucket_type:defaults/0' The job of the
+%% function is to validate the given `BucketProps' and return a two
+%% tuple `{Good, Bad}' where the first element is the list of valid
+%% properties and the second a list of `error()' tuples. Riak_Core
+%% will store the `Good' list in metadata iif the `Bad' list is the
+%% empty list. It is worth noting that on `create' we must ignore the
+%% `Existing' argument altogether.
+%%
+%% `update' means that we are either updating a bucket type or a
+%% bucket. If `Bucket' is a `binary()' or a tuple `{binary(),
+%% binary()}' then, a bucket is being updated. If `bucket' is a two
+%% tuple of `{binary(), undefined}' then a bucket type is being
+%% updated. When `validate/4' is called with `update' as the phase
+%% then `Existing' will be the set of properties stored in metadata
+%% for this bucket (the set returned as `Good' from the `create'
+%% phase) and `BucketProps' will be ONLY the properties that user has
+%% supplied as those to update (note: update may mean adding new
+%% properties.) The job of `validate/4' in this case is to validate
+%% the new properties and return a complete set of bucket properties
+%% (ie the new properties merged with the existing propeties) in
+%% `Good', riak will then persist these `Good' properties, providing
+%% `Bad' is empty.
+%%
+%% `validate/4' can be used to enforce immutable or co-invariant bucket
+%% properties, like "only non-default bucket types can have a
+%% `datatype' property", and that "`datatype' buckets must be
+%% allow_mult" and "once set, `datatype' cannot be changed".
+%%
+%% There is no way to _remove_ a property
+%%
+%% @see validate_dt_props/3, assert_no_datatype/1
 -spec validate(create | update,
                {riak_core_bucket_type:bucket_type(), undefined | binary()} | binary(),
                undefined | props(),
                props()) -> {props(), errors()}.
-validate(CreateOrUpdate, {_T, _N}, Existing0, BucketProps) when is_list(BucketProps) ->
-    Existing = case Existing0 of
-                   undefined -> [];
-                   _ -> Existing0
-               end,
-    validate_dt_props(CreateOrUpdate, Existing, BucketProps);
-validate(_CreateOrUpdate, _Bucket, _Existing, BucketProps) when is_list(BucketProps) ->
-    validate(BucketProps, [], []).
+validate(create, _Bucket, _Existing, BucketProps) when is_list(BucketProps) ->
+    validate_create_bucket_type(BucketProps);
+validate(update, {_TypeName, undefined}, Existing, New) when is_list(Existing),
+                                                            is_list(New) ->
+    validate_update_bucket_type(Existing, New);
+validate(update, {Type, Name}, Existing, New) when is_list(Existing),
+                                                   is_list(New),
+                                                   is_binary(Name),
+                                                   Type /= <<"default">> ->
+    validate_update_typed_bucket(Existing, New);
+validate(update, _Bucket, Existing, New) when is_list(Existing),
+                                             is_list(New) ->
+    validate_default_bucket(Existing, New).
 
+%% @private bucket creation time validation
+-spec validate_create_bucket_type(props()) -> {props(), errors()}.
+validate_create_bucket_type(BucketProps) ->
+    {Unvalidated, Valid, Errors} = validate_create_dt_props(BucketProps),
+    validate(Unvalidated, Valid, Errors).
+
+%% @private update phase of bucket type. Merges properties from
+%% existing with valid new properties
+-spec validate_update_bucket_type(props(), props()) -> {props(), errors()}.
+validate_update_bucket_type(Existing, New) ->
+    {Unvalidated, Valid, Errors} = validate_update_dt_props(Existing, New),
+    {Good, Bad} = validate(Unvalidated, Valid, Errors),
+    {merge(Good, Existing), Bad}.
+
+%% @private just delegates, but I added it to illustrate the many
+%% possible type of validation.
+-spec validate_update_typed_bucket(props(), props()) -> {props(), errors()}.
+validate_update_typed_bucket(Existing, New) ->
+    validate_update_bucket_type(Existing, New).
+
+%% @private as far as datatypes go, default buckets are free to do as
+%% they please, the datatypes API only works on typed buckets. Go
+%% wild!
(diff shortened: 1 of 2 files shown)


src/riak_kv_bucket.erl
```diff
@@ -36,8 +37,13 @@
                {riak_core_bucket_type:bucket_type(), undefined | binary()} | binary(),
                undefined | props(),
                props()) -> {props(), errors()}.
+validate(CreateOrUpdate, {_T, _N}, Existing0, BucketProps) when is_list(BucketProps) ->
```
jrwest: this needs something like `andalso T =/= <<"default">>` since default-type buckets may be expressed in the 2-tuple form. otherwise you can set `datatype` on buckets under the default-type.

src/riak_kv_bucket.erl
```diff
@@ -36,8 +37,13 @@
...
+    Existing = case Existing0 of
+                   undefined -> [];
+                   _ -> Existing0
+               end,
+    validate_dt_props(CreateOrUpdate, Existing, BucketProps);
 validate(_CreateOrUpdate, _Bucket, _Existing, BucketProps) when is_list(BucketProps) ->
-    lager:debug("Was called ~p~n", [BucketProps]),
```
jrwest, line 40: shouldn't we validate here that there is not a `datatype` (or the `datatype` is a counter) present in the properties? Otherwise you can set `datatype` on buckets under the default-type.

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+validate_dt_props(create, Existing, Props) ->
+    validate_data_type(proplists:get_value(datatype, Props), Existing, Props);
+validate_dt_props(update, Existing, Props) ->
+    case proplists:get_value(datatype, Existing) of
+        undefined ->
+            validate_dt_props(create, Existing, Props);
+        DType ->
+            case proplists:get_value(datatype, Props) of
+                undefined ->
+                    validate_allow_mult(Existing, Props, [], []);
```
jrwest: this allows me to create a type without a `datatype` and then set and change the`datatype` on a bucket underneath the created type. we shouldn't do that because there is no creation invariant on bucket properties under any type, correct?

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+                DType ->
+                    validate_allow_mult(Existing, Props, [{datatype, DType}], []);
+                Other ->
+                    Err = lists:flatten(io_lib:format("Cannot change datatype from ~p to ~p", [DType, Other])),
```
jrwest: maybe include something like "...because bucket type <type> is active"?

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+validate_data_type(undefined, _Existing, Props) ->
+    validate(Props, [], []);
+validate_data_type(DType, Existing, Props0) ->
+    Props = lists:keydelete(datatype, 1, Props0),
+    Mod = riak_kv_crdt:to_mod(DType),
+    case lists:member(Mod, ?V2_TOP_LEVEL_TYPES) of
+        true ->
+            validate_allow_mult(Existing, Props, [{datatype, DType}], []);
+        false ->
+            Err = lists:flatten(io_lib:format("~p not supported for bucket datatype property", [DType])),
```
jrwest: it would be nice to make this a bit clearer since its user facing. perhaps print the list of supported `datatype` values and maybe shorten the sentence to "<datatype> is not supported"?

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+            validate_allow_mult(Existing, Props, [], [{datatype, Err}])
+    end.
+
+allow_mult(true, undefined) ->
```
jrwest: add a comment about what this is validating? Its a bit unclear unless you look at the rest of the code that `false, true` will never be passed because of other validation.

src/riak_kv_bucket.erl
```diff
@@ -36,8 +37,13 @@
                {riak_core_bucket_type:bucket_type(), undefined | binary()} | binary(),
                undefined | props(),
                props()) -> {props(), errors()}.
+validate(CreateOrUpdate, {_T, _N}, Existing0, BucketProps) when is_list(BucketProps) ->
+    Existing = case Existing0 of
+                   undefined -> [];
+                   _ -> Existing0
+               end,
+    validate_dt_props(CreateOrUpdate, Existing, BucketProps);
```
jrwest: the mutual recursion between these two functions is a bit hard to follow. perhaps just return from `validate_dt_props` and then call into `validate`?

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+                Other ->
+                    Err = lists:flatten(io_lib:format("Cannot change datatype from ~p to ~p", [DType, Other])),
+                    validate_allow_mult(Existing, Props, [], [{datatype, Err}])
+            end
+    end.
+
+validate_allow_mult(Existing, Props, Valid, Errors) ->
+    case allow_mult(allow_mult(Existing), allow_mult(Props)) of
+        true ->
+            validate(lists:keydelete(allow_mult, 1, Props), [{allow_mult, true} | Valid], Errors);
```
jrwest: this `lists:keydelete` causes a problem because the props returned from `validate` are the ones stored. if I change `riak_core_bucket_type:defaults/0` [1] to `{allow_mult, false}` then I can create a bucket-type with a `datatype` but `allow_mult=false`:

```
$ dev/dev1/bin/riak-admin bucket-type create crdt1 '{"props": {"datatype": "counter"}}'
...
$ dev/dev1/bin/riak-admin bucket-type status crdt1 | grep "datatype\|allow_mult"
datatype: counter
allow_mult: false
```

Was the intention to not validate `allow_mult` again? If I change this line to `validate(Props, Valid, Errors)` and leave the change to `riak_core_bucket_type:defaults/0` then I get a validate error (as expected) when running the creation command above but if I specify `allow_mult: true` in the creation command everything still works fine.

[1] https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_type.erl#L129

jrwest:
Something I noticed while testing:

```
$ dev/dev1/bin/riak-admin bucket-type create crdt1 '{"props": {"allow_mult": 1, "datatype": "dne"}}'
                                 {badarg,
                                  [{erlang,binary_to_existing_atom,
                                    [<<"dne">>,utf8],
                                    []},
                                   {riak_kv_wm_utils,erlify_bucket_prop,1,
                                    [{file,"src/riak_kv_wm_utils.erl"},
                                     {line,360}]},
                                   {riak_kv_console,
                                    '-bucket_type_create/2-lc$^0/1-0-',1,
                                     {line,483}]},
                                   {riak_kv_console,
                                    '-bucket_type_create/2-lc$^0/1-0-',1,
                                     {line,483}]},
                                   {riak_kv_console,bucket_type_create,2,
                                     {line,483}]},
```

Its not exactly pretty and caused by https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_utils.erl#L393 but I'm fine punting for now. 

We don't really get to the `datatype` validation because of this but its easy to test by creating the atom.

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+        false ->
+            Err = io_lib:format("Data Type buckets must be allow_mult=true", []),
+            validate(lists:keydelete(allow_mult, 1,  Props), Valid, [{allow_mult, Err} | Errors])
+    end.
+
+validate_data_type(undefined, _Existing, Props) ->
+    validate(Props, [], []);
+validate_data_type(DType, Existing, Props0) ->
+    Props = lists:keydelete(datatype, 1, Props0),
+    Mod = riak_kv_crdt:to_mod(DType),
```
jrwest: `DType` here is an atom because of `erlify_bucket_prop` [1] so using`"sets"` and `"counters"` as the `datatype` value does not work. Only ones in the `?MOD_MAP` in `riak_kv_crdt:to_mod/1` do. Modify that function to work on the plural atom versions as well (or add special handling to `erlify_bucket_prop` [2]) maybe?

[1] https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_utils.erl#L393
[2] https://github.com/basho/riak_kv/pull/708#discussion_r7143687

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
 coerce_bool(_) ->
     error.
 
+%% @private riak datatype support requires a bucket type of `datatype'
+%% and `allow_mult' set to `true'. This function enforces those
+%% _before_ calling the usual validate code, since the two properties
+%% are interdependant. We take the presence of a `datatype' property
+%% as indication that this bucket type is a special type, somewhere to
+%% store CRDTs. I realise this somewhat undermines the reason for
```
jrwest: what's being undermined? you don't require anything from the `name` property just your own `datatype` property.

src/riak_kv_bucket.erl
```diff
@@ -82,6 +88,77 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+%% bucket types (no magic names) but there has to be some way to
+%% indicate intent, and that way is the "special" property name
+%% `datatype'.
+%%
+%% Since we don't ever want sibling CRDT types (though we can handle
+%% them @see riak_kv_crdt), `datatype' is an immutable property. Once
+%% you create a bucket with a certain datatype you can't change
+%% it. The `update' bucket type path enforces this. It doesn't
+%% validate the type, since it assumes that was done at creation.
```
jrwest: it validates the type hasn't changed

src/riak_kv_bucket.erl
jrwest: sorry, the comment above is actually intended for L112 where we can `valdiate_dt_props(create, Existing, Props)` if there is no existing type.

jrwest: however, looking at this line now, if the `datatype` doesn't exist in `Existing` or `Props` why are we validating `allow_mult` in the `validat_dt_*` path? Shouldn't we mutually recurse (or preferably return) at this point?

src/riak_kv_bucket.erl
russelldb: Magic names. Users may have buckets with the property called `datatype` already.

src/riak_kv_bucket.erl
russelldb: It doesn't validate the type, just that it hasn't changed.

src/riak_kv_bucket.erl
russelldb: I think I just don't understand the _meaning_ of the arguments to 'validate/4'. There's no explanation of them that I can find so I had to guess. I need to start again, it seems, understanding what values to `validate/4` mean what. What combination of arguments means a bucket's properties are being set vs. which combination mean a bucket _type_ is being created / updated?

src/riak_kv_bucket.erl
russelldb: I don't get it. Why should `"sets"` and `"counters"` work? They're not valid datatypes.

src/riak_kv_bucket.erl
russelldb: but the function doesn't care about `false, true` and that is clear from the matching in the head `_, true`

src/riak_kv_bucket.erl
russelldb: Is that a problem? so `datatype` is now a reserved word in riak? How is this different to magic prefixes?

src/riak_kv_bucket.erl
russelldb: Is this a subtle change in how validate works? It also has to perform the merge between default and passed properties now?

russelldb: In the past, validate was given the _new_ properties and returned that list split into two, `{valid, errors}` now your saying it returns the merged set of new valid and existing properties to be persisted? Don't you think that is rather a large change not to mention until now? That means this PR has to totally fix riak_kv_bucket, doesn't it? How did the changes to validate/4 from validate/1 ever get past review?

src/riak_kv_bucket.erl
```diff
@@ -82,6 +180,98 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+
+%% @private validate the boolean property, if `datatype' was present,
+%% require `allow_mult=true' even if `datatype' was invalid, as we
+%% assume the user meant to create `datatype' bucket
+-spec validate_create_dt_props(props(), props(), errors()) -> {props(), props(), errors()}.
+validate_create_dt_props(Unvalidated0, Valid, Invalid) ->
+    Unvalidated = lists:keydelete(allow_mult, 1, Unvalidated0),
+    case allow_mult(Unvalidated0) of
+        true ->
+            {Unvalidated0, [{allow_mult, true} | Valid], Invalid};
```
jrwest: The first element of the tuple should be `Unvalidated`. Otherwise we end up w/ two `allow_mult` entries in the resulting list.

src/riak_kv_bucket.erl
```diff
@@ -82,6 +180,98 @@ coerce_bool(MaybeBool) when is_list(MaybeBool) ->
...
+        false ->
```
jrwest: I think we should change this to a catchall, `_`, because invalid `allow_mult` values like `"abc"` will error w/ no matching clause.

jrwest:
I ran through the same by hand tests I did previously and only found the two minor issues that I commented on. Tests pass including the new EQC but I have yet to read it in detail. I'll do that while waiting for the minor changes, or @russelldb if you prefer, for the sake of time, I can make the changes myself.

jrwest:
For completeness, my one comment about the `binary_to_existing_atom` issue still applies, but its not specific to `datatype` or this PR, so we can move on.

src/riak_kv_bucket.erl
russelldb: Yup, good catch.

src/riak_kv_bucket.erl
russelldb: Did you manage to do that? I thought `allow_mult/1` used `coerce_bool/1`? The only way this could be anything other than `true` or `false` is if the defaults merged with new passed to create did not have `allow_mult`, you think I should guard against that? I think a crash would be preferable…though maybe matching `undefined` would give us a better clue if that ever occurred?

jrwest: Yea its easy to hit because even if we do reach `coerce_bool` we wont until after this codepath:

```
$ dev/dev1/bin/riak-admin bucket-type create crdt5 '{"props": {"datatype": "set", "allow_mult": "abc"}}'                                                                                                                                rdb-bucket-type-validator/riak | develop
                                 {{{case_clause,error},
                                   [{riak_kv_bucket,validate_create_dt_props,
                                     3,
                                     [{file,"src/riak_kv_bucket.erl"},
                                      {line,226}]},
                                    {riak_kv_bucket,
                                     validate_create_bucket_type,1,
                                     [{file,"src/riak_kv_bucket.erl"},
                                      {line,110}]},
                                    {riak_core_bucket_props,validate,6,
                                     [{file,"src/riak_core_bucket_props.erl"},
                                      {line,51}]},
                                    {riak_core_claimant,handle_call,3,
                                     [{file,"src/riak_core_claimant.erl"},
                                      {line,279}]},
                                    {gen_server,handle_msg,5,
                                     [{file,"gen_server.erl"},{line,585}]},
                                    {proc_lib,init_p_do_apply,3,
                                     [{file,"proc_lib.erl"},{line,239}]}]},
                                  {gen_server,call,
                                   [{riak_core_claimant,'dev1@127.0.0.1'},
                                    {create_bucket_type,<<"crdt5">>,
                                     [{allow_mult,<<"abc">>},
                                      {basic_quorum,false},
                                      {big_vclock,50},
                                      {chash_keyfun,
                                       {riak_core_util,chash_std_keyfun}},
                                      {datatype,set},
                                      {dw,quorum},
                                      {last_write_wins,false},
                                      {linkfun,
                                       {modfun,riak_kv_wm_link_walker,
                                        mapreduce_linkfun}},
                                      {n_val,3},
                                      {notfound_ok,true},
                                      {old_vclock,86400},
                                      {postcommit,[]},
                                      {pr,0},
                                      {precommit,[]},
                                      {pw,0},
                                      {r,quorum},
                                      {rw,quorum},
                                      {small_vclock,50},
                                      {w,quorum},
                                      {young_vclock,20}]},
                                    infinity]}}}
```

jrwest: I changed the implementation locally to use a catchall. Here is the result of running my previous example, i think the error message is sufficient. Thoughts?

```
$ dev/dev1/bin/riak-admin bucket-type create crdt5 '{"props": {"datatype": "set", "allow_mult": "abc"}}'                                                                                                                                rdb-bucket-type-validator/riak | develop
Error creating bucket type crdt5: [{allow_mult,
                                       "Data Type buckets must be allow_mult=true"}]
```

russelldb: But it _does_ hit `coerce_bool/1`but that returns `error` ,just need a case clause for that, anything else would hide boogs.

russelldb: Infact, maybe I should validate the normal properties first…man, pretty deep hole

russelldb: Yeah, fuck it, that will do. Nice catch.

russelldb: Same at 261 then?

russelldb: Yeah

```
riak@127.0.0.1)3> riak_core_bucket_type:update(<<"jets">>, [{allow_mult, "boygan"}]).
** exception exit: {{{case_clause,error},
                 [{riak_kv_bucket,validate_update_dt_props,3,
                                  [{file,"src/riak_kv_bucket.erl"},{line,256}]},
                  {riak_kv_bucket,validate_update_bucket_type,2,
                                  [{file,"src/riak_kv_bucket.erl"},{line,117}]},
                  {riak_core_bucket_props,validate,6,
                                          [{file,"src/riak_core_bucket_props.erl"},{line,51}]},
                  {riak_core_claimant,handle_call,3,
                                      [{file,"src/riak_core_claimant.erl"},{line,291}]},
                  {gen_server,handle_msg,5,
                              [{file,"gen_server.erl"},{line,585}]},
                  {proc_lib,init_p_do_apply,3,
                            [{file,"proc_lib.erl"},{line,239}]}]},
                {gen_server,call,
                            [{riak_core_claimant,'riak@127.0.0.1'},
                             {update_bucket_type,<<"jets">>,[{allow_mult,"boygan"}]}]}}
```

Same catch all fixes. I reckon the eqc needs work, then.

russelldb:
Pushed. Thanks for your thoroughness and patience on this PR.

jrwest:
Changes look good. I tested the cases I was hitting and they no longer crash, as expected. Read through the EQC test a bit more. I agree it could be expanded to cover some of the cases we just found but its still a good test. I have some code comments for the test but those are not relevant right now.

+1 merge. Nice work @russelldb. Thanks for working through it and putting up w/ the not so sane API that is `validate/4` + documenting it. I think the end result is much cleaner too :).

russelldb:
One final push for non-boolean allow_mult generation in the EQC. Please raise issue(s) for outstanding review comments that you don't want to hold up the PR for (or mail me the list of comments and I'll raise issues, I realise this has eaten a bunch of valuable time for you)

jrwest:
+1 again, EQC still passes and verified that it covers the case you fixed. @russelldb I didn't open any specific issues because I really need to sit down and spend more time with the test before I do, which I will (after we merge).

PR #709, opened 2013-10-24
Remove ability for clients to modify bucket types.

seancribbs:
See also basho/riak_pb#62 and basho/riak_api#42.

### DIFF
--- a/src/riak_kv_wm_bucket_type.erl
+++ b/src/riak_kv_wm_bucket_type.erl
@@ -34,17 +34,6 @@
 %%      "fun":FunctionName}
 %%   Where ModuleName and FunctionName are each strings representing
 %%   a module and function.
-%%
-%% POST|PUT /types/Type/props
-%%   Modify bucket-type properties.
-%%   Content-type must be application/json, and the body must have
-%%   the form:
-%%     {"props":{Prop:Val}}
-%%   Where the "props" object takes the same form as returned from
-%%   a GET of the same resource.
-%%
-%% DELETE /types/Type/props
-%%   Reset bucket-type properties back to the default settings
 
 -module(riak_kv_wm_bucket_type).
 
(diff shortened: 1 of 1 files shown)


jrwest:
I was chatting w/ @rzezeski about this yesterday and he made my opinion waver a bit. We want to remove these APIs because we won't expose them in the client unless we can do so for the full set of bucket type operations (we are currently missing create and activate). However, the reason for not exposing certain ones is because Security. In some cases a user may care more about programmatic use of these operations over security and we can give them that option by writing APIs for the missing operations but only allowing them to succeed when security is disabled (I think that would be safe,  @Vagabond?). Curious to get other people's thoughts on this.

Vagabond:
Yes, we could tie it to some non-existant permission or something.

I don't mind doing it like this as long as I don't have to implement ACL management over the riak client API (at least not yet).

lafka:
@jrwest can you elaborate on the potential security issues of exposing API for creating and updating bucket types?

I would like to have the ability to update bucket type properties programmatic, same as I would like to have the ability to update the properties of individual buckets.

seancribbs:
@jrwest @Vagabond @rzezeski I know we've been over-and-over this topic, but I'd love to discuss again post-RICON.

jrwest:
@lafka i'm going to defer to @Vagabond on the details. We are discussing supporting the API when security is disabled for now.

@seancribbs happy to chat about it whenever.

slfritchie:
Pulling PRs out of the freezer ... this doesn't have a milestone assigned to it?  Is that by design?

seancribbs:
@slfritchie We will have to decide this by 2.0, it is contentious and should be one-way-or-t'other. @jrwest @Vagabond shall we reopen this debate?

jrwest:
Yea it would be great to re-open it and get something in the API w/o security on in my opinion. I guess there is just the matter of time...What do you think @seancribbs @Vagabond?

seancribbs:
@jrwest Let's plan to discuss soon. Monday work?

jrwest:
works for me

seancribbs:
Had a mumble about it today, the consensus was:
1. Remove the "reset bucket type" operation from the API
2. Omit some of the functionality from client libraries until the API is complete with all features.

As a result, I'm closing this PR and its buddies in favor of new ones that accomplish point 1.

PR #710, opened 2013-10-24
Support for strongly consistent Riak

jtuple:
This is the `riak_kv` related pull-request for on-going work to add strong consistency to Riak.

Primarily, this pull-request integrates with [riak_ensemble](http://github.com/basho/riak_ensemble) that provides most of the heavy lifting.

This pull-request is provides a `riak_ensemble_backend` implementation that integrates with K/V vnodes, registers ensembles (consensus groups) for each distinct replica set (preflist/n-val) in Riak, and maps the existing client API onto `riak_ensemble_client` against these new K/V ensembles.

---

Objects written under a bucket type with property `consistent = true` are consistent objects and use the consistent code path, everything else uses the normal Riak code path.

Consistent operations are conditional single key atomic updates:
- If you write a key, the next successful read is guaranteed to show that write (or the result of a future write that saw the write).
- If you get/modify/put an object, Riak ensures that the object didn't change since you read it. In other words, the request will fail if a concurrent write occurred and changed the object.
- In general, timeouts are indistinguishable from partial writes in a distributed system. Thus, consistent Riak resolves partial write ambiguity at read time. If you try to write a value, and that write times out, the current state of the key is unknown -- it could be the prior value or the attempted to write value. However, the next successful read of the key tells you which occurred and ensures that decision does not change in the future. 
- For example, if you partially write to 1 replica which then becomes partitioned/goes offline. Subsequently, perform a successful read that shows the old value. It's guaranteed the old value wins. The partitioned failed--to-write value is essentially rolled back. Even once that replica comes back online and propagates the failed-to-write value, the old value still wins. Of course, in a slightly different scenario, it's entirely possible for the initial read-after-failed-write to return the new value instead. Again, either option is possible on the next read. But, after that read, things do not change.

Internally, `riak_ensemble` supports three types of writes: `put_once`, `modify`, and `overwrite`.
- A `put_once` operation only succeeds if the key doesn't already exist.
- A `modify` operation is used as the `put` in a get/modify/put sequence. A `modify` operation fails the write if the object no longer matches the base object returned in the `get` part of the sequence.
- A `overwrite` operation is a fast, unsafe, unchecked operation. `overwrite` operations overwrite the key without regard to it's existing value. This is useful when bulk loading data.

For Riak, the existing K/V API is mapped to a subset of these operations. Client puts that are missing a vector clock are mapped to `put_once` operations, while those that provide a vector clock are mapped to `modify`. There is currently no mapping to `overwrite`.

A previous version of this code mapped operations w/o vector clocks to `overwrite`, and puts with `if_none_matched` to `put_once`, but it is too easy for a client to accidentally not send vector clock data and mapping that to `overwrite` made it too easy for users to clobber data. Hence the safer mapping above.

Delete operations are similar. Riak client deletes map to a unchecked (similar to overwrite) ensemble delete. Whereas client delete_vclock operations map to safe deletes that are similar to `modify`. Specifically, the safe variant is used in the sequence: `Obj = get(), delete(Obj)` where Riak fails the delete if the object has changed since the get.

Note: All mentions of "vector clock" above are slightly different than normal Riak. Consider them opaque contexts passed back/forth to the client. In general, vector clocks for consistent data aren't really the same as with normal Riak operations. We don't main normal vector clocks at all. The consistent subsystem maintains it's own logical clock per ensemble (eg. per preflist) which is used for all consistent operations. However, for compatibility this value is converted into a "vector clock" thus the object and client requests formats stays the same.

---

Known limitations/issues (TODO: file issues)
- Consistent data does not currently support secondary indexes.
- Consistent data does not currently support object mutators.
- Consistent operations do not yet fire any stats-related events, nor keep any stats of their own.
- Consistent operations currently bypass the overload protection system and need their own overload protection to be added.
- Deleting consistent data leaves around tombstones that are not currently harvested. 
- Consistent reads currently do not use read repair, nor does normal AAE currently heal missing consistent data.
- The consensus subsystem (riak_ensemble) requires implementing modules to provide a `sync` behavior that heals potentially diverted/corrupted replicas. The plan is to use AAE for this in `riak_kv`. However, this currently is not done and will be done in a future pull-request.

Sibling pull-requests: basho/riak_core#441, basho/riak_ensemble#1

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -64,17 +64,7 @@
 get(Bucket, Key) ->
     get(Bucket, Key, []).
 
-%% @spec get(riak_object:bucket(), riak_object:key(), options()) ->
-%%       {ok, riak_object:riak_object()} |
-%%       {error, notfound} |
-%%       {error, {deleted, vclock()}} |
-%%       {error, timeout} |
-%%       {error, {n_val_violation, N::integer()}} |
-%%       {error, {r_val_unsatisfied, R::integer(), Replies::integer()}} |
-%%       {error, Err :: term()}
-%% @doc Fetch the object at Bucket/Key.  Return a value as soon as R-value for the nodes
-%%      have responded with a value or error.
-get(Bucket, Key, Options) when is_list(Options) ->
+normal_get(Bucket, Key, Options) ->    
     Me = self(),
     ReqId = mk_reqid(),
     case node() of
@@ -86,7 +76,41 @@ get(Bucket, Key, Options) when is_list(Options) ->
     end,
     %% TODO: Investigate adding a monitor here and eliminating the timeout.
     Timeout = recv_timeout(Options),
-    wait_for_reqid(ReqId, Timeout);
+    wait_for_reqid(ReqId, Timeout).
+
+consistent_get(Bucket, Key, Options) ->
+    BKey = {Bucket, Key},
+    Ensemble = ensemble(BKey),
+    Timeout = recv_timeout(Options),
+    case riak_ensemble_client:kget(Node, Ensemble, BKey, Timeout) of
+        {error, _}=Err ->
+            Err;
+        {ok, Obj} ->
+            case riak_object:get_value(Obj) of
+                notfound ->
+                    {error, notfound};
+                _ ->
+                    {ok, Obj}
+            end
+    end.
+
+%% @spec get(riak_object:bucket(), riak_object:key(), options()) ->
+%%       {ok, riak_object:riak_object()} |
+%%       {error, notfound} |
+%%       {error, {deleted, vclock()}} |
+%%       {error, timeout} |
+%%       {error, {n_val_violation, N::integer()}} |
+%%       {error, {r_val_unsatisfied, R::integer(), Replies::integer()}} |
+%%       {error, Err :: term()}
+%% @doc Fetch the object at Bucket/Key.  Return a value as soon as R-value for the nodes
+%%      have responded with a value or error.
+get(Bucket, Key, Options) when is_list(Options) ->
+    case riak_kv_util:consistent_object(Bucket) of
+        true ->
+            consistent_get(Bucket, Key, Options);
+        false ->
+            normal_get(Bucket, Key, Options)
+    end;
 
 %% @spec get(riak_object:bucket(), riak_object:key(), R :: integer()) ->
 %%       {ok, riak_object:riak_object()} |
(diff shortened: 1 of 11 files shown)


src/riak_client.erl
```diff
@@ -70,17 +70,7 @@ new(Node, ClientId) ->
...
-%%       {error, notfound} |
-%%       {error, {deleted, vclock()}} |
-%%       {error, timeout} |
-%%       {error, {n_val_violation, N::integer()}} |
-%%       {error, {r_val_unsatisfied, R::integer(), Replies::integer()}} |
-%%       {error, Err :: term()}
-%% @doc Fetch the object at Bucket/Key.  Return a value as soon as R-value for the nodes
-%%      have responded with a value or error.
-get(Bucket, Key, Options, {?MODULE, [Node, _ClientId]}) when is_list(Options) ->
+normal_get(Bucket, Key, Options, {?MODULE, [Node, _ClientId]}) ->
```
evanmcc, line 73: Instead of normal/consistent condsider switching to ec_get/sc_get (ic_ for immediate?)

src/riak_object.erl
```diff
@@ -120,6 +122,11 @@ new_int(B, K, V, MD) ->
             end
     end.
 
+is_robject(#r_object{}) ->
+    true;
+is_robject(_) ->
```
evanmcc, line 127: should this handle the binary encoded robject case?

src/riak_client.erl
```diff
@@ -172,7 +187,67 @@ put(RObj, Options, {?MODULE, [Node, ClientId]}) when is_list(Options) ->
...
+%%       {ok, riak_object:riak_object()} |
+%%       {ok, riak_object:riak_object(), details()} |
+%%       {error, notfound} |
+%%       {error, timeout} |
+%%       {error, {n_val_violation, N::integer()}} |
+%%       {error, Err :: term()}
+%%       {error, Err :: term(), details()}
+%% @doc Store RObj in the cluster.
+put(RObj, Options, {?MODULE, [Node, _ClientId]}=THIS) when is_list(Options) ->
+    case consistent_object(Node, riak_object:bucket(RObj)) of
```
evanmcc, line 243: this gets called twice on each object on the put path, once in the frontend _object files and once here.

evanmcc:
Code looks ok with an eye towards not affecting anything when disabled.  I have some perf concerns, but those can be addressed later.  Biggest worry here is that there aren't any/get/put checks for ensembles being disabled.  Might be a good idea to deny at least puts and possibly gets from consistent buckets when ensembles are disabled to avoid polluting the buckets with objects that the ensemble code isn't able to handle.

evanmcc:
note that this code won't actually build unless the ensemble backend behavior is defined, which it isn't on riak_ensemble/develop

evanmcc:
+1 for tech-preview merge, will open an issue or two.

jtuple:
As discussed internally, going ahead and merging with the understanding that strong consistency and the consensus subsystem are disabled by default. This allows us to ship an optionally-enabled beta release of strong consistency as part of the upcoming tech preview.

A full review of basho/riak_ensemble#1 is forth coming, and will involve a more detailed review of the code paths in this PR that are hit when the consensus system is enabled.

src/riak_kv_ensembles.erl
```diff
@@ -0,0 +1,127 @@
+%% -------------------------------------------------------------------
```
sumerman, line 1: Seemingly this entire module except the ensemble and peer IDs prefix (`kv` here), and the backend module name, will be copy-pasted to every ensabmles-based app. Probably it should be moved to the ensembles app with above mentioned parameters exposed as arguments for the `start_link`. BTW peer/ensemble id may be provided via backend module's non-obligatory callbacks.

PR #711, opened 2013-10-25
Add hook to Yokozuna for handoff

rzezeski:
Yokozuna needs to hook into KV handoff in order to behave in a
predictable manner during ownership change. Yokozuna doesn't have
vnodes. Instead it piggy-backs the KV vnode. When an object is written
Yokozuna's index hook is called so that Solr may be updated. When
ownership change happens this hook will be executed for writes
arriving at the new node because they are writes like any
other. However in order for successful indexing the indexes and bucket
properties must arrive at the new node first.

If the indexes don't arrive before the ownership handoff objects then
those objects will fail to index and cause incorrect query results for
the user. If the bucket properties don't arrive before then the index
hook won't be invoked at all and index entries will again be
missing. Currently Yokozuna uses some tricky code to detect a
handoff-based write and do a busy wait for the index to appear. This
is undesirable because it blocks the vnode. It also doesn't work for
the new bucket properties based on cluster metadata since they are no
longer stored in the ring.

Rather than create more hacks this patch adds a handoff hook to
Yokozuna. This way Yokozuna can make sure the indexes and properties
are sync'd before KV is allowed to send its data. This removes the
race on the new node and will allow removal of the busy-wait hack
currently in the vnode write path.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -89,6 +89,12 @@
 -define(INDEX(Obj, Reason, Partition), yz_kv:index(Obj, Reason, Partition)).
 -endif.
 
+-ifdef(TEST).
+-define(YZ_SHOULD_HANDOFF(X), true).
+-else.
+-define(YZ_SHOULD_HANDOFF(X), yz_kv:should_handoff(X)).
+-endif.
+
 -record(mrjob, {cachekey :: term(),
                 bkey :: term(),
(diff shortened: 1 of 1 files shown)


rzezeski:
Corresponding Yokozuna PR: basho/yokozuna#230.

coderoshi:
+1. basho/yokozuna#230 tests pass with this patch

PR #712, opened 2013-10-25
Use the deep equality check for riak_objects.

cmeiklejohn:
Since some of the objects stored in the r_object are items like dicts,
etc where the internal representation can change; ensure that we perform
the deep equality check.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1179,7 +1179,7 @@ put_merge(false, true, _CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=
     {newobj, UpdObj};
 put_merge(false, false, CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=false
     ResObj = riak_object:syntactic_merge(CurObj, UpdObj),
-    case ResObj =:= CurObj of
+    case riak_object:equal(ResObj, CurObj) of
         true ->
             {oldobj, CurObj};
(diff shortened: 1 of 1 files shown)


evanmcc:
lgtm but haven't had time to test.

vinoski:
+1, looks good.

russelldb:
I don't understand this change…it looks like it might be expensive…why is it needed?

russelldb:
So, syntactic merge does the deep equals test already, since the thing returned from syntactic merge is either EXACTLY the object that went in, or it is changed I think the old =:= was correct.

cmeiklejohn:
screen shot 2013-10-25 at 11 42 09 am

Since we have lack of history here, I'll update what's happening.

The metadata dictionary is implemented as a tree structure, with changes structure based on insertion order of elements, so two objects with the same value can generate siblings when they are equal unless we perform the actual deep equality check of the objects.  This is extremely problematic with a default of allow_mult=true.

This is most likely a huge performance hit in riak_kv, but produces correct behavior.

evanmcc:
Note that I've seen the convo in Core/KV, so just following up on that + my
thoughts when reviewing:

I signed off on it without doing a lot of review of context because
although it might be expensive, I had a plan to rework this entire code
path for 2.0 (since syntactic merge is already so expensive) and I thought
that correctness was more important for the preview.

It's good to hear that there are other (potentially simpler) optimizations
incoming, though.

PR #714, opened 2013-10-28
Correctly enocde the update response when return_body is true

russelldb:


src/riak_kv_pb_crdt.erl
```diff
@@ -39,9 +39,9 @@
 
 -module(riak_kv_pb_crdt).
 
--include_lib("riak_pb/include/riak_dt_pb.hrl").
--include_lib("riak_pb/include/riak_kv_pb.hrl").
--include_lib("riak_pb/include/riak_pb_kv_codec.hrl").
+-include_lib("../../riak_pb/include/riak_dt_pb.hrl").
```
seancribbs: :-1: use rebar deps_dir=..

russelldb: Addressed. I've never got to the bottom of why my flymake will not find those includes any other way. I often forget to revert that change. :(

src/riak_kv_pb_crdt.erl
```diff
@@ -340,9 +335,7 @@ upgrade_response(_, {reply, #rpbcountergetresp{value=Val}, State}) ->
 upgrade_response(#rpbcounterupdatereq{}, {reply, #rpbcounterupdateresp{}, State}) ->
     {reply, #dtupdateresp{}, State};
 upgrade_response(_, {reply, #rpbcounterupdateresp{value=Value}, State}) ->
-    {reply, update_resp(riak_kv_crdt:from_mod(?COUNTER_TYPE),
-                        Value,
-                        undefined), State};
+    {reply, #dtupdateresp{counter_value=Value}, State};
```
seancribbs, line 338: Why doesn't this use `encode_update_response`?

russelldb, line 338: No idea, I guess it doesn't need to.

seancribbs, line 338: You're right, sorry my mistake.

russelldb:
Maybe @engelsanchez could have a sneaky peeky?

seancribbs:
Sorry for the latent :+1:. It was the use of `Type` instead of `Mod` that was the major breakage, but it's also nice to cleanup.

PR #715, opened 2013-10-29
2i term regex filter 1.4

engelsanchez:
Backport the 2i term regex filter to the 1.4 series.

For convenience, the feature/2i-regex-filter-1.4 branch exists in top riak and other repos so this can be built and run with the right dependencies.

/cc @russelldb

### DIFF
--- a/src/riak_index.erl
+++ b/src/riak_index.erl
@@ -32,9 +32,8 @@
          format_failure_reason/1,
          normalize_index_field/1,
          timestamp/0,
+         to_index_query/1,
          to_index_query/2,
-         to_index_query/3,
-         to_index_query/6,
          make_continuation/1,
          return_terms/2,
          return_body/1,
@@ -52,6 +51,7 @@
 -define(TIMEOUT, 30000).
 -define(BUCKETFIELD, <<"$bucket">>).
 -define(KEYFIELD, <<"$key">>).
+-define(CURRENT_2I_VERSION, v3).
 
 %% See GH610, this default is for backwards compat, so 2i behaves as
 %% it did before the FSM timeout bug was "fixed"
@@ -73,6 +73,7 @@
 -type key() :: binary().
 -opaque continuation() :: binary(). %% encoded last_result().
 
+-type query_version() :: v1 | v2 | v3.
 mapred_index(Dest, Args) ->
     mapred_index(Dest, Args, ?TIMEOUT).
 mapred_index(_Pipe, [Bucket, Query], Timeout) ->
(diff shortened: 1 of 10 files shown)


engelsanchez:
Related PRs:

basho/riak_pb#58
basho/riak_test#411
basho/riak-erlang-client#122

src/riak_kv_vnode.erl
```diff
@@ -678,6 +679,13 @@ handle_coverage(?KV_INDEX_REQ{bucket=Bucket,
     handle_coverage_index(Bucket, ItemFilter, Query,
                           FilterVNodes, Sender, State, fun result_fun_ack/2).
 
+prepare_index_query(#riak_kv_index_v3{term_regex=RE} = Q) when
+        RE =/= undefined ->
+    {ok, CompiledRE} = re:compile(RE),
```
russelldb, line 684: Why when you compile it at the API entry points to validate?

russelldb:
Is there anything to be done for MapReduce?

src/riak_index.erl
```diff
@@ -243,89 +244,89 @@ timestamp() ->
...
+            {F, Default} <-
+            [{field, Defaults?KV_INDEX_Q.filter_field},
+             {start_term, Defaults?KV_INDEX_Q.start_term},
+             {start_inclusive, Defaults?KV_INDEX_Q.start_inclusive},
+             {end_term, Defaults?KV_INDEX_Q.end_term},
+             {end_inclusive, Defaults?KV_INDEX_Q.end_inclusive},
+             {term_regex, Defaults?KV_INDEX_Q.term_regex},
+             {return_body, Defaults?KV_INDEX_Q.return_body},
+             {continuation, undefined}]],
+    End = case End1 of undefined -> Start; _ -> End1 end,
```
russelldb, line 267: I don't like this style, but it is up to you. Reminds me of ternary operators in curly brace languages.

russelldb:
Two minor comments, and the MapReduce question, apart from that I'm +1

russelldb:
Also curious about perf costs, but only a bit.

src/riak_kv_vnode.erl
engelsanchez, line 684: Rolling upgrade paranoia: what if you are updating from one version of Riak to another that has the same feature, but the re module compiles to a slightly different format? (say, changes between Erlang releases) Then you would send that to a node and it would probably crash over there. Since it's opaque, I tried to make no assumptions.  It kind of sucks. If you have a good argument to switch to a single re compilation, I'm all ears. It would be nice to do it only once.

src/riak_index.erl
engelsanchez, line 267: Sorry, that's just me cursing and working around vimerl's indentation style and trying to avoid indenting manually and/or super long lines (if the ] = is moved up, the next section is moved all the way to the right too).  I haven't found a way to make it indent like Emacs :(

engelsanchez:
About MapReduce: I wasn't planning in allowing this filter in MR just yet. We could, but I guess it can wait until somebody asks for it.  On the other hand, I have not run the MR test suite on this to make sure that it doesn't mess with it. That would be a good thing to do.

src/riak_kv_vnode.erl
russelldb, line 684: I have no better alternative now I understand your thinking. Tempted to ask for a comment explaining just that here, though.

russelldb:
Will run MapReduce suite when I get back on later. Assuming it passes, :+1:

russelldb:
Wow, GitHub.

src/riak_kv_vnode.erl
engelsanchez, line 684: That should be explained in the code. I had convinced myself that there was a note somewhere, but it was just my memory tricking me into not doing any more work. I'll update.

russelldb:
Yarp. MapReduce passes. :+1:

PR #671, opened 2013-10-30
Streaming 2i Query Performance Penalty in 1.4

engelsanchez:
In 1.4.2, the streaming merge sort which enables pagination creates a large performance hit to queries that might not require it (streaming). 

Making the sort optional on non-paginated 2i queries would provide greatly increased performance on queries that don't require deterministic ordering.

This issue is particularly visible when returning multi-million key results. 1-10million

### DIFF
--- a/src/riak_kv_index_fsm.erl
+++ b/src/riak_kv_index_fsm.erl
@@ -51,7 +51,8 @@
 -type req_id() :: non_neg_integer().
 
 -record(state, {from :: from(),
-                merge_sort_buffer :: sms:sms(),
+                order :: sorted | unsorted,
+                merge_sort_buffer = undefined :: sms:sms() | undefined,
                 max_results :: all | pos_integer(),
                 results_per_vnode = dict:new() :: dict(),
                 results_sent = 0 :: non_neg_integer()}).
@@ -87,16 +88,27 @@ init(From={_, _, _}, [Bucket, ItemFilter, Query, Timeout]) ->
     %% atom() > number()
     init(From, [Bucket, ItemFilter, Query, Timeout, all]);
 init(From={_, _, _}, [Bucket, ItemFilter, Query, Timeout, MaxResults]) ->
+    init(From, [Bucket, ItemFilter, Query, Timeout, MaxResults, undefined]);
+init(From={_, _, _}, [Bucket, ItemFilter, Query, Timeout, MaxResults, Order0]) ->
     %% Get the bucket n_val for use in creating a coverage plan
     BucketProps = riak_core_bucket:get_bucket(Bucket),
     NVal = proplists:get_value(n_val, BucketProps),
+    Order = case is_integer(MaxResults) andalso MaxResults > 0 of
+        true -> sorted;
+        false -> case Order0 of
+                undefined -> unsorted;
+                _ -> Order0
+            end
+    end,
     %% Construct the key listing request
     Req = req(Bucket, ItemFilter, Query),
     {Req, all, NVal, 1, riak_kv, riak_kv_vnode_master, Timeout,
-     #state{from=From, max_results=MaxResults}}.
+     #state{from=From, max_results=MaxResults, order=Order}}.
 
-plan(CoverageVNodes, State) ->
-    {ok, State#state{merge_sort_buffer=sms:new(CoverageVNodes)}}.
+plan(CoverageVNodes, State = #state{order=sorted}) ->
+    {ok, State#state{merge_sort_buffer=sms:new(CoverageVNodes)}};
+plan(_CoverageVNodes, State) ->
+    {ok, State}.
 
 process_results(_VNode, {error, Reason}, _State) ->
     {error, Reason};
@@ -105,7 +117,7 @@ process_results(_VNode, {From, _Bucket, _Results}, State=#state{max_results=X, r
     {done, State};
 process_results(VNode, {From, Bucket, Results}, State) ->
     case process_results(VNode, {Bucket, Results}, State) of
-        {ok, State2} ->
+        {ok, State2 = #state{order=sorted}} ->
             #state{results_per_vnode=PerNode, max_results=MaxResults} = State2,
             VNodeCount = dict:fetch(VNode, PerNode),
             case VNodeCount < MaxResults of
@@ -116,11 +128,14 @@ process_results(VNode, {From, Bucket, Results}, State) ->
                     riak_kv_vnode:stop_fold(From),
                     {done, State2}
             end;
+        {ok, State2} ->
+            riak_kv_vnode:ack_keys(From),
+            {ok, State2};
         {done, State2} ->
             riak_kv_vnode:stop_fold(From),
             {done, State2}
     end;
-process_results(VNode, {_Bucket, Results}, State) ->
+process_results(VNode, {_Bucket, Results}, State = #state{order=sorted}) ->
     #state{merge_sort_buffer=MergeSortBuffer, results_per_vnode=PerNode,
            from={raw, ReqId, ClientPid}, results_sent=ResultsSent, max_results=MaxResults} = State,
     %% add new results to buffer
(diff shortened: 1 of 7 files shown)


reiddraper:
@engelsanchez the default is sorting _false_ now?

engelsanchez:
@reiddraper I've done a round of cleanup. The default should now only come from cuttlefish.  Since either default will  surprise some customers, I've gone ahead and made it configurable. That way the worst case (people surprised and unable to update their software quickly) has a workaround.

reiddraper:
We've decided to defer the 'default' decision a bit. It'll get merged in with 'false' as the default for now, and will just focus on review.

engelsanchez:
The comment about sort order not documented has been tweaked to remove that bit as suggested by Reid offline. This comment will likely be subject to change when we have the default value discussion.

src/riak_kv_index_fsm.erl
```diff
@@ -87,16 +88,29 @@ init(From={_, _, _}, [Bucket, ItemFilter, Query, Timeout]) ->
...
+init(From={_, _, _}, [Bucket, ItemFilter, Query, Timeout, MaxResults, Sort0]) ->
     %% Get the bucket n_val for use in creating a coverage plan
     BucketProps = riak_core_bucket:get_bucket(Bucket),
     NVal = proplists:get_value(n_val, BucketProps),
+    Paginating = is_integer(MaxResults) andalso MaxResults > 0, 
+    Sort = case {Paginating, Sort0} of
+        {true, _} ->
+            true;
+        {_, undefined} ->
+            app_helper:get_env(riak_kv, secondary_index_sort_default);
```
reiddraper, line 101: This is kind of pedantic, but I'd prefer us to grab the default out in `riak_client`, and not have to worry about threading an `undefined` very far. Up to you on this one.

src/riak_kv_wm_index.erl
```diff
@@ -153,18 +157,33 @@ malformed_request(RD, Ctx) ->
     Args2 = [list_to_binary(riak_kv_wm_utils:maybe_decode_uri(RD, X)) || X <- Args1],
     ReturnTerms0 = wrq:get_qs_value(?Q_2I_RETURNTERMS, "false", RD),
     ReturnTerms = normalize_boolean(string:to_lower(ReturnTerms0)),
+    Sort0 = wrq:get_qs_value(?Q_2I_SORT, RD),
+    Sort = case Sort0 of
+        undefined -> undefined;
```
reiddraper, line 162: re: my earlier comment, it might even be better to grab the default even further out (further from riak_client), and make riak_client require a boolean. We'd grab the default in webmachine then.

src/riak_kv_index_fsm.erl
engelsanchez, line 101: I'm not feeling strongly about this one, but it's not hard to imagine the FSM getting started by something other than the riak client in the future. See how riak_kv_index_pipe skips the client altogether to do its 2i query. If you are going to pick one and one only place to define a default, the single place where it is actually used seems more pure to me.

src/riak_kv_wm_index.erl
engelsanchez, line 162: I think you and I have very different interpretations of DRY :). Then do the same in PB and have to maintain both paths? And remember to add it to any new path we create later? Yes, we could centralize it in some method in riak_index and make sure that every entry point calls the same code. It just seems like strange dance to me.

reiddraper, line 162: > I think you and I have very different interpretations of DRY :). 

Nope, I just hadn't considered it being called from webmachine and pb. What's the argument for not putting the 'sorted' as part of the `riak_kv_index` record?

> Yes, we could centralize it in some method in riak_index and make sure that every entry point calls the same code.

Isn't that just what an API is? I'm imagining we'd just create an extra arity riak_index:to_index_query, and have the others grab the default from the app config. As I mentioned before, none of this is a big deal, but I'm enjoying the design discussion :).

engelsanchez, line 162: The problem is that it's all leaky right now. to_index_query is concerned with the internal query sent to each vnode, mostly (range parameters). The other parameters are scattered (bucket, max results, return terms, etc). Ideally, we'd have this thin translation of HTTP/PB messages to this single '2i query' message to our internal API. That would check syntax at that layer, leaving the internal API to check semantics, fill in defaults, etc. So our internal API would cleanly be this single point of reference we want. Heck, we could even have a dumb transformer to avoid all this HTTP/PB code duplication... Might be worth opening an issue to looking into a through cleaning of this.

reiddraper, line 162: Cool, I'm in complete agreement with that. Let's leave it as a separate issue for now, as you suggest.

reiddraper:
+1, but would you mind filing an issue to write a simple riak-test?

engelsanchez:
@reiddraper I'm open to suggestions about the riak test: I haven't figured out a way to verify this in a proper way. If sorted is not specified, results _shouldn't_ be sorted, but heck, they might be.

reiddraper:
> If sorted is not specified, results shouldn't be sorted, but heck, they might be.

Yeah, I wondered about that too. We can always start off by only testing that overriding the default works, which is that adding sort=true _does_ sort the results. Better than no test.

engelsanchez:
Agreed. I'm preparing that part of the test now. No reason to wait for later. It would also be worth to add a check that if the default is overriden via config, it sticks (again, only on the positive side unless someone comes up with something clever).

reiddraper:
Cool, sounds like a solid plan to me. Thanks.

engelsanchez:
@reiddraper did you realize that the order in which objects are returned is not by key, but by (term, key)? Somehow I wasn't thinking too much about that, but of course that is what's needed for pagination. This default order is really not that useful unless you use return_terms=true. I thought that might help in the default value discussion. But maybe this was clear in your mind already.

engelsanchez:
The secondary index tests were spiced up to exercise this: https://github.com/basho/riak_test/pull/444

reiddraper:
> did you realize that the order in which objects are returned is not by key, but by (term, key)? Somehow I wasn't thinking too much about that, but of course that is what's needed for pagination. This default order is really not that useful unless you use return_terms=true. I thought that might help in the default value discussion. But maybe this was clear in your mind already.

Yeah, though my bias has definitely been in thinking about the `$key` index. After reflecting on our discussion earlier, I I'm on board with making `false` the default for non-paginated requests. Let's just be clear on our documentation/messaging.

engelsanchez:
I know this is kind of closed, but to your point about messaging and documentation, I'm worried that this "sort" option is non-intuitive for all non-key/bucket, non-return_terms queries.  I suppose that as long as is documented as "sort by pagination order, that is by matching 2i term first, then object key" or something it should be OK.  If you have others thoughts about clarifying this in the code, I'm open to suggestions. Maybe pagination_sort is a better parameter name? That would force you to go "wat?" and check out the documentation to understand what it means, vs. seeing "sort" and assuming, like even I did momentarily, that it was key sort.   Coding is easy. Coding things that make sense to users is hard.

angrycub:
Since the pagination syntax has to sort on {term, index}, I think I would rather have a paginated sort default `return-terms` to true.  That way the most naive case makes sense when the data is returned -- otherwise the users are going to complain that their data isn't sorted, rather than looking at the docs.  If a more seasoned user wanted to disable terms, then we could allow them to do that with an explicit `return-terms = false` 

This might become it's own issue though, since this is a full mitigation of the performance issue.

reiddraper:
You guys are making good points, my biggest concern with changing the default to include the terms is that, I'm guessing, it will change the data-structure returned to them in the client. If this isn't the case, I'd be +1 on making it the default, but I don't want to break anyone's existing code.

reiddraper:
+1

slfritchie:
Pulling PRs out of the freezer ... wow, comment EOVERFLOW.  What remains for this ticket?

engelsanchez:
Development on this was backported and released in the 1.4 series, and the result of the backported to 2.0 here: https://github.com/basho/riak_kv/pull/776. 

Closing

PR #719, opened 2013-11-05
Avoid mutator code path when necessary.

cmeiklejohn:
Avoid the mutator code path completely when there are no registered
mutators, which prevents the mutators_applied item from being added to
the metadata dictionary when there are no mutators.

@Vagabond @engelsanchez @lordnull

### DIFF
--- a/src/riak_kv_mutator.erl
+++ b/src/riak_kv_mutator.erl
@@ -85,7 +85,7 @@
 
 -export([register/1, register/2, unregister/1]).
 -export([get/0]).
--export([mutate_put/2, mutate_get/1]).
+-export([mutate_put/3, mutate_put/2, mutate_get/1]).
 
 -callback mutate_put(Meta :: dict(), Value :: any(), ExposedMeta :: dict(), FullObject :: riak_object:riak_object(), BucketProps :: orddict:orddict()) -> {dict(), any(), dict()}.
 -callback mutate_get(FullObject :: riak_object:riak_object()) -> riak_object:riak_object() | 'notfound'.
@@ -183,6 +183,11 @@ mutate_get(Object, [Mutator | Tail]) ->
             mutate_get(Object2, Tail)
     end.
 
+-spec mutate_put(Object :: riak_object:riak_object(), BucketProps :: orddict:orddict()) -> {riak_object:riak_object(), riak_object:riak_object()}.
+mutate_put(Object, BucketProps) ->
+    {ok, Modules} = ?MODULE:get(),
+    mutate_put(Object, BucketProps, Modules).
+
 %% @doc Mutate an object in preparation to storage, returning a tuple of the
 %% object to store and the object to return to the client. For each sibling
 %% the object has {Meta, Value} pair, each mutator is called with a copy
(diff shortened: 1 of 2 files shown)


Vagabond:
Looks good, but have not run the code.

lordnull:
The mutator code path is still hit if the registered mutators list is of length 0. It might be nicer if the registered function returned [] in the case of undefined.

One could even contend that instead of putting that check in the kv code, put it in the mutator code and have mutate_put return early in the case of an empty mutator list.

cmeiklejohn:
Pushed an update which changes the function to `any_registered/0` which returns true or false.  I'd rather not call into mutate_put if we don't have to, which preserves existing behavior.

lordnull:
Eunit tests pass.

Ran this through the repl_reduced risk_test which fails the data_push test in that it creates a sibling. I doubt that's due to this change however.

I would like to have recorded more on the "why" of this. It's not saving a call to metadata (adding one for when there are mutators in fact). It's by-passing the resolver function and just using the metadata last write wins. If the goal was just to avoid adding metadata to the object if no mutators were configured, then I see reason to alter the behavior of the put in mutator; it's not like it's been shipped in a final release.

cmeiklejohn:
The motivation was to avoid a majority of the code in the `riak_kv_mutator` code path, in addition to not modifying the metadata dictionary with the `mutators_applied` entry when no mutators were registered.

lordnull:
I have two main issues that have not been addressed:
1. the existing resolver for getting mutators is ignored for no reason
2. for most enterprise customers, particularly those with repo, this will add a call to the metadata system to get the mutators; i.e. a second call to get the exact same data.

cmeiklejohn:
@lordnull Can you point me to what you mean by 1?

lordnull:
The exported `get` function: https://github.com/basho/riak_kv/blob/52e91f69f08bc8cd16beb4943480e8552336dff1/src/riak_kv_mutator.erl#L87
uses a resolver internally:
https://github.com/basho/riak_kv/blob/52e91f69f08bc8cd16beb4943480e8552336dff1/src/riak_kv_mutator.erl#L150

I have concerns about just using the default resolver for cluster metadata of 'last write wins', so I a merging resolver. If it's not useful or has potential issues in this case, some documentation stating why would be helpful.

cmeiklejohn:
Ah; I see what you mean now.  Thanks.

cmeiklejohn:
"...for most enterprise customers, particularly those with repo, this will add a call to the metadata system to get the mutators; i.e. a second call to get the exact same data."

True, but you could easily argue that for all OSS users, or rather _anyone_ using kv, not enterprise, have an additional call to a function that will perform no action, which is the larger and more common use case.

cmeiklejohn:
Amended commit with new strategy.  Thoughts?

lordnull:
Passes eunit, riak_tests that passed before continue to pass. :+1:

> True, but you could easily argue that for all OSS users, or rather anyone using kv, not enterprise, have an additional call to a function that will perform no action, which is the larger and more common use case.

So which is cheaper: a case statement or an external module function call? Since I don't know, and I'm pretty certain both are faster than querying for the the metadata (and resolving it), it's good enough to merge.

cmeiklejohn:
I'll let @evanmcc handle optimization when he gets to it.

PR #722, opened 2013-11-07
Improve the documentation for the storage_backend setting.

seancribbs:
Also cleanup that spurious 'default' option.

/cc @joedevivo @vinoski


joedevivo:
Cool. My favorite part is how there's now only one default backend! 

+1

vinoski:
+1 to that, and my favorite part is the removal of the trailing whitespace. We hates it, it burnses!

PR #726, opened 2013-11-13
Keep a per vnode counter and use it to create the "fresh" vclock

russelldb:
When a local "notfound" occurs on the get for a co-ord put, use
the vnode counter to create a fresh vclock and merge it with the
received objects vclock. See riak_kv 679 for details.

Also, needs https://github.com/basho/riak_core/pull/452

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -99,6 +99,12 @@
                 modstate :: term(),
                 mrjobs :: term(),
                 vnodeid :: undefined | binary(),
+                %% The number of writes co-ordinated by this vnode
+                counter :: non_neg_integer(),
+                %% How many co-ordinating writes before persisting the
+                %% counter to disk.
+                %% See config value `{riak_kv, counter_flush_threshold}'
+                counter_flush_threshold :: non_neg_integer(),
                 delete_mode :: keep | immediate | pos_integer(),
                 bucket_buf_size :: pos_integer(),
                 index_buf_size :: pos_integer(),
@@ -115,6 +121,8 @@
 -type state() :: #state{}.
 
 -define(DEFAULT_HASHTREE_TOKENS, 90).
+%% default value for `counter_flush_threshold' in `state'
+-define(DEFAULT_CNTR_THRESHOLD, 1000).
 
 -record(putargs, {returnbody :: boolean(),
                   coord:: boolean(),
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1738,14 +1775,34 @@ read_vnode_status(File) ->
...
+    flush_counter(Index, Counter, Threshold);
+maybe_flush_counter(_, _, _) ->
+    {ok, not_flushed}.
+
+flush_counter(Index, Counter, Threshold) ->
+    F = fun(Status) ->
+                {flushed, [{counter, {Counter, Threshold}} |
+                      proplists:delete(counter, Status)]}
+        end,
+    update_vnode_status(F, Index).
```
evanmcc, line 1805: `update_vnode_status` is potentially expensive and looks to be on the hot path (called from do_put).  Maybe do it async (and maybe bound how many times it'll try)?  If it's required we should consider writing an alternate path where we just keep the handle open and avoid expensive `filelib:ensure_dir()` calls every 1000 puts.

russelldb, line 1805: I am at your mercy here. I have no idea what is best. It _cannot_ be async. The riak_core_util code used to write the path is the code I was directed too as safe.

The threshold (1000) is still up for debate…finding the balance between performance and rapidly climbing counters. Though it seems we have 56bits for the count so 1000 seems kind of low.

What do you recommend? I'm not aware of the perf costs of this code…

engelsanchez, line 1805: I am wondering if this would show up in our perf smoke test (make perfdev && make perf) on the code with and without this change.  @evanmcc that should hit this path, right?

russelldb, line 1805: So, if we add the file handle to the state of the vnode (say), what happens if a user removes the file between flushes?

evanmcc, line 1805: It should, but without additional disk load (i.e. changing the backend to bitcask, say) and running on a system with a spinning disk, you may not see a true reflection of the actual performance effect.

evanmcc, line 1805: I think that you'd need to be able to handle it, open a new file, regenerate the status, and reflush.

russelldb, line 1805: Happy to do that based on your intuition.

engelsanchez, line 1805: Well, the whole idea of the perf smoke test is to show unexpected perf changes, not to be precise about the actual impact (which you can never be, really. It always depends). If it's not useful here, we should consider chucking it :)

russelldb:
You know, this just doesn't work. I have a test that shows that.

The problem is this. Node A coordinates a write or two for key K and they get replicated to B and C.
Allow_mult is true.
Node A loses it's data (for key K) (but keeps it's vnode id / counter.)
Node A coordinates a new write for K, the put has no vclock. It _should_ be concurrent with the old writes, but with this patch it gets a vclock of `{A, 1001}` (for example) and dominates those surviving replicas.

As it stands without this patch the scenario above loses the last PUT, with this patch it loses the last but one.

I wonder which is worse, since I can't think of a solution (beyond asking every node for the key.)

PR #727, opened 2013-11-14
moved some schema bits in from riak.schema

joedevivo:
I found some bits in the riak/rel/files/riak.schema that should be in here.

### DIFF
--- a/test/riak_kv_schema_tests.erl
+++ b/test/riak_kv_schema_tests.erl
@@ -43,6 +43,10 @@ basic_schema_test() ->
     cuttlefish_unit:assert_config(Config, "riak_kv.multi_backend", undefined),
 
     cuttlefish_unit:assert_config(Config, "riak_kv.secure_referer_check", true),
+    cuttlefish_unit:assert_config(Config, "riak_kv.warn_object_size", 5242880),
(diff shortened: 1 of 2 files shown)


test/riak_kv_schema_tests.erl
```diff
@@ -77,7 +81,11 @@ override_non_multi_backend_schema_test() ->
         {["object_format"], v0},
         {["memory_backend", "max_memory"], "8GB"},
         {["memory_backend", "ttl"], "1d"},
-        {["secure_referer_check"], off}
+        {["secure_referer_check"], off},
+        {["warn_object_size"], "10MB"},
+        {["max_object_size"], "100MB"},
+        {["warn_siblings"], 250},
+        {["max_siblings"], 1000}
```
engelsanchez, line 88: I have a great idea! If the commas were at the beginning, the merge difference would be cleaner! Let's change all teh codez!

joedevivo, line 88: commas get their own lines!

engelsanchez:
+1 I verified building a devrel after removing the bits from riak.schema. Values are set correctly, it's all good.

joedevivo:
Cool. I'll remove those in another PR that includes some other stuff later today

PR #728, opened 2013-11-16
Transform an empty context to a `undefined` for the PB protocol

russelldb:
As spotted by @broach in https://github.com/basho/riak-java-client/pull/332


seancribbs:
:+1:

PR #729, opened 2013-11-18
Fix consistent_object mixed-mode failure.

cmeiklejohn:
Ensure that if we attempt to call consistent_object on a remote node
and get a undefined function error back, this most likely means that
the node contacted is running an earlier version of Riak, in which
case strong consistency is not available.

It appears that in addition to this, we probably should have a
capability for this.

This fixes the current failure on `verify_basic_upgrade.`


cmeiklejohn:
Tagging @jtuple and adding to the kill bill.

jtuple:
This is a good change to make even with a capability, since `riak_client` can also be used as a remote Erlang client (not recommended) in which case there is no local capability negotiation. We may still want to add a capability anyway, but that's a separate issue.

+1 merge

cmeiklejohn:
Thanks.

PR #730, opened 2013-11-18
add riak_pb to apps in src/riak_kv.app.src

vinoski:
The riak_pb app is used in riak_kv but was not listed as an application
dependency in the src/riak_kv.app.src file.

Also add missing apps to riak_kv_test_util code that manages
starting/stopping application dependencies for testing.

### DIFF
--- a/src/riak_kv.app.src
+++ b/src/riak_kv.app.src
@@ -17,7 +17,8 @@
                   webmachine,
                   os_mon,
                   riak_pipe,
-                  riak_dt
+                  riak_dt,
+                  riak_pb
                  ]},
   {registered, []},
   {mod, {riak_kv_app, []}},
--- a/src/riak_kv_test_util.erl
+++ b/src/riak_kv_test_util.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_test_util: utilities for test scripts
 %%
-%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
 %%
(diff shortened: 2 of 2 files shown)


cmeiklejohn:
`eunit` stalls on me, but otherwise it looks good and builds.

```
module 'riak_kv_pb_object'

=INFO REPORT==== 18-Nov-2013::19:05:06 ===
    application: bitcask
    exited: stopped
    type: temporary
```

vinoski:
Commit amended, should be OK now.

cmeiklejohn:
Confirmed the amended commit resolves the hangs in the test suite.  

```
  All 286 tests passed.
```

cmeiklejohn:
:+1:

seancribbs:
Note `riak_pb` is a dependency of `riak_api`, should it need to be listed in the dependent apps?

vinoski:
`riak_kv` code uses `riak_pb` code directly, so I'd say yes. Seems wrong to count on picking up direct dependencies via other dependencies.

cmeiklejohn:
I agree with @vinoski, but I'm up for the discussion.

vinoski:
In fact, perhaps the `rebar.config` for `riak_kv` should explicitly list `riak_pb` as a dep as well, for the same reason.

seancribbs:
@vinoski Disagree with rebar.config change, that just makes it harder to roll versions.

vinoski:
@seancribbs Understood, and that's definitely the practical conclusion, but I think one could argue that problem is really a rebar issue...

seancribbs:
@vinoski Agree, but the day that rebar works that way is not today. :smile_cat:

PR #731, opened 2013-11-19
remove ebloom from rebar.config

vinoski:
Remove the ebloom dependency, as it's not used anywhere in riak_kv.


cmeiklejohn:
:+1: to merge.

PR #732, opened 2013-11-20
Add HTTP API for datatypes.

seancribbs:
This rewrites `riak_kv_wm_crdt` and ensures that the functionality it exposes is similar to PB. The JSON format of request/response payloads is described in detail at the top of the module, and implemented in `riak_kv_crdt_json`.

### DIFF
--- a/src/riak_kv_crdt_json.erl
+++ b/src/riak_kv_crdt_json.erl
@@ -1,0 +1,431 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_crdt_json: Codec routines for CRDTs in JSON
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(riak_kv_crdt_json).
+-export([update_request_from_json/3, fetch_response_to_json/4]).
+-compile([{inline, [bad_op/2, bad_field/1]}]).
+
+-define(FIELD_PATTERN, "^(.*)_(counter|set|register|flag|map)$").
+
+-ifdef(TEST).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+-include("riak_kv_types.hrl").
+-endif.
+
+-export_type([context/0]).
+
+%% Mostly copied from riak_pb_dt_codec
+%% Value types
+-opaque context() :: binary().
+-type map_field() :: {binary(), embedded_type()}.
+-type embedded_type() :: counter | set | register | flag | map.
+-type toplevel_type() :: counter | set | map.
+-type type_mappings() :: [{embedded_type(), module()}].
+
+%% Operations
+-type counter_op() :: increment | decrement | {increment | decrement, integer()}.
+-type simple_set_op() :: {add, binary()} | {remove, binary()} | {add_all, [binary()]} | {remove_all, [binary()]}.
+-type set_op() :: simple_set_op() | {update, [simple_set_op()]}.
+-type flag_op() :: enable | disable.
+-type register_op() :: {assign, binary()}.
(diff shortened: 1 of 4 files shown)


src/riak_kv_wm_crdt.erl
```diff
@@ -19,148 +19,155 @@
...
+
+service_available(RD, Ctx0) ->
+    Ctx = riak_kv_wm_utils:ensure_bucket_type(RD, Ctx0, #ctx.bucket_type),
+    {ok, Client} = riak_kv_wm_utils:get_riak_client(
+                     local, riak_kv_wm_utils:get_client_id(RD)),
+    {true, RD,
+     Ctx#ctx{client=Client,
+             bucket=path_segment_to_bin(bucket, RD),
+             key=path_segment_to_bin(key, RD),
+             crdt_type=proplists:get_value(crdt, wrq:path_info(RD)),
```
russelldb: I thought this came from the bucket_type now (that the path segment was just the string "datatypes")? Should this still be in the code?

src/riak_kv_wm_crdt.erl
```diff
@@ -276,174 +237,284 @@ malformed_boolean_param({Idx, Name, Default}, {Result, RD, Ctx}) ->
...
-normalize_rw_param("quorum") -> quorum;
-normalize_rw_param("all") -> all;
-normalize_rw_param(V) -> list_to_integer(V).
+malformed_timeout_param({Result, RD, Ctx}) ->
+    case wrq:get_qs_value("timeout", undefined, RD) of
+        undefined ->
+            {Result, RD, Ctx};
+        TimeoutStr when is_list(TimeoutStr) ->
+            try
+                Timeout = list_to_integer(TimeoutStr),
```
russelldb: Should we also check that it is greater than zero?

src/riak_kv_wm_crdt.erl
```diff
@@ -276,174 +237,284 @@ malformed_boolean_param({Idx, Name, Default}, {Result, RD, Ctx}) ->
...
+        true ->
+            forbidden_check_crdt_type(RD, Ctx);
+        false ->
+            handle_common_error(bucket_type_unknown, RD, Ctx)
+    end.
+
+forbidden_check_crdt_type(RD, Ctx=#ctx{bucket_type = <<"default">>,
+                                       bucket=B0,
+                                       key=K0}) ->
+    B = mochiweb_util:quote_plus(B0),
```
russelldb, line 308: I think some comments  / docs about _why_ would be great here for future me when I maintain this code.

src/riak_kv_wm_crdt.erl
seancribbs: You're right, I missed this.

src/riak_kv_wm_crdt.erl
seancribbs: Yes, except doesn't `0` disable the timeout? Either way a guard can be added here.

src/riak_kv_wm_crdt.erl
seancribbs, line 308: Right-o.

src/riak_kv_wm_crdt.erl
```diff
@@ -276,174 +236,295 @@ malformed_boolean_param({Idx, Name, Default}, {Result, RD, Ctx}) ->
...
-normalize_rw_param("quorum") -> quorum;
-normalize_rw_param("all") -> all;
-normalize_rw_param(V) -> list_to_integer(V).
+malformed_timeout_param({Result, RD, Ctx}) ->
+    case wrq:get_qs_value("timeout", undefined, RD) of
+        undefined ->
+            {Result, RD, Ctx};
+        TimeoutStr when is_list(TimeoutStr) ->
+            try list_to_integer(TimeoutStr) of
+                Timeout when is_integer(Timeout), Timeout > 0 ->
```
russelldb, line 252: Did you figure out if `> 0` is correct? I wonder if `>= 0` is better, as you say, then `0` can mean `infinity`?

src/riak_kv_crdt_json.erl
```diff
@@ -0,0 +1,431 @@
...
+    case lists:keyfind(Mod, 2, Mods) of
+        false ->
+            Field;
+        {Type, Mod} ->
+            {Name, Type}
+    end.
+
+-spec field_to_json(map_field()) -> binary().
+field_to_json({Name, Type}) when is_binary(Name), is_atom(Type) ->
+    BinType = atom_to_binary(Type, utf8),
```
russelldb, line 125: Efficiency nitpick, but I assume this mapping is static, and could be either added to `?MOD_MAP` or some similar look up. But really, a  nitpick.

russelldb:
Ran eunit tests and tested by hand using riak-erlang-http-client#28.
Looks good, works for me. As mentioned, a riak_test would be nice. We can probably generalise the existing test that uses the PB client. :+1:

src/riak_kv_wm_crdt.erl
seancribbs, line 252: I think in the other APIs, yes `0` means `infinity`, but this doesn't coerce it. Fixing.

PR #733, opened 2013-11-22
make riak_kv_test_util app startup less brittle

vinoski:
In commit c3ebdc9 I added some missing apps to the list of apps in
riak_kv_test_util that need to be started for some riak_kv
testing. Unfortunately trying to list out the entire app dependency tree in
that module is very brittle and is to easy to forget to update if a new app
dependency is added in the future. Modify that list to eliminate the lower
level apps and instead just list the obvious higher-level apps, and add a
new function that starts each application's dependencies before starting
the app itself. Also change the riak_kv_test_util setup function to return
the list of all apps started, and change the cleanup function to take that
list of apps and stop them all.

### DIFF
--- a/src/riak_kv_test_util.erl
+++ b/src/riak_kv_test_util.erl
@@ -162,12 +162,12 @@ setup(TestName, SetupFun) ->
     net_kernel:start([TestNode, longnames]),
 
     %% Start dependent applications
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
:+1:, this resolves the test stalls.

PR #734, opened 2013-11-23
Refactor FSMs to do less work

evanmcc:
And hopefully be a little bit faster.  (Cluster speed tests will happen after vacation).

Four primary things have been done.:
1. changing the code wherever possible to pile up state mutations into a single operation to best take advantage of the compiler and avoid generating garbage.
2. Debugging and tracing code has been hidden behind the `Trace` variable, which also cuts down on FSM state tuple mutations, and doing extra work on platforms where we cannot benefit from it.
3. `proplists` module calls have been replaced with keyfind operations where possible as keyfind is a BIF.
4.  In addition to the `proplists` changes, there has been some refactoring to avoid puling the default props from an ets table over and over again, they're now pulled at the start and merged with the bucket properties.

There are a few inconsistent places that I'll fix when I get back, I just wanted to get this in front of people in case someone had time to start their review before I got back.

Some simple single node performance tests done with this code:  https://gist.github.com/evanmcc/c94f5c197760e0def652 
The number before the rate type is the number of worker threads, testing was done on r2s10.bos.

```
| state/ref         | 1 g/s | 128 g/s | 1 p/s | 128 p/s |
| baseline 2        | 2635  |   51856 | 1335  |   32522 | 
| macro cleanup     | 3862  |   63320 | 1494  |   37379 | 
```

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -69,6 +69,7 @@
                 bucket_props,
                 startnow :: {non_neg_integer(), non_neg_integer(), non_neg_integer()},
                 get_usecs :: non_neg_integer(),
+                trace = false :: boolean(),
                 tracked_bucket=false :: boolean(), %% is per bucket stats enabled for this bucket
                 timing = [] :: [{atom(), erlang:timestamp()}],
                 calculated_timings :: {ResponseUSecs::non_neg_integer(),
@@ -146,15 +147,23 @@ test_link(From, Bucket, Key, GetOptions, StateProps) ->
 %% ====================================================================
 
 %% @private
-init([From, Bucket, Key, Options, Monitor]) ->
+init([From, Bucket, Key, Options0, Monitor]) ->
     StartNow = os:timestamp(),
-    StateData = add_timing(prepare, #state{from = From,
-                                           options = Options,
-                                           bkey = {Bucket, Key},
-                                           startnow = StartNow}),
+    Options = proplists:unfold(Options0),
+    StateData = #state{from = From,
+                       options = Options,
+                       bkey = {Bucket, Key},
+                       timing = riak_kv_fsm_timing:add_timing(prepare, []),
+                       startnow = StartNow},
     (Monitor =:= true) andalso riak_kv_get_put_monitor:get_fsm_spawned(self()),
-    riak_core_dtrace:put_tag(io_lib:format("~p,~p", [Bucket, Key])),
-    ?DTRACE(?C_GET_FSM_INIT, [], ["init"]),
+    Trace = application:get_env(riak_kv, fsm_trace_enabled),
+    case Trace of 
+        true ->
+            riak_core_dtrace:put_tag([Bucket, $,, Key]),
+            ?DTRACE(?C_GET_FSM_INIT, [], ["init"]);
+        _ -> 
+            ok
+    end,
     {ok, prepare, StateData, 0};
 init({test, Args, StateProps}) ->
     %% Call normal init
@@ -164,24 +173,33 @@ init({test, Args, StateProps}) ->
     Fields = record_info(fields, state),
     FieldPos = lists:zip(Fields, lists:seq(2, length(Fields)+1)),
     F = fun({Field, Value}, State0) ->
-                Pos = proplists:get_value(Field, FieldPos),
+                Pos = get_option(Field, FieldPos),
                 setelement(Pos, State0, Value)
         end,
     TestStateData = lists:foldl(F, StateData, StateProps),
-
-    %% Enter into the execute state, skipping any code that relies on the
-    %% state of the rest of the system
     {ok, validate, TestStateData, 0}.
 
 %% @private
 prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
-                                  options=Options}) ->
-    ?DTRACE(?C_GET_FSM_PREPARE, [], ["prepare"]),
+                                  options=Options,
+                                  trace=Trace}) ->
+    ?DTRACE(Trace, ?C_GET_FSM_PREPARE, [], ["prepare"]),
+    {ok, DefaultProps} = application:get_env(riak_core, 
+                                             default_bucket_props),
     BucketProps = riak_core_bucket:get_bucket(Bucket),
-    DocIdx = riak_core_util:chash_key(BKey),
-    Bucket_N = proplists:get_value(n_val,BucketProps),
-    CrdtOp = proplists:get_value(crdt_op, Options),
-    N = case proplists:get_value(n_val, Options) of
+    %% typed buckets never fall back to defaults
+    Props = 
+        case is_tuple(Bucket) of
+            false ->
+                lists:keymerge(1, lists:keysort(1, BucketProps), 
+                               lists:keysort(1, DefaultProps));
+            true ->
+                BucketProps
+        end,
+    DocIdx = riak_core_util:chash_key(BKey, BucketProps),
+    Bucket_N = get_option(n_val, BucketProps),
+    CrdtOp = get_option(crdt_op, Options),
+    N = case get_option(n_val, Options) of
             undefined ->
                 Bucket_N;
             N_val when is_integer(N_val), N_val > 0, N_val =< Bucket_N ->
@@ -195,17 +213,18 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
             StateData2 = client_reply(Error, StateData),
             {stop, normal, StateData2};
         _ ->
-            StatTracked = proplists:get_value(stat_tracked, BucketProps, false),
-            Preflist2 = case proplists:get_value(sloppy_quorum, Options, true) of
-                        true ->
-                            UpNodes = riak_core_node_watcher:nodes(riak_kv),
-                            riak_core_apl:get_apl_ann(DocIdx, N, UpNodes);
-                        false ->
-                            riak_core_apl:get_primary_apl(DocIdx, N, riak_kv)
-                    end,
+            StatTracked = get_option(stat_tracked, BucketProps, false),
+            Preflist2 = 
+                case get_option(sloppy_quorum, Options) of
+                    false ->
+                        riak_core_apl:get_primary_apl(DocIdx, N, riak_kv);
+                    _ ->
+                        UpNodes = riak_core_node_watcher:nodes(riak_kv),
+                        riak_core_apl:get_apl_ann(DocIdx, N, UpNodes)
+                end,
             new_state_timeout(validate, StateData#state{starttime=riak_core_util:moment(),
                                                 n = N,
-                                                bucket_props=BucketProps,
+                                                bucket_props=Props,
                                                 preflist2 = Preflist2,
                                                 tracked_bucket = StatTracked,
                                                 crdt_op = CrdtOp})
@@ -213,8 +232,9 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
 
 %% @private
 validate(timeout, StateData=#state{from = {raw, ReqId, _Pid}, options = Options,
-                                   n = N, bucket_props = BucketProps, preflist2 = PL2}) ->
-    ?DTRACE(?C_GET_FSM_VALIDATE, [], ["validate"]),
+                                   n = N, bucket_props = BucketProps, preflist2 = PL2,
+                                   trace=Trace}) ->
+    ?DTRACE(Trace, ?C_GET_FSM_VALIDATE, [], ["validate"]),
     AppEnvTimeout = app_helper:get_env(riak_kv, timeout),
     Timeout = case AppEnvTimeout of
                   undefined -> get_option(timeout, Options, ?DEFAULT_TIMEOUT);
@@ -240,7 +260,7 @@ validate(timeout, StateData=#state{from = {raw, ReqId, _Pid}, options = Options,
                     _ElseFalse ->
                         N - FailR + 1 % cannot ever get R 'ok' replies
                 end,
-            AllowMult = proplists:get_value(allow_mult,BucketProps),
+            AllowMult = get_option(allow_mult, BucketProps),
             NFOk0 = get_option(notfound_ok, Options, default),
             NotFoundOk = riak_kv_util:expand_value(notfound_ok, NFOk0, BucketProps),
             DeletedVClock = get_option(deletedvclock, Options, false),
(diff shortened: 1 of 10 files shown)


jtuple:
Just started reviewing this and won't be finished until tomorrow. But, one immediate thing I noticed is that this PR changes code to use a new `riak_core_util:chash_key/2` that doesn't actually exist in develop. Didn't see a sibling `riak_core` PR so just modified things locally to add the obvious:

``` diff
diff --git a/src/riak_core_util.erl b/src/riak_core_util.erl
index bbab32c..ea3d111 100644
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -32,6 +32,7 @@
          unique_id_62/0,
          str_to_node/1,
          chash_key/1,
+         chash_key/2,
          chash_std_keyfun/1,
          chash_bucketonly_keyfun/1,
          mkclientid/1,
@@ -230,8 +231,11 @@ mkclientid(RemoteNode) ->

 %% @spec chash_key(BKey :: riak_object:bkey()) -> chash:index()
 %% @doc Create a binary used for determining replica placement.
-chash_key({Bucket,Key}) ->
+chash_key(BKey={Bucket,_}) ->
     BucketProps = riak_core_bucket:get_bucket(Bucket),
+    chash_key(BKey, BucketProps).
+
+chash_key({Bucket,Key}, BucketProps) ->
     {chash_keyfun, {M, F}} = proplists:lookup(chash_keyfun, BucketProps),
     M:F({Bucket,Key}).
```

Just need to make sure we get such a change into `riak_core` before merging this PR.

evanmcc:
Yes, sorry I missed that.  Will issue the PR tomorrow, I think there might be one other minor change.

src/riak_kv_put_fsm.erl
```diff
@@ -305,15 +324,16 @@ prepare(timeout, StateData0 = #state{from = From, robj = RObj,
...
                     process_reply({error, all_nodes_down}, StateData0);
                 {_, true} ->
                     %% This node is not in the preference list
                     %% forward on to a random node
-                    ListPos = crypto:rand_uniform(1, length(Preflist2)+1),
-                    {{_Idx, CoordNode},_Type} = lists:nth(ListPos, Preflist2),
-                    _Timeout = get_option(timeout, Options, ?DEFAULT_TIMEOUT),
-                    ?DTRACE(?C_PUT_FSM_PREPARE, [1],
+                    random:seed(os:timestamp()),
+                    ListPos = random:uniform(length(Preflist2)+1),
```
jtuple: Is `random:uniform` really faster than `crypto:rand_uniform`? Also, if we are going to go with `random` use `uniform_s` rather than `seed + uniform`. Eg:

``` erlang
{ListPos, _} = random:uniform_s(length(Preflist2)+1, os:timestamp())
```

Also, if we really want to get crazy with micro-optimization, there's always the various crazy ideas I tried [here](https://github.com/basho/riak_ensemble/blob/jdb-wip-2.0/src/riak_ensemble_router.erl#L172-L185). Although, avoiding that insanity is probably best.

src/riak_kv_put_fsm.erl
```diff
@@ -477,15 +532,15 @@ execute(State=#state{options = Options, coord_pl_entry = CPL}) ->
 %% will guarantee a frontier object.
 %% N.B. Not actually a state - here in the source to make reading the flow easier
 execute_local(StateData=#state{robj=RObj, req_id = ReqId,
-                                timeout=Timeout, bkey=BKey,
-                                coord_pl_entry = {_Index, Node} = CoordPLEntry,
-                                vnode_options=VnodeOptions,
-                                starttime = StartTime}) ->
-    ?DTRACE(?C_PUT_FSM_EXECUTE_LOCAL, [], [atom2list(Node)]),
-    StateData1 = add_timing(execute_local, StateData),
```
jtuple, line 486: Looks like we've lost the timing information for `execute_local`. In `execute_remote`, you kept the timing information, but guarded it via the trace option ([link](https://github.com/basho/riak_kv/pull/734/files#diff-c5485aae316e5a86ed16db7faf1411afR600))

src/riak_kv_put_fsm.erl
```diff
@@ -526,36 +582,49 @@ waiting_local_vnode(Result, StateData = #state{putcore = PutCore}) ->
...
 %% enough responses have been received yet (i.e. if W/DW=1)
 %% N.B. Not actually a state - here in the source to make reading the flow easier
-execute_remote(StateData=#state{robj=RObj, req_id = ReqId,
-                                preflist2 = Preflist2, bkey=BKey,
-                                coord_pl_entry = CoordPLEntry,
-                                vnode_options=VnodeOptions,
-                                putcore=PutCore,
-                                starttime = StartTime}) ->
-    StateData1 = add_timing(execute_remote, StateData),
+execute_remote(StateData0=#state{robj=RObj, req_id = ReqId,
```
jtuple: Minor nitpick: why the change from `StateData, StateData1` to `StateDate0, StateData` here? We end up with a few diff lines that haven't changed at all except for the naming. And this isn't a case where we've added more variables and therefore needed to rename things regardless.

src/riak_kv_put_fsm.erl
```diff
@@ -698,69 +795,45 @@ flatten_options([{Key, Value} | Rest], Opts) ->
...
+    case {Find, ReturnBody} of
+        {false, _} ->
+            [{returnbody, ReturnBody}|Opts];
+        {{_, true}, false} ->
+            lists:keyreplace(returnbody, 1, Opts, {returnbody, true});
+        {{_, false}, true} ->
+            lists:keyreplace(returnbody, 1, Opts, {returnbody, false});
+        _ ->
+            Opts
+    end.
```
jtuple: So the craziness that is `fixup_opts/2` is actually faster than just using `lists:keystore` since `lists:keyfind` is a BIF whereas `keystore` isn't? So sad.

Also, aren't the replacement tuples in `keyreplace` calls backwards? Shouldn't it be `false / true` rather than `true / false`?

src/riak_kv_put_fsm.erl
```diff
@@ -782,26 +855,27 @@ invoke_hook(undefined, undefined, JSName, RObj) when JSName /= undefined ->
 invoke_hook(_, _, _, _) ->
     {error, {invalid_hook_def, no_hook}}.
 
--spec decode_precommit(any()) -> fail | {fail, any()} | riak_object:riak_object().
-decode_precommit({erlang, {Mod, Fun}, Result}) ->
+-spec decode_precommit(any(), atom()) -> fail | {fail, any()} | riak_object:riak_object().
```
jtuple: Is the spec for `Trace` better served as `boolean()` instead of `atom()`? Technically, the same in Erlang but better for humans.

src/riak_kv_put_fsm.erl
```diff
@@ -897,25 +975,34 @@ get_option(Name, Options) ->
...
                      Reply;
                  [] ->
                      Reply;
                  Details ->
-                     add_client_info(Reply, Details, State2)
+                     add_client_info(Reply, Details, 
+                                     State#state{timing = Timing})
              end,
     Pid ! {ReqId, Reply2},
-    add_timing(reply, State2#state{reply = Reply}).
```
jtuple, line 929: So, the original code for `client_reply/2` called `add_timing` twice. First at the beginning of the function, and a second time at the end. Your modified version only calls things once. Which makes sense. Was the old code just horribly wrong or was there some method to that madness?

src/riak_kv_put_fsm.erl
```diff
@@ -939,8 +1026,6 @@ client_info([timing | Rest], StateData = #state{timing = Timing}, Info) ->
 default_details() ->
     [timing].
 
-
-%% Add timing information to the state
 add_timing(Stage, State = #state{timing = Timing}) ->
     State#state{timing = riak_kv_fsm_timing:add_timing(Stage, Timing)}.
```
jtuple, line 1039: Any reason to remove this comment? Just an accident?

src/riak_kv_get_fsm.erl
```diff
@@ -236,7 +249,7 @@ validate(timeout, StateData=#state{from = {raw, ReqId, _Pid}, options = Options,
                     _ElseFalse ->
                         N - FailR + 1 % cannot ever get R 'ok' replies
                 end,
-            AllowMult = proplists:get_value(allow_mult,BucketProps),
+            AllowMult = get_option(allow_mult, BucketProps, default),
```
jtuple: Is returning `default` if missing what we want to do here? Seems different than the original code, and isn't what happens in the put FSM changes.

jtuple:
When using `lists:keyfind` you sometimes match the non-false case as `{the-key, Val}` and other times as `{_, Val}`. Let's pick one approach and be consistent. Since this is a performance refactoring, `{_, Val}` is theoretically faster when looking at the generated BEAM code, since it avoids extra code to re-check the key (already checked inside the `keyfind` BIF).

Related: we probably should use a utility function for this. In the get FSM we have a `get_option/3` where as in the put FSM we have `get_option/2` and `get_option/3`. Since you've already converted both over to `keyfind`, probably should make sure we have both arities in both FSMs and use `get_option` instead of `keyfind` everywhere except in the few cases where we actually match on the `false` case. Also, wonder if `get_option` should really just be moved to `riak_kv_util` or something. Maybe some other time, since that would be a pretty big diff to rename all call sites in the FSMs.

jtuple:
In some of your case statements that check the `Trace` variable you you match against `true` and `false` and in other places against `true` and `_`. Should change things to be consistent.

jtuple:
I've noticed that in a few places where we formerly checked bucket options against `BucketProps` we're now checking against the combined `BucketProps + default properties` list you create via the `keymerge`. Either by directly matching against the new `Props` value, or in later states where we're matching against `BucketProps` but the value you've stored in the state for `bucket_props` is actually `Props` not `BucketProps`. In the original code, only the `riak_kv_util:get_bucket_option` would fallback to the default properties -- now multiple direct checks inside the FSMs do (by way of using your merged property list). I wonder if this is actually safe, or if we're changing semantics.

Likewise, until I reviewed this PR I didn't even know `riak_kv_util:get_bucket_option/2` fell back to the default properties. I wonder if that (ie. even on develop w/o this PR) is safe in the world of bucket types. Typed buckets don't use the default bucket properties by design.

Any thoughts on the matter @jrwest?

**_EDIT:**_ So, after talking to @jrwest we've determined that all we need to do is change the two places in this PR where we merge in the default bucket properties to only do so for untyped buckets. For typed buckets, we don't use the default properties. And yes, things are also already broken on develop inside `riak_kv_util:get_bucket_option/2`, but the changes made in this PR (after the only-merge-for-untyped change just discussed) implicitly fixes that issue.

src/riak_kv_put_fsm.erl
```diff
@@ -417,18 +442,45 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
...
-            StateData = init_putcore(StateData3, IdxType),
-            ?DTRACE(?C_PUT_FSM_VALIDATE, [N, W, PW, DW], []),
+            RObj1 = 
+                case get_option(update_last_modified, Options) of
+                    true ->
+                        riak_object:update_last_modified(RObj0);
+                        _ ->
+                        RObj0
+                end,
+            RObj = riak_object:apply_updates(RObj1),
```
jtuple: Inlining `init_putcore` makes sense, but I think we should keep `apply_updates` as a separate function to keep the code clean. Just change `apply_updates` to do the above `RObj1, RObj` logic and return the final `RObj` that you'll end up storing in `StateData`.

src/riak_kv_put_fsm.erl
```diff
@@ -417,18 +442,45 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
...
+                    true ->
+                        {true, true};
+                    _ ->
+                        case Postcommit of
+                            [] -> 
+                                {false, false};
+                            _ -> 
+                                {true, false}
+                        end
+                end,
```
jtuple, line 459: Perhaps keep this case statement in a separate function as well to keep things clean. Maybe combine with the `fixup_opts` call below. Would need to move the `PutCore` stuff after, but that's no biggie.

Something like:

``` erlang
VNodeOpts0 = handle_options(Options, []),
{VNodeOpts, ReturnBody} = maybe_override_returnbody(Postcommit, Options, VNodeOpts0),
PutCore = ...
```

src/riak_kv_put_fsm.erl
evanmcc: the crypto module appears to only have one lock, so while any individual call is likely faster, the shared lock makes performance increasingly bad as concurrency increases.  I had a lot of issues with this wrt my client.erl test code.

src/riak_kv_put_fsm.erl
evanmcc, line 1039: either accident or I felt that it was very 

```
//add 5 to x 
X1 = X +5, 
```

but I can revert

src/riak_kv_put_fsm.erl
evanmcc: reverted

src/riak_kv_put_fsm.erl
evanmcc: I rethought this and was able to simplify the code, including getting rid of this function. I obviously wasn't thinking clearly when I wrote it.  I blame eqc fatigue.

src/riak_kv_put_fsm.erl
evanmcc, line 929: The timing is only added once in the get_fsm version of this function, so I imagine it was just broken, I couldn't see any advantage to calling it twice.  I was confused as well.

src/riak_kv_get_fsm.erl
evanmcc: I added get_option/2 which defaults to undefined, so this should now have the same behavior.

src/riak_kv_put_fsm.erl
evanmcc, line 459: I got rid of `fixup_opts`, so I think that I'll leave this in for now.

evanmcc:
pushed a commit:
- Have moved all calls to keyfind into get_option/2,3, added 2 where it was missing, so things should be more consistent there.  
- All `Trace` cases now use `_`.
- addressed bucket types comments as keymerge no longer happens in typed buckets.
- reassessed which list everything was coming from, everything should now have the same behavior as before.  The only value for which this is not true is AllowMult, which I suspect was an oversight anyway.  I cannot find where anything would handle allow_mult being undefined, at least.

Also issued the core commit: https://github.com/basho/riak_core/issues/486

evanmcc:
120 second runs:

develop 
get 1 avg = 3879, gets [3803,4064,3771]
get 128 avg = 53899, gets [54091,53791,53817]
put 1 avg = 1849, puts [1891,1903,1753]
put 128 avg = 31799, puts [31913,31752,31732]

address comments
get 1 avg = 5827, gets [6160,5656,5667]
get 128 avg = 63021, gets [63549,61989,63525]
put 1 avg = 2389, puts [2325,2501,2342]
put 128 avg = 37028, puts [37122,36899,37064]

jtuple:
Re-adding the time tracing logic to `execute_local` at the end of the function ([here](https://github.com/basho/riak_kv/blob/7877a834d568ff44f8d712d5b43fe05b06f054d5/src/riak_kv_put_fsm.erl#L550-L556)) doesn't look right, since `new_state/2` will immediately add the timing info for `waiting_local_vnode` and thus the timestamp for `execute_local` and `waiting_local_vnode` will be essentially identical.

The original code added timing [at the top](https://github.com/basho/riak_kv/blob/develop/src/riak_kv_put_fsm.erl#L486) of the function [as does](https://github.com/basho/riak_kv/blob/7877a834d568ff44f8d712d5b43fe05b06f054d5/src/riak_kv_put_fsm.erl#L608) `execute_remote` in your branch. That seems to be the more correct way, such that the timestamp is before the blocking `riak_kv_vnode:coord_put`.

jtuple:
Other than timing issue above, I'm +1 on this code. Did some basic testing, but nothing terribly involved. On the other hand, there's nothing in the newer commits that I believe would invalidate the testing I did when I reviewed this code last week.

I'd say address the timing comment above, clean-up commit history, and merge away.

evanmcc:
this PR and #734 somehow got tied together because I was tired and re-using the same directory and managed to merge them somehow.

PR #736, opened 2013-11-25
Add 2i AAE command

engelsanchez:
This change adds support for a manual command to repair secondary index
data that is out of sync with the KV objects. The command will
repair a given partition or all partitions in a given node, with an
optional speed throttle parameter. The speed throttle is a duty cycle
number such that if 50 is given, it will try to do some work, then wait
for the time that work took before the next unit of work and so on.

Each vnode will now have a special AAE tree where it will store hashes
for the 2i data of each object in the backend.
When the command runs, for each partition to be repaired it fetches all
secondary index data from the backend and builds a temporary hashtree.
It then runs an exchange between this temporary hashtree and the vnode's
2i AAE tree. Differences found will trigger an index refresh command
that will rewrite 2i data based on an object's current index terms.

A fold_indexes command has been added to index supporting backends to
fold only over 2i data, as well as a refresh index put that will only
rewrite index data that is inconsistent. This was all added to the
memory backend too, which is kind of overkill as it should never have
any 2i inconsistencies, but it makes other layers agnostic, so...
The multi backend was also updated.

Data for this operation will be stored in the anti_entropy directory in
a new '2i' folder. It should be cleaned up after each operation under
normal conditions. LevelDB is used to create the temporary 2i database.
Some tuning might be needed in the future. It would be nice to have a
mostly in-memory database that only used disk if the 2i data was too big
to fit into some amount of data specified by the operator.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -117,6 +117,7 @@
          local_compare/2,
          compare/2,
          compare/3,
+         compare/4,
          top_hash/1,
          get_bucket/3,
          key_hashes/2,
@@ -402,6 +403,9 @@ compare(Tree, Remote) ->
 -spec compare(hashtree(), remote_fun(), acc_fun(X)) -> X.
 compare(Tree, Remote, AccFun) ->
     compare(1, 0, Tree, Remote, AccFun, []).
+
+compare(Tree, Remote, AccFun, Acc) ->
+    compare(1, 0, Tree, Remote, AccFun, Acc).
 
 -spec levels(hashtree()) -> pos_integer().
 levels(#state{levels=L}) ->
--- a/src/riak_kv_2i_aae.erl
+++ b/src/riak_kv_2i_aae.erl
@@ -1,0 +1,673 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Code related to repairing 2i data.
+-module(riak_kv_2i_aae).
+-behaviour(gen_fsm).
+
+-include("riak_kv_wm_raw.hrl").
+
+-export([start/2, stop/1, get_status/0, to_report/1]).
+
+-export([next_partition/2, wait_for_aae_pid/2, wait_for_repair/3,
+         wait_for_repair/2]).
+
+%% gen_fsm callbacks
+-export([init/1,
+         handle_event/3,
+         handle_sync_event/4,
+         handle_info/3,
+         format_status/2,
+         terminate/3,
+         code_change/4]).
+
+-export_type([status/0, partition_result/0, worker_status/0]).
+
+%% How many items to scan before possibly pausing to obey speed throttle
+-define(DEFAULT_SCAN_BATCH, 1000).
+-define(MAX_DUTY_CYCLE, 100).
+-define(MIN_DUTY_CYCLE, 1).
+-define(LOCK_RETRIES, 10).
+-define(AAE_PID_TIMEOUT, 30000).
+-define(INDEX_SCAN_TIMEOUT, 300000). % 5 mins, in case fold dies timeout
+-define(INDEX_REFRESH_TIMEOUT, 60000).
+
+-type worker_status() :: starting | scanning_indexes | populating_hashtree |
+                         exchanging.
+
+-type partition_result() :: {Partition :: integer(),
+                             {ok,  term()} | {error, term()}}.
+
+-type status() :: [{speed, integer()} |
+                   {partitions, [integer()]} |
+                   {remaining, [integer()]} |
+                   {results, [partition_result()]}|
+                   {current_partition, integer()} |
+                   {current_partition_status, worker_status()} |
+                   {index_scan_count, integer()} |
+                   {hashtree_population_count, integer()} |
+                   {exchange_count, integer()}].
+
+-record(state,
+        {
+         speed :: integer(),
+         partitions :: [integer()],
+         remaining :: [integer()],
+         results = [],
+         caller :: pid(),
+         reply_to,
+         early_reply,
+         aae_pid_req_id :: pid() | undefined,
+         worker_monitor :: reference() | undefined,
+         worker_status :: worker_status(),
+         index_scan_count = 0 :: integer(),
+         hashtree_population_count = 0 :: integer(),
+         exchange_count = 0 :: integer(),
+         open_dbs = [] :: [reference()]
+        }).
+
+-record(partition_result, {
+          status :: ok | error,
+          error,
+          index_scan_count = 0 :: integer(),
+          hashtree_population_count = 0:: integer(),
+          exchange_count = 0 :: integer()
+         }).
+
+
+%% @doc Starts a process that will issue a repair for each partition in the
+%% list, sequentially.
+%% The Speed parameter controls how fast we go. 100 means full speed,
+%% 50 means to do a unit of work and pause for the duration of that unit
+%% to achieve a 50% duty cycle, etc.
+%% If a single partition is requested, this function will block until that
+%% partition has been blocked or that fails to notify the user immediately.
+-spec start([integer()], integer()) -> {ok, pid()} | {error, term()}.
+start(Partitions, Speed)
+  when Speed >= ?MIN_DUTY_CYCLE, Speed =< ?MAX_DUTY_CYCLE ->
+    Res = gen_fsm:start({local, ?MODULE}, ?MODULE,
+                        [Partitions, Speed, self()], []),
+    case {Res, length(Partitions)} of
+        {{ok, _Pid}, 1} ->
+            try
+                EarlyAck = gen_fsm:sync_send_all_state_event(?MODULE,
+                                                             early_ack,
+                                                             infinity),
+                case EarlyAck of
+                    lock_acquired ->
+                        Res;
+                    _ ->
+                        EarlyAck
+                end
+            catch
+                exit:_ ->
+                    {error, early_exit}
+            end;
+
+        _ ->
+            Res
+    end.
+
+%% @doc Will stop any worker process and try to close leveldb databases.
+stop(TimeOut) ->
+    gen_fsm:sync_send_all_state_event(?MODULE, stop, TimeOut).
+
+-spec get_status() -> [{Key :: atom(), Val :: term()}].
+get_status() ->
+    gen_fsm:sync_send_all_state_event(?MODULE, status, infinity).
+
+init([Partitions, Speed, Caller]) ->
+    process_flag(trap_exit, true),
+    lager:info("Starting 2i repair at speed ~p for partitions ~p",
+               [Speed, Partitions]),
+    {ok, next_partition, #state{partitions=Partitions, remaining=Partitions,
+                                speed=Speed, caller=Caller}, 0}.
+
+%% @doc Notifies caller for certain cases involving a single partition
+%% so the user sees an error condition from the command line.
+maybe_notify(State=#state{reply_to=undefined}, Reply) ->
+    State#state{early_reply=Reply};
+maybe_notify(State=#state{reply_to=From}, Reply) ->
+    gen_fsm:reply(From, Reply),
+    State#state{reply_to=undefined}.
+
+
+%% @doc Process next partition or finish if no more remaining.
+next_partition(timeout, State=#state{remaining=[]}) ->
+    %% We are done. Report results
+    Status = to_simple_state(State),
+    Report = to_report(Status),
+    lager:info("Finished 2i repair:\n~s", [Report]),
+    {stop, normal, State};
+next_partition(timeout, State=#state{remaining=[Partition|_]}) ->
+    ReqId = make_ref(),
+    riak_kv_vnode:request_hashtree_pid(Partition, {fsm, ReqId, self()}),
+    {next_state, wait_for_aae_pid, State#state{aae_pid_req_id=ReqId}, ?AAE_PID_TIMEOUT}.
+
+%% @doc Waiting for vnode to send back the Pid of the AAE tree process.
+wait_for_aae_pid(timeout, State) ->
+    State2 = maybe_notify(State, {error, no_aae_pid}),
+    {stop, no_aae_pid, State2};
+wait_for_aae_pid({ReqId, {error, Err}}, State=#state{aae_pid_req_id=ReqId}) ->
+    State2 = maybe_notify(State, {error, {no_aae_pid, Err}}),
+    {stop, no_aae_pid, State2};
+wait_for_aae_pid({ReqId, {ok, TreePid}},
+                 State=#state{aae_pid_req_id=ReqId, speed=Speed,
+                              remaining=[Partition|_]}) ->
+    WorkerFun =
+    fun() ->
+            Res =
+            repair_partition(Partition, Speed, ?MODULE, TreePid),
+            gen_fsm:sync_send_event(?MODULE, {repair_result, Res}, infinity)
+    end,
+    Mon = monitor(process, spawn_link(WorkerFun)),
+    {next_state, wait_for_repair, State#state{worker_monitor=Mon}}.
+
+%% @doc Waiting for a partition repair process to finish
+wait_for_repair({lock_acquired, Partition},
+                _From,
+                State=#state{remaining=[Partition|_]}) ->
+    State2 = maybe_notify(State, lock_acquired),
+    {reply, ok, wait_for_repair, State2};
+wait_for_repair({repair_result, Res},
+                _From,
+                State=#state{remaining=[Partition|Rem],
+                             results=Results,
+                             worker_monitor=Mon}) ->
+    State2 =
+    case Res of
+        {error, {lock_failed, LockErr}} ->
+            maybe_notify(State, {lock_failed, LockErr});
+        _ ->
+            State
+    end,
+    demonitor(Mon, [flush]),
+    {reply, ok, next_partition,
+     State2#state{remaining=Rem,
+                  results=[{Partition, Res}|Results],
+                  worker_monitor=undefined,
+                  worker_status=starting,
+                  index_scan_count=0,
+                  hashtree_population_count=0,
+                  exchange_count=0},
+     0}.
+
+%% @doc Process async worker status updates during a repair
+wait_for_repair({index_scan_update, Count}, State) ->
+    {next_state, wait_for_repair,
+     State#state{worker_status=scanning_indexes,
+                 index_scan_count=Count}};
+wait_for_repair({hashtree_population_update, Count}, State) ->
+    {next_state, wait_for_repair,
+     State#state{worker_status=populating_hashtree,
+                 hashtree_population_count=Count}};
+wait_for_repair({repair_count_update, Count}, State) ->
+    {next_state, wait_for_repair,
+     State#state{worker_status=exchanging,
+                 exchange_count=Count}};
+wait_for_repair({open_db, DBRef}, State = #state{open_dbs=DBs}) ->
+    {next_state, wait_for_repair,
+     State#state{open_dbs=[DBRef|DBs]}};
+wait_for_repair({close_db, DBRef}, State = #state{open_dbs=DBs}) ->
+    {next_state, wait_for_repair,
+     State#state{open_dbs=lists:delete(DBRef, DBs)}}.
+
+%% @doc Performs the actual repair work, called from a spawned process.
+-spec repair_partition(integer(), integer(), {pid(), any()}, pid()) ->
+    #partition_result{}.
+repair_partition(Partition, DutyCycle, From, TreePid) ->
+    case get_hashtree_lock(TreePid, ?LOCK_RETRIES) of
+        ok ->
+            lager:info("Acquired lock on partition ~p", [Partition]),
+            gen_fsm:sync_send_event(From, {lock_acquired, Partition}, infinity),
+            lager:info("Repairing indexes in partition ~p", [Partition]),
+            case create_index_data_db(Partition, DutyCycle) of
+                {ok, {DBDir, DBRef, IndexDBCount}} ->
+                    Res =
+                    case build_tmp_tree(Partition, DBRef, DutyCycle) of
+                        {ok, {TmpTree, TmpTreeCount}} ->
+                            Res0 = do_exchange(Partition, DBRef, TmpTree,
+                                               TreePid),
+                            remove_tmp_tree(Partition, TmpTree),
+                            case Res0 of
+                                {ok, CmpCount} ->
+                                    #partition_result{
+                                       status=ok,
+                                       index_scan_count=IndexDBCount,
+                                       hashtree_population_count=TmpTreeCount,
+                                       exchange_count=CmpCount};
+                                {error, TmpTreeErr} ->
+                                    #partition_result{
+                                       status=error,
+                                       error=TmpTreeErr,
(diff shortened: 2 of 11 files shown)


engelsanchez:
@jtuple this is the 2i AAE PR. Start reviewing at your leisure. @rzezeski feel free to take a peek. Remember that this off 1.4.  There are some eunit tests failures that I haven't tracked down. Tests in riak_kv_vnode fail because getting bucket properties from a test bucket is not returning the hash function, just a proplist with only the bucket name.

engelsanchez:
Changed to acquire hashtree lock before building the temporary 2i hashtree and reporting immediately to the user if the lock can't be acquired and a single partition is being repaired. As suggested by @jtuple

engelsanchez:
Added better handling for when AAE is disabled (return an error immediately) and the single partition case when the tree has not been built (again, reporting and returning an error immediately).  
Last time I checked, the Github UI was not showing the new commit. But it's there.

engelsanchez:
The last commit fixes the index folder in riak_kv_index_hashtree, which was expecting a riak_object instead of a binary object, which is what object fold functions actually get.

jtuple:
Trying to do a 2i repair when running Bitcask leads to following error in logs:

```
17:26:12.686 [error] gen_fsm <0.632.0> in state active terminated with reason: call to undefined function riak_kv_bitcask_backend:fold_indexes/4 from riak_kv_vnode:handle_command/3 line 559
17:26:12.686 [error] CRASH REPORT Process <0.632.0> with 1 neighbours exited with reason: call to undefined function riak_kv_bitcask_backend:fold_indexes(#Fun<riak_kv_2i_aae.7.114426996>, 0, [async_fold], {state,#Ref<0.0.0.2264>,"68507889249886074290797726533575766546371837952",[{vnode_vclocks,true},...],...}) in gen_fsm:terminate/7 line 611
17:26:12.686 [error] Supervisor riak_core_vnode_sup had child undefined started with {riak_core_vnode,start_link,undefined} at <0.632.0> exit with reason call to undefined function riak_kv_bitcask_backend:fold_indexes(#Fun<riak_kv_2i_aae.7.114426996>, 0, [async_fold], {state,#Ref<0.0.0.2264>,"68507889249886074290797726533575766546371837952",[{vnode_vclocks,true},...],...}) in context child_terminated
```

Not a real issue (let it crash!) but leads to unnecessary log spam. This error is because Bitcask doesn't support index folds. In my original prototype I [added](https://github.com/basho/riak_kv/commit/dde12254153aa82790b03625c840ec1f82389c53#diff-6b44822fed66dcddf6f5dbe0354b4f96L61) a new `index_fold` capability that could be checked to see if we support index folds or not in a given backend. Looks like that didn't get moved to this new branch. However, since both LevelDB and memory both support index folds now, probably safe to just check the existing `indexes` capability before trying to do the index fold.

In fact, perhaps it's a good idea to check during the `riak-admin` command and print a message and not even attempt things. Easy for the non multi-backend case.

engelsanchez:
The intention was to make index folds an implied property of backends supporting the indexes capability.  So Bitcask should simply report "indexes not supported". Will fix.

engelsanchez:
It seems GH lost my last comment on this PR. :(
I had mentioned that the bitcask case has a better error message now, and that this is now using similar leveldb options to regular AAE.
Another extra commit was just pushed: There is now an immediate and better error message if the repair-2i command is run while another one is already executing.

engelsanchez:
One more commit. This one should:
- Issue async hashtree pid requests and time out
- Adds status command to indicate whether repair is running
- Adds kill command to forcibly kill the current repair command

src/riak_kv_2i_aae.erl
```diff
@@ -0,0 +1,451 @@
...
+%% list, sequentially.
+%% The DutyCycle parameter controls how fast we go. 100 means full speed,
+%% 50 means to do a unit of work and pause for the duration of that unit
+%% to achieve a 50% duty cycle, etc.
+-spec start_partition_repair([integer()], integer()) -> pid().
+start_partition_repair(Partitions, DutyCycle)
+  when is_number(DutyCycle),
+       DutyCycle >= ?MIN_DUTY_CYCLE, DutyCycle =< ?MAX_DUTY_CYCLE ->
+    Parent = self(),
+    Ref = make_ref,
```
cmeiklejohn: Don't you mean `make_ref()`?

engelsanchez:
The fix for the make_ref typo and removal of some debug log statements I left behind when deleting lock files is in.
Another thing: I'm convinced that the zombie leveldb problem I saw yesterday is due to the fact that we do an index fold in one of the worker pools, which doesn't immediately GC and keeps a leveldb handle for longer than you'd wish. In production, where a lot of work would be going on in the worker pool, that's hopefully less likely. But it is not ideal.

src/riak_kv_2i_aae.erl
```diff
@@ -0,0 +1,449 @@
...
+    Hash = riak_kv_index_hashtree:hash_index_data(Indexes),
+    Tree2 = hashtree:insert(K, Hash, Tree),
+    Count2 = Count + 1,
+    StartTime2 =
+    case Count2 rem BatchSize of
+        0 -> duty_cycle_pause(WaitFactor, StartTime),
+             now();
+        _ -> StartTime
+    end,
+    tmp_tree_insert(tmp_tree_iter_move(Itr, next),
```
engelsanchez: Matthew is porting the prefetch change in eleveldb to 1.4. That changes the semantics of the 'next' move.  I will look into rewriting this code properly with an elveldb fold instead, which protects it from the semantics change and possible performance loss.

engelsanchez:
Changed leveldb iteration code to use a standard fold to protect against leveldb iterator semantics changes for 1.4 (the prev operation). And it's just the right thing to do. I'm not sure why I didn't do that in the first place.

src/hashtree.erl
```diff
@@ -403,6 +404,9 @@ compare(Tree, Remote) ->
 compare(Tree, Remote, AccFun) ->
     compare(1, 0, Tree, Remote, AccFun, []).
```
jtuple, line 405: Any reason not to update to the new `compare/4`?

``` erlang
compare(Tree, Remote, AccFun) ->
    compare(Tree, Remote, AccFun, []).
```

src/riak_kv_console.erl
```diff
@@ -436,9 +437,134 @@ run_reformat(M, F, A) ->
...
+parse_repair_2i_args(DutyCycle, Partitions) ->
+    case get_2i_repair_indexes(Partitions) of
+        {ok, IdxList} ->
+            {ok, IdxList, DutyCycle};
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+get_2i_repair_indexes([]) ->
+    AllVNodes = riak_core_vnode_manager:all_vnodes(),
```
jtuple: Prefer `riak_core_vnode_manager:all_vnodes(riak_kv_vnode)` instead. No need for extra filtering, and less garbage generated since vnode manager will do filtering inside the ETS table rather than process heap.

src/riak_kv_memory_backend.erl
```diff
@@ -202,16 +203,20 @@ put(Bucket, PrimaryKey, IndexSpecs, Val, State=#state{data_ref=DataRef,
                                                       ttl=TTL,
                                                       used_memory=UsedMemory}) ->
     Now = os:timestamp(),
-    case TTL of
-        undefined ->
-            Val1 = Val;
+    NoValChange = TTL =:= undefined orelse Val =:= undefined,
```
jtuple: Here, and other locations in this file, where you write `A =:= B orelse C =:= D` I would suggest going with `(A =:= B) orelse (C =:= D)`. Slightly easier to read. But, really, it's because Erlang fails to compile `A =:= B or C =:= D` (`or` vs `orelse`), and it's too easy for someone to accidentally (or intentionally for some future reason) change the `or` to `orelse` down the road and wonder why Erlang suddenly complains.

src/riak_kv_vnode.erl
```diff
@@ -501,10 +521,53 @@ handle_command({rehash, Bucket, Key}, _, State=#state{mod=Mod, modstate=ModState
...
+            IndexSpecs = riak_object:diff_index_data(OldIdxData, IdxData),
+            {Reply, UpModState} = 
+            case Mod:put(Bucket, Key, IndexSpecs, undefined, ModState) of
+                {ok, ModState2} ->
+                    riak_kv_stat:update(vnode_index_refresh),
+                    {ok, ModState2};
+                {error, Reason, ModState2} ->
+                    {{error, Reason}, ModState2}
+            end,
+            % TODO: Add index refresh stats
```
jtuple: I'm pretty sure you've done this TODO, no?

src/riak_kv_vnode.erl
```diff
@@ -1444,13 +1507,60 @@ do_diffobj_put({Bucket, Key}, DiffObj,
...
+                                          mod=Mod,
+                                          modstate=ModState}) ->
+    {ok, ModCaps} = Mod:capabilities(Bucket, ModState),
+    IndexCap = lists:member(indexes, ModCaps ),
+    BKey = {Bucket, Key},
+    IndexN = riak_kv_util:get_index_n(BKey),
+    IndexKey = term_to_binary(BKey),
+    IndexHash = riak_kv_index_hashtree:hash_object(BKey, RObj),
+    Item0 = {IndexN, IndexKey, IndexHash},
+    Items = case IndexCap of
```
jtuple: The word "Index" is crazily overloaded in this function. `IndexN` is properly named as the type is a 2-tuple that is `{partition-index(), n-val()}`, but `IndexKey`, `IndexHash`, (later on) `IndexKey2i, IndexData, IndexHash2i` have nothing to do with indexes. Well, the 2i stuff has to do with secondary indexes. Without 2i, `IndexKey` is still confusing, since there's no index involved. With the addition of 2i, things are even more confusing. `KeyBin`, `ObjHash` seem more correct terms for `IndexKey, IndexHash`, etc.

Or something else, dunno really. Would prefer different names as this function is very confusing to read as-is.

Also, I apologize for the unfortunate naming of `riak_kv_index_hashtree` since that likely encouraged the use of "Index" here. `riak_kv_index_hashtree` really should have been called `riak_kv_vnode_hashtree`, but we use the terms vnode, partition, and index (eg. partition index, eg. same thing as partition) interchangeably throughput the codebase. We should probably stop that someday.

src/riak_kv_2i_aae.erl
```diff
@@ -0,0 +1,434 @@
...
+%% @doc Create temporary DB holding 2i index data from disk.
+create_index_data_db(Partition, DutyCycle) ->
+    DBDir = filename:join(data_root(), "tmp_db"),
+    LockFile = filename:join(DBDir, "LOCK"),
+    case filelib:is_file(LockFile) of
+        true ->
+            file:delete(LockFile);
+        false ->
+            ok
+    end,
```
jtuple: I don't think we should be messing with LOCK files. LevelDB should eventually release the LOCK file. Yes, sometimes we'll need to wait, we've already changed the user facing interface to let them know about that.

jtuple:
Other than the comments in the diff:
- Why all the changes to `riak_kv_eleveldb_backend` and `riak_kv_memory_backend` that check for `undefined` values and whatnot? I'm not following how the changes made to support 2i AAE lead to undefined values.
- Great job adding the backend filter stuff and cleaning up the reformat fold. Not really related to this PR, but good clean up to see.
- As discussed in backchannel, `riak_kv_2i_aae` should be changed to be a `gen_fsm`. It's too large with too much bare message passing, receive loops, ad hoc monitoring, etc. As a `gen_fsm`, will be much easier to maintain.
- I'm pretty sure the use of `sync_command` isn't safe in the face of vnode crashes and/or overload protection kicking in. Of course, this isn't a "new to this PR" issue -- AAE uses `sync_command` today in Riak. I'll need to investigate/test, but we should fix that in a future PR if an issue.
- Not a huge fan of the refactoring done to `riak_kv_index_hashtree`, especially since it has little to do with this specific PR. Would have preferred the addition of `riak_kv_index_hashtree:insert_2i` or something rather than the generic `insert` interface that pushes index_n, hash calculation, etc out of `riak_kv_index_hashtree` and into `riak_kv_vnode`. At the very least, if time permitting, would prefer to see tuple tagging so at least Dialyzer can help keep us sorted. Ie:

``` erlang
    Item0 = {object, IndexN, KeyObj, HashObj},
    ...
    Items = [Item0, {index, Tree2i, Key2i, Hash2i}]
    riak_kv_index_hashtree:insert(Items, [], Trees)
```
- As discussed in backchannel, pretty sure we can make canceling a repair clean-up the LevelDB lock sooner by tracking the LevelDB reference somewhere. Even if this is as simple as having the 2i repair process put the LevelDB reference in it's process dictionary after opening it and pulling that out before killing things in `riak_console`. Something like this (with appropriate error handling, or maybe just a try/catch):

``` erlang
io:format("Will kill current 2i repair process\n", []),
Mon = monitor(process, riak_kv_2i_aae_repair),
{dictionary, PD} = process_info(Pid, dictionary),
{_, DBRef} = lists:keyfind(tmp_leveldb_ref, 1, PD),
eleveldb:close(DBRef),
exit(Pid, kill),
```

src/riak_kv_2i_aae.erl
engelsanchez: Yes, at one point I thought the existence of the file itself was the issue, but it was not. I'll remove this.

src/riak_kv_vnode.erl
engelsanchez: Yep, the put is the thing. Removing.

src/riak_kv_vnode.erl
engelsanchez: Yeah. Let me see if I can clear it up a little bit.

src/hashtree.erl
engelsanchez, line 405: Not really. I was just trying not to touch this repo-crossing file more than strictly necessary. Hence my not porting all the changes in 2.0 that came along compare/4.

engelsanchez:
The "undefined" related changes need better documentation because it's actually kind of big: you can now put an object with index specs, and if the value is undefined, you are basically telling the backend to update the index specs but not the value. It's a 2i data refresh put.

The FSM is on its way.

I like your proposed change to better guarantee leveldb cleanup, but it's a bit more complicated and would hate to try to do that at this hour. If we are still changing stuff on Monday, it will be in.  The indexDB is actually created by processes spawned from the 2i aae process, so we'll need to communicate the DB refs and/or store them globally and manage them. Easy to mess up.

engelsanchez:
@jtuple here it is. I won't even pretend I expect these many code changes to be right.  But it passes the riak_test (which has an accompanying change needed with this refactor, so pull that) and it looks more like an Erlang programmer wrote it.  Let me know what breakage you find.  I did not get to go over the riak_kv_index_hashtree changes and the riak_kv_vnode indexyIndex name cleanup.  I'm not sure if I'll have a chance to do that this weekend but I will try.

engelsanchez:
The last commit improves on the repair-2i command, which will now generate reports like: 

2i repair status is running:
    Total partitions: 64
    Finished partitions: 63
    Speed: 100
    Total 2i items scanned: 295322
    Total tree objects: 295322
    Total objects fixed: 80863

or

2i repair is not running.

The internal status returned by riak_kv_2i_aae:get_status() also has details on each partition's information.

engelsanchez:
Another commit: the 2i AAE FSM now has a stop command that will attempt to close leveldb instances that could potentially remain open for a long time if in a fold.  It is now much, much harder to hit the condition where starting another repair is not possible due to leveldb locks not dying soon enough.

engelsanchez:
Changed things a bit more to address review comments about the interface changes in riak_kv_index_hashtree and the presence of too much hashtree specific code in riak_kv_vnode that could be encapsulated in riak_kv_index_hashtree, as well as some overuse of the name 'Index' in that code.

Heads up that I'm chasing an issue found while adding deletes to the riak_test. A single deleted key seems to always be skipped for repair. It's always the same lonely key. I'm not sure yet what is causing that or whether it's a test problem.

engelsanchez:
Reverted the refactor to move code related to riak object handling to riak_kv_index_hasthree. It was wastefully copying the object from the vnode process to the hashtree process. I shouldn't be allowed to change this code anymore

I'm still chasing the delete problem.

engelsanchez:
@jtuple After some redbugging of the hashtree compare operation, it was obvious that the 2i AAE hashtree had not been updated correctly. The last commit fixes that and should make the updated riak_test pass finally. yay!  That was the last item in my list. If you see anything I've left out that you think I should address, please let me know.

src/riak_kv_2i_aae.erl
```diff
@@ -0,0 +1,669 @@
...
+    State2 = maybe_notify(State, {error, {no_aae_pid, Err}}),
+    {stop, no_aae_pid, State2};
+wait_for_aae_pid({ReqId, {ok, TreePid}},
+                 State=#state{aae_pid_req_id=ReqId, speed=Speed,
+                              remaining=[Partition|_]}) ->
+    WorkerFun =
+    fun() ->
+            Res =
+            repair_partition(Partition, Speed, ?MODULE, TreePid),
+            gen_fsm:sync_send_event(?MODULE, {repair_result, Res})
```
jtuple: Is the default 5 second timeout here desired? Do we want infinity? Something longer?

src/riak_kv_2i_aae.erl
```diff
@@ -0,0 +1,669 @@
...
+     State#state{open_dbs=lists:delete(DBRef, DBs)}}.
+
+%% @doc Performs the actual repair work, called from a spawned process.
+-spec repair_partition(integer(), integer(), {pid(), any()}, pid()) ->
+    #partition_result{}.
+repair_partition(Partition, DutyCycle, From, TreePid) ->
+    case get_hashtree_lock(TreePid, ?LOCK_RETRIES) of
+        ok ->
+            lager:info("Acquired lock on partition ~p", [Partition]),
+            gen_fsm:sync_send_event(From, {lock_acquired, Partition}),
```
jtuple: Is the default 5 second timeout here desired? Do we want infinity? Something longer?

src/riak_kv_2i_aae.erl
```diff
@@ -0,0 +1,669 @@
...
+    Status = to_simple_state(State),
+    Report = to_report(Status),
+    lager:info("Finished 2i repair:\n~s", [Report]),
+    {stop, normal, State};
+next_partition(timeout, State=#state{remaining=[Partition|_]}) ->
+    ReqId = make_ref(),
+    riak_core_vnode_master:command({Partition, node()},
+                                   {hashtree_pid, node()},
+                                   {fsm, ReqId, self()},
+                                   riak_kv_vnode_master),
```
jtuple: Probably good to keep the `request_hashtree_pid` API encapuslated in `riak_kv_vnode.erl`. We already have a precedent with `get/3, put/6, etc` of providing a `Sender` in the API call. Perhaps add a `request_hashtree_pid/2` that has a `Sender` argument.

``` erlang
%% In riak_kv_vnode.erl
request_hashtree_pid(Partition) ->
    ReqId = {hashtree_pid, Partition},
    Sender = {raw, ReqId, self()},
    request_hashtree_pid(Partition, Sender).

request_hashtree_pid(Partition, Sender) ->
    riak_core_vnode_master:command({Partition, node()},
                                   {hashtree_pid, node()},
                                   Sender,
                                   riak_kv_vnode_master).

%% In riak_kv_2i_aae.erl
next_partition(timeout, State=#state{remaining=[Partition|_]}) ->
    ReqId = make_ref(),
    riak_kv_vnode:request_hashtree_pid(Partition, {fsm, ReqId, self()}),
    ...
```

jtuple:
Unless I missed something, it appears there is no mechanism to kill the 2i repair worker pid if the `riak_kv_2i_aae` FSM process dies. Seems like the async worker would continue if the FSM process died, as well as if we ever fallback to the "force kill" approach for `riak-admin repair-2i kill` command. Should we just use a `spawn_link` for the async worker rather than a `spawn`?

jtuple:
The more I think about things, the more I worry about moving the AAE hashing to the K/V vnode. The benefit of computing the hash in the vnode is that we don't copy the riak_object to the K/V hashtree process. However, the majority of data in a `#riak_object{}` record are off-heap binaries that aren't copied between processes -- just pointer copies. Thus, copying is mostly constant regardless of actual object size. On the other hand, computing the hash is dependent on the object size. And we have known issues in the past with hashing (MD5, SHA, whatever) causing scheduler issues. That's mostly alleviated now with the incremental hash, but I still worry about having the already overloaded and not-scheduler-friendly (due to backends using NIFs) vnodes doing even more work. We also lose the ability for the hashtree process to be doing hashes in parallel with the vnode, as well as counting against a different reduction count.

I think that if we're going to make this change, we really need to do some detailed performance testing. Unless @evanmcc has some insight I don't that could help alleviate my fears. On the other hand, changing this PR back to copying the object to the hashtree process returns things to the status quo -- it can't be any worse than things already are, so no need to worry about performance testing.

jtuple:
I've ran the related `riak_test` multiple times against memory, LevelDB, and memory + LevelDB. Things worked as intended.

I've also done a lot of manual testing. Mucking with indexes and data by directly modifying LevelDB backend data via LevelDB API. All things worked as expected. Also, did some evil things like killing hashtree processes and disabling AAE while 2i repair was occuring -- we get lots of log spam (let it crash!) but, otherwise, things fail as I'd expect and nothing breaks.

Decided to extend the `riak_test` a bit as well, and pushed a the [feature/2i-aae-1.4-jdb](https://github.com/basho/riak_test/tree/feature/2i-aae-1.4-jdb) branch. Up to you if we want to merge or not. Just some things I was manually testing that was easier to just throw in the test.

Besides the above comments, nothing else to add.  +1 on merging after we decide if/how we want to address those remaining concerns.

engelsanchez:
@jtuple I must have lost the linking of the worker in the FSM refactor. That was the intention. Everything should eventually die if the FSM dies.  With that, I think the workers sync sending events to the FSM with infinity timeout will make sense. I'm OK with the object copying given that it is what used to happen anyway. I will reinstate the code I commited, then deleted that moved the hashing bits into riak_kv_index_hashtree. Hang on to your hat

engelsanchez:
The last commit should address the last batch of review comments. @jtuple most of the changes are in riak_kv_index_hashtree. Hashing has been moved there from the vnode, but maybe you have some opinions about the actual way it's done. Let me know.

src/riak_kv_index_hashtree.erl
```diff
@@ -510,36 +548,88 @@ valid_time({X,Y,Z}) when is_integer(X) and is_integer(Y) and is_integer(Z) ->
...
+
+expand_items(HasIndex, Items) ->
+    lists:foldl(fun(I, Acc) ->
+                        expand_item(HasIndex, I, Acc)
+                end, [], Items).
+
+expand_item(Has2ITree, {object, BKey, RObj}, Others) ->
+    IndexN = riak_kv_util:get_index_n(BKey),
+    BinBKey = term_to_binary(BKey),
+    ObjHash = riak_kv_index_hashtree:hash_object(BKey, RObj),
```
jtuple: Now that the code is back in `riak_kv_index_hashtree`, we should remove the `riak_kv_index_hashtree:` prefix and just call the local function.

src/riak_kv_index_hashtree.erl
```diff
@@ -510,36 +548,88 @@ valid_time({X,Y,Z}) when is_integer(X) and is_integer(Y) and is_integer(Z) ->
...
+    Item0 = {IndexN, BinBKey, ObjHash},
+    case Has2ITree of
+        false ->
+            [Item0 | Others];
+        true ->
+            IndexData = riak_object:index_data(RObj),
+            Hash2i =  riak_kv_index_hashtree:hash_index_data(IndexData),
```
jtuple: Now that the code is back in `riak_kv_index_hashtree`, we should remove the `riak_kv_index_hashtree:` prefix and just call the local function.

jtuple:
Is there any value with having the `{object, ...}` prefix in `riak_kv_index_hashtree` now that you've changed the API so that we only ever pass in the objects as items and derive the 2i index specs inside `riak_kv_index_hashtree`? Not a big deal really, just curious on your opinion.

engelsanchez:
Hmmm. You're right. This must come from a time when I thought the riak_kv_index_hashtree API would be used by the 2i AAE code to update index specs directly, instead of the implemented solution of the refresh index data command.  Man what a long circle. Let me look at a diff against what it used to be and see if I can clean it up and restored it as close as possible.

engelsanchez:
OK, as agreed we're going to forget further style changes. We can polish when we get to 2.0 :). Module prefix change is done. I hope this is it!

jtuple:
Merge away! :+1: :beer: :dancer:

engelsanchez:
This longwinded branch has been squashed. All the original commits can be found here (for a while until it's cleaned up): https://github.com/basho/riak_kv/tree/feature/2i-aae-1.4-orig. And now, the merge

PR #737, opened 2013-11-25
Use basho/sext which has the NIF

engelsanchez:
@jaredmorrow This is a largely symbolic PR. All we need to do is bring basho/sext to 1.4 and tag it as necessary, etc. I'll leave it up to you exactly how you want to do it.


jaredmorrow:
+1, will lock down to a tag when all PR's are ready.

slfritchie:
+1

jaredmorrow:
Not remembering there was a PR, I already made this change on the command line for 1.4.4.  I'm going to close this now.

PR #738, opened 2013-11-27
Make 2i pagination sort optional - 1.4 branch version

engelsanchez:
This addresses the performance penalty from the new default of sorting
2i results since pagination was added. When paginating, (max_results
set) sorting is always enabled. But when not, it will only be used if
configured in app.config or specified in the client request.

This is a backport to 1.4 of https://github.com/basho/riak_kv/pull/671 with two main differences:
- The parameter name has been changed from 'sort' to 'pagination_sort' to help avoid confusion with its meaning since we don't use a "natural" sort based on keys, but rather (term, key) as is required by pagination.
- The call to app_helper:get_env to get the default for the global switch has an in-code default of false. This is necessary so that it defaults to false without any config. The 2.0 version was relying on cuttlefish, but even there it's necessary for deployments with legacy configuration files.

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -522,9 +522,10 @@ get_index(Bucket, Query) ->
 get_index(Bucket, Query, Opts) ->
     Timeout = proplists:get_value(timeout, Opts, ?DEFAULT_TIMEOUT),
     MaxResults = proplists:get_value(max_results, Opts, all),
+    PgSort = proplists:get_value(pagination_sort, Opts),
     Me = self(),
     ReqId = mk_reqid(),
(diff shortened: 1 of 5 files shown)


russelldb:
Is there a PR for riak that adds the new app.config param to the config (even if only commented out and documented?)

engelsanchez:
Not currently. I'll add it in a bit.

engelsanchez:
@russelldb The last commit adds a special case for requests with a continuation but without max_results. Like you pointed out, the user would expect normal pagination behavior there, so the sort is applied.

russelldb:
Yup! Bit of fiddling with the riak_tests and they all still pass, can see clearly that without the new param results aren't sorted, and with they are. As discussed in hipchat, a new graph from @angrycub would be awesome. :+1:

PR #742, opened 2013-12-02
Pass crdt_op to vnode on read-repair.

seancribbs:
This allows another opportunity for CRDTs to merge, instead of simply
creating more siblings on read-repair.

/cc @russelldb


russelldb:
WFM :+1:

PR #743, opened 2013-12-02
Fix vnode sending > max_results items for 2i query

engelsanchez:
This change sends the max_results value in the 2i query to the vnode,
which then may be used as the batch size if < than the default batch
size.  Before, a batch would be sent with possibly more items than the
index FSM would ever send back to the client.

/cc THE @russelldb


russelldb:
WFM. Tests pass, code makes sense. Intuitively it is a performance improvement (less work, right?) would love to see some figures, but not so much that I'll actually run the benchmarks myself today. :+1:

PR #745, opened 2013-12-04
Add CRDT stats

seancribbs:
- [x] datatype-specific stats
- [x] segregated stats for get FSM
- [x] segregated stats for put FSM
- [x] vnode-internal stats

Requires basho/riak_dt#73.


seancribbs:
Ready for review! @russelldb

src/riak_kv_crdt.erl
```diff
@@ -91,6 +93,21 @@ crdt_value(Type, error) ->
 crdt_value(Type, {ok, {_Meta, ?CRDT{mod=Type, value=Value}}}) ->
     {get_context(Type, Value), Type:value(Value)}.
 
+crdt_stats(_, error) -> [];
+crdt_stats(Type, {ok, {_Meta, ?CRDT{mod=Type, value=Value}}}) ->
+    case lists:member({stat,2}, Type:module_info(exports)) of
+        true ->
+            EnabledStats = application:get_env(riak_kv, datatype_stats, ?DATATYPE_STATS_DEFAULTS),
```
russelldb: Is there an associated cost with calling `application:get_env` on every request? Do we know what it is? Is it too damn high?

seancribbs: That's a good question, but I believe we do it for other things. It's essentially an ETS lookup.

engelsanchez: I've done some heavy DTracing of this, and it doesn't appear to be very high. Commonly ~5 microseconds per call (it's actually two ETS lookups). That is BTW, better than the average dict lookup .  There is a plan to possibly optimize this further through the app_helper:get_env(). I would suggest to always prefer that interface to a raw application:get_env() to take advantage of any future optimizations.

seancribbs: Just checked, there are a number of things we lookup in the request path. Here's a gist (not all of these are in the request path): https://gist.github.com/seancribbs/7806229

seancribbs: Thanks @engelsanchez, I'll switch it out.

russelldb:
I'm still going on with this. And starting some testing by hand. I notice there is no addition to `riak_kv_stat_bc.erl` so these new stats won't get in the "blob"

Is that something we want? Or are these attach-and-query only stats?
Shall we add a new `/stats/crdt` endpoint just for them, rather than further embiggening the blob?

seancribbs:
@russelldb Honestly I just didn't know that was necessary to modify, thank you for bringing it to my attention.

russelldb:
@seancribbs are you going to blob up the stats? Shall I?

lenary:
I'm going to try do the backwards compat stuff.
- [x] riak_kv_stats_bc

russelldb:
Ok. I tested by hand (we should probably note you need 5+ reads to register the Actor stats) and all seems well. I removed the `vnode_dt_merge` stats and replaced them with `riak_object_merge` stats. We lose the code location of the merge, but gain a general merge time stats.

I'm plus one on the work you've done. Very nice.

Just need to you cast your eye over the changes mentioned above, too.

seancribbs:
Thanks, @russelldb!

lenary:
Why do we need 5+ reads to register the actor stats? just a rate limit so that we're not getting the stat every single time?

russelldb:
Dunno.folsom.read it and weep.

seancribbs:
@lenary Sampling.

PR #746, opened 2013-12-04
Use a DVV like approach to stop sibling explosion

russelldb:
As has been witnessed by our customers, and demonstrated by Baquero et al, sibling explosion occurs when we keep values as concurrent when they're not really.

Our vnode vclocks are not fine grained enough to describe which of a set of siblings should be overwritten by a PUT, so we keep all siblings. It is trivially easy to create a scenario where interleaving writes create an explosion of sibling values, even without failure, or concurrency (see basho/riak_test#383.)

This PR introduces a more fine grained approach to causality tracking, without going all the way and using DVV or DVVSets.

The general mechanism is to store the event (`{actor, {count, ts}}`) that a value is created at as metadata with the value. When a PUT is received, any siblings whose events are dominated by the PUT vclock are removed. Only siblings that the writer has not seen are genuinely concurrent.

In order not to drop ALL the values during other merge types, when any two merging objects both have the same event, that sibling stays.

Nothing is for free, so there is some extra complexity around `riak_object:merge`.

This PR requires the basho/riak_core#463 and can be tested with basho/riak_test#480.

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -347,7 +347,8 @@
 
 {mapping, "dvv", "riak_kv.dvv_enabled", [
   {default, on},
-  {datatype, {enum, [on, off]}}
+  {datatype, {enum, [on, off]}},
+  {level, advanced}
 ]}.
 
 { translation,
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -337,3 +337,26 @@
   {datatype, integer},
   {default, 100}
 ]}.
+
+%% @doc dvv is an enhancement to riak's logical clocks that prevents
+%% sibling explosion by storing an event in a value's metadata on
+%% creation. Riak later later uses that event information on merging
+%% conflicting writes to ensure that only genuinely concurrent values
+%% remain. DVV is enabled by default. As it is a new feature, you can
+%% switch it off.
+
+{mapping, "dvv", "riak_kv.dvv_enabled", [
+  {default, on},
+  {datatype, {enum, [on, off]}}
+]}.
+
+{ translation,
+  "riak_kv.dvv_enabled",
+  fun(Conf) ->
+    Setting = cuttlefish_util:conf_get_value("dvv", Conf),
+    case Setting of
+      on -> true;
+      off -> false;
+      _Default -> true
+    end
+  end}.
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1186,9 +1186,7 @@ handle_crdt(_, undefined, _VId, RObj) ->
 handle_crdt(true, CRDTOp, VId, RObj) ->
     riak_kv_crdt:update(RObj, VId, CRDTOp);
 handle_crdt(false, _CRDTOp, _Vid, RObj) ->
-    %% non co-ord put that was a CRDT operation? Merge the values if
-    %% there are siblings 'cos that is the point of CRDTs: no siblings
-    riak_kv_crdt:merge(RObj).
+    RObj.
 
 perform_put({fail, _, _}=Reply, State, _PutArgs) ->
     {Reply, State};
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1321,7 +1321,7 @@ put_merge(true, false, LocalObj, PutObj, VId, StartTime) -> %% Coordinating=true
             %% so merge the PUT object and the local object.
             MergedClock = vclock:merge([PutVC, LocalVC]),
             FrontierClock = vclock:increment(VId, StartTime, MergedClock),
-            Dot = vclock:get_entry(VId, FrontierClock),
+            {ok, Dot} = vclock:get_entry(VId, FrontierClock),
             %% Asign an event to the put value
             DottedPutObject = riak_object:assign_dot(PutObj, Dot),
             MergedObject = riak_object:merge(DottedPutObject, LocalObj),
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1292,6 +1292,9 @@ select_newest_content(Mult) ->
 put_merge(false, true, _CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=true
     {newobj, UpdObj};
 put_merge(false, false, CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=false
+    %% a downstream merge, or replication of a coordinated PUT
+    %% Merge the value received with local replica value
+    %% and store the value IIF it is different to what we already have
     ResObj = riak_object:syntactic_merge(CurObj, UpdObj),
     case riak_object:equal(ResObj, CurObj) of
         true ->
@@ -1301,21 +1304,28 @@ put_merge(false, false, CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=
     end;
 put_merge(true, true, _CurObj, UpdObj, VId, StartTime) -> % coord=true, LWW=true
     {newobj, riak_object:increment_vclock(UpdObj, VId, StartTime)};
-put_merge(true, false, CurObj, UpdObj, VId, StartTime) ->
-    UpdObj1 = riak_object:increment_vclock(UpdObj, VId, StartTime),
-    UpdVC = riak_object:vclock(UpdObj1),
-    CurVC = riak_object:vclock(CurObj),
+put_merge(true, false, LocalObj, PutObj, VId, StartTime) -> %% Coordinating=true, LWW=false, local replica has value for key
+    %% Get the vclock we have the object stored on disk at this replica
+    LocalVC = riak_object:vclock(LocalObj),
+    %% get the vclock from the put
+    PutVC = riak_object:vclock(PutObj),
 
-    %% Check the coord put will replace the existing object
-    case vclock:get_counter(VId, UpdVC) > vclock:get_counter(VId, CurVC) andalso
-        vclock:descends(CurVC, UpdVC) == false andalso
-        vclock:descends(UpdVC, CurVC) == true of
+    %% Optimisation: if the put object's vclock descends from the
+    %% local object's vclock, then don't merge with local value, just
+    %% increment the clock and overwrite.
+    case vclock:descends(PutVC, LocalVC) of
         true ->
-            {newobj, UpdObj1};
+            {newobj, riak_object:increment_vclock(PutObj, VId, StartTime)};
         false ->
-            %% If not, make sure it does
-            {newobj, riak_object:increment_vclock(
-                       riak_object:merge(CurObj, UpdObj1), VId, StartTime)}
+            %% The PUT object is concurrent with some other PUT,
+            %% so merge the PUT object and the local object.
+            MergedClock = vclock:merge([PutVC, LocalVC]),
+            FrontierClock = vclock:increment(VId, StartTime, MergedClock),
+            Dot = vclock:get_entry(VId, FrontierClock),
+            %% Asign an event to the put value
+            DottedPutObject = riak_object:assign_dot(PutObj, Dot),
+            MergedObject = riak_object:merge(DottedPutObject, LocalObj),
+            {newobj, riak_object:set_vclock(MergedObject, FrontierClock)}
     end.
 
 %% @private
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1292,7 +1292,7 @@ put_merge(false, true, _CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=
 put_merge(false, false, CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=false
     %% a downstream merge, or replication of a coordinated PUT
     %% Merge the value received with local replica value
-    %% and store the value IIF it is different to what we already have
+    %% and store the value IFF it is different to what we already have
     ResObj = riak_object:syntactic_merge(CurObj, UpdObj),
     case riak_object:equal(ResObj, CurObj) of
         true ->
@@ -1320,7 +1320,7 @@ put_merge(true, false, LocalObj, PutObj, VId, StartTime) -> %% Coordinating=true
             MergedClock = vclock:merge([PutVC, LocalVC]),
             FrontierClock = vclock:increment(VId, StartTime, MergedClock),
             {ok, Dot} = vclock:get_entry(VId, FrontierClock),
-            %% Asign an event to the put value
+            %% Assign an event to the put value
             DottedPutObject = riak_object:assign_dot(PutObj, Dot),
             MergedObject = riak_object:merge(DottedPutObject, LocalObj),
             {newobj, riak_object:set_vclock(MergedObject, FrontierClock)}
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -1027,6 +1027,14 @@ is_updated_test() ->
     OVu = riak_object:update_value(O, testupdate),
     ?assert(is_updated(OVu)).
 
+remove_duplicates_test() ->
+    O0 = riak_object:new(<<"test">>, <<"test">>, zero),
+    O1 = riak_object:new(<<"test">>, <<"test">>, one),
+    ND = remove_duplicate_objects([O0, O0, O1, O1, O0, O1]),
+    ?assertEqual(2, length(ND)),
+    ?assert(lists:member(O0, ND)),
+    ?assert(lists:member(O1, ND)).
+
 new_with_ctype_test() ->
     O = riak_object:new(<<"b">>, <<"k">>, <<"{\"a\":1}">>, "application/json"),
     ?assertEqual("application/json", dict:fetch(?MD_CTYPE, riak_object:get_metadata(O))).
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -286,7 +286,7 @@ merge_contents(NewContents, OldContents, NewClock, OldClock) ->
 
 
 fold_contents({r_content, Dict, _Value}=C0, {DropContents, KeepContents}, Clock) ->
-    case dict:find(?DOT, Dict) of
+    case get_dot(Dict) of
         {ok, Dot} ->
             case {vclock:descends(Clock, [Dot]), orddict:is_key(Dot, DropContents)} of
                 {true, true} ->
@@ -301,9 +301,25 @@ fold_contents({r_content, Dict, _Value}=C0, {DropContents, KeepContents}, Clock)
                     Content = convert_to_dict_list(C0),
                     {DropContents, [Content | KeepContents]}
             end;
-        error ->
+        undefined ->
             Content = convert_to_dict_list(C0),
             {DropContents, [Content |  KeepContents]}
+    end.
+
+%% @private Get the dot from the passed metadata dict
+%% (if present and valid).
+-spec get_dot(dict()) -> {ok, vclock:dot()} | undefined.
+get_dot(Dict) ->
+    case dict:find(?DOT, Dict) of
+        {ok, Dot} ->
+            case vclock:valid_entry(Dot) of
+                true ->
+                    {ok, Dot};
+                false ->
+                    undefined
+            end;
+        error ->
+            undefined
     end.
 
 %% @doc Convert a r_content's inner dictionary to a list, and sort it to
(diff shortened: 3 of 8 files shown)


src/riak_object.erl
```diff
@@ -387,12 +505,52 @@ set_vclock(Object=#r_object{}, VClock) -> Object#r_object{vclock=VClock}.
...
+dot_key() ->
+    ?DOT.
+
+%% @private assign the dot to the value only if DVV is enabled. Only
+%% call with a valid dot. Only assign dot when there is a single value
+%% in contents.
+-spec assign_dot(riak_object(), vclock:dot(), boolean()) -> riak_object().
+assign_dot(Object, Dot, true) ->
+    #r_object{contents=[C=#r_content{metadata=Meta0}]} = Object,
+    %% Dot must actually be a vclock, allbeit it one with a single entry
```
russelldb: oops, this comment was left by mistake, and needs removing

src/riak_object.erl
```diff
@@ -239,30 +263,124 @@ compare_content_dates(C1,C2) ->
...
+%% remove dominated siblings and stop fake concurrency that causes
+%% sibling explsion. Also, since every sibling is iterated over (some
+%% twice!) why not merge CRDTs here, too?
+merge_contents(NewObject, OldObject, true) ->
+    Bucket = bucket(NewObject),
+    Key = key(NewObject),
+    MergeAcc0 = prune_object_siblings(OldObject, vclock(NewObject)),
+    MergeAcc = prune_object_siblings(NewObject, vclock(OldObject), MergeAcc0),
+    #merge_acc{crdt=CRDT, error=Error} = MergeAcc,
+    riak_kv_crdt:log_merge_errors(Bucket, Key, CRDT, Error),
```
Vagabond, line 293: Do we consider a classic riak_object a CRDT, or are we just piggybacking on handy functions from the CRDT module?

russelldb, line 293: Previously, we merged CRDTs in the vnode on a CRDT Operation PUT only. This means CRDT siblings can be generated by repl, AAE, and Read Repair. CRDT values can't be dotted: There is no single PUT to overwrite previous values, instead a merge of values generates  a new value. In DVVSets they have the "anonymous list" for these generated values. In Riak it makes sense to have them undotted, as undotted siblings always survive. However, if CRDTs are undotted, then they can "explode" via Repl. To avoid this, we merge crdts in riak_object merge. It has other advantages too (mainly that yokozuna can index CRDTs). It is also a slight optimization on the CRDT path. The main thing is that this PR has DVV _and_ riak CRDT merging in riak_object merge.

russelldb, line 293: Oops, the code you refer to just logs any CRDT merge issues encountered during the merge operation. It reuses the existing CRDT code, yes.

src/riak_object.erl
```diff
@@ -168,12 +182,18 @@ equal_contents([C1|R1],[C2|R2]) ->
 ancestors(pure_baloney_to_fool_dialyzer) ->
     [#r_object{vclock = vclock:fresh()}];
 ancestors(Objects) ->
-    ToRemove = [[O2 || O2 <- Objects,
-                       vclock:descends(O1#r_object.vclock,O2#r_object.vclock),
-                       (vclock:descends(O2#r_object.vclock,O1#r_object.vclock) == false)]
```
Vagabond, line 173: This is subtly different than the vclock:dominates() call, in that dominates() checks if the vclocks are equal. This is fine, right?

russelldb, line 173: They're the same, the only way two vclocks could descend each other is if they were equal. I think `vclock:dominates/2` makes that clearer (although, IMO, equals check first might be an optimisation.)

src/riak_object.erl
```diff
@@ -182,22 +202,16 @@ ancestors(Objects) ->
 %%       X-Riak-Last-Modified header.
 -spec reconcile([riak_object()], boolean()) -> riak_object().
 reconcile(Objects, AllowMultiple) ->
```
Vagabond, line 201: I'm getting a little lost here. It would be nice to have comments for dedup and reconcile/2 and reconcile2/2. I sort of understand it, but would like to feel more confident.

russelldb:
@Vagabond pushed a commit that removes dedupe logic and adds comments.

src/riak_object.erl
```diff
@@ -182,34 +202,42 @@ ancestors(Objects) ->
...
+%% merged riak_object remains that contains all sibling values. Only
+%% called with a list of conflicting objects. Use `remove_dominated/1'
+%% to get such a list first.
+reconcile([Object]) ->
+    %% Only one object? Then we're reconciled already!
+    Object;
+reconcile([Object1, Object2 | Rest]) ->
+    %% Start reconciling merging the first two objects in the list to
+    %% get a most merged object.
+    reconcile2(Rest, syntactic_merge(Object1, Object2)).
```
Vagabond: Maybe something like this would be clearer than reconcile2/2?

``` erlang
lists:foldl(fun(Obj, MergeDest) ->
    syntactic_merge(Obj, MergeDest)
end, hd(Objects), tl(Objects))
```

russelldb: except the single object in a list case?

Vagabond: that should work:

``` erlang
lists:foldl(fun(E, A) -> E + A end, hd([1]), tl([1])).
1
```

But I meant to replace reconcile2/2 with it, I typoed originally.

russelldb: Okay, will try that. I consider it stylistic, but if it makes it easier to read, then that is good.

seancribbs: Single-object list case will still work in @Vagabond's version.

``` erlang
lists:foldl(fun syntactic_merge/2, hd(Objects), tl(Objects))
```

This is the definition of lists:foldl:

``` erlang
foldl(F, Accu, [Hd|Tail]) ->
    foldl(F, F(Hd, Accu), Tail);
foldl(F, Accu, []) when is_function(F, 2) -> Accu.
```

russelldb: Sorry, I forgot the order of the args (it is Fun, Acc, List, I thought it was Fun, List, Acc, ('cos the Fun takes Element, Acc (which is more like the order List, Acc))) I blame the lists lib :D

seancribbs: Indeed. My rule of thumb is "list last".

evanmcc:
these are n=1 numbers:
develop:
get 1 avg = 3264, gets [2865,3477,3452]
get 128 avg = 46938, gets [48297,44438,48079]
put 1 avg = 1797, puts [1788,1827,1778]
put 128 avg = 32391, puts [32608,32407,32158]

dvv @ f00aac8
get 1 avg = 4108, gets [4137,4032,4157]
get 128 avg = 56012, gets [56413,55948,55677]
put 1 avg = 1709, puts [1670,1644,1814]
put 128 avg = 33558, puts [33621,33648,33405]

dvv @ dcca904:
get 1 avg = 4090, gets [4216,4116,3938]
get 128 avg = 56423, gets [56834,56278,56158]
put 1 avg = 1806, puts [1816,1842,1760]
put 128 avg = 33463, puts [33835,33357,33198]

src/riak_object.erl
```diff
@@ -239,30 +267,127 @@ compare_content_dates(C1,C2) ->
...
+%% remove duplicate values, and merge CRDT values down to a single
+%% value.
+-spec fold_contents(r_content(), merge_acc(), vclock:vclock()) -> merge_acc().
+fold_contents({r_content, Dict, Value}=C0, MergeAcc, Clock) ->
+    #merge_acc{drop=Drop, keep=Keep, crdt=CRDT, error=Error} = MergeAcc,
+    case get_dot(Dict) of
+        {ok, Dot} ->
+            case {vclock:descends(Clock, [Dot]), orddict:is_key(Dot, Drop)} of
+                {true, true} ->
+                    %% Dot present on both sides, keep it
```
Vagabond: I know we talked about this logic, but more comments here about this logic would be nice.

src/riak_object.erl
```diff
@@ -239,30 +267,127 @@ compare_content_dates(C1,C2) ->
...
+                    MergeAcc#merge_acc{crdt=CRDT2}
+            end
+    end.
+
+%% @private Transform a `merge_acc()' to a list of `r_content()'
+-spec merge_acc_to_contents(merge_acc()) -> list(r_content()).
+merge_acc_to_contents(MergeAcc) ->
+    #merge_acc{keep=Keep, crdt=CRDTs} = MergeAcc,
+    Keep2 = lists:map(fun convert_from_dict_list/1, ordsets:to_list(Keep)),
+    orddict:fold(fun(_Type, {Meta, CRDT}, Acc) ->
```
Vagabond, line 415: _Type is always r_content, right? Should we use Type on the next line instead of having that r_content literal, or pattern match on r_content so we can't accidentally strip type information if more types are added later?

src/riak_object.erl
```diff
@@ -239,30 +267,127 @@ compare_content_dates(C1,C2) ->
...
+                         [{r_content, riak_kv_crdt:meta(Meta, CRDT),
+                           riak_kv_crdt:to_binary(CRDT)} | Acc]
```
Vagabond, line 417: I'm assuming that for regular riak_objects, the 'mod' in the CRDT record is riak_object? All these CRDT references in riak_object are kind of freaking me out.

src/riak_object.erl
```diff
@@ -239,30 +267,127 @@ compare_content_dates(C1,C2) ->
...
+                 end,
+                 Keep2,
+                 CRDTs).
+
+%% @private Get the dot from the passed metadata dict
+%% (if present and valid).
+-spec get_dot(dict()) -> {ok, vclock:dot()} | undefined.
+get_dot(Dict) ->
```
Vagabond, line 425: Are there known cases when either the dot is not valid, or is not present in the vclock? If so, can we document them? The only time I can think of is if a vclock is pruned...?

src/riak_object.erl
```diff
@@ -239,30 +267,127 @@ compare_content_dates(C1,C2) ->
...
+            end;
+        error ->
+            undefined
+    end.
 
 %% @doc Convert a r_content's inner dictionary to a list, and sort it to
 %%      ensure we can do comparsions between r_contents.
 convert_to_dict_list({r_content, Dict, Value}) ->
+    {r_content, lists:usort(dict:to_list(Dict)), Value};
+convert_to_dict_list({Dict, Value}) ->
```
Vagabond: When do we get untyped objects like this?

src/riak_object.erl
```diff
@@ -959,14 +1127,6 @@ is_updated_test() ->
     OVu = riak_object:update_value(O, testupdate),
     ?assert(is_updated(OVu)).
 
-remove_duplicates_test() ->
```
Vagabond, line 962: Was this test removed for a reason?

src/riak_object.erl
```diff
@@ -1036,4 +1196,52 @@ vclock_codec_test() ->
...
+    ?assertEqual(3, riak_object:value_count(ACZ)),
+    Contents = lists:sort(riak_object:get_contents(ACZ)),
+    verify_contents(Contents, [{{a, 2}, <<"a1">>}, {{c, 2}, <<"c1">>}, {{z, 2}, <<"z1">>}]),
+    Vclock = riak_object:vclock(ACZ),
+    ?assertEqual(2, vclock:get_counter(a, Vclock)),
+    ?assertEqual(2, vclock:get_counter(c, Vclock)),
+    ?assertEqual(2, vclock:get_counter(z, Vclock)),
+    ?assertEqual(3, length(Vclock)).
+
+mixed_merge_test() ->
```
Vagabond, line 1263: The mixed_merge tests could use some comments to clear up what we're actually testing here.

src/riak_object.erl
```diff
@@ -387,12 +512,52 @@ set_vclock(Object=#r_object{}, VClock) -> Object#r_object{vclock=VClock}.
 -spec increment_vclock(riak_object(), vclock:vclock_node()) -> riak_object().
 %% @spec increment_vclock(riak_object(), vclock:vclock_node()) -> riak_object()
 increment_vclock(Object=#r_object{}, ClientId) ->
-    Object#r_object{vclock=vclock:increment(ClientId, Object#r_object.vclock)}.
+    NewClock = vclock:increment(ClientId, Object#r_object.vclock),
+    {ok, Dot} = vclock:get_entry(ClientId, NewClock),
+    assign_dot(Object#r_object{vclock=NewClock}, Dot, dvv_enabled()).
```
Vagabond, line 567: calling assign_dot/3 here instead of assign_dot/2 skips the vclock:valid_entry check, is this intentional?

Vagabond:
It looks like the code paths where dvv is NOT enabled are not covered by the eunit suite in riak_object anymore. Should we reinstate some tests for the old codepath as well?

Vagabond:
It also looks like the ec_eqc test no longer excersises the DVV=false code path, either.

src/riak_object.erl
russelldb, line 962: The code it tests was removed.

src/riak_object.erl
russelldb, line 415: No. Type is a CRDT type (riak_dt_pncounter, riak_dt_map, riak_dt_orswot etc). The orddict that is being mapped over is a mapping of Type-> {Meta, CRDT value}. We don't need the type in this fold, so we don't use it.

src/riak_object.erl
russelldb, line 417: Regular riak objects are not gathered in the orddict. They're dealt with separately.

src/riak_object.erl
russelldb, line 425: None that I can think of. Do you think we should a) log here or b) just let it crash at the point the dot is compared to the clock? c) crash here?

src/riak_object.erl
russelldb: Never, it is from a previous version. I'll remove it.

src/riak_object.erl
russelldb, line 567: Yes. The comment explains it. assign/2 is an exported function, so it can't trust the dot. assign/3 is called inside riak_object, and only with a valid dot that has just been created. It says so in the comments.

src/riak_object.erl
russelldb, line 417: Can you explain your issue with CRDT code in riak_object? We have CRDTs now. There is no pure separation between riak_object and CRDTs. The merging of contents is what this function does. CRDT just means "non-opaque content item". Merge has extended what it does. I don't understand the issue. Is it some "law of demeter" thing, or a general squeamish-ness? We could merge the riak object contents, then pass the object to riak_kv_crdt and merge it again. The munging has occurred for optimisation reasons (like the comments says.) If we're iterating over the contents, don't do it again in riak_kv_crdt.  This commit comment, hopefully explains it well enough https://github.com/basho/riak_kv/commit/18fae0511b56ee9e65dcb74c21c401e44a7e32e8

russelldb:
I've addressed the review comments in the last commit.

Looking at coverage there is still no test for merging CRDTs in riak_object (although the CRDT merge logic is tested elsewhere), even though we are short on time I think we really need this, and maybe even an explicit eqc test for `reconcile/2`, `merge/2`, and `syntactic_merge/2`. I'll let the reviewer decide if that test is needed _before_ we merge this and freeze on Friday, or after.

src/riak_object.erl
seancribbs, line 173: Suggested simplification (multiple generators are same as nested comprehensions):

``` erlang
[ O2 || O1 <- Objects, O2 <- Objects, strict_descendant(O1, O2) ]
```

src/riak_object.erl
```diff
@@ -182,34 +202,33 @@ ancestors(Objects) ->
...
-    end.
+    sets:to_list(sets:subtract(All, Del)).
+
+%% @private pairwise merge the objects in the list so that a single,
+%% merged riak_object remains that contains all sibling values. Only
+%% called with a list of conflicting objects. Use `remove_dominated/1'
+%% to get such a list first.
+reconcile(Objects) ->
+    lists:foldl(fun(O, Acc) ->
+                        syntactic_merge(O, Acc)
```
seancribbs: If you're not changing argument order, this is shorter and clearer:

``` erlang
fun syntactic_merge/2
```

priv/riak_kv.schema
```diff
@@ -337,3 +337,26 @@
...
+%% @doc dvv is an enhancement to riak's logical clocks that prevents
+%% sibling explosion by storing an event in a value's metadata on
+%% creation. Riak later later uses that event information on merging
+%% conflicting writes to ensure that only genuinely concurrent values
+%% remain. DVV is enabled by default. As it is a new feature, you can
+%% switch it off.
+
+{mapping, "dvv", "riak_kv.dvv_enabled", [
+  {default, on},
+  {datatype, {enum, [on, off]}}
```
seancribbs: Please add `{level, advanced}`. For new clusters, we want this behavior on and not directly exposed to new users.

src/riak_kv_vnode.erl
```diff
@@ -1292,6 +1290,9 @@ select_newest_content(Mult) ->
 put_merge(false, true, _CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=true
     {newobj, UpdObj};
 put_merge(false, false, CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=false
+    %% a downstream merge, or replication of a coordinated PUT
+    %% Merge the value received with local replica value
+    %% and store the value IIF it is different to what we already have
```
seancribbs: "IIF" -> "IFF"

src/riak_kv_vnode.erl
```diff
@@ -1301,21 +1302,28 @@ put_merge(false, false, CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=
...
         false ->
-            %% If not, make sure it does
-            {newobj, riak_object:increment_vclock(
-                       riak_object:merge(CurObj, UpdObj1), VId, StartTime)}
+            %% The PUT object is concurrent with some other PUT,
+            %% so merge the PUT object and the local object.
+            MergedClock = vclock:merge([PutVC, LocalVC]),
+            FrontierClock = vclock:increment(VId, StartTime, MergedClock),
+            {ok, Dot} = vclock:get_entry(VId, FrontierClock),
+            %% Asign an event to the put value
```
seancribbs: "Assign"

src/riak_object.erl
seancribbs, line 173: If you chose the simplification, you also don't have to flatten, since it generates a flat list already.

src/riak_object.erl
```diff
@@ -239,26 +258,187 @@ compare_content_dates(C1,C2) ->
...
+-spec prune_object_siblings(riak_object(), vclock:vclock(), merge_acc()) -> merge_acc().
+prune_object_siblings(Object, Clock, MergeAcc) ->
+    lists:foldl(fun(Content, Acc) ->
+                        fold_contents(Content, Acc, Clock)
+                end,
+                MergeAcc,
+                Object#r_object.contents).
+
+%% @private called once for each content entry for each object being
+%% merged. It's job is to drop siblings that are causally dominated,
```
seancribbs: It's / Its

src/riak_object.erl
```diff
@@ -239,26 +258,187 @@ compare_content_dates(C1,C2) ->
...
+%% Nuno Preguiça, Carlos Bauqero, Paulo Sérgio Almeida, Victor Fonte,
+%% and Ricardo Gonçalves. 2012
+-spec fold_contents(r_content(), merge_acc(), vclock:vclock()) -> merge_acc().
+fold_contents({r_content, Dict, Value}=C0, MergeAcc, Clock) ->
+    #merge_acc{drop=Drop, keep=Keep, crdt=CRDT, error=Error} = MergeAcc,
+    case get_dot(Dict) of
+        {ok, Dot} ->
+            case {vclock:descends(Clock, [Dot]), orddict:is_key(Dot, Drop)} of
+                {true, true} ->
+                    %% When the exact same dot present is present in
```
seancribbs: "present" twice

src/riak_object.erl
```diff
@@ -239,26 +258,187 @@ compare_content_dates(C1,C2) ->
...
+                    %% both objects siblings, we keep that
+                    %% value. Without this merging the an object with
+                    %% itself would result in an object with no value
+                    %% at all, since an object's vector clock always
+                    %% dominates all its dots.
+                    Content = convert_to_dict_list(C0),
+                    MergeAcc#merge_acc{keep=ordsets:add_element(Content, Keep)};
+                {true, false} ->
+                    %% The `dot' os dominated by the other object's
```
seancribbs: "os" / "is"

test/ec_eqc.erl
```diff
@@ -791,7 +802,7 @@ kv_vnode(#msg{from = From, c = {coord_put, ReqId, NewObj, Timestamp}},
     [Pri1, Pri2] = lists:sort([next_pri(), next_pri()]),
     WMsg = new_msg(Name, From, {w, Idx, ReqId}, Pri1),
     NodeId = {vc, Start, Node},
-    NewObj1 = riak_object:increment_vclock(NewObj, NodeId, ReqId),
+    NewObj1 = NewObj,%%riak_object:increment_vclock(NewObj, NodeId, ReqId),
```
seancribbs: Rebinding this is kind of pointless, why not just rename the variable through the rest of the function?

test/ec_eqc.erl
```diff
@@ -882,21 +893,30 @@ syntactic_put_merge(CurObj, UpdObj) ->
...
 
-coord_put_merge(undefined, UpdObj, _NodeId, _Timestamp) ->
-    UpdObj;
-coord_put_merge(CurObj, UpdObj, NodeId, Timestamp) ->
-    %% Make sure UpdObj descends from CurObj and that NodeId is greater
-    CurVC = riak_object:vclock(CurObj),
-    UpdVC = riak_object:vclock(UpdObj),
+coord_put_merge(undefined, UpdObj, NodeId, Timestamp) ->
+    riak_object:increment_vclock(UpdObj, NodeId, Timestamp);
+coord_put_merge(LocalObj, PutObj, NodeId, Timestamp) ->
```
seancribbs, line 897: I guess this is ok, but it feels like it mirrors the actual code too much.

seancribbs:
Some dialyzer warnings related to these changes that are worth fixing:

```
riak_object.erl:336: The call vclock:descends(Clock::vclock:vclock(),[vclock:dot(),...]) contains an opaque term as 2nd argument when terms of different types are expected in these positions
```

This seems to be the construction of a `vclock()` type from a `dot()` outside of the `vclock` module. (Note both `vclock()` and `dot()` are opaque.) vclock should have an exported function which is specifically for dots, or constructs a minimal clock from a dot.

```
riak_object.erl:445: Function convert_to_dict_list/1 has no local return
riak_object.erl:446: Record construction #r_content{metadata::[{_,_}]} violates the declared type of field metadata::'undefined' | dict()
riak_object.erl:449: Function convert_from_dict_list/1 has no local return
riak_object.erl:450: The call dict:from_list(Dict::'undefined' | dict()) contains an opaque term as 1st argument when a structured term of type [{_,_}] is expected
```

Ignore the "no local return" errors, this is trying to construct a dict from the atom `undefined` in some cases. Not necessarily your fault, but it is called by your changes.

```
riak_object.erl:597: Type specification riak_object:dot_key() -> binary() is a supertype of the success typing: riak_object:dot_key() -> <<_:24>>
```

Can this function just be a macro, avoiding a function call where it is used?

Aside from those dialyzer issues and the other things Andrew and I commented on, I'm :+1: to this change. Tests pass, code size is decreased and simplified overall.

Vagabond:
I'm generally +1 on this, I asked Sean to be a second reviewer, time permitting, so once you make Sean happy with this too, I think you're cleared for merge.

test/ec_eqc.erl
russelldb, line 897: Man, believe me when I say "I know." But do you really think I should re-write `ec_eqc` for this change? Jon and I discussed this, there are "issues" with this test in general. But this is not the time. If you feel strongly, open an issue, please.

russelldb:
Will fix the dialyzer problems, but please be aware that _all_ but the `dot` issues were added by some other code typist, m'kay? (I agree that it is right to fix these, but better still would be consistent review so one code typist didn't need to clean up after another as part of their general work.) I find your reviews to be amongst the most thorough (talking to you @seancribbs), can you jot down a list of how you perform a review, please? I think it would be valuable.

russelldb:
@seancribbs with regard to 

```
riak_object.erl:597: Type specification riak_object:dot_key() -> binary() is a supertype of the success typing:    riak_object:dot_key() -> <<_:24>>
```

> > Can this function just be a macro, avoiding a function call where it is used?

Do you mean a macro that is shared outside riak_object?

russelldb:
Hopefully that covers it. Thanks for the thorough review, gentlemen.

test/ec_eqc.erl
seancribbs, line 897: Yeah, it wasn't so much a "this has to change for this PR to land" but more like "ugh".

seancribbs:
@rdb I didn't realize the Dot key had to be shared outside riak_object. Where else is it called?

seancribbs:
@russelldb I'll send you some notes. I'm not sure I've codified it, really, but there are some typical things I do.

russelldb:
@seancribbs the ?DOT is used in riak_kv_crdt. CRDTs drop the dot.

seancribbs:
@russelldb I see now, thanks!

russelldb:
@Vagabond @seancribbs happy?

seancribbs:
@russelldb :+1:

Vagabond:
+1

PR #747, opened 2013-12-05
validate consistent bucket type properties

jrwest:
cc/ @jtuple @russelldb 

A bucket type that is consistent must remain so (the property is
immutable) and the n_val of the type may not be changed (for the type
or its buckets). A type that has consistent=true is consisdered to be
strongly-cosnsitent regardless of the values of other properties
including the datatype property -- strong-consistency may use this
property for its own purposes in the future. Untyped buckets (default
bucket type) cannot be strongly consistent.

The eqc tests have been updated to reflect the changes.

Other Notes:
- The consistent property is validated in two paths depending on if the
  type is consistent or not. This is a bit meh.
- It is assumed that the value of allow_mult does not matter to strong
  consistency and thus it is not validated (besides being a valid boolean)
  for this path
- The datatype property is not validated if the type is strongly-cosnsitent.
  This is because, what the property will mean is still undefined.
- I may have butchered prop_merges() a bit (the most at least). As validation
  grows we may want to re-think how we build the expected set in the property

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -107,14 +107,29 @@ validate(update, _Bucket, Existing, New) when is_list(Existing),
 %% @private bucket creation time validation
 -spec validate_create_bucket_type(props()) -> {props(), errors()}.
 validate_create_bucket_type(BucketProps) ->
-    {Unvalidated, Valid, Errors} = validate_create_dt_props(BucketProps),
+    case proplists:get_value(consistent, BucketProps) of
+        %% type is explicitly or implicitly not intended to be consistent
+        Consistent when Consistent =:= false orelse
+                        Consistent =:= undefined ->
+            {Unvalidated, Valid, Errors} = validate_create_dt_props(BucketProps);
+        %% type may be consistent (the value may not be valid)
+        Consistent ->
+            {Unvalidated, Valid, Errors} = validate_create_consistent_props(Consistent, BucketProps)
+    end,
     validate(Unvalidated, Valid, Errors).
 
 %% @private update phase of bucket type. Merges properties from
 %% existing with valid new properties
 -spec validate_update_bucket_type(props(), props()) -> {props(), errors()}.
 validate_update_bucket_type(Existing, New) ->
-    {Unvalidated, Valid, Errors} = validate_update_dt_props(Existing, New),
+    case proplists:get_value(consistent, Existing) of
+        %% type is explicitly or implicitly not already consistent
+        Consistent when Consistent =:= false orelse
+                        Consistent =:= undefined ->
+            {Unvalidated, Valid, Errors} = validate_update_dt_props(Existing, New);
+        _Consistent ->
+            {Unvalidated, Valid, Errors} = validate_update_consistent_props(Existing, New)
+    end,
     {Good, Bad} = validate(Unvalidated, Valid, Errors),
     {merge(Good, Existing), Bad}.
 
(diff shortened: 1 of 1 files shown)


jrwest:
Also, should we just merge https://github.com/basho/riak_kv/pull/667 into this PR?

src/riak_kv_bucket.erl
```diff
@@ -366,32 +444,93 @@ prop_create_valid_datatype() ->
...
-                Expected = case {default_bucket(Bucket), has_allow_mult(New), allow_mult(New), has_datatype(Existing)} of
-                               {true, _, Mult,  _} when Mult /= error ->
+                DefaultBucket = default_bucket(Bucket),
+                HasAllowMult = has_allow_mult(New),
+                AllowMult = allow_mult(New),
+                HasDatatype = has_datatype(Existing),
+                NValChanged = n_val_changed(Existing, New),
+                IsConsistent = is_consistent(Existing),
+                NewConsistent = proplists:get_value(consistent, New),
+                Expected = case {DefaultBucket, HasAllowMult, AllowMult, HasDatatype,
```
russelldb, line 477: Wow, this is quite the truth table now. I know it is my fault for structuring it that way. Oh boy.

russelldb:
Tests pass, and it seems to work with simple tests by hand. We need to add an issue to refactor that EQC test. It can probably even happen after the freeze since it is a test. Imma :+1:

russelldb:
And I agree, we should merge in #667, too.

jrwest:
@jtuple would you mind taking a look still?

@russelldb thanks! issue opened: https://github.com/basho/riak_kv/issues/750. I'll merge #667 into this PR and test before merging.

src/riak_kv_bucket.erl
jrwest, line 477: yea I know. I made it pretty bad. As we generate more properties it will continue to get worse too. I had a few thoughts on how to fix but none are really concrete. We need a better way of determining what should be in the `Expected` set, but most soln's I can come up w/ mirror most of the validation or other properties. I wonder if we could instead know ahead of time which properties we generated are invalid (or we expect to be invalid rather) and then building the `Expected` set is just a matter of removing the invalid entries the generator told us about...not sure how clear that is...thoughts?

jtuple:
Since @russelldb already did a more detailed review, I just built this branch and played around with things as a user would. As far as I can tell, things work as they should.

I'm still not sure if data types should or shouldn't be mutable on a consistent bucket. I need to revisit my consistent data types branch soon and determine if it gracefully handles mixed types in a bucket / to the same key. But, that's outside the scope of this pull-request.

I did notice an issue when passing in a mispelled datatype that is unrelated to this PR. Get a pretty lame/nasty error message when trying to create unknown atom. But, not sure we can easily fix this; and imagine it's a general problem with how we erlify JSON. Eg.

```
./dev/dev1/bin/riak-admin bucket-type update strong '{"props": {"datatype": "setz"}}' 
                                 {badarg,
                                  [{erlang,binary_to_existing_atom,
                                    [<<"setz">>,utf8],
                                    []},
                                   {riak_kv_wm_utils,erlify_bucket_prop,1,
                                    [{file,"src/riak_kv_wm_utils.erl"},
                                     {line,360}]},
                                   {riak_kv_console,
                                    '-bucket_type_update/2-lc$^0/1-0-',1,
                                     {line,506}]},
                                   {riak_kv_console,bucket_type_update,2,
                                     {line,506}]},
```

In any case, +1 from me.

jrwest:
@jtuple yea I didn't try to do anything around `datatype` when `consistent=true` really. as for the error you found, that was another issue i hoped to address by the command line changes that I unfortunately did not get to. Will merge this with #667 as part of it as promised shortly.

jrwest:
closing for rebased/merged version in #770

PR #748, opened 2013-12-05
Integrate the Exometer metrics package

uwiger:
This PR is part of a set of PRs aimed at integrating the [Exometer](https://github.com/Feuerlabs/exometer) metrics package into Riak.

From our measurements so far, Exometer offers both better throughput and lower footprint than the previous metrics management, and at the same time offers more flexible and uniform handling and better extensibility.

In addition to maintaining the console command `riak-admin status` and the HTTP JSON report (which aim to be backwards-compatible), a new console command, `riak-admin stat <cmd>` has been added, for selective reporting of statistics as well as some management (ability to enable/disable metrics on the fly).

Other noteworthy changes:
- Exometer entry names are always lists. In riak, only atoms and numbers should be used as list elements.
- Internally in riak, stats are referred to symbolically using the same (tuple-based) names as before. These often refer to more than one low-level metric, so it seemed reasonable to keep this naming scheme.
- Exometer provides similar functionality as 'sidejob' and the riak_core stat cache, so these are no longer used for stats management (although sidejob still maintains some stats on its own, which are accessible via Exometer). The function `riak_kv_stat:get_stats()` now queries the Exometer registry directly (through the riak_core_stat API).


src/riak_kv_stat.erl
```diff
@@ -432,11 +445,11 @@ vnode_status(Idx) ->
     [{Idx, [Status]}] = riak_kv_vnode:vnode_status(PList),
     Status.
 
-leveldb_read_block_errors({backend_status, riak_kv_eleveldb_backend, Status}) ->
```
evanmcc, line 564: would prefer something like _int or _pvt instead of _ if you have to dup the fun name.

src/riak_kv_stat.erl
```diff
@@ -60,39 +60,28 @@ start_link() ->
...
+       
 
 %% Creation of a dynamic stat _must_ be serialized.
 register_stat(Name, Type) ->
-    gen_server:call(?SERVER, {register, Name, Type}).
+    do_register_stat(Name, Type).
+    %% gen_server:call(?SERVER, {register, Name, Type}).
 
 update(Arg) ->
-    case erlang:module_loaded(riak_kv_stat_sj) of
```
evanmcc, line 83: please revert this.  it's been our experience that doing sync stat updates results in an unacceptable latency/throughput hit under certain workloads.

src/riak_kv_app.erl
```diff
@@ -53,8 +53,13 @@ start(_Type, _StartArgs) ->
         undefined ->
             ok;
         _ ->
+	    Base = [riak_core_stat:prefix(), riak_kv],
             sidejob:new_resource(riak_kv_put_fsm_sj, sidejob_supervisor, FSM_Limit),
-            sidejob:new_resource(riak_kv_get_fsm_sj, sidejob_supervisor, FSM_Limit)
+	    riak_kv_exometer_sidejob:new_entry(Base ++ [put_fsm, sidejob],
```
evanmcc: I can't get riak to start without commenting out both of these new_entry() calls.  I get the following error:

{"Kernel pid terminated",application_controller,"{application_start_failure,riak_kv,{bad_return,{{riak_kv_app,start,[normal,[]]},{'EXIT',{{case_clause,{error,undef,[{sidejob_stat,new,[[riak,riak_kv,put_fsm,sidejob],sidejob,[{sj_name,riak_kv_put_fsm_sj}]],[]},{exometer,create_entry,1,[{file,\"src/exometer.erl\"},{line,674}]},{exometer_admin,handle_call,3,[{file,\"src/exometer_admin.erl\"},{line,116}]},{gen_server,handle_msg,5,[{file,\"gen_server.erl\"},{line,585}]},{proc_lib,init_p_do_apply,3,[{file,\"proc_lib.erl\"},{line,239}]}]}},[{exometer_admin,new_entry,3,[{file,\"src/exometer_admin.erl\"},{line,67}]},{riak_kv_app,start,2,[{file,\"src/riak_kv_app.erl\"},{line,58}]},{application_master,start_it_old,4,[{file,\"application_master.erl\"},{line,269}]}]}}}}}"}

src/riak_kv_stat.erl
uwiger, line 83: I can do that, but will just note what I wrote on the mailing list Sep 26 when I first suggested this:
"The exometer API takes pains to run all costly updates asynchronously, so devices like sidejob should not be needed."
Given that exometer histogram updates are already asynchronous, also going through sidejob might increase latency without adding much benefit. However, I haven't verified that it gives acceptable throughput in all use cases.

src/riak_kv_app.erl
uwiger: I just successfull cloned and built a fresh system based on riak,
feuerlabs-exometer.
What happens when you try that?

2013/12/7 Evan Vigil-McClanahan notifications@github.com

evanmcc: Same thing.  Disabling sidejob  with fsm_limit undefined (using app.config, because riak.conf doesn't support that just yet) allows it to start, but then, of course, sidejob is disabled.

src/riak_kv_stat.erl
```diff
@@ -168,75 +170,99 @@ code_change(_OldVsn, State, _Extra) ->
 
 %% @doc Update the given stat
 do_update({vnode_get, Idx, USecs}) ->
-    folsom_metrics:notify_existing_metric({?APP, vnode, gets}, 1, spiral),
-    create_or_update({?APP, vnode, gets, time}, USecs, histogram),
-    do_per_index(gets, Idx, USecs);
+    P = riak_core_stat:prefix(),
+    exometer:update([P, ?APP, vnode, gets], 1),
+    create_or_update([P, ?APP, vnode, gets, time], USecs, histogram),
+    do_per_index(puts, Idx, USecs);
```
evanmcc: copy/paste error here: this should be gets

src/riak_kv_stat.erl
```diff
@@ -281,13 +309,14 @@ do_get_bucket(true, {Bucket, Microsecs, Stages, NumSiblings, ObjSize}=Args) ->
...
+    case exometer:update([P, ?APP, node, puts, Bucket], 1) of
         ok ->
-            folsom_metrics:notify_existing_metric({?APP, node, puts, time, Bucket}, Microsecs, histogram),
-            do_stages([?APP, node, puts, time, Bucket], Stages);
-        {'EXIT', _} ->
-            register_stat({?APP, node, puts, Bucket}, spiral),
-            register_stat({?APP, node, puts, time, Bucket}, histogram),
+	    exometer:update([P, ?APP, node, puts, time, Bucket], Microsecs),
+            do_stages([P, ?APP, node, puts, time, Bucket], Stages);
+	{error, _} ->
```
evanmcc: our whitespace standards are all spaces, 4 space per indent level.

src/riak_kv_status.erl
```diff
@@ -80,3 +81,154 @@ are_indexes_fixed(riak_kv_eleveldb_backend, Status) ->
 are_indexes_fixed(riak_kv_multi_backend, {_Idx, [{backend_status,_,Status}]}) ->
     Statuses = [S || {_, S} <- Status, lists:member({mod, riak_kv_eleveldb_backend}, Status)],
     fixed_index_status(riak_kv_eleveldb_backend, Statuses).
+
+
+
+get_stats(web) ->
```
evanmcc, line 87: lots of duplication here and and the webmachine stats endpoint.  Can that be cleaned up into a set of common functions?

src/riak_kv_exometer_sidejob.erl
```diff
@@ -0,0 +1,47 @@
+-module(riak_kv_exometer_sidejob).
```
evanmcc, line 29: needs license header

jburwell:
This PR has been superseded by [PR 1006](https://github.com/basho/riak_kv/issues/1006).

PR #754, opened 2013-12-10
Add timer:sleep()-based throttle to riak_kv_exchange_fsm:read_repair_keydiff()

slfritchie:
Note: this PR is against the `1.4` branch.

The throttle part is easy: a 1-liner.  The rest is the infrastructure
to poll each node for its maximum vnode mailbox queue length, pick the
worst one, calculate a sleep time based on a list of {Q_length, SleepMSec},
and communicate via time-honored application env vars.

The only hot code path that reads the ETS table for app env vars is
riak_kv_exchange_fsm:read_repair_keydiff() itself, and since it is
doing a full KV get FSM plus other stuff, the extra heat on the hot
path is insignificant.

@joedevivo I'd appreciate some cuttlefish advice on how to go about
crafting the value of the {riak_kv, aae_throttle_limits} in a cuttlefish
config: `list({integer(), non_neg_integer()})`.  Once I have that figured out,
then I can create an expanded PR for `develop`.

### DIFF
--- a/src/riak_kv_cinfo.erl
+++ b/src/riak_kv_cinfo.erl
@@ -48,12 +48,13 @@ cluster_info_generator_funs() ->
 aae_throttle(CPid) ->
     {Throttle, ThrottleLimits} = get_aae_throttle(),
     cluster_info:format(CPid, "Current throttle: ~p msec\n", [Throttle]),
-    cluster_info:format(CPid, "Limit AAT throttle parameters: ~p\n",
+    cluster_info:format(CPid, "Limit AAE throttle parameters: ~p\n",
                         [ThrottleLimits]).
 
 get_aae_throttle() ->
-    {riak_kv_entropy_manager:get_aae_throttle(),
-     riak_kv_entropy_manager:get_aae_throttle_limits()}.
+    [{throttle_kill, riak_kv_entropy_manager:get_aae_throttle_kill()},
+     {current_throttle, riak_kv_entropy_manager:get_aae_throttle()},
+     {limits, riak_kv_entropy_manager:get_aae_throttle_limits()}].
 
 status(CPid) -> % CPid is the data collector's pid.
     cluster_info:format(CPid, "~p\n", [riak_kv_status:statistics()]).
--- a/src/riak_kv_cinfo.erl
+++ b/src/riak_kv_cinfo.erl
@@ -22,6 +22,7 @@
 
 -module(riak_kv_cinfo).
 -export([cluster_info_init/0, cluster_info_generator_funs/0]).
+-export([get_aae_throttle/0]).
 
 %% @spec () -> term()
 %% @doc Required callback function for cluster_info: initialization.
@@ -40,8 +41,19 @@ cluster_info_generator_funs() ->
     [
      {"Riak KV status", fun status/1},
      {"Riak KV ringready", fun ringready/1},
-     {"Riak KV transfers", fun transfers/1}
+     {"Riak KV transfers", fun transfers/1},
+     {"Riak KV anti-entropy throttle", fun aae_throttle/1}
     ].
+
+aae_throttle(CPid) ->
+    {Throttle, ThrottleLimits} = get_aae_throttle(),
+    cluster_info:format(CPid, "Current throttle: ~p msec\n", [Throttle]),
+    cluster_info:format(CPid, "Limit AAT throttle parameters: ~p\n",
+                        [ThrottleLimits]).
+
+get_aae_throttle() ->
+    {riak_kv_entropy_manager:get_aae_throttle(),
+     riak_kv_entropy_manager:get_aae_throttle_limits()}.
 
 status(CPid) -> % CPid is the data collector's pid.
     cluster_info:format(CPid, "~p\n", [riak_kv_status:statistics()]).
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -42,13 +42,13 @@
          get_aae_throttle_limits/0,
          set_aae_throttle_limits/1,
          get_max_local_vnodeq/0]).
+-export([multicall/5]).                         % for meck twiddle-testing
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
 -ifdef(TEST).
--export([multicall/5]).                         % for meck twiddle-testing
 -export([query_and_set_aae_throttle/1]).        % for eunit twiddle-testing
 -export([make_state/0, get_last_throttle/1]).   % for eunit twiddle-testing
 -include_lib("eunit/include/eunit.hrl").
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -39,6 +39,8 @@
 -export([all_pairwise_exchanges/2]).
 -export([get_aae_throttle/0,
          set_aae_throttle/1,
+         get_aae_throttle_kill/0,
+         set_aae_throttle_kill/1,
          get_aae_throttle_limits/0,
          set_aae_throttle_limits/1,
          get_max_local_vnodeq/0]).
@@ -75,6 +77,7 @@
 -define(DEFAULT_CONCURRENCY, 2).
 -define(DEFAULT_BUILD_LIMIT, {1, 3600000}). %% Once per hour
 -define(AAE_THROTTLE_ENV_KEY, aae_throttle_sleep_time).
+-define(AAE_THROTTLE_KILL_ENV_KEY, aae_throttle_kill).
 
 %%%===================================================================
 %%% API
@@ -712,6 +715,17 @@ get_aae_throttle() ->
 set_aae_throttle(Milliseconds) when is_integer(Milliseconds), Milliseconds >= 0 ->
     application:set_env(riak_kv, ?AAE_THROTTLE_ENV_KEY, Milliseconds).
 
+get_aae_throttle_kill() ->
+    case app_helper:get_env(riak_kv, ?AAE_THROTTLE_KILL_ENV_KEY, undefined) of
+        true ->
+            true;
+        _ ->
+            false
+    end.
+
+set_aae_throttle_kill(Bool) when Bool == true; Bool == false ->
+    application:set_env(riak_kv, ?AAE_THROTTLE_KILL_ENV_KEY, Bool).
+
 get_max_local_vnodeq() ->
     try
         {element(2,lists:keyfind(riak_kv_vnodeq_max, 1, riak_core_stat:get_stats())), node()}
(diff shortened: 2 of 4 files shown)


jonmeredith:
Interested in thoughts from @jtuple and @engelsanchez.

seancribbs:
@slfritchie `anti_entropy.queue_delay.$length` with a datatype of `integer` seems appropriate. However, this should probably be an "advanced" setting too, putting sensible defaults in .app.src.

src/riak_kv_cinfo.erl
```diff
@@ -40,9 +41,20 @@ cluster_info_generator_funs() ->
...
      {"Riak KV ringready", fun ringready/1},
-     {"Riak KV transfers", fun transfers/1}
+     {"Riak KV transfers", fun transfers/1},
+     {"Riak KV anti-entropy throttle", fun aae_throttle/1}
     ].
 
+aae_throttle(CPid) ->
+    {Throttle, ThrottleLimits} = get_aae_throttle(),
+    cluster_info:format(CPid, "Current throttle: ~p msec\n", [Throttle]),
+    cluster_info:format(CPid, "Limit AAT throttle parameters: ~p\n",
```
joedevivo: s/AAT/AAE ?

test/riak_kv_entropy_manager_test.erl
```diff
@@ -0,0 +1,87 @@
...
+        V <- [5,6]].
+
+set_aae_throttle_limits_test() ->
+    {error, _} = ?TM:set_aae_throttle_limits([]),
+    {error, _} = ?TM:set_aae_throttle_limits([{5,7}]),
+    {error, _} = ?TM:set_aae_throttle_limits([{-1,x}]),
+    {error, _} = ?TM:set_aae_throttle_limits([{-1,0}, {x,7}]),
+    ok = ?TM:set_aae_throttle_limits([{-1,0}, {100, 500}, {100, 500},
+                                      {100, 500}, {100, 500}, {100, 500}]).
+
```
joedevivo, line 49: If I read correctly, the minimum required list is `[{-1,0}]`. might as well test for that, right?

src/riak_kv_entropy_manager.erl
```diff
@@ -685,3 +705,86 @@ requeue_exchange(LocalIdx, RemoteIdx, IndexN, State) ->
...
+
+get_aae_throttle_limits() ->
+    %% init() should have already set a sane default, so the default below should never be used.
+    app_helper:get_env(riak_kv, aae_throttle_limits, [{-1, 0}]).
+
+%% @doc Set AAE throttle limits list
+%%
+%% Limit list = [{max_vnode_q_len, per-repair-delay}]
+%% A tuple with max_vnode_q_len=-1 must be present.
+%% List sorting is not required: the throttle is robust with any ordering.
```
joedevivo, line 755: What happens if somebody configures this with the same Q_len twice? e.g. `[{-1,0}, {500, 100}, {500, 200}, {1000, 500}]`

joedevivo:
@seancribbs is right about the cuttlefish mappings. You could also use the datatype `duration` for cool units. You'll need a single translation as well, to assemble the list. The following should work, but is untested, may require a tweak or two, but the general idea is:

``` erlang
{mapping, "cool.user_friendly_name.$length", "riak_kv.aae_throttle_limits", [
  {datatype, {duraton, ms}},
  {include_default, "-1"},
  {level, advanced},
  {default, "0ms"}
]}.

{translation,
 "riak_kv.aae_throttle_limits",
 fun(Conf) ->
   Values = cuttlefish_util:filter_by_variable_starts_with("cool.user_friendly_name"),
   [ {QLen, SleepMillis} || {["cool", "user_friendly_name", QLen], SleepMillis} <- Values]
 end
}.
```

Then your riak.conf would generate none of this, but you could modify it by adding:

```
cool.user_friendly_name.-1 = 0ms
cool.user_friendly_name.500 = 10ms
cool.user_friendly_name.1000 = 1s
```

If you specified nothing, the default that's generated would be: `[{-1, 0}]`

kthxbye

test/riak_kv_entropy_manager_test.erl
slfritchie, line 49: Actually, {-1, non_neg_integer()} is the minimum required.  But I'm happy with the coverage, given the `andalso` that's in the sanity check.

slfritchie, line 49: Derp, not `andalso`, but IMO good enough.

src/riak_kv_entropy_manager.erl
slfritchie, line 755: Doctor, it hurts.  You get the larger tuple: results are sorted, reversed, and take the first one.

slfritchie:
Thoughts on the basics of this approach, @jtuple @engelsanchez @evanmcc @jrwest?  Cuttlefish work for 2.0 isn't done yet, sorry.

evanmcc:
Why is this targeted against 1.4? I think that it's more than a bit late to get into 1.4.4.  Is there going to be another 1.4.X release (please no)? I thought that there was going to be some other throttling approach for 2.0?

slfritchie:
Because I didn't want to tangle with cuttlefish right away and because I expect to need it for customer delivery on 1.4.anything (and thus also doesn't need cuttlefish) before Riak 2.0 needs it.

There is no throttling approach that I'm aware of for Riak 2.0 that adds sufficient backpressure to riak_kv_exchange_fsm:read_repair_keydiff().  If a minority of vnodes are slow, the Erlang client `get()` will not slow down, and the `rehash` command isn't a synchronous command.

matthewvon:
FYI:  AAE was intentionally made exempt to leveldb's write throttle.  Might this be accomplished by removing the exemption?

jrwest:
I was under the impression @jtuple is preparing a PR that includes a similar change among others. Is there a reason this was done separately?

jtuple:
@matthewvon This PR adds throttling to the repair side of AAE -- the component that repairs keys that have been identified as divergent. This is a read-only code path, write throttling doesn't come into play. Also, these are high level K/V reads that hit whatever backend is configured -- eg. if this is a Bitcask user, than LevelDB isn't doing much work (read or otherwise) in this code path. 

There is separate work to add throttling to AAE rebuilds. This has already been committed in a branch that will be landing in a different PR. Rebuilds issue read-only traffic against the K/V backend, and read/write traffic against the AAE LevelDB instance. Write throttling could help here, but it's unclear if it's needed after I add my throttling. My throttling needs to be added regardless, since the primary thing I'm trying to throttle is the read-only traffic against the K/V backend (which could be LevelDB, Bitcask, etc).

@jrwest As I just mentioned, this (Scott's) PR adds throttling to the repair side of AAE. My throttling work adds a throttle to the AAE rebuild path. These are disjoint and unrelated. Both appear to be necessary, and aim to fix different problems. This PR is a bit more complex since it's throttling traffic that spans multiple nodes. No real way to reuse anything from either approach.

@slfritchie I've glanced over the PR and things look good on first pass. I'd still want to run/test things, something I won't get to until Monday. But, I don't have any major reservations against this PR. Will need to do a more detailed review though before final +1.

My initial reaction was that I was in favor of a very simple static throttle without any of the vnode message queue scaling. Just have users/CSEs manually change the setting when trouble strikes. But, I see the benefit in the dynamic approach. It's just more code we need to be extra critical of and heavily test before merging.

I highly suggest we add an additional app.config setting that we can set that completely disables the dynamic scaling -- just skips the `query_and_set_aae_throttle` call in the manager tick and stick with the lowest value. We'd ship with the dynamic scaling, but have a "get out of jail" card to use if something shows up in production.

I think the ship has sailed on getting this into 1.4.4, but I'm adding it to my list of things to review next week.

jrwest:
@jtuple thanks for the clarification

slfritchie:
```
  vn1    vn2    vn3    vn4    vn5
------  -----  -----  -----  -----
         AAE    AAE           AAE
         src -> dst <-------- src
quick   quick  slow   quick  quick
|<- get FSM    ->|
    vnode ops    |<- get FSM  ->|
    originating      vnode ops originating
    from vn2's       from vn5's 
    client           client     
```

If an AAE exchange & repair is running between vnode 2 -> vnode3 and also vnode5 -> vnode3, the get FSMs that are running on vnode 2's host and vnode 5's host will get no backpressure from the slow replies of vnode 3, if it is the only vnode that is having performance problems.  Note that the destination of the exchanges doesn't matter (the diagram has the case that is arguably the worst case): the overlap of the preflists used for the get FSMs matters more.

In this PR's implementation, it's possible for long vnode mailbox queues over on machines that host none of vnodes 1-5 would slow down the exchanges happening with vnode 3.  But figuring that out is Hard(™), so I'm taking the easy & safer way out for now.  More smarts can be added later.

A much simpler way to get the backpressure we want would be to use r={n_val}/all.  However, as diagrammed, the failure of vnode 4 would make the vnode 5 -> 3 exchange & repair impossible ... and that sounds less agreeable to me than this PR.

The get-out-of-jail is a setting of [{-1,0}], which means that a queue length >= -1 would get a sleep time of 0 seconds.  That sounds as effective as a kill switch without any additional conditional code to make a complete bypass.

Thoughts?

src/riak_kv_entropy_manager.erl
```diff
@@ -685,3 +705,86 @@ requeue_exchange(LocalIdx, RemoteIdx, IndexN, State) ->
...
+query_and_set_aae_throttle(#state{last_throttle=LastThrottle} = State) ->
+    Limits = lists:sort(get_aae_throttle_limits()),
+    {MaxNds, BadNds} = ?MODULE:multicall([node()|nodes()], ?MODULE, get_max_local_vnodeq, [], 10*1000),
+    %% If a node is really hosed, then this RPC call is going to fail for that node.
+    %% We might also delay the 'tick' processing by several seconds.  But the tick
+    %% processing is OK if it's delayed while we wait for slow nodes here.
+    {WorstVMax, WorstNode} =
+        case {BadNds, lists:reverse(lists:sort(MaxNds))} of
+            {[], [{VMax, Node}|_]} ->
+                {VMax, Node};
```
jtuple, line 802: In a mixed cluster, such as during a rolling upgrade, older nodes that do not implement `get_max_local_vnodeq` will return `{badrpc, _}` that will match this clause but return unexpected values.

We should either validate that we don't get back any unexpected values, change `get_max_local_vnodeq` to return something like `{ok, VMax}` that we can match on, or make the entire AAE throttling code path guarded by a new capbility -- and not use any of the new stuff until after the capability is active.

jtuple:
I went ahead and implemented a riak_test for this code, basho/riak_test#491

Two final comments.

First, I still believe we should consider an app var based kill switch that disables the vnode querying + throttle adjustment code in the entropy manager. Seems unwise to have every node RPC every other node at a fixed tick without the ability to turn that off if trouble strikes. Specific fear: flaky network with frequent netsplits that leads to lots of stalling on net kernel reconnects + eventual timeouts.

Second, and related concern, is the RPCs blocking the entropy manager. The entire AAE system as well as how it's integrated into other areas of Riak is designed and validated assuming an entropy manager that never blocks. Even a 10 second stall may be too much. Or rather, it's non-trivial to prove that it doesn't matter.

Much easier to just make the query non-blocking. Have the entropy manager spawn a process that does the RPC query, and then sends the results back to the entropy manager via a cast. Have the manager track the pid it spawned in it's state. If a pid is still alive the next tick, don't spawn it again. Super simple and removes any concerns about blocking manager.

Let me know if you don't have time to do either of these changes before US Friday (Dec 20). I'm willing to make the changes myself to make sure we can wrap this up and get it merged in. We still need to port over to develop for the 2.0 code freeze as well, but that's very straightforward.

src/riak_kv_cinfo.erl
slfritchie: Derp, thanks, corrected in to-be-pushed commit with review fixes.

slfritchie:
> Much easier to just make the query non-blocking. 

Uff da, that makes a mess of the current EUnit testing method.  /me needs thinking cap.

slfritchie:
Joe, Sean, and Joe, many thanks for your help & feedback.  The non-Cuttlefish items have been addressed (I hope) by my last two commits, abc3063 and d32b021.

For the develop branch, I squashed the commits from this branch and then added (I hope) all necessary for Cuttlefish.  See https://github.com/basho/riak_kv/compare/slf-aae-throttle-develop?expand=1  If there's a +1 on the Cuttlefish+develop branch, then I'll merge that one and then close this PR until it's necessary for the 1.4 branch.

seancribbs:
@slfritchie I left you some comments on the second commit.

slfritchie:
@seancribbs Didn't I get the cuttlefish config correct?

jtuple:
The new commits address my concerns. Although, it looks like a mixed-cluster still causes the worst cased throttle to be chosen, rather than distinguishing between a `nodedown` and `undef` call to a function that doesn't exist on older nodes. IMO, since we don't throttle things today, we shouldn't throttle in a mixed cluster either. But, that's something we can debate + (not) change in the future.

The eunit test passes. As does the riak_test I wrote and mentioned earlier in this PR.

I manually increased the throttle and played around with changing the kill switch (`riak_kv_entropy_manager:set_aae_throttle_kill/1`) and saw the desired behavior. The throttle would drop to 0, and redbug tracing of the status query code would show no more calls. Disabling kill switch and throttle would rise again, and status calls resume.

Of course, I should add kill switch testing to the riak_test, but that's for another day.

I'm +1 on these changes for 1.4. I'm now going to review the 2.0 against `develop`'s current state and report back.

jtuple:
Alright. Merged [slf-aae-throttle-develop](https://github.com/basho/riak_kv/compare/slf-aae-throttle-develop?expand=1) into develop locally and tested things out a bit.

Amazingly, zero merge conflicts. You touched the few parts of AAE that haven't been changed by 2-3 other PRs that merged today. Code as merged looks correct for develop even with latest changes.

Performed some basic manual smoke testing. Things appear to work as well as the 1.4 version.

The eunit from this PR, basho/riak_test#491 for this PR, as well as basho/riak_test#481 which tests several new additions to AAE all work as intended after merging.

I'm +1 for merging to develop. Please make a pull-request off the new branch, reference this pull-request, and go ahead and merge.

If @seancribbs has remaining Cufflefish concerns, we can fix this in a future PR before 2.0.

PR #755, opened 2013-12-10
Added a clause to from_binary/3 to handle bucket type information

bowrocker:
This change is needed for MDC replication to be able to handle typed-bucket replication. The "sink" side runs this function after un-marshaling a replicated object, and type information is necessary.

This will depend on Joe Blomstedt's addition of conditional post commit hooks, in this branch:

https://github.com/basho/riak_kv/tree/jdb-conditional-postcommit

The Riak Test for typed bucket replication is here:

https://github.com/basho/riak_test/tree/jra_bucket_types

Additionally, there is a design discussion here:

https://gist.github.com/bowrocker/7f0d9d6879493f1ac0e9

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -31,6 +31,7 @@
 -export_type([riak_object/0, bucket/0, key/0, value/0, binary_version/0]).
 
 -type key() :: binary().
+-type type() :: binary().
 -type bucket() :: binary() | {binary(), binary()}.
 %% -type bkey() :: {bucket(), key()}.
 -type value() :: term().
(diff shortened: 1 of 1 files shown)


src/riak_object.erl
```diff
@@ -557,7 +558,17 @@ binary_version(<<131,_/binary>>) -> v0;
 binary_version(<<?MAGIC:8/integer, 1:8/integer, _/binary>>) -> v1.
 
 %% @doc Convert binary object to riak object
--spec from_binary(bucket(),key(),binary()) -> riak_object().
+-spec from_binary({type(), bucket()},key(),binary()) -> riak_object().
```
seancribbs, line 561: I wouldn't change the spec here, the `bucket()` type is already defined as `binary() | {binary(), binary()}`.

seancribbs, line 561: In which case, doesn't this clause just duplicate the functionality of the third?

bowrocker, line 561: I don't think so, since the third clause does not set the Type in the bucket field of the r_object:

#r_object{bucket={T,B},key=K,contents=Contents,vclock=Vclock};

Right?

seancribbs, line 561: Ummm. `B` will match a two-tuple just as well as a binary, there's no guard on it.

bowrocker, line 561: Also: When I tried that third clause after replicating an object, I got no type info out of it, so though it may be specified to include it, it doesn't seem to.

engelsanchez, line 561: Sean is very correct. This extra clause doesn't really add anything new. Maybe you had a different problem. The B will match a tuple or binary and place whatever it finds in the bucket field.

jrwest, line 561: I agree w/ @seancribbs and @engelsanchez. The bucket/key are not encoded in the `v1` encoding of the object [1]. The reason it is passed as the first argument to `from_binary/3` is so we can properly reconstitute the full `riak_object`. Whether its a 2-tuple or a binary should not matter as the others have mentioned. In the `v0` case we use `binary_to_term/term_to_binary` so again it should be irrelevant. 

[1] https://gist.github.com/jrwest/d525738de67a40292757

bowrocker, line 561: Understood, thanks everyone. Apparently it solved a problem I still have somewhere in repl -- I will find it.

PR #756, opened 2013-12-10
Clean up two security holes: link walking and arbitrary erlang MR

Vagabond:
Link walking is deprecated in 2.0, so we won't make it work with
security. Also, Erlang mapreduce phases and inputs must now use modules
that reside in an add_paths directory in the config file. This is to
prevent arbitrary use of the Erlang standard library to do malicious
things.


seancribbs:
:+1: riak_tests and manual tests pass, code is straightforward

PR #758, opened 2013-12-12
Merge CRDTs inside riak_object merge

russelldb:
Why? Well it means no CRDT siblings on disk. Previously CRDTs
were merged only when a CRDT Operation was performed from the CRDT
API endpoints. This means read repair, AAE, and repl, could all cause
siblings on disk.

With the addition of DVV style causality, we need to make a choice
about CRDTs. We don't have DVVSets and the anonymous list. So CRDTs
either have to merge Dots into a mini-version-vector, or not have dots.

If CRDTs don't have dots, so their values always survive a merge,
then they are subject to repl induced sibling-explosion (possibly)

Merging in riak_object:merge solves that issue.
As an added benefit it also means that Yokozuna indexing of CRDTs
will just work.

I'd really appreciate a performance comparision between this and
the plain DVV branch, and the pre-DVV stuff for 2.0.

### DIFF
--- a/src/riak_kv_crdt.erl
+++ b/src/riak_kv_crdt.erl
@@ -25,6 +25,7 @@
 -export([update/3, merge/1, value/2, new/3,
          supported/1, to_mod/1, from_mod/1]).
 -export([to_binary/2, to_binary/1, from_binary/1]).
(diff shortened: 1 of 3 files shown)


russelldb:
Decided to merge this to the branch and have both as a single PR for DVV in Riak

PR #759, opened 2013-12-12
Fix CS buckets 2i query

engelsanchez:
This change makes sure that those always use pagination sort and fixes a
problem with start/end parameters after the refactor for the 1.4.4 2i
changes.  This uncovered some inconsistencies with the internal 2i query
parameters. This makes sure that bucket queries don't set start/end terms
and quality always have start=end and return_terms=false, among other
things. All 2i riak tests pass with this.

@russelldb my man, if you could please review this, it would be spectacular.


russelldb:
All tests pass, code makes sense. WFM :+1:

engelsanchez:
Thanks for looking over this @russelldb!  BTW, I'm re-reading the PR description I wrote now... I guess I was drunk.

PR #760, opened 2013-12-13
make it possible to disable overload protection

evanmcc:
now that https://github.com/basho/cuttlefish/pull/89 has landed, make it possible to disable fsm limits.  

thanks, @joedevivo


joedevivo:
+1

PR #762, opened 2013-12-13
Disconnect external/internal return terms 2i param

engelsanchez:
Unfortunately I got confused yesterday when refactoring the 2i
parameters. There is an external return terms param that controls
whether the user will receive (term, key) pairs or just key values.
There is also an internal return terms parameter that controls whether
(term, key) values will be passed from the vnode to the FSM. That
internal one is only off for equality queries or when using the fake 2i
keys that simply fold over regular keys ($bucket, $key). This change
decouples them in the http interface.

This should make the secondary_index_test pass with both memory and leveldb backends in the 1.4 branch


jaredmorrow:
+1, riak_tests pass

Issue #753, opened 2013-12-10
riak_kv_vnode:handle_overload_command() does not handle everything


slfritchie:
The offending clause is here: https://github.com/basho/riak_kv/blob/c00bdb4e1cccad553920c3119655f1083b4fe680/src/riak_kv_vnode.erl#L406

When we're overloaded, we can't have riak_core_vnode_proxy throwing away commands without any response at all.  For example, problems reported with KV backend status, 2I operations, bucket fold operations; all hanging or timing out in unexpected ways.

runesl:
IMHO, the desired behaviour would be to always complete KV_VNODE_STATUS_REQ. I.e.also in overload situations. The operators need their stats to return, so they know if the node is alive and well or needs restart.

The command types serving user requests should return an overload response like is already done for PUT and GET.

rzezeski:
+1 to this. We should be returning an error response to callers so they at least know something went wrong. I imagine the current implementation could cause process count to grow and message queues to back up as things wait for a response.

One specific thing I wonder about is AAE tree building during overload (as it requires folding over KV data). IIRC when AAE was first written it made assumptions that some operations would always make progress. I think we eventually fixed that by adding timeouts everywhere but if we didn't a bug like this could cause issues for AAE.

evanmcc:
in: https://github.com/basho/riak_kv/blob/develop/src/riak_kv_get_fsm.erl#L107-L124
and: https://github.com/basho/riak_kv/blob/develop/src/riak_kv_put_fsm.erl#L142-L159

There is another potential overload path.  These overloads come from sidejob keeping watch on the number of get and put fsms that are live at any one time (https://github.com/basho/sidejob/blob/develop/src/sidejob_supervisor.erl#L52-L58).  It'd be nice to differentiate these, perhaps by changing the plain {error, overload} message to be specific to the kind of overload that is being encountered.

cc @angrycub @jtuple

andrewjstone:
I propose changing the last clause [here](https://github.com/basho/riak_kv/blob/c00bdb4e1cccad553920c3119655f1083b4fe680/src/riak_kv_vnode.erl#L406) to the following

``` erlang
handle_overload_command(_, Sender, _) ->
    riak_core_vnode:reply(Sender, {error, mailbox_overload}).
```

[...]

PR #763, opened 2013-12-14
Make vnode_status overload safe

evanmcc:
- add overload for vnode_status commands
- add handling for overload messages

This stems from the stats lockup issue.  While this will work on its own, it will occasionally break stats without the tagging stuff in https://github.com/basho/riak_core/pull/482, so these should likely be reviewed as a pair.

Testing notes:

I tested this two different ways: 
- using `meck:expect(riak_kv_vnode, vnode_status, fun(_) -> recieve ok -> ok end end).` which simulates the vnode command getting dropped silently by overload
- setting `{vnode_overload_threshold,`},` in riak_core's app.config section or advanced.config 

after 10 or so seconds you should see `riak-admin status | grep level` and `curl -s localhost:10018/stats | grep level` start returning `"stale: 0"` for the leveldb block read errors.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -403,6 +403,8 @@ handle_overload_command(?KV_PUT_REQ{}, Sender, Idx) ->
     riak_core_vnode:reply(Sender, {fail, Idx, overload});
 handle_overload_command(?KV_GET_REQ{req_id=ReqID}, Sender, Idx) ->
     riak_core_vnode:reply(Sender, {r, {error, overload}, Idx, ReqID});
+handle_overload_command(?KV_VNODE_STATUS_REQ{}, Sender, _Idx) ->
+    riak_core_vnode:reply(Sender, {error, overload});
 handle_overload_command(_, _, _) ->
     %% not handled yet
     ok.
@@ -1777,6 +1779,8 @@ wait_for_vnode_status_results(PrefLists, ReqId, Acc) ->
             wait_for_vnode_status_results(UpdPrefLists,
                                           ReqId,
                                           [{Index, Status} | Acc]);
+        {ReqId, {error, overload}} ->
+            {error, overload};
          _ ->
             wait_for_vnode_status_results(PrefLists, ReqId, Acc)
     end.
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -403,8 +403,8 @@ handle_overload_command(?KV_PUT_REQ{}, Sender, Idx) ->
     riak_core_vnode:reply(Sender, {fail, Idx, overload});
 handle_overload_command(?KV_GET_REQ{req_id=ReqID}, Sender, Idx) ->
     riak_core_vnode:reply(Sender, {r, {error, overload}, Idx, ReqID});
-handle_overload_command(?KV_VNODE_STATUS_REQ{}, Sender, _Idx) ->
-    riak_core_vnode:reply(Sender, {error, overload});
+handle_overload_command(?KV_VNODE_STATUS_REQ{}, Sender, Idx) ->
+    riak_core_vnode:reply(Sender, {vnode_status, Idx, [{error, overload}]});
 handle_overload_command(_, _, _) ->
     %% not handled yet
     ok.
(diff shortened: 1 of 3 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1777,6 +1779,8 @@ wait_for_vnode_status_results(PrefLists, ReqId, Acc) ->
             wait_for_vnode_status_results(UpdPrefLists,
                                           ReqId,
                                           [{Index, Status} | Acc]);
+        {ReqId, {error, overload}} ->
+            {error, overload};
```
slfritchie: This commit's edit to riak_kv_stat.erl uses a preflist of length 1, which I suppose would be fine for its use.  But riak_kv_status.erl uses a (usually) longer preflist in `vnode_status/0` at https://github.com/basho/riak_kv/blob/a5a5e8c988defd319406fc4fc81ff21ad40f6e74/src/riak_kv_status.erl#L51  ... I'm wondering that it's probably not a good idea to both return a non-list and also to stop receiving other status results?  The `lists:foldl/3` call at line 70 will not like a non-list. ... Hrm, but to keep track of the indexes that have responded (and to put the index # into the {Index#, StatusPropListThingie} tuple), the `{error, overload}` message will need an index number in it.  And then that ripples out to touch more code?

src/riak_kv_vnode.erl
```diff
@@ -403,6 +403,8 @@ handle_overload_command(?KV_PUT_REQ{}, Sender, Idx) ->
     riak_core_vnode:reply(Sender, {fail, Idx, overload});
 handle_overload_command(?KV_GET_REQ{req_id=ReqID}, Sender, Idx) ->
     riak_core_vnode:reply(Sender, {r, {error, overload}, Idx, ReqID});
+handle_overload_command(?KV_VNODE_STATUS_REQ{}, Sender, _Idx) ->
+    riak_core_vnode:reply(Sender, {error, overload});
 handle_overload_command(_, _, _) ->
```
kuenishi, line 408: in `riak_kv_vnode.hrl` , we have other requests such as 

```
-define(KV_PUT_REQ, #riak_kv_put_req_v1).
-define(KV_GET_REQ, #riak_kv_get_req_v1).
-define(KV_LISTBUCKETS_REQ, #riak_kv_listbuckets_req_v1).
-define(KV_LISTKEYS_REQ, #riak_kv_listkeys_req_v4).
-define(KV_INDEX_REQ, #riak_kv_index_req_v2).
-define(KV_VNODE_STATUS_REQ, #riak_kv_vnode_status_req_v1).
-define(KV_DELETE_REQ, #riak_kv_delete_req_v1).
-define(KV_MAP_REQ, #riak_kv_map_req_v1).
-define(KV_VCLOCK_REQ, #riak_kv_vclock_req_v1).
```

So I think we should handle all of these requests here. Or other code to handle these requests in somewhere else?

russelldb, line 408: I agree with @kuenishi. However, I don't think that that is part of this. The issue riak_kv#753 tracks the general problem.

src/riak_kv_vnode.erl
russelldb: @evanmcc I agree with @slfritchie here. For example https://github.com/basho/riak_kv/blob/develop/src/riak_kv_console.erl#L188 expects a list of statuses.

I'll have a play at how to do this better.

russelldb:
After a bit of poking it seems that the overload protection does not even apply to the only other call to vnode_status.

`riak_kv_console` calls `riak_kv_status` which calls `riak_kv_vnode:vnode_status/1` with a PrefList of `[{partion(), pid()}]` rather than a PrefList of `[{partition(), node()}]`. The latter gets dropped by overload protection, the former does not.

I imagine that means we need to either raise an issue against riak_core for better overload protection, or change the way `riak_kv_status` behaves.

russelldb:
I pushed a commit that I think addresses @slfritchie's concerns, and allows @evanmcc's changes to still be valuable and work. Tomatoes?

russelldb:
Also, I always see the value as `<<stale: undefined>>`, I'm looking into that, too.

Aaaand it was my derpy-ness, not removing the threshold of 0.

russelldb:
I'm plus one on this change, but since the latest commit is mine that's probably not good enough. @slfritchie comments?

slfritchie:
+1 'cause Russell likes it and I think it looks much better than the current situation.

evanmcc:
Took a look at russel's changes and retested.  Merging.

src/riak_kv_vnode.erl
kuenishi, line 408: OH I didn't know another issue #753 already opened.

PR #765, opened 2013-12-16
Numerous AAE improvements

jtuple:
This pull request adds multiple improves to the AAE subsystem. Each improvement is a separate isolated commit. Reviewing this PR commit by commit would be the easiest approach.

The changes includes the following:
1. A new throttle that can be used to throttle AAE rebuilds to prevent rebuilds from monopolizing I/O.
2. A change that causes AAE trees for empty vnodes to be build instantly, rather than waiting for being scheduled in the future and performing an empty fold. With this change, newly provisioned clusters come up with AAE trees already built.
3. The ability to disable AAE expiration by setting the expiration time to `never`.
4. A bug fix that ensures in-memory AAE buffers are flushed to LevelDB before closing the AAE trees. Also, cleanly close AAE trees in `riak_kv_index_hashtree:terminate/2`.
5. When switching from automatic to manual exchange mode, the existing queue of exchanges (filled by the automatic mode) is now cleared.
6. AAE trees are no longer deleted immediately when expired. Instead, trees are marked as expired, but not deleted until right before they are going to be rebuilt (ie. after being scheduled and acquiring all necessary locks). The previous behavior was to throw all trees away, even though most would not be re-built until hours/days later.

For more information, more detailed commit message exist for most of these changes.

Sibling pull request: basho/riak_core#483, basho/riak_test#481

### DIFF
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -202,8 +202,15 @@ init([]) ->
     schedule_reset_build_tokens(),
     {ok, State2}.
 
-handle_call({set_mode, Mode}, _From, State) ->
-    {reply, ok, State#state{mode=Mode}};
+handle_call({set_mode, Mode}, _From, State=#state{mode=CurrentMode}) ->
+    State2 = case {CurrentMode, Mode} of
+                 {auto, manual} ->
+                     %% Clear exchange queue when switching to manual mode
+                     State#state{exchange_queue=[]};
+                 _ ->
+                     State
+             end,
+    {reply, ok, State2#state{mode=Mode}};
 handle_call({manual_exchange, Exchange}, _From, State) ->
     State2 = enqueue_exchange(Exchange, State),
     {reply, ok, State2};
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -204,7 +204,7 @@ init([]) ->
 
 handle_call({set_mode, Mode}, _From, State=#state{mode=CurrentMode}) ->
     State2 = case {CurrentMode, Mode} of
-                 {auto, manual} ->
+                 {automatic, manual} ->
                      %% Clear exchange queue when switching to manual mode
                      State#state{exchange_queue=[]};
                  _ ->
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -50,6 +50,7 @@
          insert_object/3,
          async_insert_object/3,
          stop/1,
+         clear/1,
          destroy/1]).
 
 -export([poke/1,
@@ -206,6 +207,10 @@ stop(Tree) ->
 destroy(Tree) ->
     gen_server:call(Tree, destroy, infinity).
 
+%% @doc Clear the specified index_hashtree, clearing all associated hashtrees
+clear(Tree) ->
+    gen_server:call(Tree, clear, infinity).
+
 %%%===================================================================
 %%% gen_server callbacks
 %%%===================================================================
@@ -297,6 +302,10 @@ handle_call({compare, Id, Remote, AccFun, Acc}, From, State) ->
 handle_call(destroy, _From, State) ->
     State2 = destroy_trees(State),
     {stop, normal, ok, State2};
+
+handle_call(clear, _From, State) ->
+    State2 = clear_tree(State),
+    {reply, ok, State2};
 
 handle_call(_Request, _From, State) ->
     Reply = ok,
@@ -653,7 +662,7 @@ maybe_clear(State=#state{lock=undefined, built=true}) ->
                                 anti_entropy_expire,
                                 ?DEFAULT_EXPIRE),
     %% Need to convert from millsec to microsec
-    case Diff > (Expire * 1000) of
+    case (Expire =/= never) andalso (Diff > (Expire * 1000)) of
         true ->
             clear_tree(State);
         false ->
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -30,7 +30,7 @@
 -include_lib("riak_kv_vnode.hrl").
 
 %% API
--export([start/3, start_link/3]).
+-export([start/4, start_link/4]).
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
@@ -81,17 +81,17 @@
 
 %% @doc Spawn an index_hashtree process that manages the hashtrees (one
 %%      for each `index_n') for the specified partition index.
--spec start(index(), nonempty_list(index_n()), pid()) -> {ok, pid()} | 
-                                                         {error, term()}.
-start(Index, IndexNs=[_|_], VNPid) ->
-    gen_server:start(?MODULE, [Index, IndexNs, VNPid], []).
+-spec start(index(), nonempty_list(index_n()), pid(), boolean()) -> {ok, pid()} | 
+                                                                    {error, term()}.
+start(Index, IndexNs=[_|_], VNPid, VNEmpty) ->
+    gen_server:start(?MODULE, [Index, IndexNs, VNPid, VNEmpty], []).
 
 %% @doc Spawn an index_hashtree process that manages the hashtrees (one
 %%      for each `index_n') for the specified partition index.
--spec start_link(index(), nonempty_list(index_n()), pid()) -> {ok, pid()} |
-                                                              {error, term()}.
-start_link(Index, IndexNs=[_|_], VNPid) ->
-    gen_server:start_link(?MODULE, [Index, IndexNs, VNPid], []).
+-spec start_link(index(), nonempty_list(index_n()), pid(), boolean()) -> {ok, pid()} |
+                                                                         {error, term()}.
+start_link(Index, IndexNs=[_|_], VNPid, VNEmpty) ->
+    gen_server:start_link(?MODULE, [Index, IndexNs, VNPid, VNEmpty], []).
 
 %% @doc Add a key/hash pair to the tree identified by the given tree id
 %%      that is managed by the provided index_hashtree pid.
@@ -210,7 +210,7 @@ destroy(Tree) ->
 %%% gen_server callbacks
 %%%===================================================================
 
-init([Index, IndexNs, VNPid]) ->
+init([Index, IndexNs, VNPid, VNEmpty]) ->
     case determine_data_root() of
         undefined ->
             case riak_kv_entropy_manager:enabled() of
@@ -233,6 +233,11 @@ init([Index, IndexNs, VNPid]) ->
                            built=false,
                            path=Path},
             State2 = init_trees(IndexNs, State),
+            %% If vnode is empty, mark tree as built without performing fold
+            [begin
+                 lager:debug("Built empty AAE tree for ~p", [Index]),
+                 gen_server:cast(self(), build_finished)
+             end || VNEmpty =:= true],
             {ok, State2}
     end.
 
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -409,14 +409,25 @@ hash_object({Bucket, Key}, RObjBin) ->
 %% key/hash pair will be ignored.
 -spec fold_keys(index(), pid()) -> ok.
 fold_keys(Partition, Tree) ->
+    {Limit, Wait} = app_helper:get_env(riak_kv,
+                                       anti_entropy_build_throttle,
+                                       {1000000, 100}),
     Req = riak_core_util:make_fold_req(
-            fun(BKey={Bucket,Key}, RObj, _) ->
+            fun(BKey={Bucket,Key}, RObjBin, Acc) ->
+                    ObjSize = byte_size(RObjBin),
+                    if (Limit =/= 0) andalso ((Acc + ObjSize) > Limit) ->
+                            lager:debug("Throttling AAE rebuild for ~b ms", [Wait]),
+                            Acc2 = 0,
+                            timer:sleep(Wait);
+                       true ->
+                            Acc2 = Acc + ObjSize
+                    end,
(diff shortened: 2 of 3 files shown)


jtuple:
I've rebased this branch on top of develop and force pushed to once again make this merge clean. The only conflict was in `riak_kv_index_hashtree.erl` with the changes made to integrate AAE with the new background manager vnode lock. This pull-request is now updated to account for the background manager changes. Only the "Change AAE to no longer delete trees..." commit changed in this rebase.

I've maintained the old pre-rebase branch at [jdb-aae-improvements.old](https://github.com/basho/riak_kv/tree/jdb-aae-improvements.old) for reference.

For the curious, here's the [old version](https://github.com/basho/riak_kv/commit/da434093073e0d6f9a1d48e8afce21b5ca84cf9c) of the lock change commit vs the [new version](https://github.com/basho/riak_kv/commit/5c346c347beb7459fe70e53275f958a68633f5fb) (as seen in this PR).

engelsanchez:
The throttle Limit/Wait was very simple to implement. But it's hard to think about it from an operator's point of view.  It would be nice to think about it in terms of something like a  duty cycle ("as fast as you can" or a percentage of that) or even better, total bandwith (AAE Should not be cost me more than X MB/s) implemented over some internal (per second perhaps).  If not total, which sounds like a pain to implement, perhaps at least per build. Thinking per build X number of allowed concurrent builds is still not bad. What do you think?

jtuple:
So, it's not just about being simple to implement. It's about having fine grained control.

I don't like duty cycle approach, as it loses absolute units. When taking about disk, I think bandwidth is the important metric. MB/s is the right approach, and that's what I implemented via Limit/Wait. The current default at 1 MB/100ms = 10 MB/s max rate per build. 

Sure, a user could just say 10 MB/s, but we lose fine grained control. There's different ways I can express 10 MB/s: 1MB / 100ms, 5 MB / 500ms, 10 MB / 1000 ms, etc. I think the user should be able to tune that, as it's going to depend on their workload. It's not terribly hard as an operator to say (assuming we provide docs/examples): "I want 25 MB/s for rebuild rate per build. I can get that with 2.5 MB/100 ms, 5 MB/200 ms, etc. Let's try each one and see what works best and gives us least latency impact"

MB/s = limit_mb \* 1000 / wait_ms

engelsanchez:
I think it's still too subtle. Having the max rate exposed through cuttlefish and the two units in advanced configs would probably be a good compromise. There would be a lot of variation with the rate calculated that way for an operator to make sense of. The actual rate will tend to be a lower that the upper bound given by s/wait_ms, only approaching it with very fast disk or other low latency scenarios. The good thing about duty cycle is that it adapts better to high disk usage, horrible latencies without too many calculations.  If reading those 5MB is taking multiple seconds, with a duty cycle of %50 you will make it wait multiple seconds before it tries again. Good in case of an emergency when a calculator isn't handy :). In that same scenario with a size/wait pair, you will tend to spend little time waiting when you should spend more. But rates are definitely more user friendly. If you find me somebody who, amongst the sea of options we've given them, has time to tune this pair as is for their workload, I will owe you a case of beer. Somehow I don't think that's going to happen.

engelsanchez:
Would it be worth setting the default expiration defined value to never? I understand that having it change behavior is icky, but if our current assessment is that tree rebuilds are perceived problematic, maybe it's worth it.

jtuple:
The hope is that the rebuild throttle makes rebuilds less problematic. Hopefully we can get some good testing in during the QA cycle and make a call. If things are still an issue, I'm open to setting to `never`.

The problem is that without rebuilds we may never repair data in certain corner cases. If a user's only option is `off` or `on + never`, then occasionally wrong AAE is the better option. But, for general usage, it's a sad option. We have users today that run AAE and tolerate rebuilds, so I'm keen to lean towards ship with best AAE, let users reduce it's usefulness if needed. Hopefully some performance testing makes this a clear cut decision.

engelsanchez:
Ugh, fml. I just realized that clear_tree does not use the original list of index_n pairs passed in, but calls responsible_preflists again. That means that for 2i AAE the "special" tree is not retained. Ugh, blergh

src/riak_kv_index_hashtree.erl
```diff
@@ -675,52 +711,98 @@ maybe_build(State) ->
...
-    State#state{trees=Trees2}.
+    Trees2 = [begin
+                  NewTree = try
+                                hashtree:flush_buffer(Tree)
+                            catch _:_ ->
+                                    Tree
+                            end,
+                  {IdxN, NewTree}
+              end || {IdxN, Tree} <- Trees],
+    Trees3 = [{IdxN, hashtree:close(Tree)} || {IdxN, Tree} <- Trees2],
```
engelsanchez, line 788: I can see that the current hashtree:close() operations is pretty crash safe. But code does change. Would it make sense to wrap the try catch around the flush and close? I would think whatever makes flush fail, will make close fail, even though right now it would probably just fail silently.

src/riak_kv_entropy_manager.erl
```diff
@@ -202,8 +202,15 @@ init([]) ->
     schedule_reset_build_tokens(),
     {ok, State2}.
 
-handle_call({set_mode, Mode}, _From, State) ->
-    {reply, ok, State#state{mode=Mode}};
+handle_call({set_mode, Mode}, _From, State=#state{mode=CurrentMode}) ->
+    State2 = case {CurrentMode, Mode} of
+                 {auto, manual} ->
```
engelsanchez: This should be `automatic` and not `auto`, right?

src/riak_kv_index_hashtree.erl
```diff
@@ -65,6 +67,7 @@
 -record(state, {index,
                 vnode_pid,
                 built,
+                expired :: boolean(),
```
engelsanchez, line 70: Sounds like we're starting to go down a path where have a state enumeration (not_built, built, expired) or something might be cleaner than the separate boolean flags. What do you think?

src/riak_kv_index_hashtree.erl
```diff
@@ -625,25 +662,24 @@ do_compare(Id, Remote, AccFun, Acc, From, State) ->
...
-    State1 = maybe_clear(State),
+    State1 = maybe_rebuild(maybe_expire(State)),
     State2 = maybe_build(State1),
     State2.
 
-%% If past expiration, clear all hashtrees.
--spec maybe_clear(state()) -> state().
-maybe_clear(State=#state{lock=undefined, built=true}) ->
+-spec maybe_expire(state()) -> state().
+maybe_expire(State=#state{lock=undefined, built=true}) ->
```
engelsanchez: It might be worth adding a clause for the expired=true case.

src/riak_kv_index_hashtree.erl
```diff
@@ -675,52 +711,98 @@ maybe_build(State) ->
...
-                    [hashtree:rehash_tree(T) || {_,T} <- Trees],
-                    lager:debug("Finished rehash (a): ~p", [Index]),
-                    gen_server:cast(Self, build_finished);
-                {_Error, _} ->
-                    gen_server:cast(Self, build_failed)
-            end;
-        max_concurrency ->
-            %% Something else is already doing a fold on this vnode.
+    Locked = get_all_locks(Type, Index, self()),
+    build_or_rehash(Self, Locked, Type, State).
```
engelsanchez, line 737: It's a tiny bit confusing to have the incoming pid be called Self, but using your real self in the line above. Since it's important to understand that pids end up associated with the locks, it might be good to rename?

src/riak_kv_index_hashtree.erl
jtuple, line 788: This try/catch is here because flush followed by close fails...often. I've never seen close fail. I have no idea why flush + close fails, but flush + close at Riak shutdown leads to flush error spam. Sometimes. Other times, all is well. Since we don't flush at all today, I figured attempting to flush and ignoring failure if we failed is better than nothing. Perhaps someday we'll have time to figure out why flush + close fails.

buddhisthead:
With respect to the build rate limiting, I think no matter how we offer it, or how simple it looks, we'll not be able to make it back off in terms of actual load until we have some metric of load available at runtime. That would make a dynamic rate limiter possible. Until then, I feel most comfortable with the limit based on bytes/sec where we can define both bytes and time as currently implemented.

Is it possible in cutlefish to offer two ways to address the same setting? One could be the current fine-grained version and one could hide the details with a simple "rate" that we decompose into {Limit,Time}. Advanced and intermediate offerings?

I really feel that we should be moving to dynamic configurations because as @engelsanchez implied, it's just onerous to tune all of our settings.

src/riak_kv_index_hashtree.erl
engelsanchez, line 788: Nasty. Maybe we are dancing around a LevelDB race on closing immediately following a batch of writes. Oh well.

engelsanchez:
@buddhisthead We do have the latency of the fold reads as a rough but very good indicator of overall load.  If the fold has only read 100K in 2 seconds, for example, you can make a pretty good guess that we are under load and that background tasks such as yourself should back off a little.

engelsanchez:
@jtuple, what is the plan for the cuttlefish changes related to this? I'm going to assume that for this release it's advanced.config only. For example, the 'never' value for anti-entropy expiration and stuff.

src/riak_kv_index_hashtree.erl
```diff
@@ -625,25 +662,24 @@ do_compare(Id, Remote, AccFun, Acc, From, State) ->
...
     Diff = timer:now_diff(os:timestamp(), State#state.build_time),
     Expire = app_helper:get_env(riak_kv,
                                 anti_entropy_expire,
                                 ?DEFAULT_EXPIRE),
     %% Need to convert from millsec to microsec
-    case Diff > (Expire * 1000) of
+    case (Expire =/= never) andalso (Diff > (Expire * 1000)) of
         true ->
-            clear_tree(State);
+            State#state{expired=true};
```
engelsanchez, line 695: It would be useful to add a debug message here informing of the expiration.

engelsanchez:
I've finished testing these changes manually.  If that auto atom is changed to automatic, everything seems to work as far as I can tell:
- Starting fresh on leveldb/bitcask immediately marked all trees as built.
- Setting expiration in advance.config to never seemed work. I didn't wait an infinite amount of time though
- Changing mode to manual from automatic clears the exchange queue in the state
- Tweaked expiration and observe trees stay around and do exchanges and live their normal lives
- When specified in advance.config, the throttle triggers the debug messages, seems to work.

I'm done with reviewing this for now. I expect future changes to cuttlefish post freeze to expose the new configuration and hopefully make the throttle one simple.

jtuple:
Address all review comments.

Before shipping 2.0 I want to expose the `never` change to rebuild time in `riak.conf`. When I started this branch, cufflefish didn't support type unions: `integer() | never`. If that gets added, I'll add `never`.

For throttle limits, I think the `Limit/Wait` is going to be `advanced.config` setting. Although, I'd like to provide a `Bytes/s` setting in `riak.conf` that we translate to `Limit/Wait` likely using fixed `100ms` wait.

jtuple:
I've merged this branch to develop offline. There was some merge conflicts.

The main changes when merging to `riak_kv_index_hashtree` was conflicts against the new 2i AAE logic just merged. This changes how folds are done. This leads to the throttle being moved into the new new fold functions, with the following minimal diff for reference (diff against develop per-merge):

``` diff
@@ -410,32 +445,52 @@ hash_index_data(IndexData) when is_list(IndexData) ->
 fold_keys(Partition, Tree, HasIndexTree) ->
     FoldFun = fold_fun(Tree, HasIndexTree),
     Req = riak_core_util:make_fold_req(FoldFun,
-            ok, false, [aae_reconstruction]),
+                                       0, false, [aae_reconstruction]),
     riak_core_vnode_master:sync_command({Partition, node()},
                                         Req,
                                         riak_kv_vnode_master, infinity),
     ok.

 %% @doc Generate the folding function
 %% for a riak fold_req
 -spec fold_fun(pid(), boolean()) -> fun().
 fold_fun(Tree, _HasIndexTree = false) ->
     ObjectFoldFun = object_fold_fun(Tree),
-    fun(BKey, RObj, _) ->
+    {Limit, Wait} = get_build_throttle(),
+    fun(BKey, RObj, Acc) ->
             BinBKey = term_to_binary(BKey),
             ObjectFoldFun(BKey, RObj, BinBKey),
-            ok
+            Acc2 = maybe_throttle_build(RObj, Limit, Wait, Acc),
+            Acc2
 fold_fun(Tree, _HasIndexTree = true) ->
     %% Index AAE backend, so hash the indexes
     ObjectFoldFun = object_fold_fun(Tree),
     IndexFoldFun = index_fold_fun(Tree),
-    fun(BKey = {Bucket, Key}, BinObj, _) ->
+    {Limit, Wait} = get_build_throttle(),
+    fun(BKey = {Bucket, Key}, BinObj, Acc) ->
             RObj = riak_object:from_binary(Bucket, Key, BinObj),
             BinBKey = term_to_binary(BKey),
             ObjectFoldFun(BKey, RObj, BinBKey),
             IndexFoldFun(RObj, BinBKey),
-            ok
+            Acc2 = maybe_throttle_build(RObj, Limit, Wait, Acc),
+            Acc2
     end.
```

Likewise, the `build_or_rehash` logic that was modified in this branch also conflicted with a minor 2i AAE change. Everything merged towards the approach in this branch (expired, then rebuild), but there was one line conflict that was ported over from develop. Here's the minimal diff for reference (this diff is against this branch):

``` diff
@@ -740,7 +832,7 @@ build_or_rehash(Self, Locked, Type, #state{index=Index, trees=Trees}) ->
     case {Locked, Type} of
         {true, build} ->
             lager:debug("Starting build: ~p", [Index]),
-            fold_keys(Index, Self),
+            fold_keys(Index, Self, has_index_tree(Trees)),
             lager:debug("Finished build (a): ~p", [Index]), 
             gen_server:cast(Self, build_finished);
         {true, rehash} ->
```

Otherwise, everything merged automatically (or as you expected for simple stuff: new exports, new state elements, etc).

With that said, going to push out the merged branch and close out this pull-request.

PR #766, opened 2013-12-16
Fix error when hashing index data in tree builds

engelsanchez:
/cc @jtuple 

The riak test for this: https://github.com/basho/riak_test/pull/483


bookshelfdave:
+1, dependent on riak_test results

PR #769, opened 2013-12-18
Integrate Background Manager with Handoff And AAE Tree Rebuilds

buddhisthead:
NOTE: DO NOT MERGE YET (this PR requires code from #475 or https://github.com/basho/riak_core/pull/484)

The riak_core handoff portion is here: https://github.com/basho/riak_core/pull/484 (includes #475)
The riak_core background manager PR is here: basho/riak_core#475
The riak_repl PR is here: basho/riak_repl#488

The background manager allows subsystems to coordinate with locks (and tokens) in a non-blocking API when desired. In our case, we've seen several customer support issues related to a production cluster being overloaded by the combined weight of MDC fullsync replication in combination with either handoff, or AAE tree rebuilds (or all three!).

Each of these subsystems is being taught, for 2.0, to "take" a lock on the kv vnode/partition before it does a fold over that partition. Locks are returned when the process dies or exists normally. Each subsystem (at this point in time) is coded to keep working on other partitions if one partition is denied a lock via the return value of max_concurrency, which means progress is made even when one partition is busy.

Use of this feature can be bypassed in production. Besides a global skill switch on the background manager itself, the riak_core configuration supports two new parameters: aae_skip_bg_manager and handoff_skip_bg_manager, which defaults to false. Setting it to true will not call the background manager and proceed like it used to.

I chose not to separate the AAE and Handoff into two separate PRs because we already have too many, but if we determine that we want to exclude the handoff portion, I'll spit them up.

### DIFF
--- a/include/riak_kv_vnode.hrl
+++ b/include/riak_kv_vnode.hrl
@@ -68,4 +68,4 @@
 %% @doc vnode_lock(PartitionIndex) is a kv per-vnode lock, used possibly,
 %% by AAE tree rebuilds, fullsync, and handoff.
 %% See @link riak_core_background_mgr:get_lock/1
--define(VNODE_LOCK(Idx), {vnode_lock, Idx}).
+-define(KV_VNODE_LOCK(Idx), {vnode_lock, Idx}).
--- a/include/riak_kv_vnode.hrl
+++ b/include/riak_kv_vnode.hrl
@@ -64,3 +64,8 @@
 -define(KV_DELETE_REQ, #riak_kv_delete_req_v1).
 -define(KV_MAP_REQ, #riak_kv_map_req_v1).
 -define(KV_VCLOCK_REQ, #riak_kv_vclock_req_v1).
+
+%% @doc vnode_lock(PartitionIndex) is a kv per-vnode lock, used possibly,
+%% by AAE tree rebuilds, fullsync, and handoff.
+%% See @link riak_core_background_mgr:get_lock/1
+-define(VNODE_LOCK(Idx), {vnode_lock, Idx}).
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -28,6 +28,7 @@
 -behaviour(gen_server).
 
 -include_lib("riak_kv_vnode.hrl").
+-include_lib("riak_core/include/riak_core_locks.hrl").
 
 %% API
 -export([start/3, start_link/3]).
@@ -680,22 +681,60 @@ build_or_rehash(Self, State=#state{index=Index, trees=Trees}) ->
                false -> build;
                true  -> rehash
            end,
-    Lock = riak_kv_entropy_manager:get_lock(Type),
-    case {Lock, Type} of
-        {ok, build} ->
-            lager:debug("Starting build: ~p", [Index]),
-            fold_keys(Index, Self),
-            lager:debug("Finished build (a): ~p", [Index]), 
-            gen_server:cast(Self, build_finished);
-        {ok, rehash} ->
-            lager:debug("Starting rehash: ~p", [Index]),
-            [hashtree:rehash_tree(T) || {_,T} <- Trees],
-            lager:debug("Finished rehash (a): ~p", [Index]),
-            gen_server:cast(Self, build_finished);
-        {_Error, _} ->
+    case maybe_get_vnode_lock(Index) of
+        ok ->
+            Lock = riak_kv_entropy_manager:get_lock(Type),
+            case {Lock, Type} of
+                {ok, build} ->
+                    lager:debug("Starting build: ~p", [Index]),
+                    fold_keys(Index, Self),
+                    lager:debug("Finished build (a): ~p", [Index]), 
+                    gen_server:cast(Self, build_finished);
+                {ok, rehash} ->
+                    lager:debug("Starting rehash: ~p", [Index]),
+                    [hashtree:rehash_tree(T) || {_,T} <- Trees],
+                    lager:debug("Finished rehash (a): ~p", [Index]),
+                    gen_server:cast(Self, build_finished);
+                {_Error, _} ->
+                    gen_server:cast(Self, build_failed)
+            end;
+        max_concurrency ->
+            %% Something else is already doing a fold on this vnode.
             gen_server:cast(Self, build_failed)
     end.
 
 close_trees(State=#state{trees=Trees}) ->
     Trees2 = [{IdxN, hashtree:close(Tree)} || {IdxN, Tree} <- Trees],
     State#state{trees=Trees2}.
+
+%% @private
+%% @doc Return true iff neither of the "skip background manager" configuration
+%%      settings are defined as anything other than false. 'skip_background_manager'
+%%      turns off all use of bg-mgr. 'aae_skip_background_manager' only stops
+%%      AAE tree rebuilds from using bg-mgr.
+-spec use_bg_mgr() -> boolean().
+use_bg_mgr() ->
+    %% note we're tolerant here of any non-boolean value as well. Iff it's 'false', we don't skip.
+    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager, false) =/= false,
+    AAESkip = not app_helper:get_env(riak_core, aae_skip_background_manager, false) =/= false,
+    not (GlobalSkip orelse AAESkip).
+
+%% @private
+%% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
+%%      Sets our task meta-data in the lock as 'aae_rebuild', which is useful for
+%%      seeing what's holding the lock via @link riak_core_background_mgr:ps/0.
+-spec maybe_get_vnode_lock(SrcPartition::integer()) -> ok | max_concurrency.
+maybe_get_vnode_lock(SrcPartition) ->
+    Lock = ?VNODE_LOCK(SrcPartition),
+    case use_bg_mgr() of
+        true  ->
+            case riak_core_bg_manager:get_lock(Lock, self(), [{task, aae_rebuild}]) of
+                {ok, _Ref} -> ok;
+                max_concurrency -> max_concurrency
+            end;
+        false ->
+            lager:debug("Handoff is skipping the background manager vnode lock: ~p", [Lock]),
+            ok
+    end.
+
+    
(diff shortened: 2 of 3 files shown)


src/riak_kv_index_hashtree.erl
```diff
@@ -680,22 +681,60 @@ build_or_rehash(Self, State=#state{index=Index, trees=Trees}) ->
...
     Trees2 = [{IdxN, hashtree:close(Tree)} || {IdxN, Tree} <- Trees],
     State#state{trees=Trees2}.
+
+%% @private
+%% @doc Return true iff neither of the "skip background manager" configuration
+%%      settings are defined as anything other than false. 'skip_background_manager'
+%%      turns off all use of bg-mgr. 'aae_skip_background_manager' only stops
+%%      AAE tree rebuilds from using bg-mgr.
+-spec use_bg_mgr() -> boolean().
+use_bg_mgr() ->
```
cmeiklejohn: This function is duplicated in both `riak_kv` and `riak_repl`.  Can't this be in core, and take a parameter instead of what property to inspect in the environment, if required?

src/riak_kv_index_hashtree.erl
```diff
@@ -680,22 +681,60 @@ build_or_rehash(Self, State=#state{index=Index, trees=Trees}) ->
...
+    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager, false) =/= false,
+    AAESkip = app_helper:get_env(riak_core, aae_skip_background_manager, false) =/= false,
+    not (GlobalSkip orelse AAESkip).
+
+%% @private
+%% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
+%%      Sets our task meta-data in the lock as 'aae_rebuild', which is useful for
+%%      seeing what's holding the lock via @link riak_core_background_mgr:ps/0.
+-spec maybe_get_vnode_lock(SrcPartition::integer()) -> ok | max_concurrency.
+maybe_get_vnode_lock(SrcPartition) ->
```
cmeiklejohn, line 714: This function is also duplicated verbatim in `riak_repl`.  Can't this be in `riak_core` and be parameterized?

buddhisthead, line 714: You mean as a helper function? I'm not sure where it would go. I don't think it belongs in the background manager. We could have guarded all the entry points with checks to the global switch, but that would undermine the module itself and make it harder to test. Likewise, each subsystem wants its own switch to be able to ignore the bg-mgr. It's a slippery slope. We'd need to parameterize for this special case of using our task name in the meta data, e.g. `{task, aae_rebuild}`. I agree that it's nice to reduce duplicated code, though. I wonder where we could put it? Somewhere in KV, perhaps? It's kind of specific to the set of consumers that happen to be locking on a KV node lock. I'm going to bring @jrwest into this to help. He has a good feel for this kind of question.

src/riak_kv_index_hashtree.erl
buddhisthead: I think for the global switch, you have a good point. I'll add this entry to the background manager.

src/riak_kv_index_hashtree.erl
jrwest, line 714: At least for the "global flag", `skip_background_manager` we should move that into `get_lock`. In psuedo-erlang:

``` erlang
%% in bg manager change get lock to
get_lock(Lock, Pid, Meta) ->
   case app_helper:get_env(riak_core, skip_background_manager, false) of
        false -> %% rest of code
        true -> max_concurrency
   end.

```

From there its not too hard to take that a step further and make a `get_lock/4` like:

``` erlang
get_lock(Lock, Pid, Meta, SubsysSkip) ->
    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager),
    DoSkip = (SubsysSkip =:= true) orelse (GlobalSkip =:= true)
    case DoSkip of
        false -> %% proceed
        true -> max_concurrency
    end.
```

This would get rid of the need for `maybe_get_vnode_locks` at each callsite and replace it w/ e.g.:

``` erlang
riak_core_bg_manager:get_lock({riak_kv_vnode, SrcPartition}, self(), 
              [{task, aae_rebuild}], app_helper:get_env(riak_kv, aae_skip_background_manager))
```

Abstracting over the flag using the function argument is a bit wierd since passing true basically means "call this function but don't" but I think its better than duplicating this code in several subsystems.

In addition, `aae_skip_background_manager` should be a `riak_kv` application variable not a `riak_core` one. 

I'd also love to see them re-worded in the positive e.g. `use_background_manager` and `aae_use_background_manager` (which would change my implementations above a bit as well obviously).

buddhisthead, line 714: While I don't mind moving the global check into get_lock, I don't like moving this into the arguments of get_lock, unless we're not planning on writing more eqc tests for it. Having a function in bg-mgr that does the checks for you is cool, though. Otherwise, it's going to get messy. We'd need to add these to more of the API, at least get_token, and set_concurrency_limit. More surface area.

I propose that I create two additional helper functions in bg-mgr that do the config checking, but the call points still do the tests.

jrwest, line 714: We're going to need to test it anyways whether it's at the call or definition site and modifying the eqc test for this should be pretty straight forward. Plus the eqc test already exists. It seems to me that the new surface area is either spread over the call sites or the definition site (bg mgr) since the latter is new code IMO vs changing established code even more it makes more sense to localize the changes to the bg mgr and reduce repetition at least to me.

All in all the, I'm somewhat indifferent if it continues to work.

buddhisthead, line 714: Ok, well I've tried it both ways and the code is no prettier with the checks in the bg_manager. In fact, it complicates the function type signatures because we have different meanings for "ok, we're not using this API, so continue anyhow"  vs "ok and here is your lock reference". So, I'm voting to add convenience functions to the bg_mgr for getting switch values. That does reduce code in the subsystems. But not moving the checks inside of the calls themselves.

And I changed meaning of the switches to be expressed in the positive.

src/riak_kv_vnode.erl
```diff
@@ -1937,6 +1950,58 @@ sanitize_bkey({{<<"default">>, B}, K}) ->
...
+                      N when is_integer(N) -> N;
+                      _NotNumber -> 1
+                  end,
+    try_set_concurrency_limit(?VNODE_LOCK(Idx), Concurrency).
+
+try_set_concurrency_limit(Lock, Limit) ->
+    try_set_concurrency_limit(Lock, Limit, riak_core_bg_manager:use_bg_mgr()).
+
+try_set_concurrency_limit(_Lock, _Limit, false) ->
+    lager:info("Skipping background manager."),
```
jrwest: this is probably going to get verbose as an `info` statement since its per-partition

buddhisthead: Oh heck. I meant to make that a debug. I wonder if we even need it all.

jrwest: i don't think we need it

src/riak_kv_vnode.erl
```diff
@@ -1937,6 +1950,58 @@ sanitize_bkey({{<<"default">>, B}, K}) ->
 sanitize_bkey(BKey) ->
     BKey.
 
+%% @private
+%% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
+%%      Sets our task meta-data in the lock as 'handoff', which is useful for
+%%      seeing what's holding the lock via @link riak_core_background_mgr:ps/0.
+-spec maybe_get_vnode_lock(SrcPartition::integer(), pid()) -> ok | max_concurrency.
+maybe_get_vnode_lock(SrcPartition, Pid) ->
+    case riak_core_bg_manager:use_bg_mgr(riak_kv, handoff_skip_background_manager) of
```
jrwest: this per-subsystem switch wasn't changed to be "in the positive", e.g. `handoff_use_backgroun_manager`. so right now `handoff_skip_background_manager=true` mean "use the background manager".

buddhisthead: Yep. Must have been tired last night. I'll change that.

include/riak_kv_vnode.hrl
```diff
@@ -64,3 +64,8 @@
 -define(KV_DELETE_REQ, #riak_kv_delete_req_v1).
 -define(KV_MAP_REQ, #riak_kv_map_req_v1).
 -define(KV_VCLOCK_REQ, #riak_kv_vclock_req_v1).
+
+%% @doc vnode_lock(PartitionIndex) is a kv per-vnode lock, used possibly,
+%% by AAE tree rebuilds, fullsync, and handoff.
+%% See @link riak_core_background_mgr:get_lock/1
+-define(VNODE_LOCK(Idx), {vnode_lock, Idx}).
```
jrwest: its probably not a big deal but now that this is moved naming it something kv specific might be nice.

src/riak_kv_index_hashtree.erl
jrwest, line 714: i find this method very verbose, still. Looking at my version it seems it would get rid of the need for  `maybe_get_vnode_lock` entirely since you can just call `get_lock`directly above. If the problem is there is no ref, make a fake one and pattern match on `{ok, _}` instead of `ok` above at the call-site. At this point though its probably not worth changing.

jrwest:
+1 merge after fixing the `handoff_skip_background_manager` config and removing the overly verbose logging commented on above. 

I tested this mostly by hand because unfortunately there aren't many riak_test's that cover this. I verified that AAE rebuilds and handoff's don't stomp on each other via redbug (`max_concurrency` returned as expected, only one `?FOLD_REQ` handled, etc) in addition to grabbing some of the partition locks myself and ensuring that neither subsystem could do any work (a nice trick btw). Also checked that turning off the subsystems' use of the background manager or global use of it allowed them to proceed anyways.

Final Comments:
-  A riak_test that does some of the above would be a nice addition to the test suite
- I think `try_set_concurrency_limit` is exactly why we should move away from the table manager in the future and use a `public` ets table. Having one of those in each subsystem (or some general version) is going to get pretty annoying.
- Along w/ that, I think we should re-consider my comments above about moving the subsystem enabled checking into a `get_lock/4`. Otherwise, each subsystem we add code to is going to need a similar `maybe_lock_something` function.
- we still need cuttlefish mappings for the various new configs added

src/riak_kv_vnode.erl
buddhisthead: Addressed by a78aee5

src/riak_kv_vnode.erl
buddhisthead: Addressed by a78aee5

include/riak_kv_vnode.hrl
buddhisthead: Addressed by a78aee5

PR #771, opened 2013-12-19
Add support for conditional postcommit hooks

jtuple:
This commit adds support for conditional postcommit hooks. Currently, these
are implemented solely for typed buckets, but may be expanded to untyped
buckets in the future.

This addition is motivated by the lack of support for "bucket fixups" for
typed buckets. For traditional buckets, fixups are the primary mechanism
used in Riak to ensure hooks are installed on buckets as necessary.

Unlike traditional postcommits hooks, that are registered per-bucket,
conditional hooks are registered globally (per node). Each time a PUT is
triggered, all registered conditional postcommit hooks are triggered. The
hook function is passed the relevant bucket, key, and bucket properties and
should return either false or a list of postcommit hooks that should be
triggered. If any hooks are returned, they are invoked the same as normal
postcommit hooks.

Test pull-request: basho/riak_test#485

### DIFF
--- a/src/riak_kv_hooks.erl
+++ b/src/riak_kv_hooks.erl
@@ -1,0 +1,118 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2012-2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(riak_kv_hooks).
+
+%% API
+-export([add_conditional_postcommit/1,
+         del_conditional_postcommit/1,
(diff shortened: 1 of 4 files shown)


src/riak_kv_hooks.erl
```diff
@@ -0,0 +1,118 @@
...
+%% @doc
+%% This function invokes each registered conditional postcommit
+%% hook. Each hook will return either `false' or a list of active
+%% hooks. This function then returns the combined list of active hooks.
+-spec get_conditional_postcommit({bucket(), key()}, bucket_props()) -> [any()].
+get_conditional_postcommit({{BucketType, Bucket}, _Key}, BucketProps) ->
+    Hooks = get_hooks(conditional_postcommit),
+    ActiveHooks =
+        [ActualHook || {Mod, Fun} <- Hooks,
+                       ActualHook <- [Mod:Fun(BucketType, Bucket, BucketProps)],
```
jrwest, line 75: The other functions that take bucket type and bucket, including `get_conditional_postcommit`, typically pass them as a 2-tuple. If we were to extend `get_conditional_postcommit` to untyped buckets I think we'd probably want to make this a 2-arity function.

jrwest:
changes look good. I like the public ets table approach used here w/ the backup to appenv. I left one comment but I don't see that as necessary to change before merge. Test in https://github.com/basho/riak_test/pull/485 was used to verify the changes. In addition, I hacked up the test a bit to ensure that we only run the postcommit hook once if it is in both bucket props and in the conditional hooks as well as a few other cases. 

+1 merge.

jtuple:
Yeah, probably should change things to a 2-tuple in that one case, but since the relevant riak_repl PR that adds Riak's first ever conditional hook is coded to the current approach, I'm going to leave things as-is since we're trying to wrap up things for code freeze.

If/when we get around to extending to untyped buckets, we'll make the change and update existing conditional hooks as necessary.

PR #772, opened 2013-12-19
Fix broken return terms handling

engelsanchez:
This problem was introduced in the previous hotfix for mixed cluster
queries.  Queries not specifying the return_terms parameter were
failing.


russelldb:
- [x] all 2i tests eleveldb
- [x] all 2i tests memory
- [x] mixed cluster tests from https://github.com/basho/riak_kv/pull/767 by hand

russelldb:
All tests pass on both backends and also those mixed tests from riak_kv#767.

Would love it if those tests were riak_tests using both clients. Anyway. +1. Thanks.

PR #773, opened 2013-12-19
Remove reset functionality from bucket-type API.

seancribbs:
Pursuant to yesterday's discussion. @jrwest @Vagabond @broach

### DIFF
--- a/src/riak_kv_wm_bucket_type.erl
+++ b/src/riak_kv_wm_bucket_type.erl
@@ -42,9 +42,6 @@
 %%     {"props":{Prop:Val}}
 %%   Where the "props" object takes the same form as returned from
 %%   a GET of the same resource.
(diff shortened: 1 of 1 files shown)


jrwest:
[The check for `'DELETE'` here](https://github.com/basho/riak_kv/blob/2eea9e5315fab62644a81f59ffdca77d390d8a11/src/riak_kv_wm_bucket_type.erl#L150) is extraneous, although removing it would only serve for clarity. Obviously this is caught earlier by `allowed_methods`.

jrwest:
+1 merge. verified w/ `http_bucket_types`. test attempting to do a reset fails as expected. checking out `refactor/sdc/remove-bucket-type-reset` in `riak_test` makes the test happy, also as expected.

PR #775, opened 2013-12-20
Make memory backend obey 2i return terms properly

engelsanchez:
Otherwise it might send term/key pairs to an old node, which wouldn't
know what to do with them. Found with verify_2i_mixed_cluster test.

Compare this to similar code in riak_kv_eleveldb_backend.erl

/cc @jaredmorrow @jonmeredith


engelsanchez:
Found with this riak test by configuring it to pick up 1.2.1 as legacy and the memory backend.

engelsanchez:
Ugh, I meant this: https://github.com/basho/riak_test/pull/489

russelldb:
I need to run basho/riak_test#489 with 1.2.1 as legacy without this patch, then with this patch to verify it works, correct?

engelsanchez:
You could try any release < 1.4 I believe. Anything before return terms should have the same problem.

russelldb:
OK, finally able to get basho/riak_test#489 to fail without this patch, and pass with. So. Plus One. Again.

engelsanchez:
It's déjà vu all over again

PR #778, opened 2013-12-23
Fix bug with stats for legacy counters.

seancribbs:
With legacy counters, the operation sent to the vnode is simply an
integer increment, not a `crdt_op` record.


Vagabond:
+1, I discovered this and tested the fix.

PR #780, opened 2013-12-23
Fix lost 2i AAE tree on rebuild

engelsanchez:
This is a fix to avoid 2i AAE trees to be dropped on a rebuild.

/cc @jtuple @jonmeredith

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -69,7 +69,8 @@
                 lock :: undefined | reference(),
                 path,
                 build_time,
-                trees}).
+                trees,
+                use_2i = false :: boolean()}).
 
 -type state() :: #state{}.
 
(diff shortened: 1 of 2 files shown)


jtuple:
Should we make [handle_unexpected_key/3](https://github.com/basho/riak_kv/blob/b7eaba87e09dddfe93bac65f407eb67cbd6afdad/src/riak_kv_index_hashtree.erl#L628) smarter as well? Implicitly we should never send an unexpected key in, but who knows what this code is going to do in the future.

In general, I think the notion of needing/not-needing an index tree should be first-class concept in `riak_kv_index_hashtree`. I imagine such a change is for a different pull-request, since we're trying to get this one out the door. But, seriously, current approach is kinda error-prone.

Would rather have `riak_kv_index_hashtree` be told when created if it needs an index tree or not. Likewise, move away from using `responsible_preflists` directly and have a new wrapper function `responsible_trees(State)` that returns the preflists + index tree (if needed). Could then extend to other special tree types we one day add as well.

Food for thought. Likely for another day.

Haven't got around to testing this yet. But code looks good. Will run any new riak_tests when they become available.

engelsanchez:
Yes, I went for the smallest change possible for this since it will sent out as a beam for a customer. But I'm all for making it cleaner for 2.0, hopefully with the next pre.

engelsanchez:
The last commit cleans up the 2i setup in riak_kv_index_hashtree. It passes the updated 2i AAE test with the tree rebuild.  The first commit on this PR is still a better candidate for a patch to 1.4.6 as it involves a single beam with a single function changed. The functionality is the same.

jtuple:
Reviewed both commits independently. And both address the problem. Ran the improved riak_test and things fail on the current 1.4 branch, and pass against both commits in this pull-request. Also, I manually played with tree expiration and used `sys:get_status` to look at the state of the `riak_kv_index_hashtree` processes. I verified that on 1.4, the `{0,0}` tree is indeed missing after expiration/rebuild, whereas the tree is preserved in both of the new commits.

Only 1 code style comment. Both commits have code of the form:

``` erlang
IndexNs = riak_kv_util:responsible_preflists(Index) ++
[?INDEX_2I_N || has_index_tree(Trees)],
```

in which it's easy to overlook that the list comprehension is an extension of the previous line, since it's at the same indentation level. I suggest indenting the second line, or slightly rewriting code to keep things on 1 line.

Eg. indention (this is what emacs mode would do):

``` erlang
responsible_preflists(#state{index=Partition, use_2i=Use2i}) ->
    RP = riak_kv_util:responsible_preflists(Partition) ++
        [?INDEX_2I_N || Use2i],
    RP.
```

Eg. minor rewrite:

``` erlang
responsible_preflists(#state{index=Partition, use_2i=Use2i}) ->
    Maybe2i = [?INDEX_2I_N || Use2i],
    RP = riak_kv_util:responsible_preflists(Partition) ++ Maybe2i,
    RP.
```

Other than the style suggestion, +1 / ship it.

engelsanchez:
I agree with the style comment. My bad: I recently updated to vim 7.4 which makes indentation closer to Emacs, except for this oddity, which I haven't managed to configure/hack properly so I stop sprinkling code with that indentation style.  I pinky swear I'll fix that. For now, I'm merging as is to stop messing with this branch (one can dream).

PR #699, opened 2013-10-10
Add metadata cache

evanmcc:
NB: this patch has merged https://github.com/basho/riak_kv/pull/695 and should not be merged until after that patch has landed (and rebased if that patch changes).

This change adds an ETS table per vnode to act as a cache for the vclocks and index specs of any values that have recently been gotten or put.  This is meant to save round-trips to the disk in common get-then-put workflows on slower disks.

The cache is size limited, and setting the size to 0 will disable it, since it is a slight performance hit in certain situations (e.g. using the memory backend, when all of your data fits into RAM, etc).

PR #781, opened 2013-12-23
Add metadata cache (as rebased by jtuple)

evanmcc:
This change adds an ETS table per vnode to act as a cache for the vclocks and index specs of any values that have recently been gotten or put. This is meant to save round-trips to the disk in common get-then-put workflows on slower disks.

The cache is size limited, and setting the size to 0 will disable it, since it may be slight performance hit in certain situations (e.g. using the memory backend, when all of your data fits into RAM, etc).

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -359,7 +359,17 @@ init([Index]) ->
     DeleteMode = app_helper:get_env(riak_kv, delete_mode, 3000),
     AsyncFolding = app_helper:get_env(riak_kv, async_folds, true) == true,
     MDCacheSize = app_helper:get_env(riak_kv, vnode_md_cache_size), 
-    MDCache = new_md_cache(VId), 
+    MDCache = 
+        case MDCacheSize of
+            N when is_integer(N),
+                   N > 0 ->
+                lager:info("Initializing metadata cache with size limit: ~p bytes", 
+                           [MDCacheSize]),
+                new_md_cache(VId);
+            _ -> 
+                lager:info("No metadata cache size defined, not starting"),
+                undefined
+        end,
     case catch Mod:start(Index, Configuration) of
         {ok, ModState} ->
             %% Get the backend capabilities
@@ -878,7 +888,7 @@ terminate(_Reason, #state{mod=Mod, modstate=ModState}) ->
 
 handle_info(recreate_md_cache, State=#state{vnodeid=VId}) ->
     lager:info("md_cache died and was recreated"),
-    State#state{md_cache=new_md_cache(VId)};
+    {ok, State#state{md_cache=new_md_cache(VId)}};
 handle_info(retry_create_hashtree, State=#state{hashtrees=undefined}) ->
     State2 = maybe_create_hashtrees(State),
     case State2#state.hashtrees of
@@ -1022,7 +1032,13 @@ prepare_put(#state{vnodeid=VId,
                              prunetime=PruneTime,
                              crdt_op = CRDTOp},
             IndexBackend) ->
-    {CacheClock, CacheSpecs} = check_md_cache(MDCache, BKey),
+    {CacheClock, CacheSpecs} = 
+        case MDCache of
+            undefined ->
+                {undefined, undefined};
+            _ ->
+                check_md_cache(MDCache, BKey)
+        end,
     RequiresGet = 
         case CacheClock of
             undefined ->
@@ -1152,9 +1168,14 @@ perform_put({true, Obj},
                      reqid=ReqID,
                      index_specs=IndexSpecs}) ->
     {Obj2, Fake} = riak_kv_mutator:mutate_put(Obj, BProps),
-    VClock = riak_object:vclock(Obj2),
-    Specs = riak_object:index_specs(Obj2),
-    insert_md_cache(MDCache, MDCacheSize, BKey, {VClock, Specs}),
+    case MDCache of 
+        undefined ->
+            ok;
+        _ ->
+            VClock = riak_object:vclock(Obj2),
+            Specs = riak_object:index_specs(Obj2),
+            insert_md_cache(MDCache, MDCacheSize, BKey, {VClock, Specs})
+    end,
     case encode_and_put(Obj2, Mod, Bucket, Key, IndexSpecs, ModState) of
         {{ok, UpdModState}, EncodedVal} ->
             update_hashtree(Bucket, Key, EncodedVal, State),
@@ -1254,8 +1275,10 @@ do_get(_Sender, BKey, ReqID,
                     md_cache=MDCache, md_cache_size=MDCacheSize}) ->
     StartTS = os:timestamp(),
     Retval = do_get_term(BKey, Mod, ModState),                
-    case Retval of
-        {ok, Obj} ->
+    case {MDCache, Retval} of
+        {undefined, _} ->
+            ok;
+        {_, {ok, Obj}} ->
             VClock = riak_object:vclock(Obj),
             Specs = riak_object:index_specs(Obj),
             insert_md_cache(MDCache, MDCacheSize, BKey, {VClock, Specs});
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -104,13 +104,16 @@
                 key_buf_size :: pos_integer(),
                 async_folding :: boolean(),
                 in_handoff = false :: boolean(),
-                hashtrees :: pid() }).
+                hashtrees :: pid(),
+                md_cache :: ets:tab(), 
+                md_cache_size :: pos_integer() }).
 
 -type index_op() :: add | remove.
 -type index_value() :: integer() | binary().
 -type index() :: non_neg_integer().
 -type state() :: #state{}.
 
+-define(MD_CACHE_BASE, "riak_kv_vnode_md_cache").
 -define(DEFAULT_HASHTREE_TOKENS, 90).
 
 -record(putargs, {returnbody :: boolean(),
@@ -355,6 +358,8 @@ init([Index]) ->
     {ok, VId} = get_vnodeid(Index),
     DeleteMode = app_helper:get_env(riak_kv, delete_mode, 3000),
     AsyncFolding = app_helper:get_env(riak_kv, async_folds, true) == true,
+    MDCacheSize = app_helper:get_env(riak_kv, vnode_md_cache_size), 
+    MDCache = new_md_cache(VId), 
     case catch Mod:start(Index, Configuration) of
         {ok, ModState} ->
             %% Get the backend capabilities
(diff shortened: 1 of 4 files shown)


evanmcc:
see #699 for some other comments.

jtuple:
I could be wrong, but I've convinced myself that we should be caching `index_data` not `index_specs`, and changing `riak_object:diff_specs_direct/2` to take in `index_data` similar to how `diff_index_specs/2` does internally. Index specs are a transient set of operations. Given object A and object B with different index data, the specs will add missing items, delete no-longer-relevant items, etc. Granted, the base `index_specs/1` object basically returns `index_data` with `add` prepended to each element, as it's the A=undefined, B=new object case. But, not sure it's a great idea to rely upon that behavior. Esp. when all we do is ultimately pass the cached specs into `diff_specs_direct` that immediately does a list comprehension to filter out the operations.

AFIAK, making this change includes:
1. Modifying [riak_kv_vnode.erl#L1213-L1215](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1213-L1215) to use `index_data` rather than `index_specs`.
2. Modifying [riak_kv_vnode.erl#L1313](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1313) and [riak_kv_vnode.erl#L1430](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1430) to use `index_data` not `index_specs`.
3. Modifying `riak_object:diff_specs_direct`.

jtuple:
Since we insert into the metadata cache into two places, probably a good idea to move that into a function.

Something like:

``` erlang
maybe_cache_object(BKey, Obj, #state{md_cache=MDCache,
                                     md_cache_size=MDCacheSize}) ->
    case MDCache of 
        undefined ->
            ok;
        _ ->
            VClock = riak_object:vclock(Obj),
            Specs = riak_object:index_specs(Obj),
            insert_md_cache(MDCache, MDCacheSize, BKey, {VClock, Specs})
    end.
```

Which is then used at [riak_kv_vnode.erl#L1308-L1315](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1308-L1315) and [riak_kv_vnode.erl#L1425-L1434](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1425-L1434)

jtuple:
You have a bunch of defensive checks around ETS tables not existing. Is it even possible for these tables to not exist? Have you seen this happen? These are private, unnamed tables, that should exist until the vnode exits. I've never seen such tables go missing. Not sure it's possible in Erlang short of calling `ets:delete`.

If we are going to keep this defensive coding, I hear Scott is pushing us to prefer `try/catch` over `catch` for performance reasons. Not sure it matters here since the `catch` is exceptional (ie. I don't even think it's possible). But, perhaps it'll lead to better code for the case statement since it doesn't need to match on the `EXIT` case anymore.

jtuple:
I wonder about the performance impact of `trim_md_cache`. Esp. for different pathological cases. Such has having a large cache where all the recent items are huge (eg. large index data), but all the old items are tiny. Would end up recursing a bunch to evict all the old tiny items before reaching the larger recent items.

In general, there probably many ways we can improve the trimming as well as the caching approach in general. Perhaps for a future PR.

In the simplest case, I wonder if we could just add a size estimate to each cache entry, fetch multiple objects at once (eg. `ets:match` w/ limit parameter), and then delete multiple objects based on sum of size estimates. Obvs, still check actual ets size, but could get close with fewer calls.

Again, for a future PR.

Probably replace ETS approach with a NIF or something anyway. I have plenty of really fast, unpublished caching NIFs from last December sitting on my laptop.

src/riak_kv_vnode.erl
```diff
@@ -2021,6 +2098,59 @@ try_set_concurrency_limit(Lock, Limit, true) ->
...
+            case Size > MaxSize of
+                true ->
+                    trim_md_cache(Table, MaxSize);
+                false ->
+                    ok
+            end
+    end.
+        
+new_md_cache(VId) ->                
+    MDCacheName = list_to_atom(?MD_CACHE_BASE ++ integer_to_list(binary:decode_unsigned(VId))),
```
jtuple, line 2254: The ETS name doesn't need to be unique since we're using unnamed tables. I can have all the `ets:new(abc, [])` tables I want. Are we spending the time constructing the unique name here so that `ets:i()` shows the name? Makes sense. Just annoying that `ets:i().` truncates table names. And `ets:all().` just gives you tids for unnamed tables. Wonder if there's any way to access full name for these really long names.

jtuple, line 2254: Oh. Nevermind. Looks like `ets:i()` doesn't truncate table names. Cool. +1 to naming things for `ets:i().` benefit.

jtuple:
I _think_ the creation of a fake object with the cached metadata only when the new object descends from the cached object is correct. But, I wonder if @russelldb could ponder a bit since by timezone magic he has more time before we code freeze tomorrow morning.

Russell, any concerns with [these lines](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1159-L1184)?

This basically collpases down to changing things from.

``` erlang
%% Current logic on develop pseudo-code
OldObj = do_backend_get(BKey),
do_normal_put_merge_unchanged_from_develop(RObj, OldObj).
```

to

``` erlang
%% New logic with metadata cache
OldObj = case vclock:descends(riak_object:vclock(RObj), CachedClock) of
             true ->
                 FakeObj0 = riak_object:new(Bucket, Key, <<>>),
                 FakeObj = riak_object:set_vclock(FakeObj0, CacheClock),
                 FakeObj;
             _ ->
                 do_backend_get(BKey)
         end,
do_normal_put_merge_unchanged_from_develop(RObj, OldObj)
```

Oh, we also handle updating 2i indexes specially in the cached case, hence why we don't need the 2i index data added to the fake object. Although, that makes me wonder if we would have been better off just adding the cached index data to the fake object as well, and not special cashing the index update code path. Oh well.

src/riak_kv_vnode.erl
```diff
@@ -1244,9 +1299,20 @@ perform_put({true, Obj},
...
+                        modstate=ModState,
+                        md_cache=MDCache,
+                        md_cache_size=MDCacheSize}) ->
+    case MDCache of 
+        undefined ->
+            ok;
+        _ ->
+            VClock = riak_object:vclock(Obj),
+            Specs = riak_object:index_specs(Obj),
+            insert_md_cache(MDCache, MDCacheSize, BKey, {VClock, Specs})
```
jtuple: I wonder if we should be using `MaybeFake` here instead of `Obj`. I noticed this change when I performed the merge and preserve the original PR semantics, but I'm not sure if it's right. `Obj` and `MaybeFake` come from the `mutate_get` call earlier in `perform_put`. `Obj` is the mutated object. The thing we're writing to disk. Clients never see this object. Likewise, we unmutate this object on the get/put path after reading from the backend (well, there's actually a bug where we don't in all code paths, but that's separate from this PR ). Thus, high level Riak only ever sees the unmutated object. The `MaybeFake` object here is basically that object -- hence why the `MaybeFake` object is the thing we return back to the put FSM. Does it follow that it's more correct to use the `MaybeFake` (the faked real object) object rather than `Obj` (the mutated obj)? I'm pretty sure mutators must preserve the vclock, but I'm not certain they must perserve the index specs. Anyone with more mutator knowledge than I have an opinion?

jtuple:
Do we need to add caching logic to remove entries from the cache in [riak_kv_vnode:do_backend_delete](https://github.com/basho/riak_kv/blob/e51b60cf15fd5b3821c96d3383dddef86dc031d9/src/riak_kv_vnode.erl#L1086)?

src/riak_kv_vnode.erl
evanmcc: I am fairly certain that at this point all that the mutators are doing is adding some metadata related to repl.  there's no conceivable reason that they should be altering index specs.  That said, I think that since we can expect the same vclock and index data for both, that it shouldn't matter, but that using `MaybeFake` is likely more correct.

This makes me wonder about that `hashtree_update()` call as well, shouldn't it be using `MaybeFake`?

evanmcc:
updated to address review comments and fix a few potential bugs.

Have done a few memory backend tests, in which get performance is pretty bad in the low-concurrency case.  Currently the feature is disabled by default, we'll need to do some testing to make sure it's safe and effective before we flip the switch back.

I'll open up another issue for adding a `nocache` backend capability to the memory and yessir backends.

src/riak_kv_vnode.erl
```diff
@@ -1084,6 +1101,7 @@ handle_info({final_delete, BKey, RObjHash}, State = #state{mod=Mod, modstate=Mod
                    {ok, RObj} ->
                        case delete_hash(RObj) of
                            RObjHash ->
+                               maybe_cache_evict(BKey, State),
                                do_backend_delete(BKey, RObj, State);
```
jtuple, line 1105: In addition to this line, `do_backend_delete` can also be called [here](https://github.com/basho/riak_kv/blob/8b814dafe95bf63ad4bfb7662a040333fcc38ba9/src/riak_kv_vnode.erl#L1727). I think it's safer to move the `maybe_cache_evict` call into `do_backend_delete`.

jtuple:
(Looks like we accidentally merged this. Oh well. Continuing PR discussion here)

So, we really, really need to test this more before we ship 2.0. And ponder the causality implications. I hope Russell gets a chance to chime in/etc. BTW, good catch on making sure to evict the cache in `diffobj_put` -- that would have clobbered data.

Since this is disabled by default now, and the code path when disabled isn't changed at all, I'm +1 for merging after the "move cache evict to do_backend_delete" change.

We have the two above linked issues + lots of performance testing + a detailed riak_test to write that hits the specific code paths changed when using the metadata cache. That can all happen during the QA cycle given the lack of time before code freeze.

Depending on how we feel in the end, we have a few options by 2.0: leave as is (disabled), enable the code by default (yay!), or hardcode the code path to be disabled always. The choice depends on everything we do between now and then.

evanmcc:
pushed 9cda989 to develop to address the `backend_delete` comment.

src/riak_kv_vnode.erl
```diff
@@ -1222,12 +1241,33 @@ prepare_put(#state{vnodeid=VId,
                              prunetime=PruneTime,
                              crdt_op = CRDTOp},
             IndexBackend) ->
+    {CacheClock, CacheData} = maybe_check_md_cache(MDCache, BKey),
+
+    RequiresGet = 
+        case CacheClock of
+            undefined ->
+                true;
+            Clock ->
```
russelldb, line 1250: @jtuple asked me to comment. I know he wasn't asking me to comment on style but why not just `not vclock:descends(RObj, Clock)` here, rather than another case statement?

russelldb:
Assuming that the cache never lags the metadata on disk the fake object with dominated vclock approach looks fine to me.

PR #785, opened 2013-12-24
Add timer:sleep()-based throttle to riak_kv_exchange_fsm:read_repair_keydiff()

slfritchie:
Develop branch version of https://github.com/basho/riak_kv/pull/754

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -122,6 +122,48 @@
   {datatype, integer}
 ]}.
 
+%% @doc If on, this item will disable the AAE distributed throttle mechanism
+%% used by riak_kv_exchange_fsm.erl.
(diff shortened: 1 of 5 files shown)


slfritchie:
Derp, I didn't see Sean's comments from two days ago.  GitHub's notifications don't work like I think they do, sorry.

slfritchie:
Per @jtuple, I'm merging now.  I'll address the Cuttlefish naming thing later, sorry, @seancribbs!  There's a matter of having 2 throttles now, one for tree building and this one for exchange, and perhaps they ought to be disambiguated code-wise for future maintenance?  But if I go changing function names in riak_kv_entropy_manager.erl, then there's a big mess o' renaming to do ... am willing to debate the merits of doing that, when we return from end-of-year breaks.

seancribbs:
@slfritchie It's fine, we have plans to do more renaming anyway. Just wanted to save us a little work down the road...

PR #787, opened 2013-12-24
Fix test-compile target, skipping forced recompile if EQC is not present.

seancribbs:
@rzezeski


rzezeski:
Worked for me. +1 to merge.

PR #789, opened 2013-12-26
Fix logging call

rzezeski:


slfritchie:
+1.  I'll do the same for the 1.4 branch also, thanks.

PR #792, opened 2013-12-30
Unfold the put options at the start

russelldb:
Unfold put options at init so that `asis` is respected.


Vagabond:
This improves the behaviour of the verify_dvv_repl test, although it is still failing about 50% of the time for me. That, however, is better than the 100% fail rate without the test.

I think this is a good isolated change. +1

russelldb:
@Vagabond can you gist me your failures? I ran 100+ today no fails.

PR #801, opened 2014-01-14
Able to better handle sibling objects on mutate_get

lordnull:
Also strips the mutators_applied metadata key on mutate_get. It's not
used outside of mutators, and thus is not meant to be exposed.


lordnull:
With mutator system removed, this is no longer relevant.

PR #808, opened 2014-01-17
Use tools.mk in Makefile

Vagabond:
We intentionally let the Makefile override the 'test' target from
tools.mk, because it does weird stuff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,11 @@ clean:
 distclean: clean
 	./rebar delete-deps
 
+DIALYZER_APPS = kernel stdlib sasl erts ssl tools os_mon runtime_tools crypto inets \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -26,35 +31,3 @@ endif
 
 test: all test-compile
```
seancribbs: Is it possible to remove the warning by simply adding test-compile as a dependency but remove the command?

seancribbs: Yep, this seems to work:

``` make
test: test-compile
```

Vagabond: I don't understand what that means. Is that adding a dep to an existing target?

seancribbs: Yes.

seancribbs:
Other than fixing the `test` target, :+1:

seancribbs:
:+1: again

PR #813, opened 2014-01-22
Handle all return values of merge.

cmeiklejohn:
In the event that two objects, which are strict ancestors of themselves
are compared, handle that as if there are no strict ancestors available.
This prevents a crash of the kv_vnode during a put request.

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -725,12 +725,13 @@ syntactic_merge(CurrentObject, NewObject) ->
                   end,
 
     case ancestors([UpdatedCurr, UpdatedNew]) of
-        [] -> merge(UpdatedCurr, UpdatedNew);
         [Ancestor] ->
             case equal(Ancestor, UpdatedCurr) of
                 true  -> UpdatedNew;
                 false -> UpdatedCurr
-            end
+            end;
+        _ ->
+            merge(UpdatedCurr, UpdatedNew)
     end.
 
 %% @doc Get an approximation of object size by adding together the bucket, key,


src/riak_object.erl
```diff
@@ -725,12 +725,13 @@ syntactic_merge(CurrentObject, NewObject) ->
...
     case ancestors([UpdatedCurr, UpdatedNew]) of
-        [] -> merge(UpdatedCurr, UpdatedNew);
         [Ancestor] ->
             case equal(Ancestor, UpdatedCurr) of
                 true  -> UpdatedNew;
                 false -> UpdatedCurr
-            end
+            end;
+        _ ->
+            merge(UpdatedCurr, UpdatedNew)
```
russelldb, line 734: I think DVV merge will break. This can only happen when two vclocks descend each other: that means they are causally equal but with different timestamp(s) against some counter. The DVV merge will drop any value that is both dominated and not in both objects. If A and B have a dot {id, {1, TSA}} and {id, {1, TSB}} and a vclock [{id, {1, TSA|TSB}}] both dots will be dropped as both are dominated, and neither is in both objects.

russelldb, line 734: Actually, on reflection I think this change is bad.

For this clause to fire A would have to dominate B AND B dominate A. That is impossible. If it happens then crashing here is the best thing, as any further behaviour is completely unknown. I say crash as near the site of "impossible" bugs as possible.

reiddraper:
Is this a 2.0 blocker? It needs a milestone assigned, but i'm not familiar enough with the domain to know what it should be. @russelldb, @cmeiklejohn?

cmeiklejohn:
I don't think this is needed anymore, right @russelldb?

russelldb:
Not needed anymore.

reiddraper:
Thanks @russelldb.

PR #814, opened 2014-01-23
Fix DVV merge to handle "skewed dots"

russelldb:
A "skewed dot" is any pair of dots that have the same actor and
counter, but a different timestamp (technically, they can have the
same timestamp too) and a different riak object contents.

To ensure that we don't drop both dots we use only the {actor, count}
pair as a key for drop candidates, and keep _both_ dotted values if
the {actor, count} are the same. In the normal case the value will
be the same, and collection the dotted values in set de-dupes them.

NOTE: causality in Riak can be very odd, where vclocks with the
same actors, count, tinestamp entries can denote different
causal histories.

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -330,23 +330,35 @@ fold_contents({r_content, Dict, Value}=C0, MergeAcc, Clock) ->
     #merge_acc{drop=Drop, keep=Keep, crdt=CRDT, error=Error} = MergeAcc,
     case get_dot(Dict) of
         {ok, Dot} ->
-            case {vclock:descends_dot(Clock, Dot), orddict:is_key(Dot, Drop)} of
+            case {vclock:descends_dot(Clock, Dot), is_drop_candidate(Dot, Drop)} of
                 {true, true} ->
                     %% When the exact same dot is present in both
                     %% objects siblings, we keep that value. Without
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
@seancribbs @russelldb failures on PR branch, which do not occur on the develop branch:

```
riak_object: weird_clocks_weird_dots_test...*failed*
in function riak_object:syntactic_merge/2 (src/riak_object.erl, line 758)
in call from riak_object:weird_clocks_weird_dots_test/0 (src/riak_object.erl, line 1322)
**error:{case_clause,[{r_object,<<"b">>,<<"k">>,
                        [{r_content,{dict,1,16,16,8,80,48,{[],...},{...}},
                                    <<"a">>}],
                        [{a,{1,1000}}],
                        {dict,1,16,16,8,80,48,{[],[],...},{{...}}},
                        undefined},
              {r_object,<<"b">>,<<"k">>,
                        [{r_content,{dict,1,16,16,8,80,48,{...},...},<<"a">>}],
                        [{a,{1,100}}],
                        {dict,1,16,16,8,80,48,{[],...},{...}},
                        undefined}]}
```

cmeiklejohn:
Missed the dependency on basho/riak_core#510.

cmeiklejohn:
:+1: to merge, looks good.

PR #815, opened 2014-01-23
Setting "object.format" should not be in the default config file. [JIRA: RIAK-3372]

seancribbs:
This setting is only for use in upgrades/downgrades from versions 1.3 and earlier.
- [x] CSE approval
- [x] PM approval

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -297,7 +297,8 @@ end
 %% * 1: New format for more compact storage of small values.
 {mapping, "object.format", "riak_kv.object_format", [
   {default, 1},
-  {datatype, [{integer, 1}, {integer, 0}]}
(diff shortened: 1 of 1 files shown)


seancribbs:
@gburd @angrycub

seancribbs:
This should be delayed until 2.1. Users upgrading from 1.3 might not be able to downgrade easily without this setting being more explicitly set to '0', if the upgrade to 2.0 does not go well.

angrycub:
Agreed, @seancribbs.  I agree with the PR, just not the timing.

gburd:
seems ok to me, +1

seancribbs:
Milestone is set on this then, we'll revisit after 2.0 release.

angrycub:
Create jira issue

yorkshire-steve:

PR #817, opened 2014-01-25
Bring 1.4.7 2i AAE fixes to 2.0

engelsanchez:


engelsanchez:
This is a port of https://github.com/basho/riak_kv/pull/780.

The main issue is that the way that AAE uses to regenerate the list of trees for a partition doesn't take into account what was passed in first.  Because of this, the changes in 2i AAE to add a special tree for 2i data are not enough to make that tree survive a rebuild round. As soon as the trees expire and rebuild, you lose the ability to do 2i repairs.

jonmeredith:
@engelsanchez is this up to date for 1.4.8?

engelsanchez:
No, this is just 1.4.7. I have an item in the Kill Bill for 1.4.8 that's not yet in a PR.

src/riak_kv_index_hashtree.erl
```diff
@@ -689,9 +690,15 @@ do_delete_expanded([{Id, Key}|Rest], State=#state{trees=Trees}) ->
     end,
     do_delete_expanded(Rest, State2).
 
+-spec responsible_preflists(#state{}) -> [index_n()].
+responsible_preflists(#state{index=Partition, use_2i=Use2i}) ->
+    RP = riak_kv_util:responsible_preflists(Partition) ++
```
kellymclaughlin, line 695: Why have the assignment here instead of just returning the result of the expression directly?

kellymclaughlin:
- [x] Verified these changes represent the same changes applied to the 1.4 branch
- [x] Verified that verify_2i_aae test passes with these changes and fails without them
- [x] Verified changes appear to address the intended issue from code inspection

kellymclaughlin:
One minor comment here and one on the r_t PR, but they do not represent functional issues. Given this has already been reviewed once for 1.4.7, it appears to be doing what it is intended to do, and the tests are passing I am :+1: for merging when you are ready.

src/riak_kv_index_hashtree.erl
engelsanchez, line 695: I didn't use to do that. I've seen the style used in a lot of Basho code lately, and I'm a desperate wannabe :-|

seancribbs, line 695: Yeah, I haven't seen that until recently; I don't see what it clarifies.

PR #818, opened 2014-01-28
Add missing riak_object:is_robject function

Vagabond:
Previously backported features did not backport is_robject, which makes AAE object hashing fall back to using a default hash.

cc @jtuple @cmeiklejohn


cmeiklejohn:
:+1:

PR #819, opened 2014-01-28
Log an error when catching AAE obj hash exception

engelsanchez:


src/riak_kv_index_hashtree.erl
```diff
@@ -381,7 +381,9 @@ hash_object({Bucket, Key}, RObj0) ->
         end,
         Hash = riak_object:hash(RObj),
         term_to_binary(Hash)
-    catch _:_ ->
+    catch
```
reiddraper: is it also worth narrowing the scope of the try..catch block? Which specific line of code are we worried is going to throw an exception potentially?

engelsanchez: yes, it would be nice to do that. I believe this started with the binary corruption problems and handoff. So probably what we want is simply detect an exception from witin from_binary.  Let me check

engelsanchez:
Well, I have narrowed it down to only the from_binary clause. That also makes it very hard to test, unfortunately.

reiddraper:
Moving the scope of the try..catch makes it harder to test?

engelsanchez:
I should say, it makes this change harder to verify.  If you mock the from_binary call, it fails elsewhere. I had to add a throw in the code to verify that the bits were working, then wait for an AAE exchange, which is the only path I've found that actually sends binaries to this function during normal riak operations. Other paths send riak objects. I think I will add a small unit test to verify that bad binaries return the default hash without crashing at least.

reiddraper:
Sounds good to me.

jtuple:
Are we planning to merge this into another 1.4 release? Do we want to forward port this to 2.0 before beta?

reiddraper:
Is this a 2.0 blocking issue? What should the milestone be @engelsanchez?

engelsanchez:
This doesn't have any 2.0 tags Reid. It is also against the 1.4 branch.

yorkshire-steve:

PR #820, opened 2014-01-28
Remved mutator system.

lordnull:
Removed the mutator code, tests, and calls to mutators in vnode, and the
capability.


cmeiklejohn:
:+1:, I've run the eunit which looks good, as well as looked at the diffs.  Let me know if I should be looking at anything else.

lordnull:
Thanks! I'm running a couple of r_t's to ensure nothing is horribly broken.

lordnull:
Running r_t's shows no new errors caused by it's removal.

PR #822, opened 2014-01-28
Ajs overload plumbing

andrewjstone:
Process overload for coverage queries and other queries that were previously not handled. Return `{error, mailbox_overload}` when the riak_core_vnode_proxy tells us we are overloaded.

Part of a fix for #753 
Should be merged together with https://github.com/basho/riak_core/pull/513

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -461,9 +461,9 @@ list_keys(Bucket, Timeout, {?MODULE, [_Node, _ClientId]}=THIS) ->
 %%      Key lists are updated asynchronously, so this may be slightly
 %%      out of date if called immediately after a put or delete.
 list_keys(Bucket, Filter, Timeout0, {?MODULE, [Node, _ClientId]}) ->
-    Timeout = 
+    Timeout =
         case Timeout0 of
-            T when is_integer(T) -> T; 
+            T when is_integer(T) -> T;
             _ -> ?DEFAULT_TIMEOUT*8
         end,
     Me = self(),
@@ -588,9 +588,9 @@ list_buckets(Filter, Timeout, {?MODULE, [_Node, _ClientId]}=THIS) ->
 list_buckets(Filter, Timeout, Type, {?MODULE, [Node, _ClientId]}) ->
     Me = self(),
     ReqId = mk_reqid(),
-    {ok, _Pid} = riak_kv_buckets_fsm_sup:start_buckets_fsm(Node, 
-                                                           [{raw, ReqId, Me}, 
-                                                            [Filter, Timeout, 
+    {ok, _Pid} = riak_kv_buckets_fsm_sup:start_buckets_fsm(Node,
+                                                           [{raw, ReqId, Me},
+                                                            [Filter, Timeout,
                                                              false, Type]]),
     wait_for_listbuckets(ReqId).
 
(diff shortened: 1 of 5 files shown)


russelldb:
I feel you on the whitespace issues, really I do. We had this discussion a while back and I was told not to mix whitespace fixes with functional changes. And now I can see why: it is hard to tell what has changed for what reason.

andrewjstone:
@russelldb I definitely was not trying to make a whitespace point. I just have 'delete-trailing-whitespace' on and that's what happened when I saved the files. I've committed numerous files just like this and this is the first I'm hearing of the recommendation. I can clean this up if necessary, but it still seems like an improvement to me over no change to the whitespace.

russelldb:
@andrewjstone oops. Ok. So sometime ago we had this discussion on hipchat. I only brought it up 'cos I used to have 'delete-trailing-whitespace' as on and that was happened when I saved files. I was all like "it needs doing and it is better than before" but there was wailing, and gnashing of teeth. And Jon was like "turn it off". So that is why I brought it up. But if we're revisiting it then let's do this. And get rid of trailing whitespace and dodgy formatting. Since precedent says it is not a problem, I'm going to turn mine back on, and point to this discussion (I hope you'll back me up.)

evanmcc:
You can change the diff settings to make the review easier (I think that
adding ?w=0 to the url will work), but that said, if we're going to do
this, we all kind of need to do this, and to make an effort to make the
whole codebase whitespace clean to make the default settings minimally
irritating.

andrewjstone:
@russelldb I've got your back. But I also anticipate more gnashing later. Maybe we can fight it out as allies and convince the angry invading hoards of reviewers.

andrewjstone:
@evanmcc Yep that works. I had no idea. Thank you.

jrwest:
fwiw, `git add -i` is also useful when you have `delete-trailing-whitespace` on but don't want to commit the changes. imo, its useful to have proper formatting but we should do it all in on PR in some automated way across all files once. It mucks w/ the commit history/blame/etc otherwise.

evanmcc:
+1 to having this be an automated process, followed by an ongoing
commitment to keeping the codebase clean.

andrewjstone:
Ok. So should I change _this_ PR, or just not do this next time?

russelldb:
My opinion is first offence, let off with a warning (as in don't change this PR, but don't do it next time)

andrewjstone:
Me too. we have better things to do.

jonmeredith:
Since git-blame and github both have options for ignoring white space changes, I care less than i used too.

kuenishi:
I opened a [suggestion and place to discuss code formatting at Riak CS repos](https://github.com/basho/riak_cs/pull/783). If you like, come and discuss please.

kellymclaughlin:
Sorry to disrupt the flow of whitespace bikeshedding, but...

I modified the overload riak_test to use a PB connection instead of the local client for the key and bucket listing requests and noticed that the change to `riak_kv_keys_fsm` and `riak_kv_buckets_fsm` to return an error tuple instead of just the error reason yields results like the following returned from the PB client: 

```
{error,<<"{error,mailbox_overload}">>}}
```

A small change to the `process_stream` function clauses for error handling in `riak_kv_pb_bucket` should fix that up.

Maybe we could also change the test to use the PB client for the requests to exercise that code path and verify the change.

kellymclaughlin:
No regressions in eunit or eqc tests and riak_test testing looks good. Nice job. :+1:

PR #823, opened 2014-01-30
Store bucket type hash for new riak_objects in typed buckets

kellymclaughlin:
Calculate and store a bucket type hash for all new riak_objects
created for typed buckets. This primary consumer of this information
is Riak's replication system, but it may have general use and is
benign when not used.


kellymclaughlin:
The module tests in `riak_object` and `riak_object_eqc` do not test typed buckets. I will be pushing a commit to add that testing, but opened the PR so that the review can commence while I do that work.

kellymclaughlin:
I updated and refactored some of the tests in `riak_object` so that there is now some testing with typed buckets there. Also added some generators for typed buckets for used in `riak_object_eqc`.

bowrocker:
Replication riak_tests are passing for me.

src/riak_object.erl
```diff
@@ -498,15 +501,58 @@ apply_updates(Object=#r_object{}) ->
...
+-spec bucket_type_hashes(riak_object()) -> undefined | [integer()].
+bucket_type_hashes(RObj) ->
+    get_metadatas_entry(bucket_type_hash, RObj).
+
+%% @doc Return a list of values for a particular key contained in
+%% the metadata dictionaries associated with each value in a
+%% riak_object.
+-spec get_metadatas_entry(term(), riak_object()) -> [term()].
+get_metadatas_entry(Key, #r_object{contents=Contents}) ->
+    ordsets:to_list(
```
andrewjstone, line 540: No need to do an explicit conversion here, as ordsets are always implemented as lists. An argument can be made that it makes the code clearer, but it still adds unnecessary runtime overhead.

evanmcc, line 540: it adds one reduction, it looks like:

```
-spec to_list(Ordset) -> List when
      Ordset :: ordset(T),
      List :: [T].

to_list(S) -> S.
```

kellymclaughlin, line 540: Sure I can change that. Hooray for reliance on implementation details! #lolerlang

andrewjstone, line 540: What's your preference @evanmcc? I've seen code written both ways. Not sure why I picked this PR to bring it up. You're welcome @kellymclaughlin

evanmcc, line 540: Leave it.  Don't break encapsulation if you don't have to.

andrewjstone, line 540: I'm fine with that here. Especially since it means not changing this PR. Bikeshedding over drinks in denver sounds better.

andrewjstone:
- [x] eunit + eqc
- [x] xref
- [x] dialyzer
- [x] riak_test

Riak test taking forever. Some pass, some fail. PR seems good.

src/riak_object.erl
```diff
@@ -498,15 +501,58 @@ apply_updates(Object=#r_object{}) ->
...
+      lists:foldl(get_metadatas_entry_fold_fun(Key), ordsets:new(), Contents)).
+
+-type metadatas_entry_fold_fun() :: fun((term(), ordsets:ordset()) -> ordsets:ordset()).
+-spec get_metadatas_entry_fold_fun(term()) -> metadatas_entry_fold_fun().
+get_metadatas_entry_fold_fun(Key) ->
+    fun(Content, Acc) ->
+            case dict:find(Key, Content#r_content.metadata) of
```
andrewjstone, line 547: The type of `#r_content.metadata` is `dict() | list()`. Are we sure this function only gets called with Dict type data?

kellymclaughlin, line 547: I think that spec is wrong, but I do see one place at [L135](https://github.com/basho/riak_kv/blob/feature/riak-object-btype-hash-metadata/src/riak_object.erl#L135) where you could get something other than a `dict` stored for that field. If someone used `new/4` to specify metadata and the function clauses at [L118](https://github.com/basho/riak_kv/blob/feature/riak-object-btype-hash-metadata/src/riak_object.erl#L118) or [L120](https://github.com/basho/riak_kv/blob/feature/riak-object-btype-hash-metadata/src/riak_object.erl#L120) when creating the object they could pass any sort of non-list structure in for `MD`. The `is_list` guards in previous function clauses would not let a list make it that far though so that gives more credence to the idea that the spec is wrong.

andrewjstone, line 547: It does seem like the spec is wrong and those guards help prevent badness, but the opportunity does still exist. Maybe a separate Issue should be opened. No change required for this PR IMO.

andrewjstone:
:+1:

PR #824, opened 2014-02-01
display default type properties in console status output

jrwest:
prior to this commit default properties were not displayed because
they cannot be read directly via the claimant.

the `bucket_type_raw_props` function should actually, in some form, be
available in `riak_core_bucket_type`, however, it is left here in riak_kv
to minimize changes.

addresses https://github.com/basho/riak_core/issues/517


jrwest:
in addtion to the above, I have also added one additional change: including the default type in `riak-admin bucket-type list`

coderoshi:
+1

PR #825, opened 2014-02-01
revert allow_mult default only for untyped (default type) buckets

jrwest:
as discussed.

confbal -- I attempted to update the description for `allow_mult` but am realizing now that this is probably insufficient given all default bucket properties in `riak.conf` only affect untyped buckets and we have yet to decide if "untyped buckets" is in fact what we want to call them.

Before Merge:
- [x] verify test changes are correct and pass
- [ ] CSE +1
- [x] PM +1
- [ ] Have a plan re: testing (specifically `riak_test`) as discussed -- ideally testing both setting values, or at a minimum preserve `allow_mult=true` in tests that have been updated to account for it

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -488,11 +488,11 @@ end
  hidden
 ]}.
 
-%% @doc Whether or not siblings are allowed.  Note: See Vector Clocks
-%% for a discussion of sibling resolution.
+%% @doc Whether or not siblings are allowed, by default, for untyped buckets.
+%% Note: See Vector Clocks for a discussion of sibling resolution.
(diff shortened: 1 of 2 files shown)


jrwest:
for the reviewer: typed bucket defaults can (unfortunately?) be found [here](https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_type.erl#L119)

seancribbs:
:+1: on schema tests.

gburd:
I'm not a fan of reverting this to default to "false" :poop:
That said, +1

kellymclaughlin:
:+1:

jonmeredith:
+1 merge - accepting giddyup breakage.  Will need to fix tests to go from beta to rc.

PR #827, opened 2014-02-05
warn about inability to downgrade after activating bucket type

jrwest:
As discussed.

This implementation is less ideal than potentially writing a flag, serialized through the claimant like other bucket type operations, and checking that. A potential degenerate case is a user creating thousands of bucket types without activating any of them. However, after creating 50k+ bucket types without activation the time to create or activate a type in scientific"" tests on my machine didn't take noticeably longer than on a node with only a few types.

Example:

User first has no types:

```
$ dev/dev1/bin/riak-admin bucket-type list
default (active)
```

User first creates a type:

```
$ dev/dev1/bin/riak-admin bucket-type create type1 '{"props": {}}'
type1 created

WARNING: After activating type1, nodes in this cluster
can no longer be downgraded to a version of Riak prior to 2.0
```

User creates another type, but no types have been activated, so warning still emitted. Same
when `create` is called again to change the properties on the type.

```
$ dev/dev1/bin/riak-admin bucket-type create type2 '{"props": {}}'
type2 created

WARNING: After activating type2, nodes in this cluster
can no longer be downgraded to a version of Riak prior to 2.0
$ dev/dev1/bin/riak-admin bucket-type create type2 '{"props": {"allow_mult": false}}'
type2 created

WARNING: After activating type2, nodes in this cluster
can no longer be downgraded to a version of Riak prior to 2.0
```

User activates a type, a final warning is emitted:

```
$ dev/dev1/bin/riak-admin bucket-type activate type1
type1 has been activated

WARNING: Nodes in this cluster can no longer be
downgraded to a version of Riak prior to 2.0
```

Future operations have no warning:

```
$ dev/dev1/bin/riak-admin bucket-type create type2 '{"props": {"allow_mult": true}}'
type2 created
$ dev/dev1/bin/riak-admin bucket-type activate type2
type2 has been activated
$ dev/dev1/bin/riak-admin bucket-type create type3 '{"props": {"datatype": "set"}}'
type3 created
$ dev/dev1/bin/riak-admin bucket-type activate type3
type3 has been activated
$ dev/dev1/bin/riak-admin bucket-type list
default (active)
type3 (active)
type1 (active)
type2 (active)
```

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -471,13 +471,22 @@ bucket_type_print_props(Props) ->
 
 bucket_type_activate([TypeStr]) ->
     Type = list_to_binary(TypeStr),
-    bucket_type_print_activate_result(Type, riak_core_bucket_type:activate(Type)).
+    IsFirst = bucket_type_is_first(),
+    bucket_type_print_activate_result(Type, riak_core_bucket_type:activate(Type), IsFirst).
 
-bucket_type_print_activate_result(Type, ok) ->
-    io:format("~s has been activated~n", [Type]);
-bucket_type_print_activate_result(Type, {error, undefined}) ->
+bucket_type_print_activate_result(Type, ok, IsFirst) ->
+    io:format("~s has been activated~n", [Type]),
+    case IsFirst of
+        true ->
+            io:format("~n"),
+            io:format("WARNING: Nodes in this cluster can no longer be~n"
+                      "downgraded to a version of Riak prior to 2.0~n");
+        false ->
+            ok
+    end;
+bucket_type_print_activate_result(Type, {error, undefined}, _IsFirst) ->
     bucket_type_print_status(Type, undefined);
-bucket_type_print_activate_result(Type, {error, not_ready}) ->
+bucket_type_print_activate_result(Type, {error, not_ready}, _IsFirst) ->
     bucket_type_print_status(Type, created).
 
 bucket_type_create([TypeStr, PropsStr]) ->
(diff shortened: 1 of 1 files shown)


jrwest:
tried 100k types as well w/o seeing too much issue

src/riak_kv_console.erl
```diff
@@ -471,13 +471,22 @@ bucket_type_print_props(Props) ->
...
 bucket_type_activate([TypeStr]) ->
     Type = list_to_binary(TypeStr),
-    bucket_type_print_activate_result(Type, riak_core_bucket_type:activate(Type)).
+    IsFirst = bucket_type_is_first(),
+    bucket_type_print_activate_result(Type, riak_core_bucket_type:activate(Type), IsFirst).
 
-bucket_type_print_activate_result(Type, ok) ->
-    io:format("~s has been activated~n", [Type]);
-bucket_type_print_activate_result(Type, {error, undefined}) ->
+bucket_type_print_activate_result(Type, ok, IsFirst) ->
```
andrewjstone, line 477: Instead of passing in `IsFirst` as a new argument, the entire case clause below should be abstracted into a `maybe_print_warning/1` function that takes a `Type` parameter. This same function can be used in `bucket_type_print_create_result/2` below. The `/3` versions of these functions will no longer be needed.

jrwest, line 477: The reason `IsFirst` is passed through here is because `bucket_type_is_first()` must be called before the operation has completed otherwise the warning is never printed (the check finds the just activated type, which is the condition for skipping the warning). But perhaps there is still a way to clean that up...

andrewjstone, line 477: That makes sense. In that case don't worry about it. I was thinking about maybe just abstracting the message itself, but that's no big deal.

andrewjstone:
- [x] code inspection
- [x] dialyzer
- [x] manual testing

:+1:

gburd:
+1

Issue #740, opened 2013-11-28
dropped_vnode_requests_total missing in riak-admin status output


runesl:
dropped_vnode_requests_total is in /stats output, but not in the output from riak-admin status when calling from command line.

Observed on 1.4.2 and 2.0.0pre5

In addition, it would be nice if this stat also had a 1 minute-version "dropped_vnode_requests"  like most of the other total counters do.

slfritchie:
NOTE: There are a number of other stats that are missing from "riak-admin status" output, alas, that do indeed appear in the HTTP server's /stats JSON blob, e.g. riak_kv_vnodeq_max.

russelldb:
@slfritchie list 'em and I'll fix 'em :D

slfritchie:
Here's what I've found with brute-force sed & diff, run on today's `develop` branch of Riak.

```
diff --git a/tmp/stats.riak-admin-status b/tmp/stats.slash-stats
index 3a508ed..4ba85c7 100644
--- a/tmp/stats.riak-admin-status
+++ b/tmp/stats.GET-slash-stats
@@ -1,10 +1,14 @@
-1-minute stats for 'riak@127.0.0.1' says:
+GET /stats says:
 asn1_version
 basho_stats_version
 bitcask_version
 cluster_info_version
 compiler_version
 connected_nodes
+converge_delay_last
+converge_delay_max
+converge_delay_mean
+converge_delay_min
 coord_redirs_total
 counter_actor_counts_100
 counter_actor_counts_95
@@ -17,12 +21,16 @@ cpu_avg5
 cpu_nprocs
 crypto_version
 disk
+dropped_vnode_requests_total
 eleveldb_version
 erlang_js_version
 erlydtl_version
 executing_mappers
 goldrush_version
+gossip_received
+handoff_timeouts
 ibrowse_version
+ignored_gossip_total
 index_fsm_active
 index_fsm_create
 index_fsm_create_error
@@ -213,22 +221,42 @@ read_repairs_map_total
 read_repairs_set
 read_repairs_set_total
 read_repairs_total
+rebalance_delay_last
+rebalance_delay_max
+rebalance_delay_mean
+rebalance_delay_min
+rejected_handoffs
 riak_api_version
 riak_auth_mods_version
 riak_control_version
+riak_core_stat_ts
 riak_core_version
 riak_dt_version
 riak_kv_stat_ts
 riak_kv_version
+riak_kv_vnodeq_max
+riak_kv_vnodeq_mean
[...]

PR #828, opened 2014-02-05
Add riak_core stats to riak-admin status

russelldb:
See https://github.com/basho/riak_kv/issues/740

### DIFF
--- a/src/riak_kv_status.erl
+++ b/src/riak_kv_status.erl
@@ -35,7 +35,7 @@
 
 -spec(statistics() -> [any()]).
 statistics() ->
-    riak_kv_stat:get_stats().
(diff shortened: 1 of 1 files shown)


slfritchie:
That's all?  Wow, yay!  +1

kuenishi:
This is not directly related, but I found `riak_client:get_stats/1` only using `riak_kv_stats:get_stats`, which is called from `riak_snmp` . Though I don't know whether it's used or not, +1 from me, too.

russelldb:
@kuenishi I think snmp/jmx are different issues. Hoping that the exometer work ultimately solves the brittleness around stat export/consumption.

PR #830, opened 2014-02-06
Ajs iterator refresh

andrewjstone:
Support iterator_refresh option for eleveldb backend

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -59,7 +59,8 @@
 -endif.
 
 -define(API_VERSION, 1).
--define(CAPABILITIES, [async_fold, indexes, index_reformat, size]).
+-define(CAPABILITIES, [async_fold, indexes, index_reformat, size,
+        iterator_refresh]).
 -define(FIXED_INDEXES_KEY, fixed_indexes).
 
 -record(state, {ref :: reference(),
@@ -466,9 +467,15 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{fold_opts=FoldOpts,
            true            -> undefined
         end,
 
+    IteratorRefresh =
+        case lists:keyfind(iterator_refresh, 1, Opts) of
+            false -> [];
+            Tuple -> [Tuple]
+        end,
+
     %% Set up the fold...
     FirstKey = to_first_key(Limiter),
-    FoldOpts1 = [{first_key, FirstKey} | FoldOpts],
+    FoldOpts1 = IteratorRefresh ++ [{first_key, FirstKey} | FoldOpts],
     FoldFun = fold_objects_fun(FoldObjectsFun, Limiter),
 
     ObjectFolder =
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -41,6 +41,7 @@
          fold_keys/4,
          fold_objects/4,
          fold_indexes/4,
+         handoff_started/2,
          is_empty/1,
          status/1,
          callback/3]).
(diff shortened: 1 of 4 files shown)


andrewjstone:
Should be merged with these other PRs:
- https://github.com/basho/riak_core/pull/520
- https://github.com/basho/riak_repl/pull/521

jrwest:
discussed in hipchat: the `riak_kv_multi_backend` should probably have the option wired up for level backends

reiddraper:
> discussed in hipchat: the riak_kv_multi_backend should probably have the option wired up for level backends

Anything the Riak CS multibackend needs to have done then?

andrewjstone:
TBH I'm actually a little confused about what needs to be done for
multi-backends considering only leveldb actually handles the option. It
looks like capabilities are inherited from the sub-backends so that likely
isn't it. Maybe a mumble to cover it.

jrwest:
@reiddraper i think the cs backend would need updating as well

jrwest:
mumbled w/ @andrewjstone about this a bit. Using https://github.com/basho/riak_kv/blob/develop/src/riak_kv_multi_backend.erl#L55 it should be possible for the multi-backend to just advertise the capability and pass it blindly to the sub-backends -- assuming the other sub-backends (memory and bitcask) just ignore the option while level uses it. Otherwise, the code will have to be further modified to check the capability of the sub-backend _again_ when determining whether or not to pass the option.

jrwest:
as discussed on hipchat, the commit I pushed addresses the issue that riak_core shouldn't be passing riak_kv backend specific options.

jrwest:
some comments regarding the implications of this change:

handoff:
- point-in-time-snapshots are not something assumed by handoff -- see memory backend
- [`riak_kv_eleveldb_backend:data_size/1`](https://github.com/basho/riak_kv/blob/ajs-iterator-refresh/src/riak_kv_eleveldb_backend.erl#L547) may no longer be a very accurate estimate given it is only called when the handoff is started. This was fine when leveldb iterators were not refreshed but with this change its possible that more or less data than the initial estimate will be iterated over. We could modify `riak_kv_eleveldb_backend:data_size/1` to work like [`riak_kv_memory_backend:data_size/1`](https://github.com/basho/riak_kv/blob/develop/src/riak_kv_memory_backend.erl#L396-L404), which allows the estimate to be periodically updated when the backend does not provide snapshot iteration. The problem with this is how we estimate the data size for leveldb is not tied in any way to the refreshed iterators and the time the estimate is queried may not be in sync with the refresh leading to a pretty inaccurate estimate.
- It is already known that handoff may cause false siblings (link needed), and its suspected that this change will increase that possibility
- Following with the above, the chance of handoff failing (in delete heavy workloads) slightly increases as well because iterator refresh may fail. We need to verify that this failure does not crash the vnode. handoff can tolerate this failure because it will be rescheduled, but this further increases the chance of false siblings

AAE: @jtuple thinks it should be fine

rzezeski:
> Following with the above, the chance of handoff failing (in delete heavy workloads) slightly increases as well because iterator refresh may fail. We need to verify that this failure does not crash the vnode. handoff can tolerate this failure because it will be rescheduled, but this further increases the chance of false siblings

This should be taken very seriously as restarting handoff for large partitions is very costly and can aggravate users. If the probability of iterator refresh failing is high enough then bigger partitions fail to completely handoff indefinitely wasting precious disk and network I/O and space. I was reminded of just how painful this "start from the beginning" behavior is recently when talking with Kyle about his handoff woes. If the probability of iterator refresh failure is low and we truly are not worried about it then so be it but this could prove very painful otherwise.

jrwest:
As I understand it, the probability is assumed to be very low, and can be made lower with additional changes planned for a future release. I'd like to get a better idea of how likely it is before my +1. I have to think about it a bit more, but I can imagine some degenerate case like users who load and then delete a bunch of data repeatedly; it is not unheard of, albeit not the best model.

jrwest:
I had a chance to speak with @matthewvon and apparently this is not actually an issue. The `Seek()` API will not fail as originally expected.

rzezeski:
> I had a chance to speak with @matthewvon and apparently this is not actually an issue. The Seek() API will not fail as originally expected.

Great.

jrwest:
I'm +1 on this. The only outstanding "known issue" is the `riak-admin transfers` data size estimation is going to be off for `eleveldb` a bit more than we'd like. Against the benefits of this change, I say we merge and either open an issue or make a note in the release notes (or both).

andrewjstone:
#852 merged instead

PR #831, opened 2014-02-06
A fix and optimization for the HTTP interface to riak_kv

nathanaschbacher:
The fix is just dropping the ability to have the user set "content-encoding" and subsequently stop storing it on the riak_object.  It seems to have possibly be previously conflated as another sort of "content-type" metadata, but the HTTP RFC disagrees with us.

The optimization is to avoid the current code path on PUT that does a full client replica GET request at the coordinator before performing the actual PUT operation.  Testing done by @deanproctor showed that on a 100% PUT load we could cut CPU, network, and disk utilization almost in half.

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -545,26 +545,45 @@ content_types_accepted(RD, Ctx) ->
 %%      Documents exists if a read request to Riak returns {ok, riak_object()},
 %%      and either no vtag query parameter was specified, or the value of the
 %%      vtag param matches the vtag of some value of the Riak object.
+resource_exists(RD, #ctx{method=Method}=Ctx) when Method =:= 'PUT';
+                                                  Method =:= 'POST' ->
+    case req_is_conditional(RD) of
+        false ->
+            {false, RD, Ctx};
+        true ->
+            check_resource(RD, Ctx)
+    end;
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
This seems like a good change, but when removing user facing functionality we probably should deprecate it first in case there are users relying on the behavior and remove it for good in a subsequent release. In the meantime you could add in a configuration knob to circumvent the problem code path, but have the old code path be selected by default to ensure no existing apps become sad.

nathanaschbacher:
I don't think actually impacts existing end-user applications regardless.  If you had previously been attempting to set some random `Content-Encoding` value and tried to fetch the same key you'd end up getting a `HTTP 406` code back from Mochiweb anyway.

While I never put it past someone out there to want a highly-available, scalable `HTTP 406` generator... I'm also not sure we should be rewarding them.  :-)

Even if you tried to set RFC valid values of `deflate` or `compress` it causes a whole different weird behavior where when you try to request the key you get a code `0` back from Mochiweb, and the only way to get yourself unstuck from that condition is to put another value to the same key that contains a supported `Content-Encoding`.

tl;dr I'm pretty sure the whole storing `Content-Encoding` thing is a bug and there's no good way to have even accidentally exploited it as a piece of functionality from what I can see.

yorkshire-steve:

PR #833, opened 2014-02-11
Move metadata settings to advanced level.

evanmcc:
move md cache cuttlefish tunings into advanced so people are less
likely to play with it.  as written it is hard to make it improve
performance in a predictable way.
- [x] Eng +1
- [ ] PM +1
- [ ] CSE +1

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -318,9 +318,12 @@ end
 %% of the ETS table, rather than the actual data, to keep the size
 %% calculation simple, thus more space may be used than the simple
 %% size * vnode_count calculation would imply.
+%%
+%% Caution: Do not use without extensive benchmarking.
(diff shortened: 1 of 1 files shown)


lordnull:
Same generated value before and after change, only difference is the metadata_cache_size is not visible in the riak.conf file.

:+1:

seancribbs:
This did not get CSE/PM approval. :frowning:

evanmcc:
_shrug_

It should have never been in there in the first place, since the feature
was never fully baked.  I'll be sure to ping everyone next time.

jonmeredith:
It was discussed on the Core/KV mumble and I think GregB was there.  But I could have remembered wrong.

@seancribbs did it break something?

seancribbs:
@jonmeredith Nope, just wanted to be sure the process was clear.

PR #834, opened 2014-02-12
Export riak_kv_vnode:get/4

engelsanchez:
The new AAE test needs to get values directly from vnodes to be able to
verify the state of all replicas of an object. The function that
performs the get with an arbitrary caller (not an FSM) was unfortunately
not exported.


evanmcc:
:+1:

PR #836, opened 2014-02-13
Backport object size/sibling limits to 1.4

engelsanchez:
- Objects larger than warn_object_size will log a warning on read or
  write. The default is 5MiB
- Objects larger than max_object_size will fail to be written, logging
  an error message. By default no limit.
- Objects with a sibling count larger than warn_siblings will log a
  warning on write. By default 25.
- Objects with a sibling count larger than max_siblings will fail to
  write, logging an error message. By default no limit.

Note: in 2.0, there will be default values for max size and siblings, but I didn't want to risk failing operations for users in a point release.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1291,6 +1291,16 @@ do_get_object(Bucket, Key, Mod, ModState) ->
         false ->
             case do_get_binary(Bucket, Key, Mod, ModState) of
                 {ok, ObjBin, _UpdModState} ->
+                    ObjSize = size(ObjBin),
+                    WarnObjSize = app_helper:get_env(riak_kv, warn_object_size),
+                    case ObjSize > WarnObjSize of
+                        true ->
+                            lager:warning("Reading large object of size ~p from ~p/~p",
+                                          [ObjSize, Bucket, Key]);
+                        false ->
+                            ok
+                    end,
+
                     try
                         case riak_object:from_binary(Bucket, Key, ObjBin) of
                             {error, Reason} ->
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1768,13 +1778,53 @@ return_encoded_binary_object(Method, EncodedObject) ->
...
+                true ->
+                    lager:warning("Too many siblings for object ~p/~p (~p)",
+                                  [Bucket, Key, NumSiblings]);
+                false ->
+                    ok
+            end,
+            encode_and_put_no_sib_check(Obj, Mod, Bucket, Key, IndexSpecs, ModState)
+    end.
+
+encode_and_put_no_sib_check(Obj, Mod, Bucket, Key, IndexSpecs, ModState) ->
```
jonmeredith: Make sure handoff goes through this path - enforcing sibling / object size limits should only be done on paths controlled by the application where it could refetch/do something about it.  On the handoff path it would lead to data loss (downside is that if not enforced it will still lead to creation of an undesirably large object).

src/riak_kv_vnode.erl
```diff
@@ -1768,13 +1778,53 @@ return_encoded_binary_object(Method, EncodedObject) ->
            {{error, Reason::term(), UpdModState::term()}, EncodedObj::binary()}.
 
 encode_and_put(Obj, Mod, Bucket, Key, IndexSpecs, ModState) ->
+    NumSiblings = riak_object:value_count(Obj),
+    case NumSiblings > app_helper:get_env(riak_kv, max_siblings) of
+        true ->
+            lager:error("Too many siblings for object ~p/~p (~p)",
```
jonmeredith: Warning and failure messages are too close - we should be clear if the put went ahead or not.

jonmeredith:
+1 merge.

r_t's pass, verify_handoff passes - looked with code coverage to see code was hit.

Please don't forget to apply to 2.0.

engelsanchez:
Applying to 2.0 is already in the Kill Bill. No way it can be forgotten, right? right?

PR #838, opened 2014-02-13
Forwarding coordinated puts leads to siblings from a single put

russelldb:
With advanced causality tracking (DVV-ish dot stuff) it's possible to generate siblings on a single put. If the vnode is handing off, for example, it will both do the put (one dot) and forward (another dot.) Without dots riak_object merge would call these identical siblings a single value, with dots the concurrency is correctly detected. However, our customers my find this change confusing. To that end, this PR changes the DVV merge function to drop the dots on any identical siblings so that a single value is left (though the dots are gone, so it seems there is some slight risk of sibling explosion-ish behaviour)

Also this merges @jtuple's optimised merge branch. I reckon either @metadave or @jtuple could review this. Probably @jtuple has the best understanding of the merge.

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -385,7 +385,31 @@ fold_contents({r_content, Dict, Value}=C0, MergeAcc, Clock) ->
                     %% dominate) this sibling's `dot'. That means this
                     %% is a genuinely concurrent (or sibling) value
                     %% and should be kept for the user to reconcile.
-                    MergeAcc#merge_acc{keep=[C0|Keep]}
+                    %% UNLESS there is already a value that is
+                    %% _IDENTICAL_ to this in everyway (except the
+                    %% dot) in the accumulator, then we drop the dots
+                    %% on both values, knowing that a umerge later
+                    %% will bring them down to a single value. WHY?
+                    %% When a PUT is forwarded we actually do it
+                    %% twice, it is coordinated by two vnodes so it is
+                    %% actually concurrent with itself, we're trying
+                    %% to avoid that. (At what risk? I don't know,
+                    %% since this breaks the DVV model it is very hard
+                    %% to test.)
+                    case lists:keytake(Value, 3, Keep) of
+                        {value, {r_content, D1, Value}, Keep2} ->
+                            %% drop dots from both dicts and compare
+                            %% them, if equal, store a single, dotless
+                            %% value
+                            case lists:usort(dict:to_list(dict:erase(?DOT, D1))) ==
+                                lists:usort(dict:to_list(dict:erase(?DOT, Dict))) of
+                                true ->
+                                    MergeAcc#merge_acc{keep=[{r_content, dict:erase(?DOT, D1), Value} | Keep2]};
+                                false ->
+                                    MergeAcc#merge_acc{keep=[C0 | Keep]}
+                            end;
+                        false -> MergeAcc#merge_acc{keep=[C0 | Keep]}
+                    end
             end;
         undefined ->
             %% Both CRDTs and legacy data don't have dots. Legacy data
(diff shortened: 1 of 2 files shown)


jtuple:
We need to make a decision on if this is something we're going to go with or not. I think the last we talked about it, we decided not to do this. But, I'm not certain. Assigning to 2.0-beta milestone so we'll notice and make a decision. If we do decide to punt/not fix, and we haven't merged in the optimized merge logic we should at least merge that in as a separate pull-request.

russelldb:
Oh I'm happy for this branch to get merged. It just needs review. It doesn't need to be you @jtuple .

bookshelfdave:
I'll take a look

bookshelfdave:
is this something that could be configurable? We could ship with a default that will least likely surprise the user. I suppose it's a question for the PM team

russelldb:
It comes down to how you view the behaviour. If you think it is surprising, and worth _every_ merge paying the price, then this change is OK. If you think people who resolve siblings anyway (only an issue with allow_mult=true) won't even notice, then, why change?

Some users (Netic for example) sort of freak out when they get a sibling. Some customers (NHS for example) go to GREAT lengths to avoid any siblings, and might get pretty annoyed if riak just makes them anyway.

@gburd @michellep opinions?

bookshelfdave:
whatever the decision, we need to make sure repl works as expected.
@Vagabond @cmeiklejohn

bookshelfdave:
hm, some tests are failing. Can you rebase, this PR is auld

russelldb:
Rebased, do you want a force push, or a new branch / new PR? I still get two test failures with `cuttlefish`, but I get them on `develop`, too.

cmeiklejohn:
I would just force-push it to this branch and we'll continue the review.

russelldb:
Done.

cmeiklejohn:
Not to be annoying, but could this commit be squashed?

russelldb:
Sure, when it is plus one'd

src/riak_object.erl
```diff
@@ -294,6 +294,29 @@ merge_contents(NewObject, OldObject, true) ->
     riak_kv_crdt:log_merge_errors(Bucket, Key, CRDT, Error),
     merge_acc_to_contents(MergeAcc).
 
+%% Return true if A =< B, false otherwise
```
cmeiklejohn: I'd like to see a comment here explicitly about why this code is obtuse, specifically because it's an optimization.  It should be clear so readers understand why.

src/riak_object.erl
```diff
@@ -294,6 +294,29 @@ merge_contents(NewObject, OldObject, true) ->
...
+    if VA < VB ->
+            true;
+       VA > VB ->
+            false;
+       true ->
+            %% values are equal, compare metadata
+            compare_metadata(A, B)
+    end.
+
+compare_metadata(#r_content{metadata=MA}, #r_content{metadata=MB}) ->
```
cmeiklejohn, line 320: Same here; clarify this function operates in a particular manner because of an optimization.

src/riak_object.erl
```diff
@@ -363,8 +385,31 @@ fold_contents({r_content, Dict, Value}=C0, MergeAcc, Clock) ->
...
+                    %% actually concurrent with itself, we're trying
+                    %% to avoid that. (At what risk? I don't know,
+                    %% since this breaks the DVV model it is very hard
+                    %% to test.)
+                    case lists:keytake(Value, 3, Keep) of
+                        {value, {r_content, D1, Value}, Keep2} ->
+                            %% drop dots from both dicts and compare
+                            %% them, if equal, store a single, dotless
+                            %% value
+                            case lists:sort(dict:to_list(dict:erase(?DOT, D1))) ==
```
cmeiklejohn, line 416: Does deep equality matter here in any of the cases?

cmeiklejohn:
Overall, I'm :+1: once squashed.

Changeset seems to make sense, eunits are passing and are coherent.  Left minor concern about an equality check, and comments regarding documentation.

I'm still personally on the fence about the change.  Identifying and representing the concurrency in the objects is the correct solution, but I'm not sure it's the desired one.

src/riak_object.erl
russelldb, line 416: I don't know what you mean. Can you explain?

russelldb:
I don't understand your reservations, or maybe you haven't expressed them, but I do have my own. 

This is a hack. When I asked Nuno about I got a dismissive hand wave "they're concurrent". We're generating concurrency to ensure durability. Then we're trying to pretend we didn't, if we detect it. I'm on the fence too. I think if we're generating siblings, we should just expose them _or_ we should say that for _any and all_ equal sibling values, regardless of `time`, get collapsed into a single value.

This is a halfway house: for certain siblings that Riak generates (surprisingly) on the users behalf, we will merge them to a single value.

I don't feel high enough up the chain to call it. But I'll address your comments and squash to one commit, and let @jonmeredith @gburd @michellep make a call.

cmeiklejohn:
Yes, I was there when this conversation happened with Nuno.  I share your exact concerns, they are concurrent and we should represent it that way by generating the siblings.  I do not like collapsing them to a single value.

russelldb:
Occurred to me this morning that for one major customer I know of, a sibling is an error condition, and involves human interaction with the system. For that one reason alone, my attitude is toward including this.

russelldb:
Added comments as per review. Once your happy I'll squash this down and force push for final decision.

cmeiklejohn:
We discussed this a bit last night and would like to hold a further discussion about this with @jonmeredith, @gburd, @michellep, @jtuple and @russelldb.  Does the feature mumble tomorrow work?

The general thinking here is that DVV should be disabled for legacy buckets, and for typed buckets should be enabled, or an option to enable/disable.  This preserves 1.4 behavior for older users and upgrade users, but allows new clusters using typed buckets to start with both the dvv.enabled and allow_mult=true settings enabled, where siblings are expected.

That would result in this code not merging, but would require a separate PR for the optimized merge code.

cc: @kellymclaughlin

jonmeredith:
Resolved by basho/riak_core#568 and basho/riak_kv#891

PR #841, opened 2014-02-14
Backwards compat for users running map reduce jobs that use riak_kv_counter

russelldb:
Also add similar convenience functions for 2.0+ data types to riak_kv_crdt.

### DIFF
--- a/src/riak_kv_counter.erl
+++ b/src/riak_kv_counter.erl
@@ -1,0 +1,43 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_counter: Backwards compatibile access to counters for
+%% customer MR
+%%
+%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Backwads compatibility with 1.4 counters `value' function as
+%% used in the CRDT cookbook.
+%%
+%% @see riak_kv_crdt.erl
+%% @end
+
+-module(riak_kv_counter).
+
+-export([value/1]).
+
+-include("riak_kv_wm_raw.hrl").
+-include_lib("riak_kv_types.hrl").
+
+%% @doc Get the value of V1 (1.4) Counter from an Object. Backwards
+%% compatability with 1.4 for MapReduce.
+-spec value(riak_object:riak_object()) ->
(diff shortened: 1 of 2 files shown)


src/riak_kv_counter.erl
```diff
@@ -0,0 +1,43 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Backwads compatibility with 1.4 counters `value' function as
```
seancribbs: typo "Backwards"

src/riak_kv_counter.erl
```diff
@@ -0,0 +1,43 @@
...
+
+-include("riak_kv_wm_raw.hrl").
+-include_lib("riak_kv_types.hrl").
+
+%% @doc Get the value of V1 (1.4) Counter from an Object. Backwards
+%% compatability with 1.4 for MapReduce.
+-spec value(riak_object:riak_object()) ->
+                   integer().
+value(RObj) ->
+    {{_Ctx, Count}, _Stats} = riak_kv_crdt:value(RObj, ?V1_COUNTER_TYPE),
```
seancribbs, line 42: Will this raise an error if the type is a 2.0 counter instead of a 1.4 counter? What happens after the upgrade completes?

russelldb, line 42: no, you'll get a zero, like if you asked for a counter and it was a set, map, or regular riak_object

seancribbs:
So these are helpful, but I'd really love to see `riak_kv_mapreduce` wrappers around these so you can actually use them without writing your own functions. That said, it's a "nice-to-have" and can wait.

I made this module to test with:

``` erlang
-module(crdtmr).
-export([map_value/3]).
-include_lib("riak_kv/include/riak_kv_types.hrl").

map_value(RObj, <<"legacy_counter">>, _) ->
    [ riak_kv_counter:value(RObj) ];
map_value(RObj, _, <<"legacy_counter">>) ->
    [ riak_kv_counter:value(RObj) ];
map_value(RObj, _, _) ->
    {Type, Value} = riak_kv_crdt:value(RObj),
    [ riak_kv_crdt_json:fetch_response_to_json(Type, Value, undefined, ?MOD_MAP) ].
```

Here's the JSON of a sample job I sent from the Ruby client:

``` javascript
{"inputs":["maps","test"],"query":[{"map":{"language":"erlang","keep":true,"module":"crdtmr","function":"map_value"}}]}
```

...and here was the output:

``` ruby
[{"type"=>"map", "value"=>{"likes_counter"=>10, "friends_set"=>["Joe", "Justin"]}}]
```

Running over counters with the 'legacy_counter' arg turned on verified that for a non-legacy counter you'll just get 0:

``` ruby
irb(main):068:0> mr.query.first.arg = 'legacy_counter'
=> "legacy_counter"
irb(main):069:0> mr.run
=> [0]
irb(main):070:0> mr.query.first.arg = nil
=> nil
irb(main):071:0> mr.run
=> [{"type"=>"counter", "value"=>13}]
```

Other than the missing functionality, this works as advertised. :+1:

russelldb:
I see what you mean (re MR wrappers.) The main aim here was to not break existing code, and provide similar support for migrating existing code to 2.0 data types. I'll try and add some funs to riak_kv_mapreduce as time permits though.

Thanks!

PR #848, opened 2014-02-19
Fix incorrect specification.

cmeiklejohn:
Add missing return information and correct incorrect spec.

@seancribbs


seancribbs:
:+1: Warning about that line is gone. So much more still present though!

PR #852, opened 2014-02-20
Use iterator_refresh backend option for some folds

andrewjstone:
- Add iterator_refresh capability to riak_kv_elveldb_backend and
  riak_kv_multi_backend
- Handle iterator_refresh option/capability for ?FOLD_REQs in
  riak_kv_vnode
- support iterator_refresh option for AAE (riak_kv_index_hashtree)
- add riak_kv_vnode:fold/4
- return handoff fold options from handoff_started
  - handoff_started has been modified to require this
- Get value of iterator_refresh from env. Default to true.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -37,6 +37,7 @@
          readrepair/6,
          list_keys/4,
          fold/3,
+         fold/4,
          get_vclocks/2,
          vnode_status/1,
          ack_keys/1,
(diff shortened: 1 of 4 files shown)


andrewjstone:
Replaces #830 . See that for discussion.

jrwest:
re-iterating my +1. any `eunit` failures @andrewjstone and I came across (`riak_kv_schema_tests` and `keys_fsm_eqc` seem to be unrelated)

src/riak_kv_vnode.erl
```diff
@@ -1735,6 +1740,25 @@ do_fold(Fun, Acc0, Sender, ReqOpts, State=#state{async_folding=AsyncFolding,
...
+    case AsyncFolding andalso AsyncBackend of
+        true ->
+            [async_fold|Opts];
+        false ->
+            Opts
+    end.
+
+%% @private
+maybe_enable_iterator_refresh(Capabilities, Opts) ->
+    case lists:member(iterator_refresh, Capabilities) of
```
jrwest, line 1754: fwiw, we could have dumped the config check in here instead of in two places but meh

andrewjstone:
#853 addresses @jrwest's comment above

PR #856, opened 2014-02-25
Fix timeout handling for key and bucket listing via HTTP

kellymclaughlin:
Commit 32060f0230b00d3b6ecd6ac7fe4a6b4f71744d82 changed error handling in
riak_kv_keys_fsm and riak_kv_buckets_fsm to return error tuples instead of
just the error reason term. This broke the error message expectations for
timeout messages in riak_kv_wm_keylist and riak_kv_wm_buckets. This change
updates the error handling in those modules to be consistent with the current
form sent by the fsm modules.

This can be tested using the verify_api_timeouts riak_test
module. Without this change the test and the associated keylisting
request hangs attempting to verify the timeout for keylisting request
works correctly. With this change the test works as expected and
passes.


andrewjstone:
Code looks good. Running `verify_api_timeouts` now.

andrewjstone:
Confirmed that `verify_api_timeouts` hangs on develop but passes with this fix.

:+1:

PR #857, opened 2014-02-28
Make get_put_monitor_eqc pass deterministically

andrewjstone:
- Change check_state to be the optional invariant callback.
- Just return true from postcondition since it was only used to check
  the state.
- remove unneccessary `timer:sleep/1` calls
- Poll for folsom stat changes instead of relying on timer:sleep. This will fix
  the test on slow machines such as the builders. It also makes the test
  run a lot faster as it no longer sleeps arbitrarily.
- Move test initialization out of the initial_state callback
- Log the metric that failed for invariants
- Poll active stats during get and put fsm shutdown/exit since the
  count needs to be decremented
- swap ordering of monitoring the pid and sending it a kill msg in
  `end_and_wait/2`

### DIFF
--- a/test/get_put_monitor_eqc.erl
+++ b/test/get_put_monitor_eqc.erl
@@ -6,6 +6,11 @@
 -include_lib("eqc/include/eqc_statem.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
+-define(PUTS_ACTIVE, {riak_kv, node, puts, fsm, active}).
+-define(GETS_ACTIVE, {riak_kv, node, gets, fsm, active}).
+-define(PUTS_ERRORS, {riak_kv, node, puts, fsm, errors}).
+-define(GETS_ERRORS, {riak_kv, node, gets, fsm, errors}).
+
 -compile([export_all]).
 
 -record(state, {
@@ -23,7 +28,7 @@ eqc_test_() ->
                     ok
             end,
      fun(_) ->
-             [maybe_stop_and_wait(Server) || Server <- [riak_core_stat_cache, riak_kv_stat]]
+             ok
      end, [
            {timeout, 120, ?_assertEqual(true, quickcheck(numtests(100, ?QC_OUT(prop()))))}
           ]}.
@@ -38,9 +43,10 @@ check() ->
     check(prop(), current_counterexample()).
 
 prop() ->
+    error_logger:tty(false),
+    crypto:start(),
     ?FORALL(Cmds, commands(?MODULE), begin
-        crypto:start(),
-        application:start(folsom),        
+        reset_test_state(),
         {_,State,Res} = run_commands(?MODULE, Cmds),
         case Res of
             ok ->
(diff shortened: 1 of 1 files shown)


jrwest:
Huge improvement but not 100% there yet. Still seeing a regular failure due to a race when running 10k iterations. the test passes using `eqc:check` every time afterwards. 

```
Failed Metric = {riak_kv,node,gets,fsm,active}
QC result: {invariant,
               {'EXIT',
                   {{assertEqual_failed,
                        [{module,get_put_monitor_eqc},
                         {line,163},
                         {expression,"Val"},
                         {expected,3},
                         {value,4}]},
                    [{get_put_monitor_eqc,'-invariant/1-fun-0-',2,
                         [{file,"test/get_put_monitor_eqc.erl"},{line,163}]},
                     {get_put_monitor_eqc,'-invariant/1-lc$^0/1-0-',1,
                         [{file,"test/get_put_monitor_eqc.erl"},{line,163}]},
                     {get_put_monitor_eqc,invariant,1,
                         [{file,"test/get_put_monitor_eqc.erl"},{line,157}]},
                     {eqc_statem,run_commands,2,
                         [{file,"../src/eqc_statem.erl"},{line,704}]},
                     {get_put_monitor_eqc,'-prop/0-fun-0-',1,
                         [{file,"test/get_put_monitor_eqc.erl"},
                          {line,50}]}]}}}
Shrinking.(1 times)
[{set,{var,1},{call,get_put_monitor_eqc,get_fsm_started,[]}},
 {set,{var,2},{call,get_put_monitor_eqc,put_fsm_started,[]}},
 {set,{var,3},{call,get_put_monitor_eqc,put_fsm_noproc,[]}},
 {set,{var,4},{call,get_put_monitor_eqc,put_fsm_exit_normal,[put,{var,2}]}},
 {set,{var,5},{call,get_put_monitor_eqc,get_fsm_started,[]}},
 {set,{var,6},{call,get_put_monitor_eqc,get_fsm_started,[]}},
 {set,{var,7},{call,get_put_monitor_eqc,get_fsm_noproc,[]}}]
```

test/get_put_monitor_eqc.erl
```diff
@@ -38,9 +43,10 @@ check() ->
     check(prop(), current_counterexample()).
 
 prop() ->
+    error_logger:tty(false),
+    crypto:start(),
     ?FORALL(Cmds, commands(?MODULE), begin
-        crypto:start(),
-        application:start(folsom),        
+        reset_test_state(),
```
jrwest, line 50: its good to do clean up after the iteration runs, also. That way the last iteration doesn't leave around artifacts that mess with tests that run later in the suite.

test/get_put_monitor_eqc.erl
```diff
@@ -38,9 +43,10 @@ check() ->
...
         {_,State,Res} = run_commands(?MODULE, Cmds),
         case Res of
```
jrwest: also while we are cleaning this up, it would be nice to move away from this pattern match and use `?WHENFAIL` instead.

test/get_put_monitor_eqc.erl
```diff
@@ -129,67 +139,40 @@ next_state(S, _Res, {call, _, _, [put, Pid]}) ->
...
+        {?GETS_ERRORS, [{count, GetErrCount}, {one, GetErrCount}]}
     ],
 
     [ begin
-        ?assertEqual(Expected, folsom_metrics:get_metric_value(Metric))
-    end || {Metric, Expected} <- MetricExpects],
+        Val = folsom_metrics:get_metric_value(Metric),
+        case Val of 
+            Expected -> ok;
+            _ -> io:format(user, "Failed Metric = ~p~n", [Metric])
```
jrwest: this `io:format` can get lost in the noise or contribute to it during shrinking. In addition the use of `?assertEqual` below (noticed this was carried over from the original version) makes failures a bit noiser than they need to be and isn't very idiomatic imo. The invariant can return any value besides `true` to indicate failure. Perhaps do something like:

``` erlang
Zipped = [{Metric, Excepted, folsom_metrics:get_metric_value(Metric)} || {Metric, Expected} <- MetricExpects],
Bad = [{Metric, {expected, Expected}, {actual, Actual}} || {Metric, Expected, Actual} <- Zipped, Expected /= Actual],
case Bad of
  [] -> true;
  _ -> Bad
end.
```

andrewjstone:
I used `eqc_statem:pretty_commands/4` instead of `?WHENFAIL`

jrwest:
+1 merge. 110k iterations passed. nice work.

PR #860, opened 2014-03-04
Fix use of fsm timeouts in 2i AAE

engelsanchez:
NOTE: This is all the work that went into riak_kv from 1.4.7 to 1.48, forward ported to 2.0

Joe pointed out some problems with FSM time outs in this code. It is
possible to set a timeout for the next FSM state, then receive a message
passed to handle_info and lose that timeout if not set again. Basically,
they are useless.
Refactored code to use gen_fsm:send_event_after() instead and managed
timer for the AAE pid request manually.
Also, a zero timeout transition was avoided by refactoring the code
around the next_partition state, which is now a simple function
directly executed from other states.
Also, if a request for an AAE pid every failed, the whole process would
have died silently. Instead, that is handled now like any other
partition error. So 2i AAE can happily go finishing the other
partitions.

Testing notes: Joe and I tested this manually by modifying riak_kv_vnode to drop replies to the hashtree pid request and verifying that 2i AAE could finish. I don't have time right now to add that to verify_2i_aae, so if a kind sould could take care of that, it would be great.

### DIFF
--- a/src/riak_kv_2i_aae.erl
+++ b/src/riak_kv_2i_aae.erl
@@ -74,14 +74,14 @@
          caller :: pid(),
          reply_to,
          early_reply,
-         aae_pid_req_id :: pid() | undefined,
+         aae_pid_req_id :: reference() | undefined,
          aae_pid_timer :: reference() | undefined,
          worker_monitor :: reference() | undefined,
          worker_status :: worker_status(),
          index_scan_count = 0 :: integer(),
          hashtree_population_count = 0 :: integer(),
          exchange_count = 0 :: integer(),
-         open_dbs = [] :: [reference()]
+         open_dbs = [] :: [eleveldb:db_ref()]
         }).
 
 -record(partition_result, {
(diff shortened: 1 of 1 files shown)


andrewjstone:
- [x] code inspection
- [x] verify_2i_aae
- [x] make test
- [x] dialyzer
- [x] xref

andrewjstone:
Seeing the following dialyzer warnings:

``` erlang
riak_kv_2i_aae.erl:181: Expression produces a value of type 'false' | non_neg_integer(), but this value is unmatched
riak_kv_2i_aae.erl:191: Expression produces a value of type 'false' | non_neg_integer(), but this value is unmatched
riak_kv_2i_aae.erl:193: The created fun has no local return
riak_kv_2i_aae.erl:195: The call riak_kv_2i_aae:repair_partition(Partition::integer(),Speed::'undefined' | integer(),'riak_kv_2i_aae',TreePid::any()) breaks the contract (integer(),integer(),{pid(),any()},pid()) -> #partition_result{}
```

engelsanchez:
Added some dialyzer fixes. I still some that make absolutely no sense to me. If you have any idea of how to resolve them, please share yo.

seancribbs:
The warning on L195 is differing by the second argument, which means the spec on repair_partition/4 is probably wrong. If 'undefined' is actually allowed, you'll need to widen the spec on that function.

andrewjstone:
:+1:

engelsanchez:
The one you mentioned was fixed that way @seancribbs.

PR #866, opened 2014-03-11
Make bucket type interactions utf8-friendly

macintux:
Can now create, enable, etc unicode bucket type names. Have verified that adding and retrieving keys work correctly via both HTTP and PB, but other than that and the simple test added to `riak_test` (below) this has not been extensively tested.

Corresponding `riak_test` PR: basho/riak_test#549

/cc @jrwest


macintux:
closing in favor of #868

PR #867, opened 2014-03-13
Make get_put_monitor_eqc pass during `make test`

andrewjstone:
Calling code:purge/1 and code:delete/1 were needed to make the test
pass. They effectively skirt using a sidejob worker to update folsom
stats, and do it directly instead. Since riak_kv is not running casts to
update folsom were being dropped on the floor that were needed by the
test. This only occurred in `make test` and not when running
get_put_monitor_eqc standalone because riak_kv was started and shutdown
before this test ran during `make test`.

Additionally, I added timeouts for the polling in case of a regression,
made the precondition more specific and ran this test in a slave node.
While these changes aren't strictly necessary they provide for cleaner
code and a better example for future tests.

I also removed some dead code.

### DIFF
--- a/test/get_put_monitor_eqc.erl
+++ b/test/get_put_monitor_eqc.erl
@@ -10,6 +10,7 @@
 -define(GETS_ACTIVE, {riak_kv, node, gets, fsm, active}).
 -define(PUTS_ERRORS, {riak_kv, node, puts, fsm, errors}).
 -define(GETS_ERRORS, {riak_kv, node, gets, fsm, errors}).
+-define(POLL_TIMEOUT, 10). %% seconds
 
 -compile([export_all]).
 
@@ -24,17 +25,17 @@
         eqc:on_output(fun(Str, Args) -> io:format(user, Str, Args) end, P)).
 
 eqc_test_() ->
-    {setup, fun() ->
-                error_logger:tty(false),
-                crypto:start(),
-                ok
-            end,
-     fun(_) ->
-             cleanup() 
-     end, [
-           {timeout, 360, ?_assertEqual(true, quickcheck(numtests(5000,
-                           ?QC_OUT(prop()))))}
-          ]}.
+    error_logger:tty(false),
+    catch code:purge(riak_kv_stat_sj),
+    catch code:delete(riak_kv_stat_sj),
+    net_kernel:stop(),
+    net_kernel:start(['test-master@127.0.0.1', 'longnames']),
+    erlang:set_cookie(node(), ayo),
+    {node, 'tester@127.0.0.1', "-setcookie ayo", 
+     [
+         {timeout, 120, [?_assertEqual(true, quickcheck(eqc:testing_time(90,
+                           ?QC_OUT(prop()))))]}
+       ]}.
 
 test() ->
     test(100).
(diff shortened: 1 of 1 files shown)


test/get_put_monitor_eqc.erl
```diff
@@ -24,17 +25,17 @@
...
-           {timeout, 360, ?_assertEqual(true, quickcheck(numtests(5000,
-                           ?QC_OUT(prop()))))}
-          ]}.
+    error_logger:tty(false),
+    catch code:purge(riak_kv_stat_sj),
+    catch code:delete(riak_kv_stat_sj),
+    net_kernel:stop(),
+    net_kernel:start(['test-master@127.0.0.1', 'longnames']),
+    erlang:set_cookie(node(), ayo),
+    {node, 'tester@127.0.0.1', "-setcookie ayo",
```
Vagabond: Usually you don't do anything before returning an eunit fixture. You can't do a {setup, test-master@127.0.0.1, fun() -> ... here?

Vagabond: I'm not sure that the fixture that is returned is run right after, basicially.

andrewjstone: I was having trouble trying to wrap the node fixture in a setup figure which is how I ended up with this code. It appears to work that way from running it. Slave nodes must inherit the prior environment or something. I can rework it to just not use the slave node if you want.

Vagabond: My main concern is that this is undefined behaviour.

andrewjstone: Ok, since it's not necessary to run in a slave node, I'll just rework it to use a setup fixture. We can leave slave nodes for the future as necessary.

andrewjstone:
@Vagabond Can you give this another look. It seems to have gotten lost in the shuffle.

reiddraper:
I'm on it.

test/get_put_monitor_eqc.erl
```diff
@@ -174,11 +167,21 @@ invariant(S) ->
...
+
+poll_stat_change(Metric, OriginalValue, ExpireSecs) ->
+    {_, Secs, _} = os:timestamp(),
+    case Secs > ExpireSecs of
+        true ->
+            throw({error, {expired, Metric, OriginalValue}});
+        false ->
+            case folsom_metrics:get_metric_value(Metric) of
+                OriginalValue -> 
+                    poll_stat_change(Metric, OriginalValue, ExpireSecs);
```
reiddraper, line 181: Is there any downside to just hammering on this metric? Would it behoove us to sleep for a bit? Right now we're just spinning.

andrewjstone, line 181: A sleep would work at reducing the spinning. I figured we'd get maximum
speed from doing it this way though, since erlang scheduler would just do
it's magic and the change probably happens on another processor. I'm fine
either way.

reiddraper:
My test results:

```
Failed: 6.  Skipped: 0.  Passed: 311.
```

andrewjstone:
My guess is those are tests that either are still broken or fixed on develop. This PR just fixes `get_put_monitor_eqc` when run with `make test`.

reiddraper:
Something bizarre is going on, on another run:

```
Failed: 0.  Skipped: 0.  Passed: 85.
```

andrewjstone:
that is weird

reiddraper:
Yay nondeterminism:

```
Failed: 2.  Skipped: 0.  Passed: 315.
```

reiddraper:
+1, there are several other failures, but all appear unrelated.

andrewjstone:
yay

PR #868, opened 2014-03-13
a couple bucket type console command improvements

macintux:
### Allow utf-8 bucket types

Can now create, enable, etc unicode bucket type names. Have verified that adding and retrieving keys work correctly via both HTTP and PB, but other than that and the tests added to `riak_test` (below) this has not been extensively tested.

Corresponding `riak_test` PR: basho/riak_test#549

(Replaces #866, which included unnecessary whitespace removals)
### don't require empty json for default bucket type properties on create
- addresses https://github.com/basho/riak/issues/423
- riak-admin changes in https://github.com/basho/riak/pull/532

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -455,16 +455,12 @@ bucket_type_raw_props(Type) ->
     riak_core_claimant:get_bucket_type(Type, undefined, false).
 
 bucket_type_print_status(Type, undefined) ->
-    io:setopts([{encoding, unicode}]),
     io:format("~ts is not an existing bucket type~n", [Type]);
 bucket_type_print_status(Type, created) ->
-    io:setopts([{encoding, unicode}]),
     io:format("~ts has been created but cannot be activated yet~n", [Type]);
 bucket_type_print_status(Type, ready) ->
-    io:setopts([{encoding, unicode}]),
     io:format("~ts has been created and may be activated~n", [Type]);
 bucket_type_print_status(Type, active) ->
-    io:setopts([{encoding, unicode}]),
     io:format("~ts is active~n", [Type]).
 
 bucket_type_print_props(undefined) ->
@@ -479,7 +475,6 @@ bucket_type_activate([TypeStr]) ->
     bucket_type_print_activate_result(Type, riak_core_bucket_type:activate(Type), IsFirst).
 
 bucket_type_print_activate_result(Type, ok, IsFirst) ->
-    io:setopts([{encoding, unicode}]),
     io:format("~ts has been activated~n", [Type]),
     case IsFirst of
         true ->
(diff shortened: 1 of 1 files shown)


macintux:
@jrwest will create a separate PR for nodetool to handle the I/O properties so we don't have to specify unicode output everywhere, so this branch will be modified to suit. Also will be creating a separate PR for `riak_core` to handle at least two places in `riak_core_console` where `list_to_binary` is invoked on bucket type names.

macintux:
- Dropped repeated calls to `io:setopts` in favor of basho/node_package#119
- Created basho/riak_core#558 to complete this set of changes.

reiddraper:
+1

src/riak_kv_console.erl
```diff
@@ -444,7 +444,7 @@ run_reformat(M, F, A) ->
     end.
 
 bucket_type_status([TypeStr]) ->
-    Type = list_to_binary(TypeStr),
+    Type = unicode:characters_to_binary(TypeStr, utf8),
```
jrwest: we should probably change to using the 3-arity version here since the second argument only specifies the encoding of input data that contains binaries [1] 

EDIT: by here, i mean everywhere we call this function.

[1] http://www.erlang.org/doc/man/unicode.html#characters_to_binary-3

seancribbs: One can also check the encoding of the input using `io:getopts/0,1`, then use that to feed the input encoding in `unicode:characters_to_binary/3`. 

http://www.erlang.org/doc/man/io.html#getopts-1

macintux: @seancribbs We're forcing the input to be encoded in utf-8 via basho/node_package#119. Interestingly, `getopts` doesn't appear to be useful for shell input; as you can see below, it reports `latin1` but the argument is properly handled as utf-8 on Mac (but on Linux it would be represented as 8-bit integers unless `+fnu` is passed to the VM).

```
Have-Riak-Will-Travel:bt-unicode John$ cat mine.erl


main([Args]) ->
    io:format("~p~n", [io:getopts()]),
    io:format("~w~n", [Args]).
Have-Riak-Will-Travel:bt-unicode John$ escript mine.erl سلام
[{binary,false},{encoding,latin1}]
[1587,1604,1575,1605]
```

macintux:
@jrwest I'm out of time, still fighting another set of PRs before I fly to SF in a few hours. If we can approve/merge this with the current 2-arity version, I'll replace them at a later point.

src/riak_kv_console.erl
jrwest: fwiw, the `getopts`, `encoding` here is referring to `stdin/out/err` and not the command-line arguments, which is why it doesn't apply.

jrwest:
@macintux given this is slated for `2.0-RC`, and we have the time, I would rather see us get it fixed before merging. otherwise I'm +1.

jrwest:
@jaredmorrow +1 here as well?

jaredmorrow:
I saw the previous +1's, so didn't try out weird encodings, but things did work with the empty props for create and update / list / status continued to work.

So union of @reiddraper @macintux and myself I think this is +1 good to go.

PR #869, opened 2014-03-17
Send reply in event of a premature exit (via throw) from fold.

slfritchie:
Allow a folder to throw its accumulator to send it to the fold client, e.g. hinted/repair/ownership handoff.


reiddraper:
@borshop merge

PR #874, opened 2014-03-21
Fix list_to_binary invocations on permissions errors

macintux:
Change to `unicode:characters_to_binary`. This is required if the forthcoming `riak_core` PR is accepted, but is recommended even if not.

/cc @Vagabond


lordnull:
associated riak_test's pass w/ this branch, but fail without, so :+1:

PR #879, opened 2014-03-25
Export AAE format functions for reuse

rzezeski:
Export the internal AAE format functions so that they may be reused by
Yokozuna.


rzezeski:
Needed for basho/yokozuna#349.

jaredmorrow:
+1, it worked for me

PR #881, opened 2014-03-25
Export riak_client type.

cmeiklejohn:
External dependencies such as repl use riak_client in type signatures.
Export this type so dialyzer doesn't complain when it's unavailable when
analyzing this repo.

@Vagabond


Vagabond:
+1, do it

PR #882, opened 2014-03-25
Provide a minimal type for query_def.

cmeiklejohn:
query_def is specified across kv and repl, but the type no longer
exists, provided a minimal type to use for the meantime.  We can't use
typer to derive the type, because riak_index is so incorrectly
specified, typer believes all of the functions return no value.

cc: @Vagabond


Vagabond:
I'm not really thrilled with this PR, should we maybe just merge the repl PR and wait for KV to be fixed right?

Vagabond:
Fine whatever, +1.

PR #883, opened 2014-03-27
multi_backend.schema doesn't handle error cases of cuttlefish_generator:map/2

joedevivo:
Cuttlefish is pretty awesome. One of the awesome things it does is phased error handling. If there's an error in any phase, it returns the 3-tuple `{error, phase_name, [{error, "Message"}]}`

multi_backend is the only translation that recurses back in to recycle the translations and validators from individual backend configurations. If something in there caused a problem, it throws back that awesome error tuple, but doesn't handle it, generating an `[error, function_clause]` instead. this can be fixed by checking for the 3-tuple and passing the error back upstream.

### DIFF
--- a/priv/multi_backend.schema
+++ b/priv/multi_backend.schema
@@ -22,8 +22,6 @@
 {translation,
  "riak_kv.multi_backend",
  fun(Conf, Schema) ->
-
-
   GenerateSubConfig = fun(Name, Prefix, ProplistKey, ModuleName) ->
       BackendConfigName = ["multi_backend", Name],
       BackendConfigPrefix = BackendConfigName ++ [Prefix],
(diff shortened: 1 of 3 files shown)


priv/multi_backend.schema
```diff
@@ -22,6 +22,29 @@
...
+  GenerateSubConfig = fun(Name, Prefix, ProplistKey, ModuleName) ->
+      BackendConfigName = ["multi_backend", Name],
+      BackendConfigPrefix = BackendConfigName ++ [Prefix],
+      SubConf = [ begin
+          {Key -- BackendConfigName, Value}
+      end || {Key, Value} <- cuttlefish_variable:filter_by_prefix(BackendConfigPrefix, Conf)],
+
+      case cuttlefish_generator:map(Schema, SubConf) of
+        {error, _Phase, _Errors} ->
+            cuttlefish:invalid("Error running recursive call to cuttlefish_generator:map/2");
```
seancribbs: Could this be a bit more descriptive to the user?

priv/multi_backend.schema
```diff
@@ -22,6 +22,29 @@
 {translation,
  "riak_kv.multi_backend",
  fun(Conf, Schema) ->
+
+
+  GenerateSubConfig = fun(Name, Prefix, ProplistKey, ModuleName) ->
+      BackendConfigName = ["multi_backend", Name],
+      BackendConfigPrefix = BackendConfigName ++ [Prefix],
+      SubConf = [ begin
+          {Key -- BackendConfigName, Value}
```
seancribbs: Why isn't this just `lists:nthtail(2, Key)`?

joedevivo: like all lists:Functions, I never expect it to exist and I have a proclivity for the Little Caesar's of operators aka "operator operator".

priv/multi_backend.schema
joedevivo: See basho/cuttlefish#145 for future work to make that possible.

When this goes bad, which is rarely, and only for developers, there are multiple errors that need to be incorporated into the calling version of `map/2`'s error list.

seancribbs:
:+1: when build completes

PR #888, opened 2014-04-02
Fix handling of bucket+type with MR keyfilters

engelsanchez:
Roach found this while writing tests for the Java client. If using the
JS syntax for using bucket types and keyfilters together through the PB
interface, this was broken. In the http case without security, the path
is not touched so it didn't show before.

/cc @broach can you verify with the tests you are building?

### DIFF
--- a/src/riak_kv_mapred_term.erl
+++ b/src/riak_kv_mapred_term.erl
@@ -190,6 +190,11 @@ get_required_permissions(Inputs, _Query) ->
                                is_list(Filters) ->
             [{"riak_kv.list_keys", {<<"default">>, Bucket}},
              {"riak_kv.mapreduce", {<<"default">>, Bucket}}];
+        {{Type, Bucket}, Filters} when is_binary(Type),
+                                       is_binary(Bucket),
+                                       is_list(Filters) ->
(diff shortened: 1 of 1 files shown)


engelsanchez:
ping @broach, what happened with the MR tests that were failing? Did you have a chance to test this fix against them?

engelsanchez:
Re-ping @broach. What happened with this? Were you able to verify that this fixed the Java client tests?

engelsanchez:
This fix can be verified by https://github.com/basho/riak_test/pull/623. A new commit was added to fix another problem for bucket types + keyfilters in the parsing stage

kellymclaughlin:
@borshop merge

broach:
@engelsanchez Yeah, sorry - I had this open in a tab as a nagging reminder that I hadn't gotten to it and it hadn't updated. Thanks.

engelsanchez:
Actually @broach, it would be a good idea to make sure that your Java client tests do pass after the fix to make sure it's the exact code path you were hitting. Let me know if anything doesn't work and we can open a new issue.

broach:
@engelsanchez Finally got here. The test now passes. :+1:

PR #889, opened 2014-04-04
Merge jdb's merge optimisations

russelldb:
As requested, breaking the DVV bucket props work into two, independent, PRs. This one merges @jtuple's merge optimisation work (and adds some comments.)

PR #890, opened 2014-04-04
Make `dvv_enabled` a bucket property

russelldb:
Set it's default to `false` on legacy/default buckets
Set it's default to `true` on typed buckets

Update all the tests and associated places this change touches

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -414,12 +414,12 @@ end
 %% @doc The strategy used when merging objects that potentially have
 %% conflicts.
 %%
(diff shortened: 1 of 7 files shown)


russelldb:
Last one for @cmeiklejohn , I pwomise.

cmeiklejohn:
@rdb Isn't this the one I was supposed to test?  You referenced it in #889.

russelldb:
Re-opened as https://github.com/basho/riak_kv/pull/891 pointing at `develop` rather that the optimisations branch.

cmeiklejohn:
OK.

PR #891, opened 2014-04-04
Make `dvv_enabled` a bucket property

russelldb:
Set it's default to false on legacy/default buckets
Set it's default to true on typed buckets

Update all the tests and associated places this change touches

### DIFF
--- a/src/riak_object.erl
+++ b/src/riak_object.erl
@@ -294,7 +294,13 @@ merge_contents(NewObject, OldObject, true) ->
     riak_kv_crdt:log_merge_errors(Bucket, Key, CRDT, Error),
     merge_acc_to_contents(MergeAcc).
 
-%% Return true if A =< B, false otherwise
+%% Optimisation. To save converting every meta dict to a list sorting,
+%% comapring, and coverting back again, we use this optimisation, that
+%% shortcuts out the meta_data comparison, if the obects aren't equal,
+%% we needn't compare meta_data at all. `compare/2' is an "ordering
+%% fun". Return true if A =< B, false otherwise.
+%%
+%% @see lists:usort/2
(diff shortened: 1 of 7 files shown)


russelldb:
Sorry @cmeiklejohn do not have the time/patience to put the optimisations and bucket prop stuff as separate things. Feel free to do it yourself, though.

cmeiklejohn:
Do any of the tests specifically test the false case?  The case where the bucket is mecked to return [] and [dvv_enabled] both end up executing the true case, since the default of the get_value is true.

cmeiklejohn:
Other than that, I'm :+1: on this changeset.

russelldb:
I didn't write new tests, no. I can write some on Monday, or you can add some if you like.

cmeiklejohn:
Cool, just wanted to make sure you didn't think something was tested that wasn't.  I'd say merge.

russelldb:
Oh, it needs https://github.com/basho/riak_core/pull/568

src/riak_object.erl
```diff
@@ -294,6 +295,41 @@ merge_contents(NewObject, OldObject, true) ->
     riak_kv_crdt:log_merge_errors(Bucket, Key, CRDT, Error),
     merge_acc_to_contents(MergeAcc).
 
+%% Optimisation. To save converting every meta dict to a list sorting,
+%% comapring, and coverting back again, we use this optimisation, that
```
jplock, line 299: comparing?

russelldb, line 299: Late to the game, but well spotted. New branch and PR? Or live with it like "referer"?

jplock, line 299: https://github.com/basho/riak_kv/pull/892

PR #898, opened 2014-04-10
Bound testing time for riak_object_dvv_statem

reiddraper:
On my box, this test was timing before it reached the required 1k tests.
This change uses testing_time instead of a fixed number of iterations.
On my box, I find that I still get close to 1k iteratons in.


russelldb:
Runs as part of `make test` and stand alone with `./rebar suites=riak_object_dvv_statem`

+1 a3b36d4

reiddraper:
@borshop isn't running on this repo yet, so I'm merging manually.

Issue #903, opened 2014-04-14
Violating vclock:dot opaque type


reiddraper:
`riak_object:is_drop_candidate/2` peeks inside of a `dot()`, and dialzyer complains about this:

```
riak_object.erl:368: The call riak_object:is_drop_candidate(Dot::vclock:dot(),Drop::[{_,_}]) contains an opaque term as 1st argument when terms of different types are expected in these positions
```

Would it make sense for this function to just be moved into the vclock module?

russelldb:
I know :(

russelldb:
@reiddraper I don't think so, or I don't know, let me look into it. It was hack because riak is broken (two vclock entries can have the same {id, count} but a different timestamp) maybe we should fix that rather fundamental flaw, and leave this dialyzer bug as a open wound to remind us?

russelldb:
But yes, I think some refactoring, to get `vclock` to destructure the dot would be good. I'll do it.

russelldb:
Can't push to github, otherwise I'd push the fix to this.

reiddraper:
Closing now that #910 was merged.

PR #910, opened 2014-04-16
Use vclock encapsulation of dot/pure_dot

russelldb:
Addresses #903 needs basho/riak_core#573


reiddraper:
+1. This removes the dialyzer warnings and the code looks good with regard to the changes in basho/riak_core#573. Thanks.

PR #911, opened 2014-04-16
Ignore all current dialyzer errors

reiddraper:
The exit code from `make xref dialyzer` is now 0.
- Don't produce unmatched_returns at all
- Add all current errors to the ignore file

This is a temporary solution so that we can get the result of `make xref
dialyzer test` completely green, and get the repository running on bors.
Once this is done, this should not allow any _new_ errors into the repo.
Concurrently, we'll continue to fix dialyzer errors and _remove_ them
from the ignore file.

cc/ @Vagabond 

Here's to hoping this build is SUPERGREEN!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .PHONY: deps test
 
+DIALYZER_FLAGS =
+
 all: deps compile
 
 compile: deps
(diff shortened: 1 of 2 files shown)


Vagabond:
+1

PR #913, opened 2014-04-17
Ignore {ack,_,now_executing} that arrive late

### LINKED ISSUE #912: put fsm can get a remote coordinator ack in any state and must be able to handle it.
In perf regression testing I was running into an issue on a node that has some sort of disk problem causing high latencies and iowait.  This was causing a scheduler collapse on the node, which was causing extremely long sidejob worker queues.  

This new failure case is caused by the changes in https://github.com/basho/riak_kv/pull/627

The logic works something like this:
1. Long sidejob queues/collapse on node A are causing a > 3s delay on the remote spawn and ack process.
2. Node B, which got the original request, is # 4 on the preflist (for example).  It removes node A from the preflist, and becomes # 3, and so handles coordination itself, see: https://github.com/basho/riak_kv/blob/2.0.0beta1/src/riak_kv_put_fsm.erl#L325-L327
3. Node A finally processes the spawn and sends the ack message to B.
4. B is now no longer waiting for that particular message, which isn't handled else where in the fsm, causing the crash abbreviated below.

```
2014-04-15 04:53:16 =CRASH REPORT====
  crasher:
    initial call: riak_kv_put_fsm:init/1
    pid: <0.6330.1029>
    registered_name: []
    exception exit: {badmsg,[{gen_fsm,terminate,7,[{file,"gen_fsm.erl"},{line,622}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,239}]}]}
    ancestors: [riak_kv_put_fsm_sj_2,riak_kv_put_fsm_sj_worker_sup,riak_kv_put_fsm_sj,sidejob_sup,<0.463.0>]
    messages: []
    links: [<0.502.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 2586
    stack_size: 27
    reductions: 1164
  neighbours:
2014-04-15 04:53:16 =ERROR REPORT====
** State machine <0.2474.1029> terminating 
** Last message in was {ack,'riak@10.0.24.40',now_executing}
** When State == validate
**      Data  == {state,{raw,7430680,<0.30606.990>},{r_object,<<"testbucket">>,<<0,32,48,191>>,[  ]}
** Reason for termination = 
** badmsg

```

also, it seems that we can get this message in more or less any state:

```
$ grep -h 'State ==' log-10.0.24.44/crash.log*  | sort | uniq
** When State == finish
** When State == postcommit
** When State == precommit
** When State == validate
** When State == waiting_local_vnode
** When State == waiting_remote_vnode
```

slfritchie:
Fixes #912 

I don't like the style of 'crash if you get an unexpected message' ... but I'm not changing that particular handle_info() clause here.


evanmcc:
Log the offending node? Getting these messages is indicative of deep
sadness, we shouldn't totally ignore them.

slfritchie:
Agreed and disagreed.  The problem is that there can be dozens/hundreds of them per second.  Lager isn't smart enough IIRC to ignore those messages and _not_ ignore unrelated but important messages being logged at/near the same instant.

evanmcc:
I don't follow, I'm sorry.

If you're talking about rate limiting, my understanding is that only messages coming from error_logger are rate limited (see @Vagabond here: https://github.com/basho/riak/issues/517#issuecomment-38921318), so even if there's a flurry of these, it shouldn't be a problem.  That said, my experience here is that these are relatively uncommon as they rely on some special circumstances to come about.

Edit by @slfritchie: @evanmcc I don't want a spam of hundreds of messages/sec from the new handle_info() clause (all of which will be coming from different PIDs) to cause lager to suppress messages from other parts of Riak.

Vagabond:
Lager will slow down the caller if a caller tries to flood the system with logs.

Vagabond:
It might be better to have a counter for these in stats?

slfritchie:
@evanmcc I edited your question above.

@Vagabond Er, new stats are fine with me.  I've not invented a new one in a few years, is there any magic I should be aware of?  (A pointer is fine.)

Vagabond:
I don't have any good advice on the stat stuff, although I think the put_fsm has a stat for failed precommit/postcommit hooks you might want to look at? It is a similar issue, where if the precommit hook goes bad, it can spam the node, so we 'log' failures to a stat.

slfritchie:
@Vagabond and @evanmcc Thoughts on commit ded4f04?

Vagabond:
Passes the eyeball test for me.

evanmcc:
Looks good to me as well.  I'll find some time to test it over the coming
week.

evanmcc:
+1 @ 47acefd
Still don't really like that there's no indication which node is causing the error.  The rate is really low, at least in my setup.  Willing to let it go for now.

slfritchie:
The no-op function `late_put_fsm_coordinator_ack()` is there for that purpose: the folsom stats counters won't tell you which node, but dbg/redbug tracing of that function will tell you.

slfritchie:
@borshop +1 47acefd

slfritchie:
@borshop merge 47acefd

reiddraper:
@borshop merge

PR #915, opened 2014-04-17
Fixed erroring binary_to_exiting_atom

lordnull:
When attempting to do an add_all or remove_all on a set via the http
interface, there check for a valid verb was attempting to do
binary_to_exiting_atom/2 on the json binary verb. Becuase the atom version
of the verb was never used before, it was erroring it. The use of
binary_to_existing was redundant anyway as the guard expression for the
function clause the check was in already checked for valid content.

Before the fix: badarg error would probogate up, and the action would
fail. After fix: everything works as expected with the same guarentees.


cmeiklejohn:
I think this is missing some tags, and the 2.0-RC milestone as well.  I'll add the milestone, but it should be tagged accordingly, we don't want to miss it!

reiddraper:
Anyone already reviewing this? @kellymclaughlin?

kellymclaughlin:
> Anyone already reviewing this? @kellymclaughlin?

Yep, I'm on it

seancribbs:
This has already been fixed.

seancribbs:
c.f. #933

Issue #921, opened 2014-04-21
riak_kv_entropy_info dialyzer errors


reiddraper:
```
riak_kv_entropy_info.erl:78: Function exchange_complete/4 has no local return
riak_kv_entropy_info.erl:83: Function exchange_complete/5 has no local return
riak_kv_entropy_info.erl:84: The call riak_kv_entropy_info:update_index_info({_,_},{'exchange_complete',_,_,_}) will never return since it differs in the 2nd argument from the success typing arguments: ({_,_},{'tree_built',_})
riak_kv_entropy_info.erl:161: Record construction #index_info{exchanges::[]} violates the declared type of field repaired::pos_integer() and last_exchange::{non_neg_integer(),{non_neg_integer(),pos_integer()}}
riak_kv_entropy_info.erl:194: The call riak_kv_entropy_info:update_simple_stat(Repaired::'undefined' | pos_integer(),pos_integer()) will never return since it differs in the 2nd argument from the success typing arguments: (any(),'undefined' | #simple_stat{count::number(),sum::number()})
riak_kv_entropy_info.erl:208: Function compute_exchange_info/4 has no local return
riak_kv_entropy_info.erl:218: The call riak_kv_entropy_info:stat_tuple(Repaired::pos_integer()) will never return since it differs in the 1st argument from the success typing arguments: ('undefined' | #simple_stat{count::integer(),sum::integer()})
riak_kv_entropy_info.erl:225: Function update_simple_stat/2 has no local return
riak_kv_entropy_info.erl:225: The pattern <Value, 'undefined'> can never match the type <'undefined' | pos_integer(),pos_integer()>
riak_kv_entropy_info.erl:227: The pattern <Value, Stat = {'simple_stat', _, Min, Max, Cnt, Sum}> can never match the type <'undefined' | pos_integer(),pos_integer()>
riak_kv_entropy_info.erl:234: Function stat_tuple/1 has no local return
riak_kv_entropy_info.erl:234: The pattern 'undefined' can never match the type pos_integer()
riak_kv_entropy_info.erl:236: The pattern {'simple_stat', Last, Min, Max, Cnt, Sum} can never match the type pos_integer()
```

There were some domain-specific types I couldn't figure out here. These are also affecting several other modules, but I suspect that will be resolved once the errors are fixed here. I'm happy to help, but I also need someone with expertise in this module to take a look.

reiddraper:
Looks like most of this is best looked at by @jtuple or @rzezeski.

[...]

PR #924, opened 2014-04-22
Fix some AAE dialyzer errors

rzezeski:
This commit fixes all the dialyzer errors for `riak_kv_entropy_info`
as well as an error for `riak_kv_exchange_fsm`.  The main change was
to fix the spec for `#index_info.repaired` which is actually a
`#simple_stat{}` not a `pos_integer()`. However, even after that
dialyzer was not happy with the explicit default values of `undefined`
without corresponding entries in the type definition. This seemed like
a waste to me given that Erlang will implicitly default records values
to `undefined` and insert the appropriate type spec. This is discussed
in the following reference manual:

http://www.erlang.org/doc/reference_manual/typespec.html#typeinrecords

Here is the full list of dialyzer errors which have been fixed:

```
riak_kv_entropy_info.erl:78: Function exchange_complete/4 has no local
return
riak_kv_entropy_info.erl:83: Function exchange_complete/5 has no local
return
riak_kv_entropy_info.erl:84: The call
riak_kv_entropy_info:update_index_info({_,_},{'exchange_complete',_,_,_})
will never return since it differs in the 2nd argument fr
riak_kv_entropy_info.erl:161: Record construction
repaired::pos_integer() and last_exchange::{non_neg_inte
riak_kv_entropy_info.erl:194: The call
riak_kv_entropy_info:update_simple_stat(Repaired::'undefined' |
pos_integer(),pos_integer()) will never return since it differs in
riak_kv_entropy_info.erl:208: Function compute_exchange_info/4 has no
local return
riak_kv_entropy_info.erl:218: The call
riak_kv_entropy_info:stat_tuple(Repaired::pos_integer()) will never
return since it differs in the 1st argument from the success t
riak_kv_entropy_info.erl:225: Function update_simple_stat/2 has no local
return
riak_kv_entropy_info.erl:225: The pattern <Value, 'undefined'> can never
match the type <'undefined' | pos_integer(),pos_integer()>
riak_kv_entropy_info.erl:227: The pattern <Value, Stat = {'simple_stat',
_, Min, Max, Cnt, Sum}> can never match the type <'undefined' |
pos_integer(),pos_integer()>
riak_kv_entropy_info.erl:234: Function stat_tuple/1 has no local return
riak_kv_entropy_info.erl:234: The pattern 'undefined' can never match
the type pos_integer()
riak_kv_entropy_info.erl:236: The pattern {'simple_stat', Last, Min,
Max, Cnt, Sum} can never match the type pos_integer()
riak_kv_exchange_fsm.erl:379: Function exchange_complete/4 has no local
return
```

### DIFF
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -18,20 +18,6 @@ riak_core_pb.erl:201: The pattern <Binary, 'string'> can never match the type <_
 riak_core_pb.erl:221: The pattern <Types, [{Fnum, Bytes} | Tail], Acc> can never match the type <_,[],[{_,_}]>
 riak_kv_bitcask_backend.erl:600: Guard test Dir1TS::nonempty_string() == [] can never succeed
 riak_kv_bitcask_backend.erl:602: Guard test Dir2TS::nonempty_string() == [] can never succeed
-riak_kv_entropy_info.erl:78: Function exchange_complete/4 has no local return
-riak_kv_entropy_info.erl:83: Function exchange_complete/5 has no local return
-riak_kv_entropy_info.erl:84: The call riak_kv_entropy_info:update_index_info({_,_},{'exchange_complete',_,_,_}) will never return since it differs in the 2nd argument from the success typing arguments: ({_,_},{'tree_built',_})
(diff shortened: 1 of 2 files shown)


rzezeski:
This addresses #921.

/cc @reiddraper

reiddraper:
@borshop: retry yet again please and thank you

kellymclaughlin:
@borshop: retry

reiddraper:
@borshop merge por favor

PR #923, opened 2014-04-22
kv_vnode calling coverage_filter with 'all'

reiddraper:
In [this code](https://github.com/basho/riak_kv/blob/e66f3ffdc637d0e2385301f74d984c0632029493/src/riak_kv_vnode.erl#L766), kv_vnode calls riak_kv_coverage:build_filter with the atom `all` as the first argument. The type spec for this function expects a `binary()`. Digging through the call stack, I couldn't find any cases where we pattern match on an atom or specifically the atom `all`. Any ideas?

cc @kellymclaughlin, who I believe wrote this code.

### DIFF
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -64,11 +64,10 @@ riak_kv_vnode.erl:1189: The pattern <{Idx, Node}, Key, Obj> can never match the 
 riak_kv_vnode.erl:1229: Function do_backend_delete/3 has no local return
 riak_kv_vnode.erl:1655: Function result_fun/1 will never be called
 riak_kv_vnode.erl:1656: The created fun has no local return
-riak_kv_vnode.erl:1680: The pattern <_, _, _> can never match since previous clauses completely covered the type <'keys','undefined' | integer(),_>
-riak_kv_vnode.erl:1886: Function delete_from_hashtree/3 has no local return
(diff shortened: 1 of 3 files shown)


kellymclaughlin:
Yes, because the third parameter on that line you pointed out is `undefined` then only one of the first two guard expression of [this](https://github.com/basho/riak_kv/blob/e66f3ffdc637d0e2385301f74d984c0632029493/src/riak_kv_coverage_filter.erl#L60) `if` block will ever match in this case. So it's really just a junk parameter. You could change the spec or change the atom to be more clear. The best solution might be to add a `build_filter/2` function though this would require slightly more net changes than any other option.

reiddraper:
Aha, perfect. I should be able to make a PR to fix this now. Thanks.

reiddraper:
In [this particular call-site](https://github.com/basho/riak_kv/blob/e66f3ffdc637d0e2385301f74d984c0632029493/src/riak_kv_vnode.erl#L766), we could avoid even calling a function and simply use `ItemFilter`, or would that break too much of an abstraction?

reiddraper:
Blah, ignore that comment. I missed the call to `build_item_filter`, which we _could_ just call directly, were it exported.

reiddraper:
Just pushed cd46b2222f0db477f4c523a60588fa74e9c9ba99 to address this.

kellymclaughlin:
+1 cd46b2222f0db477f4c523a60588fa74e9c9ba99

reiddraper:
I'll retry this build once #937 is merged in, I think the tests should be green.

reiddraper:
@borshop merge please

PR #920, opened 2014-04-22
riak_core_security:check_permission/2 usage

reiddraper:
The types in `riak_core_security:check_permission/2` expect strings. But [we call it with binaries](https://github.com/basho/riak_kv/blob/e66f3ffdc637d0e2385301f74d984c0632029493/src/riak_kv_wm_counter.erl#L185). I can't tell if the `bucket/0` type in `riak_core_security` should be updated to accept binaries, or the code in riak_kv should be changed.

cc @Vagabond


reiddraper:
basho/riak_core#577 is merged now, so this can be reviewed whenever.

Vagabond:
+1 d83c43d

reiddraper:
@borshop merge

PR #925, opened 2014-04-23
Add Bitcask tombstone2 upgrade procedure

engelsanchez:
With the new tombstone changes in Bitcask (basho/bitcask#156), legacy tombtones will only be
removed on a merge when all files preceding the file containing the
tombstone has been merged already. Anticipating that many customers are
sensitive to disk not being reclaimed quickly enough, this change
triggers a migration that merges all Bitcask data files so that all the
legacy tombstones are completely removed.

At startup, the backend looks at the Bitcask app version and a version
written in the data directory, in version.txt.  If the pair require an
upgrade, an upgrade file is written to the directory and a regular check
is scheduled to periodically check on its progress. When it determines
it has finished, the new version file is written and the update file
deleted.

Upgrading to Bitcask 1.7.0 schedules a merge for all existing files by
writing a merge.txt with the list of files in the data directory.
Bitcask's need to merge (starting in 1.7.0) will read that and use the
list for merging, deleting the file when all files have been merged.
The file contains empty lines every 5 lines, so that merges happen to at
most 5 files at a time.

Also:

Remove old migration code from when Bitcask directories were
timestamped.

Add a cuttlefish configuration parameter for the Bitcask merge check
interval.

Add jitter to bitcask merge check interval to avoid lockstep merging.
bitcask.merge_check_jitter is given as a percentage of the merge check
interval. So if bitcask.merge_check_interval is set to 10m and
bitcask.merge_check_jitter is set to 30%, merges will be spaced between
7m and 13m apart.

Add bitcask.max_merge_size config to cap the total size of the input
files to merge. This can help further throttle a complete merge.
Setting it to 10GB will make sure each merge doesn't read more than that
amount of data. Setting it too low may prevent all merges!

Skip merge check if KV service is not yet up. We've had problems
recently with merges slowing things down when Riak is just starting up
and having bad side effects.

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -49,6 +49,12 @@
 -include_lib("bitcask/include/bitcask.hrl").
 
 -define(MERGE_CHECK_INTERVAL, timer:minutes(3)).
+-define(MERGE_CHECK_JITTER, 0.3).
+-define(UPGRADE_CHECK_INTERVAL, timer:seconds(10)).
+-define(UPGRADE_MERGE_BATCH_SIZE, 5).
+-define(UPGRADE_FILE, "upgrade.txt").
+-define(MERGE_FILE, "merge.txt").
+-define(VERSION_FILE, "version.txt").
 -define(API_VERSION, 1).
 -define(CAPABILITIES, [async_fold,size]).
 
@@ -67,6 +73,7 @@
 
 -type state() :: #state{}.
 -type config() :: [{atom(), term()}].
+-type version() :: {non_neg_integer(), non_neg_integer(), non_neg_integer()}.
 
 %% ===================================================================
 %% Public API
@@ -129,6 +136,7 @@ make_bk(1, Bucket, Key) ->
 %% @doc Start the bitcask backend
 -spec start(integer(), config()) -> {ok, state()} | {error, term()}.
 start(Partition, Config0) ->
+    random:seed(erlang:now()),
     {Config, KeyVsn} = 
         case app_helper:get_prop_or_env(small_keys, Config0, bitcask) of
             false -> 
@@ -151,12 +159,15 @@ start(Partition, Config0) ->
             PartitionStr = integer_to_list(Partition),
             case get_data_dir(DataRoot, PartitionStr) of
                 {ok, DataDir} ->
+                    BitcaskDir = filename:join(DataRoot, DataDir),
+                    UpgradeRet = maybe_start_upgrade(BitcaskDir),
                     BitcaskOpts = set_mode(read_write, Config),
-                    case bitcask:open(filename:join(DataRoot, DataDir), BitcaskOpts) of 
+                    case bitcask:open(BitcaskDir, BitcaskOpts) of
                         Ref when is_reference(Ref) ->
                             check_fcntl(),
                             schedule_merge(Ref),
                             maybe_schedule_sync(Ref),
+                            maybe_schedule_upgrade_check(Ref, UpgradeRet),
                             {ok, #state{ref=Ref,
                                         data_dir=DataDir,
                                         root=DataRoot,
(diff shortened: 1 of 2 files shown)


src/riak_kv_bitcask_backend.erl
```diff
@@ -429,15 +423,40 @@ callback(Ref,
...
+                               [version_to_str(Vsn), BitcaskRoot]),
+                    {ok, State};
+                {error, EndErr} ->
+                    lager:error("Finalizing backend upgrade : ~p", [EndErr]),
+                    {ok, State}
+            end;
+        pending ->
+            _ = schedule_upgrade_check(Ref),
+            {ok, State};
+        {error, Reason} ->
```
evanmcc, line 456: What is error going to mean here?  Try again later?  Shut the node down and do some manual cleanup?  It isn't clear to me that log-and-give-up is enough.

src/riak_kv_bitcask_backend.erl
```diff
@@ -713,6 +632,221 @@ set_mode(read_write, Config) ->
...
+            lager:error("Failed to check for bitcask files in ~s."
+                        " Can not determine if an upgrade is needed : ~p",
+                        [Dir, Err]),
+            no_upgrade
+    end.
+
+% @doc Start the upgrade process for the given old/new version pair.
+-spec start_upgrade(string(), version() | undefined, version()) ->
+    ok | {error, term()}.
+start_upgrade(Dir, OldVsn, NewVsn)
```
evanmcc, line 751: Is this logic unique to the way that the riak backend is using bitcask?  If not, shouldn't it be moved into bitcask so that non-riak applications don't have to reimplement it?

engelsanchez, line 751: The upgrade is mostly a Riak artifact. Partly because it involves the periodic merge check done entirely in the backend code. If the periodic merge mechanism was contained in Bitcask, I would have definitely put this along with it.  Honestly I think that's a better architecture, but... The upgrade (which is nothing more than a scheduled merge) is also not strictly necessary for other Bitcask apps. If you let it be, the old data will be merged away and old tombstones will eventually disapper now that we have the intermediate tombstone type.  The assumption here is that merging everything is better for Riak users sensitive to disk used up by tombstones not being freed soon enough. If we don't have to worry about that, really this whole PR could be canned.  But PMs were worried about that.

src/riak_kv_bitcask_backend.erl
engelsanchez, line 456: Good question. Currently the only error reason that could be returned here is "Ayo, wtf did you do with my upgrade file?.  That would most likely mean user shenanigans, so maybe the best thing to do would be to spell out in the error message (printed in line 798) something along the lines of "we expected to see upgrade.txt in the data directory, but it was removed. You may need to restart the node to retry the upgrade"

reiddraper:
@engelsanchez @evanmcc what's the status here? In need of review?

engelsanchez:
Review in process @reiddraper.  Bitcaskers have had partial availability in the last few days.

reiddraper:
@engelsanchez thanks, and no problem. I'm just going through all of the issues and asking. Would someone who is reviewing mind adding themselves as the assignee? @evanmcc?

slfritchie:
@evanmcc and @engelsanchez Is the review of this ticket finished and thus just waiting for https://github.com/basho/bitcask/pull/156 ?

engelsanchez:
No @slfritchie. Evan has been looking at it, but is also involved in a number of other fixes and reviews. If you have some cycles to join the review, it would help.  It seems that this no longer merges cleanly, so I will look into it and merge/rebase if necessary.

evanmcc:
I'll re-review and test this tomorrow.

evanmcc:
@engelsanchez ready to +1 this once there is a xref and dialyzer clean commit

engelsanchez:
Thanks @evanmcc. I hit the rebuild button on buildbot. The xref and dialyzer error was simply a new method in the Bitcask PR which has now been merged. Let's see what happens.

evanmcc:
+1 46f5f43 in that case

engelsanchez:
@borshop merge

PR #933, opened 2014-05-09
Use binary_to_atom/2 instead of binary_to_existing_atom/2.

seancribbs:
There's a possibility that the riak_dt code that contains the set operation
atoms hasn't been loaded when riak_kv_crdt_json:set_op_from_json/1 is invoked,
meaning binary_to_existing_atom/2 will raise a badarg. We are already checking 
the incoming operation as a binary via a guard on the relevant function
clause, so it is safe to use binary_to_atom/2 instead.


Vagabond:
+1 2bab35b Fixes the issue I found.

Issue #918, opened 2014-04-20
Deleting a Non-existent Key Fails over HTTP with Security On


hazen:
If I send the following request to `riak-2.0.0beta1`:

```
curl -X DELETE --ipv4 -v -H 'Authorization: Basic dGVzdHVzZXI6dGVzdHBhc3N3b3Jk' -H 'X-Riak-Vclock: a85hYGBgzGDKBVIcR4M2cgcHK7NkMCUy5rEyLF1//AxfFgA=' --cacert /Users/hazen/dev/clients/python/riak-python-client/ssl_tests/certs/cacert.pem https://localhost:10018/buckets/gzddqkrfwfgg/keys/cvufoswlysft
```

I get an internal server error:

```
<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request:<br><pre>[{webmachine_decision_core,'-decision/1-lc$^1/1-1-',
     [{error,
          {error,function_clause,
              [{riak_kv_wm_object,select_doc,
                   [{ctx,2,<<"default">>,<<"gzddqkrfwfgg">>,
                        <<"cvufoswlysft">>,
                        {riak_client,['dev1@127.0.0.1',undefined]},
                        default,default,default,default,default,default,
                        default,default,false,"riak",local,
                        {error,notfound},
                        undefined,undefined,undefined,undefined,'DELETE',
                        undefined,
                        {context,<<"testuser">>,
                            [{any,
                                 ["riak_core.get_bucket",
                                  "riak_core.get_bucket_type",
                                  "riak_core.set_bucket",
                                  "riak_core.set_bucket_type",
                                  "riak_kv.delete","riak_kv.get",
                                  "riak_kv.index","riak_kv.list_buckets",
                                  "riak_kv.list_keys","riak_kv.mapreduce",
                                  "riak_kv.put"]},
                             {<<"index">>,["search.admin","search.query"]},
                             {<<"schema">>,["search.admin","search.query"]}],
                            {1397,962670,771542}}}],
                   [{file,"src/riak_kv_wm_object.erl"},{line,928}]},
               {riak_kv_wm_object,content_types_provided,2,
                   [{file,"src/riak_kv_wm_object.erl"},{line,551}]},
               {webmachine_resource,resource_call,3,
                   [{file,"src/webmachine_resource.erl"},{line,186}]},
               {webmachine_resource,do,3,
                   [{file,"src/webmachine_resource.erl"},{line,142}]},
[...]

Issue #917, opened 2014-04-20
No Permission to Delete Bucket Props if Security is Enabled


hazen:
If I send the below message to **riak-2.0.0beta1**, I get an internal server error:

```
curl -X DELETE --ipv4 -v -H 'Authorization: Basic dGVzdHVzZXI6dGVzdHBhc3N3b3Jk' -H 'Content-Type: application/json' -H 'Accept: multipart/mixed, application/json, */*;q=0.5' --cacert /Users/hazen/dev/clients/python/riak-python-client/ssl_tests/certs/cacert.pem https://localhost:10018/buckets/sibsbucket/props

<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request:<br><pre>{error,{case_clause,'DELETE'},
       [{riak_kv_wm_props,forbidden,2,
                          [{file,"src/riak_kv_wm_props.erl"},{line,150}]},
        {webmachine_resource,resource_call,3,
                             [{file,"src/webmachine_resource.erl"},
                              {line,186}]},
        {webmachine_resource,do,3,
                             [{file,"src/webmachine_resource.erl"},
                              {line,142}]},
        {webmachine_decision_core,resource_call,1,
                                  [{file,"src/webmachine_decision_core.erl"},
                                   {line,48}]},
        {webmachine_decision_core,decision,1,
                                  [{file,"src/webmachine_decision_core.erl"},
                                   {line,221}]},
        {webmachine_decision_core,handle_request,2,
                                  [{file,"src/webmachine_decision_core.erl"},
                                   {line,33}]},
        {webmachine_mochiweb,loop,2,
                             [{file,"src/webmachine_mochiweb.erl"},{line,74}]},
        {mochiweb_http,parse_headers,5,
* Connection #0 to host localhost left intact
                       [{file,"src/mochiweb_http.erl"},{line,180}]}]}</pre><P><HR><ADDRESS>mochiweb+webmachine web server</ADDRESS></body></html>
```

I tested adding these two lines to `forbidden/2` in `src/riak_kv_wm_props.erl`

```
                'DELETE' ->
                    "riak_core.set_bucket"
```

which seems to fix the problem for this test case.

reiddraper:
Any updates on this? @javajolt @Vagabond @macintux?

reiddraper:
@macintux any chance you can take a look at this? If not, I can take a stab.

macintux:
[...]

PR #952, opened 2014-05-20
Fix security-enabled delete handling for nonexistent resources.

bowrocker:
Addresses issues:

https://github.com/basho/riak_kv/issues/918
https://github.com/basho/riak_kv/issues/917

### DIFF
--- a/src/riak_kv_wm_object.erl
+++ b/src/riak_kv_wm_object.erl
@@ -219,69 +219,61 @@ is_authorized(ReqData, Ctx) ->
                     "instead.">>, ReqData), Ctx}
     end.
 
-forbidden(RD, Ctx=#ctx{security=undefined}) ->
+-spec forbidden(#wm_reqdata{}, context()) -> term().
+forbidden(RD, Ctx) ->
     case riak_kv_wm_utils:is_forbidden(RD) of
         true ->
             {true, RD, Ctx};
         false ->
-            %% If it is a put or a post, don't require the object to already
-            %% exist, but the bucket type must exist.
-            case Ctx#ctx.method of
-                'POST' ->
-                    forbidden_check_bucket_type(RD, Ctx);
-                'PUT' ->
-                    forbidden_check_bucket_type(RD, Ctx);
-                _ ->
-                    forbidden_check_doc(RD, Ctx)
-            end
-    end;
-forbidden(RD, Ctx=#ctx{security=Security}) ->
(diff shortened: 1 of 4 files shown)


src/riak_kv_wm_object.erl
```diff
@@ -264,14 +264,15 @@ forbidden(RD, Ctx=#ctx{security=Security}) ->
                              unicode:characters_to_binary(Error, utf8, utf8),
                              RD1), Ctx};
                 {true, _} ->
-                    case Perm of
-                        "riak_kv.get" ->
+                    case (Perm == "riak_kv.get" orelse
```
macintux: Almost entirely inconsequential, but I've seen commentary that `=:=` should always be used except when `==` is desired to allow for integer/float comparisons

kellymclaughlin: I think either is ok here since we are definitely comparing two strings and the result will be the same.

kellymclaughlin:
I verified that this fix resolves the reported issues using the changes in [riak_test 612](https://github.com/basho/riak_test/pull/612). 

I suggested to Jon to also do a bit of code cleanup related to some of these changes so I will withhold  final approval until that is completed.

src/riak_kv_wm_object.erl
```diff
@@ -219,68 +219,61 @@ is_authorized(ReqData, Ctx) ->
...
+    %% we do this early as it used to be done in the
+    %% malformed check, so the rest of the resource
+    %% assumes that the key is present.
+    validate_doc(RD, Ctx);
+validate_resource(RD, Ctx, _Perm) ->
+    %% Ensure the bucket type exists, otherwise 404 early.
+    validate_bucket_type(RD, Ctx).
+
+-spec method_to_perm(atom()) -> string().
+method_to_perm(Method) when Method == 'POST' ->
```
kellymclaughlin: I would rewrite these function clauses to pattern match on the different method atoms in the parameter instead of using the guard to check. 

_e.g._ `method_to_perm('POST') ->`

kellymclaughlin: How about moving this function to `riak_kv_wm_utils` and using it here and in `riak_kv_wm_props`. There may be other places that case statement makes an appearance as well.

kellymclaughlin:
Refactoring looks good and dialyzer output is clean.

:+1:

bowrocker:
@borshop merge

kellymclaughlin:
+1 40b04a1

PR #953, opened 2014-05-21
Add basic Erlang-VM-style tracing utilities

slfritchie:
This is a cousin of the https://github.com/basho/bitcask/issues/51 request over in the Bitcask repo.

Add basic Erlang VM tracing utilities.  Their source is in the `priv/tracers` directory.  The `rebar.config.script` file was edited to kludge Rebar into recognizing & compiling those files -- an entry in `Makefile` isn't sufficient while compiling Riak or RiakEE.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -1,3 +1,4 @@
+CONFIG2 =
 case erlang:system_info(otp_release) =< "R15B01" of
     true  ->
         HashDefine = [{d,old_hash}],
@@ -8,4 +9,13 @@ case erlang:system_info(otp_release) =< "R15B01" of
                 CONFIG ++ [{erl_opts, HashDefine}]
         end;
     false -> CONFIG
-end.
+end,
+
+%% Add all of the Erlang source modules in the priv/tracers directory
+%% to rebar's list of stuff to compile.
+
+First = proplists:get_value(erl_first_files, CONFIG2, []),
+TracerSrc = filelib:wildcard("priv/tracers/*.erl"),
+%% io:format("Ts: ~p\n", [TracerSrc]),
+lists:keyreplace(erl_first_files, 1, CONFIG2, {erl_first_files,
+                                               First ++ TracerSrc}).
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -10,12 +10,4 @@ case erlang:system_info(otp_release) =< "R15B01" of
         end;
     false -> CONFIG
 end,
-
-%% Add all of the Erlang source modules in the priv/tracers directory
-%% to rebar's list of stuff to compile.
-
-First = proplists:get_value(erl_first_files, CONFIG2, []),
-TracerSrc = filelib:wildcard("priv/tracers/*.erl"),
-%% io:format("Ts: ~p\n", [TracerSrc]),
-lists:keyreplace(erl_first_files, 1, CONFIG2, {erl_first_files,
-                                               First ++ TracerSrc}).
+CONFIG2.
--- a/priv/tracers/tracer_accumulating_time.erl
+++ b/priv/tracers/tracer_accumulating_time.erl
@@ -1,0 +1,77 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2014 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+-module(tracer_accumulating_time).
+-compile(export_all).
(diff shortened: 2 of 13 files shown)


reiddraper:
Are we trying to get this in for 2.0? It's pretty late in the game for that. How much testing/production use have these scripts had? Any thoughts @jonmeredith?

evanmcc:
IMO it's better to have them there already to ease deployment pain.  I've
personally run at least the backend latency tracer on a prod cluster
without bad results.  I'm sure that scott has used more in CS bug-hunting.

slfritchie:
I'm not the original author of some of them, so I'm not 100% sure, but I believe that all of these tracers have seen use in customer environments.  I've recently edited some of them to clean up ugly cruft.  I may have introduced bugs?  I'd lightly tested all of the changes that I'd made.

I'm in agreement with @evanmcc that it's useful to have commonly used scripts bundled along with the Riak package.  (Avoiding the cross-repo sync problem is the reason that I didn't put them in both the 'riak' and 'riak_ee' repos.)  None of that code is going to be executed without explicit hackery at the command shell.

If it'd make people happier, I could try to figure out how to get the packaging goop to put them in a directory separate from the code loader's default paths, so that there'd be an extra `code:add_pathz("/path/to/somewhere")` prerequisite step before using them.

reiddraper:
Okay, well I have no objection to adding them, was just making sure we thought they were important enough to add in now, and it sounds like the answer is yes. I haven't reviewed the tracers specifically, but +1 otherwise.

slfritchie:
Any review on the tracers themselves?  The one change that I made to most that weren't already using them was adding the `timestamp` option to the tracing, so that the VM adds timestamps to each trace event (rather than having the trace event receiver process adding timestamps).  It's conceivable that that optimization might not be optimal for cases where the strictly-increasing time clock (used by `erlang:now()`) might cause problems with large-numbers-of-cores systems?

On the other hand, if the trace event consumer gets bogged down because it's doing too much work per event, then that's also a big problem (and bigger historically-speaking).  Don't trace!  Long live tracing!

slfritchie:
@reiddraper or @evanmcc Any additional comments?

rebar.config.script
```diff
@@ -8,4 +9,13 @@ case erlang:system_info(otp_release) =< "R15B01" of
                 CONFIG ++ [{erl_opts, HashDefine}]
         end;
     false -> CONFIG
-end.
+end,
+
+%% Add all of the Erlang source modules in the priv/tracers directory
+%% to rebar's list of stuff to compile.
+
```
evanmcc: Is there an advantage to doing this instead of adding the directory to `src_dirs`?

evanmcc:
These all look reasonable, I'll try to run each one briefly today.

cmeiklejohn:
This necessary for RC or not?

evanmcc:
@cmeiklejohn I dunno.  It'd be nice to have them there already, but this argument has been had above.  Maybe @jonmeredith wants to make a call.

@slfritchie I've run almost all of them and gotten sensible returns from basic tracing.  Several times I've been able to nudge the node into doing this:

```
** tracer received garbage: print_report
** tracer received garbage: print_report
** tracer received garbage: print_report
** tracer received garbage: print_report
** tracer received garbage: print_report
** tracer received garbage: print_report
```

but that doesn't look like something that's easily fixable.

There's some cleanup and common code elimination refactoring that could be done, but I think that can wait until later.  Other than the one comment about building, +1 to these changes.

rebar.config.script
slfritchie: I had no idea that `src_dirs` existed.  I'll remove the `rebar.config.script` hackery in favor of it, if I can get it to work.  Simply using:

```
{src_dirs, ["./priv/tracers", "./src"]}.
```

... as the first line in `rebar.config` doesn't appear to behave.  :-(  I'll twiddle a bit more to try to understand what Rebar is (not) doing.

slfritchie: I'm an idiot who needs to RTFM 3x minimum.

slfritchie:
Evan, the tracers themselves don't have anything resembling `tracer received garbage` in them.

Ah, but it does appear in `dbg.erl`.  Hooray for comments!  

```
    Other ->
        %%% Is this really a good idea?
        io:format(user,"** tracer received garbage: ~p~n", [Other]),
        recv_all_traces(Suspended0, Handler, Hdata, Traces)
```

AFAIK, the use of `timer:send_interval/2` by those tracers means that it's possible that the process that called `send_interval/2` might still be alive.  There's a chicken-and-egg problem for how to get the interval `reference()` thingie into the tracer's state.  Using something safe like `exit(self(), please)` on the CLI can finish the job that the tracer's `stop()` function can't always do 100% reliably.

evanmcc:
Scott, what I would see is the tracer getting overloaded somehow, returning
that on the report tick, then somehow righting itself after 10 or 12
seconds, printing out some old stuff rapidly, then behaving normally
again.  Sounds like some sort of queue backup to me, but I didn't look into
it.  I can reproduce it on command, if needed, I am pretty sure.

jonmeredith:
As long as without any user action for the console they do no harm, we can merge them.

reiddraper:
oh, @evanmcc commented in hipchat that this still might be missing a commit. Maybe re: the rebar config change? I'll hold off on asking bors to merge until then.

slfritchie:
Things work a heck of a lot better when the code is committed and pushed, sorry.

evanmcc:
+1 456b0b0

reiddraper:
I think we're ready for a borshop merge here ,yeah?

evanmcc:
imagine so

reiddraper:
@borshop merge please and thank you

Issue #875, opened 2014-03-23
PB List Keys to missing bucket type causes crash, hangs client,


broach:
If you issue a list keys command via PB, and the type specified doesn't exist, the following appears in the console.log:

2014-03-22 22:11:30.620 [error] <0.24921.0> CRASH REPORT Process <0.24921.0> with 0 neighbours exited with reason: no function clause matching proplists:get_value(n_val, {error,no_type}, undefined) line 225 in gen_fsm:init_it/6 line 379

2014-03-22 22:11:30.620 [error] <0.355.0> Supervisor riak_kv_keys_fsm_sup had child undefined started with {riak_core_coverage_fsm,start_link,undefined} at <0.24921.0> exit with reason no function clause matching proplists:get_value(n_val, {error,no_type}, undefined) line 225 in context child_terminated

In addition, the client is the hung because it's waiting for a reply that never comes.

jrwest:
moved to `2.0-RC`. we won't hold up beta for it, but this is a real issue we need to fix.

kellymclaughlin:
I'll take this. It seems like an obvious case to add to the riak_test to verify so I'll do that in tandem.

jburwell:
Implemented a fix and associated riak_tests.  During the course of testing, discovered that listing buckets for an undefined bucket type was also not operating as expected.  The following are the fix PRs:
- riak_core PR: https://github.com/basho/riak_core/pull/594
- riak-kv PR: https://github.com/basho/riak_kv/pull/958
- riak-test PR: https://github.com/basho/riak_test/pull/616

jburwell:
Solution proposed by @kellymclaughlin to reduce risk to the 2.0 risk has been submitted in the following PRs:
- riak_kv: https://github.com/basho/riak_kv/pull/970
- riak_test: https://github.com/basho/riak_test/pull/627

I have also opened [riak_kv ticket 965](https://github.com/basho/riak_kv/issues/965) to consider pushing these checks down into the associated FSMs during the 2.1 cycle.

PR #955, opened 2014-05-23
Fix PB List Keys to missing bucket type causes crash, hangs client

jburwell:
Per defect #875, list keys and buckets for an undefined bucket type failed to operate as expected.  Listing keys for an undefined bucket type caused a client hang while listing buckets for an undefined bucket type yielded an {ok, []}.  With this patch, both scenarios return {error, no_type}, and do not hang the client waiting for a reply.

This PR requires modifications to the riak_core_coverage_fsm provided by https://github.com/basho/riak_core/pull/594.

### DIFF
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -84,3 +84,4 @@ Unknown types:
   riak_core_apl:preflist2/0
   riak_dt:operation/0
   riak_dt:value/0
+riak_kv_wm_utils.erl:289: The pattern {Scheme, _, Host, Port, _, _} can never match the type {'error','no_scheme' | {_,atom(),_}}
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -84,4 +84,4 @@ Unknown types:
   riak_core_apl:preflist2/0
   riak_dt:operation/0
   riak_dt:value/0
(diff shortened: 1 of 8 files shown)


seancribbs:
I don't see why this needs to touch the client/fsm modules. The client API should do the validation.

reiddraper:
Is this 2.0 work? We should apply the 2.0-RC milestone, if so.

reiddraper:
If we go forward with this approach, kindly squash down to a commit or two, please.

kellymclaughlin:
@seancribbs Could you elaborate on your comment? Are you thinking something similar to the solution in https://github.com/basho/riak_api/pull/62?

seancribbs:
@kellymclaughlin Yes, exactly. This is client validation and should be handled before even trying to start the FSM. 

I'm also concerned about unnecessarily removing explicit matches (e.g. [here](https://github.com/basho/riak_kv/pull/955/files#diff-3ac0e13ed171d03a458be62c38f1c065R653)).

reiddraper:
Suppose we should wait for @jburwell to chime in.

jburwell:
In order to properly squash commits, a new branch was required.  Therefore, PR https://github.com/basho/riak_kv/pull/958 was created a replacement.

Issue #893, opened 2014-04-07
fix validation error output in riak-admin bucket-type commands


jrwest:
example speaks for itself I think but to clarify, its not user friendly or easily readable:

```
$ riak-admin bucket-type create any '{"props": {"n_val": "abc", "consistent": "abc"}}'
Error creating bucket type any: [{n_val,not_integer},
                                 {consistent,
                                     "<<\"abc\">> is not a valid value for consistent. use \"true\" or \"false\""},
                                 {reserved_name,
                                     "The name 'any' may not be used for bucket types"}]
```

reiddraper:
Anyone working on this?

macintux:
I wrote similar error message cleanup code for `riak_core_console` IIRC; I'll take a look.

Here are the changes I applied to core; will take the same approach here.

https://github.com/basho/riak_core/pull/546/files

macintux:
Digging through code looking for error tuples. I have no confidence that this is a comprehensive list, and it's possible to define additional validators in the Riak configuration, so the solution will have to be flexible with what it accepts.

``` erlang
{active, "Cannot modify active property"}
{reserved_name, "The name 'any'..."}

{claimant, "No claimant details found in existing properties"}
{claimant, "Cannot modify claimant property"}
{claimant, "Cannot update claimant property"}
{claimant, "Invalid claimant"}

{IntProp, not_integer}
{QProp, not_valid_quorum}

{consistent, "cannot update consistent property"}
{consistent, "~p is not a valid value for consistent. use \"true\" or \"false\""}

{BoolProp, not_boolean}
```

jrwest:
all of the tuples should live in one of the registered bucket validtors + `riak_core_bucket_props`. I agree we need a general sol'n either way -- if someone adds a tuple but doesn't change the output, things should not break.

I think it would be fine to have something like:

```
do_print({IntProp, not_integer}) ->
   ....;
do_print({claimant, Error}) ->
   ....;
do_print({X,Y}) ->
  io:format("~p: ~p~n", [X, Y]).
```

macintux:
Yeah, that's pretty much what I defined for security errors.

PR #956, opened 2014-05-27
Cleaner output for bucket type creation errors

macintux:
Could probably be further enhanced, but handles the problem reported by @jrwest in #893 much more gracefully.

```
$ ./riak-admin bucket-type create foo '{"props": {"r": "fred", "n_val": "abc", "consistent": "abc", "allow_mult": "x"}}'
Error creating bucket type foo:
r must be an integer or (one|quorum|all)
n_val must be an integer
allow_mult should be true or false
<<"abc">> is not a valid value for consistent. Use "true" or "false"
```

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -268,7 +268,7 @@ validate_create_consistent_props(false, New) ->
 validate_create_consistent_props(undefined, New) ->
     {New, [], []};
 validate_create_consistent_props(Invalid, New) ->
-    Err = lists:flatten(io_lib:format("~p is not a valid value for consistent. use \"true\" or \"false\"", [Invalid])),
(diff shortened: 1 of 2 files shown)


jrwest:
`bucket-type update` has the same issue:

```
$ dev/dev1/bin/riak-admin bucket-type update abc '{"props": {"n_val": "def", "allow_mult": "yaya"}}' 
Error updating bucket type abc: [{allow_mult,not_boolean},{n_val,not_integer}]
```

macintux:
Noted, will revise and extend.

macintux:
I'd made a mental note to update this function too, sigh. Anyway, updated, ran more tests, looked for any other functions that might need this.

jrwest:
+1 ed9a17a

macintux:
@borshop merge

PR #940, opened 2014-05-12
Fix mem backend to honor various settings correctly.

evanmcc:
@angrycub is testing.

test grid spreadsheet (basho-only):
https://docs.google.com/a/basho.com/spreadsheet/ccc?key=0AkSRjImmlw_HdG5TeVZ2dFRtUFpiZjI2YThQRHFGRWc#gid=0

Without this patch, the memory backend can 'run out of space' in that all of its memory is marked as used, while there is plenty of memory available.  Expiry can do similar things.  This patch makes sure that the state tuple is preserved through all vnode actions, as well.

This should be ported to 2.0 prior to the RC.  

cc @reiddraper re #717

PR #957, opened 2014-05-28
Bugfix/2i refresh mod state

engelsanchez:
The signature of do_get_term changed with the fix to the memory backend
TTL and memory settings. The 2i rehash command was failing due to an
outdated case statement due to that change.

This was caught by the verify_2i_aae test, which can be used to verify this change.

_NOTE_: When the memory backend fix makes it to 2.0, this change needs to be applied there too.


jaredmorrow:
Just to link the two together, this code was broken in https://github.com/basho/riak_kv/pull/940

evanmcc:
+1, code looks good.  there don't seem to be any additional uses of `do_get_term` to cover, and it makes the test pass.

PR #958, opened 2014-05-28
Fix PB List Keys to missing bucket type causes crash, hangs client

jburwell:
Per defect #875, list keys and buckets for an undefined bucket type failed to operate as expected.  Listing keys for an undefined bucket type caused a client hang while listing buckets for an undefined bucket type yielded an {ok, []}.  With this patch, both scenarios return {error, no_type}, and do not hang the client waiting for a reply.

This PR requires modifications to the riak_core_coverage_fsm provided by https://github.com/basho/riak_core/pull/594.

In order to properly squash the commits for this fix, a new branch was created -- requiring a new PR.  Therefore, this PR supersedes https://github.com/basho/riak_kv/pull/955.

/cc @kellymclaughlin @reiddraper @seancribbs

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -650,11 +650,11 @@ stream_list_buckets(Filter, Timeout, Type,
 stream_list_buckets(Filter, Timeout, Client, Type,
                     {?MODULE, [Node, _ClientId]}) ->
     ReqId = mk_reqid(),
-    {ok, _Pid} = riak_kv_buckets_fsm_sup:start_buckets_fsm(Node,
-                                                           [{raw, ReqId,
-                                                             Client},
-                                                            [Filter, Timeout,
-                                                             true, Type]]),
+    riak_kv_buckets_fsm_sup:start_buckets_fsm(Node,
+                                              [{raw, ReqId,
+                                                Client},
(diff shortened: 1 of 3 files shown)


reiddraper:
Assigning to 2.0-RC milestone.

src/riak_client.erl
```diff
@@ -650,11 +650,11 @@ stream_list_buckets(Filter, Timeout, Type,
 stream_list_buckets(Filter, Timeout, Client, Type,
                     {?MODULE, [Node, _ClientId]}) ->
     ReqId = mk_reqid(),
-    {ok, _Pid} = riak_kv_buckets_fsm_sup:start_buckets_fsm(Node,
-                                                           [{raw, ReqId,
-                                                             Client},
-                                                            [Filter, Timeout,
-                                                             true, Type]]),
+    riak_kv_buckets_fsm_sup:start_buckets_fsm(Node,
```
seancribbs, line 653: Removing the match here is bad practice and will introduce dialyzer warnings.

jburwell:
Superseded by client only implementation in [966](https://github.com/basho/riak_kv/pull/966).

Issue #846, opened 2014-02-18
AAE + ownership change race = 5 minute AAE stall


jtuple:
If the remote tree in an AAE exchange is for a partition that is changing ownership, it's possible for the exchange FSM to stall when trying to grab the remote lock, eventually timing out. As the default AAE timeout is 5 minutes, this is a noticeable stall.

Technically, this is "working as designed". The timeout triggers, the exchange fails, and the world moves on.

The main issue is the strong consistency case, which needs to perform AAE exchanges to move peers into the trusted state (when running in paranoid mode). If this race is hit, the stall can prevent an ensemble from becoming available for ~5 minutes, leaving that portion of the key space unavailable to client requests.

It's unclear how easy it is to fix this scenario, but I wanted to open an issue to document/record this issue.

The race is straightforward. The `riak_kv_exchange_fsm` calls `start_exchange_remote` on the `riak_kv_entropy_manager`. The manager handles the request, and sees that there is still a valid hashtree pid for the given tree index, gets the local concurrency lock, then forwards the request to the pid (via a cast) to get a tree lock. However, the ownership transition completes concurrently causing the hashtree pid to either already be dead or be in the process of shutting down -- either way, ignoring the message. The exchange then stalls until the timeout triggers.

Again, working as designed. But, we can probably do better.

---

The user-facing observed behavior is `riak-admin aae-status` showing missing (or highly delayed) exchange time for the partition that is going to timeout. And, ultimately, a message in the log acknowledging the timeout.

As an example.

Note the missing time for the All column for partition 3653...:

```
================================== Exchanges ==================================
Index                                              Last (ago)    All (ago)   
-------------------------------------------------------------------------------
0                                                  2.7 min       2.9 min     
365375409332725729550921208179070754913983135744   3.0 min       --          
730750818665451459101842416358141509827966271488   2.8 min       2.8 min     
1004782375664995756265033322492444576013453623296  2.7 min       2.8 min     
1096126227998177188652763624537212264741949407232  2.7 min       2.8 min     
1370157784997721485815954530671515330927436759040  2.8 min       2.9 min     
```

Eventually, we timeout:

```
[info] <0.1169.0>@riak_kv_exchange_fsm:do_timeout:386 Timeout during
exchange between
(local)
{365375409332725729550921208179070754913983135744,'dev1@127.0.0.1'} and
(remote)
{548063113999088594326381812268606132370974703616,'dev1@127.0.0.1'},
(preflist)
[...]

PR #971, opened 2014-06-04
Improve on 5 min AAE+ownership change stall

engelsanchez:
This is a proposed improvement to address basho/riak_kv#846

The hard coded default has been changed from 5 minutes to 1 minute.
Also, usage of the FSM timeout mechanism that could lead to the FSM
ending up stuck forever has been changed to gen_fsm:send_event_after

### DIFF
--- a/src/riak_kv_exchange_fsm.erl
+++ b/src/riak_kv_exchange_fsm.erl
@@ -43,11 +43,12 @@
                 local_tree  :: pid(),
                 remote_tree :: pid(),
                 built       :: non_neg_integer(),
+                timer       :: reference(),
                 timeout     :: pos_integer()
                }).
 
 %% Per state transition timeout used by certain transitions
--define(DEFAULT_ACTION_TIMEOUT, 300000). %% 5 minutes
+-define(DEFAULT_ACTION_TIMEOUT, 60000). %% 1 minute
 
 %% Use occasional calls to disk_log:log() for some backpressure periodically
 -define(LOG_BATCH_SIZE, 5000).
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
I'll take review of this one.

jtuple:
+1 95b076e9c5e827fef4ae862a255c0725198e99d9

I think we should consider making this timeout configurable (advanced.config only). However, a 1 minute hardcoded timeout seems entirely safe, and is a significant improvement over the 5 minute timeout.

I agree with the changes to using explicit timers rather state transition timeouts. I'm certain the way AAE is written/used today there is no issue, but inactivity timeouts are extremely brittle and easy for us to unknowingly break in future unrelated changes. So, +1 on that change.

We don't have a test for this specific condition, and right now isn't the time to invest in it. However, we should consider extending existing AAE tests to cover the timeout case in the future.

In any case, I ran the existing set of AAE `riak_tests` against this change, as well as manually tested the timeout by suspending several `riak_kv_index_hashtree` processes using `erlang:suspend_process`. The timeout works as expected, and shows up after 1 minute. So, all good.

I think this is good enough for now, at least for this current pull-request. Although, adding an explicit test, making the timeout configurable, and properly fixing the underlying issue behind basho/riak_kv#846 are all reasonable future work items.

engelsanchez:
Thanks Joe. Notice that it is already configurable and I'm just changing the hard coded default: riak_kv,anti_entropy_timeout.

engelsanchez:
@borshop merge

PR #972, opened 2014-06-04
Rename security permission from riak_search.query to search.query

hazen:
The current name of the permission is `search.query`
See https://github.com/basho/yokozuna/blob/develop/include/yokozuna.hrl#L148


seancribbs:
:+1: 512371c

reiddraper:
Ready to merge?

hazen:
@borshop merge

PR #973, opened 2014-06-05
Resolves GH #859 by setting tmpdir for sorting AAE disk logs.

### LINKED ISSUE #859: Temporary Files Being Created in Application Directory
I'd been wondering why I'd been getting so many writes and reads to my USB root disk, and I'd tracked down the culprit.

In my setup, I have riak set up in `/opt/riak` -- `/opt/riak/log`, `/opt/riak/data`, `/opt/riak/etc` are symlinks to a data partition, ensuring that about all dynamic changes are writtten to a data disk.

Turns out that there are temporary files being written to `/opt/riak` -- this should really be either in `/tmp` or in `/opt/riak` if more persistent.

```
fs_riak@altostratus.fabric_47889_1393_795878_730984.17
fs_riak@altostratus.fabric_47889_1393_795878_730984.34
fs_riak@altostratus.fabric_47889_1393_795878_730984.51
fs_riak@altostratus.fabric_47889_1393_795878_730984.52
fs_riak@altostratus.fabric_47889_1393_795878_730984.53
fs_riak@altostratus.fabric_47889_1393_795878_730984.54
```

jonmeredith:
Break out a function for determining the temporary directory used by AAE and ensuring it's creation.  Pass the output from the new tmp_dir() function to file_sorter.  Behavior around file_sorter crashing is still the same as before (in an unchecked try), but decided that was acceptable - if it's due to perms problems on the tmpdir the process will crash in key_exchange instead.

Resolves #859

### DIFF
--- a/src/riak_kv_exchange_fsm.erl
+++ b/src/riak_kv_exchange_fsm.erl
@@ -179,14 +179,12 @@ key_exchange(timeout, State=#state{local=LocalVN,
     lager:debug("Starting key exchange between ~p and ~p", [LocalVN, RemoteVN]),
     lager:debug("Exchanging hashes for preflist ~p", [IndexN]),
 
-    TmpDir = app_helper:get_env(riak_core, platform_data_dir, "/tmp"),
+    TmpDir = tmp_dir(),
     {NA, NB, NC} = Now = WriteLog = now(),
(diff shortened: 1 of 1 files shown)


russelldb:
:+1: 2336457

russelldb:
@borshop merge

PR #976, opened 2014-06-09
Fake Bitcask version for tests

engelsanchez:
If the application has not started, it uses a high number so it works
and does not attempt to trigger an upgrade.

This should fix some failing tests crashing trying to figure out the Bitcask version.

/cc @reiddraper


slfritchie:
+1 nice hack

reiddraper:
@borshop merge

PR #977, opened 2014-06-10
Fix reip loading of riak_core

### LINKED ISSUE #554: riak-admin reip not working in riak-2.0.0beta6
Reip, when run with the node down, is throwing the below error:

```
{badmatch,{error,{already_loaded,riak_core}}}
```

riak_kv_console:reip/1 attempts to load riak_core here: https://github.com/basho/riak_kv/blob/develop/src/riak_kv_console.erl#L219

coderoshi:
Fixes basho/riak#554, reip throwing an a badmatch when attempting to load riak_core, which was already loaded by cuttlefish.

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -216,7 +216,11 @@ reip([OldNode, NewNode]) ->
         %%
         %% Do *not* convert to use riak_core_ring_manager:ring_trans.
         %%
-        ok = application:load(riak_core),
+        case application:load(riak_core) of
+            %% a process, such as cuttlefish, may have already loaded riak_core
+            {error,{already_loaded,riak_core}} -> ok;
+            ok -> ok
+        end,
         RingStateDir = app_helper:get_env(riak_core, ring_state_dir),
         {ok, RingFile} = riak_core_ring_manager:find_latest_ringfile(),
         BackupFN = filename:join([RingStateDir, filename:basename(RingFile)++".BAK"]),
@@ -229,9 +233,7 @@ reip([OldNode, NewNode]) ->
             [element(2, riak_core_ring_manager:find_latest_ringfile())])
     catch
         Exception:Reason ->
-            lager:error("Reip failed ~p:~p", [Exception,
-                    Reason]),
-            io:format("Reip failed, see log for details~n"),
+            io:format("Reip failed ~p:~p", [Exception, Reason]),
             error
     end.
 


src/riak_kv_console.erl
```diff
@@ -216,7 +216,11 @@ reip([OldNode, NewNode]) ->
         %%
         %% Do *not* convert to use riak_core_ring_manager:ring_trans.
         %%
-        ok = application:load(riak_core),
+        case application:load(riak_core) of
+            %% a process, such as cuttlefish, may have already loaded riak_core
+            {error,{already_loaded,riak_core}} -> ok;
+            ok -> ok
```
jburwell, line 222: As I understand it, we can safely ignore {error, {already_loaded, riak_core}}.  However, the  [application:load/1](http://www.erlang.org/doc/apps/kernel/application.html#load-1) function specifies {error, Reason }.  How should {error, _} be handled?

seancribbs:
Wasn't there a decision to remove reip for 2.0? "cluster replace" does the same thing, but correctly.

jonmeredith:
We decided not to - the case it doesn't handle is a restore from backup
where you have no original nodes live and need to replace node names before
you can restart.

On Tue, Jun 10, 2014 at 10:18 PM, Sean Cribbs notifications@github.com
wrote:

> Wasn't there a decision to remove reip for 2.0? "cluster replace" does the
> same thing, but correctly.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/basho/riak_kv/pull/977#issuecomment-45673264.

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

src/riak_kv_console.erl
coderoshi, line 222: I'm going to allow any other error to just break and bubble to the surface. Adding the {error, _} case gives us two options: set `ok` and continue, which would just push the problem to line 225. Or, throw a different error, which is immediately caught by this function anyway.

jburwell, line 222: @coderoshi seems reasonable to me.  +1

jcapricebasho:
@coderoshi When testing I noticed errors weren't bubbling up to the console.  With reip being run when the node is down, the lager:error call in this catch won't be written: https://github.com/basho/riak_kv/blob/bug/er/reip-fix/src/riak_kv_console.erl#L236.  Can this be changed to a single io:format, outputting the exception and reason, as part of this PR?

coderoshi:
Agreed @jcapricebasho. Also, rebased

jcapricebasho:
Just for record, this is reproducible in at least 2.0.0beta6, which is where I found the issue.  It can be reproduced by building a devrel, stopping a node, and attempting to run a `riak-admin reip`.  Based on the most recent commit from Eric, the badmatch from ok = application:load(riak_core) actually returning  {error, {already_loaded, riak_core}} is not visible in the console.

jburwell:
Code looks good, and test passed on my laptop. +1.

coderoshi:
@jburwell want to +1 for bors? `+1 574f42a`

jburwell:
+1 574f42a

Sorry for the n00b mistake.

coderoshi:
@borshop merge

PR #979, opened 2014-06-11
Add console commands to inspect ensemble system

jtuple:
This pull-request adds commands to inspect the ensemble/consensus sub-system to `riak_kv_console` (which can be use by `riak-admin`).

riak_kv_console:ensemble_status([]) prints an overview of the consensus system, including an overview of all running ensembles.

riak_kv_console:ensemble_status([Label) prints details about a specific ensemble, with Label corresponding to the printed label for that ensemble in the overview output.

Examples of command output in various scenarios can be printed by running riak_kv_ensemble_console:print_variations().

Sibling pull-requests: basho/riak_ensemble#26, basho/riak#558, basho/riak_core#602

Addresses issues: basho/riak_ensemble#9

---

Here's some example output (the same as printed by `print_variations`) for the overview command.

Output when strong consistency/consensus is not enabled in `riak.conf / app.config`:

```
============================== Consensus System ===============================
Enabled:     false
Active:      false
Ring Ready:  true
Paranoia:    medium (AAE syncing required)
AAE enabled: false

Note: The consensus subsystem is not enabled.

================================== Ensembles ==================================
There are no active ensembles.
```

Output when consensus is enabled, but has not yet been activated because there are too few nodes in the cluster:

```
============================== Consensus System ===============================
Enabled:     true
Active:      false
Ring Ready:  true
Paranoia:    medium (AAE syncing required)
AAE enabled: false

Note: The consensus subsystem will not be activated until there are more
      than three nodes in this cluster.

================================== Ensembles ==================================
There are no active ensembles.
```

Output when both configured and active, but when AAE is disabled when paranoia mode is medium or higher:

```
============================== Consensus System ===============================
Enabled:     true
Active:      false
Ring Ready:  true
Paranoia:    medium (AAE syncing required)
AAE enabled: false

Warning: Paranoia level is medium, but AAE is not enabled. Ensembles will be
         unable to become trusted and will never reach quorum.

================================== Ensembles ==================================
There are no active ensembles.
```

Output during normal operation:

```
============================== Consensus System ===============================
Enabled:     true
Active:      true
Ring Ready:  true
Paranoia:    medium (AAE syncing required)
AAE enabled: true

================================== Ensembles ==================================
 Ensemble     Quorum        Nodes      Leader
-------------------------------------------------------------------------------
    1         2 / 3         2 / 3      dev1@127.0.0.1
    2         0 / 3         2 / 3      --
```

Output during a view change (eg. ensemble is transitioning between two separate memberships):

```
============================== Consensus System ===============================
Enabled:     true
Active:      true
Ring Ready:  true
Paranoia:    medium (AAE syncing required)
AAE enabled: true

================================== Ensembles ==================================
 Ensemble     Quorum        Nodes      Leader
-------------------------------------------------------------------------------
    1         2 / 3         2 / 3      dev1@127.0.0.1
              2 / 3         3 / 3  
    2         0 / 3         2 / 3      --
```

---

Here's some example output (the same as printed by `print_variations`) for the ensemble-specific command.

Output under normal conditions:

```
================================= Ensemble #1 =================================
Id:       test
Leader:   dev1@127.0.0.1

==================================== Peers ====================================
 Peer  Status     Trusted          Epoch         Node
-------------------------------------------------------------------------------
  1     leading     yes             1            dev1@127.0.0.1
  2    following    yes             1            dev2@127.0.0.1
  3    following    yes             1            dev3@127.0.0.1
```

Output when a node is offline:

```
================================= Ensemble #1 =================================
Id:       test
Leader:   dev1@127.0.0.1

==================================== Peers ====================================
 Peer  Status     Trusted          Epoch         Node
-------------------------------------------------------------------------------
  1     leading     yes             1            dev1@127.0.0.1
  2    (offline)    --              --           dev2@127.0.0.1
  3    following    yes             1            dev3@127.0.0.1
```

Output during a view change:

```
================================= Ensemble #1 =================================
Id:       test
Leader:   dev1@127.0.0.1

==================================== Peers ====================================
 Peer  Status     Trusted          Epoch         Node
-------------------------------------------------------------------------------
  1     leading     yes             1            dev1@127.0.0.1
  2    following    yes             1            dev2@127.0.0.1
  3    following    yes             1            dev3@127.0.0.1
-------------------------------------------------------------------------------
  1     leading     yes             1            dev1@127.0.0.1
  2    following    yes             1            dev2@127.0.0.1
  4    following    yes             1            dev4@127.0.0.1
```

### DIFF
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -56,7 +56,7 @@ cluster_info:dump_all_connected/1
 cluster_info:dump_nodes/2
 cluster_info:format/3
 cluster_info:register_app/1
-riak_kv_console.erl:152: The pattern {'error', 'legacy_mode'} can never match the type {'error','is_up' | 'not_member' | 'only_member'}
+riak_kv_console.erl:154: The pattern {'error', 'legacy_mode'} can never match the type {'error','is_up' | 'not_member' | 'only_member'}
 # Callback info not available
 Callback info about the riak_pipe_vnode_worker behaviour is not available
 Callback info about the riak_core_vnode_worker behaviour is not available
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -705,8 +705,13 @@ ensemble_status([]) ->
 ensemble_status(["root"]) ->
     riak_kv_ensemble_console:ensemble_detail(root);
 ensemble_status([Str]) ->
-    N = list_to_integer(Str),
-    riak_kv_ensemble_console:ensemble_detail(N).
+    N = parse_int(Str),
+    case N of
+        undefined ->
+            io:format("No such ensemble: ~s~n", [Str]);
+        _ ->
+            riak_kv_ensemble_console:ensemble_detail(N)
+    end.
 
 %%%===================================================================
 %%% Private
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -46,6 +46,8 @@
          bucket_type_reset/1,
          bucket_type_list/1]).
 
+-export([ensemble_status/1]).
+
 %% Reused by Yokozuna for printing AAE status.
 -export([aae_exchange_status/1,
          aae_repair_status/1,
@@ -697,6 +699,14 @@ repair_2i(Args) ->
                       "node are repaired\n", []),
             error
     end.
+
+ensemble_status([]) ->
+    riak_kv_ensemble_console:ensemble_overview();
+ensemble_status(["root"]) ->
+    riak_kv_ensemble_console:ensemble_detail(root);
+ensemble_status([Str]) ->
+    N = list_to_integer(Str),
+    riak_kv_ensemble_console:ensemble_detail(N).
 
 %%%===================================================================
 %%% Private
--- a/src/riak_kv_ensemble_console.erl
+++ b/src/riak_kv_ensemble_console.erl
@@ -238,11 +238,16 @@ format_info(Label, Peer, PeerInfo) ->
 
 get_details() ->
     {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
-    #details{enabled     = riak_core_sup:ensembles_enabled(),
+    Enabled = riak_core_sup:ensembles_enabled(),
+    Ensembles = case Enabled of
+                    true -> ordered_ensembles();
+                    _ -> []
+                end,
+    #details{enabled     = Enabled,
              active      = riak_ensemble_manager:enabled(),
              paranoia    = medium,
              aae_enabled = riak_kv_entropy_manager:enabled(),
-             ensembles   = ordered_ensembles(),
+             ensembles   = Ensembles,
              quorums     = [],
              nodes       = riak_core_ring:ready_members(Ring),
              ring_ready  = riak_core_ring:ring_ready(Ring)}.
--- a/src/riak_kv_ensemble_console.erl
+++ b/src/riak_kv_ensemble_console.erl
@@ -1,0 +1,488 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2014 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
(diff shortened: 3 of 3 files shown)


engelsanchez:
I'm taking the review for this and related PRs

engelsanchez:
Added EE version in basho/riak_ee#254

engelsanchez:
On a cluster with strong consistency off, you get this ugly response. It would be nice to clean up:

```
$ bin/riak-admin ensemble-status
                                 {badarg,
                                  [{ets,lookup_element,
                                    [riak_ensemble_manager,cluster_state,2],
                                    []},
                                   {riak_ensemble_manager,get_cluster_state,
                                    0,
                                    [{file,"src/riak_ensemble_manager.erl"},
                                     {line,198}]},
                                   {riak_kv_ensemble_console,all_ensembles,0,
                                    [{file,"src/riak_kv_ensemble_console.erl"},
                                     {line,266}]},
                                   {riak_kv_ensemble_console,
                                    ordered_ensembles,0,
                                    [{file,"src/riak_kv_ensemble_console.erl"},
                                     {line,252}]},
                                   {riak_kv_ensemble_console,get_details,0,
                                    [{file,"src/riak_kv_ensemble_console.erl"},
                                     {line,245}]},
                                   {riak_kv_ensemble_console,
                                    ensemble_overview,0,
                                    [{file,"src/riak_kv_ensemble_console.erl"},
                                     {line,51}]},
```

engelsanchez:
This note is nice:

```
Note: The consensus subsystem will not be activated until there are more
      than three nodes in this cluster.
```

One small thing: it won't show up if the cluster has 3 nodes but one has strong consistency off, for example.

engelsanchez:
It would be nice to print usage if the given input is not root or an ensemble number:

```
$ bin/riak-admin ensemble-status 3a
                                 {badarg,
                                  [{erlang,list_to_integer,["3a"],[]},
                                   {riak_kv_console,ensemble_status,1,
                                     {line,708}]},
```

engelsanchez:
The warning about AAE uses the local AAE status.  In case a single node has AAE off, for example, you might see ensemble peers remaining untrusted without the warning. Handling that case would be another nice to have

jtuple:
I'm torn on adding checks that look at all nodes in the cluster. All nodes having the same uniform configuration is pretty much already assumed to be a given in the design of strong consistency itself. Just like it is with a million other things in Riak.

If strong consistency isn't enabled the same on all nodes, things pretty much break. Horribly. The same happens for Search if it's not uniformly enabled, for multi-backends if configuration isn't identical across the cluster, etc.

I'd rather not solve it specifically for the strong consistency subsystem when Riak in general has this problem. Rather, we should solve this properly for all subsystems in a future release (perhaps 2.1?). 

In past discussions on this issue, we've decided to move configuration that must be uniform across the cluster out of `app.config / riak.conf`. Perhaps putting in cluster metadata. Then giving users the ability to change these settings through something like `riak-admin config`. Thus, we can enforce uniform settings and remove user error. I think this is a fantastic idea, but obviously something that's not happening for 2.0.

We can certainly add checks for 2.0 in the `ensemble-status` command, but that's extra work that only protects us against some subset of misconfigurations. False security and all that. Plus, we would probably need to add actual checks in strong consistency sub-system itself so that it doesn't do bad things when misconfigured (which it currently does).

engelsanchez:
Consider the notes above about different node configuration as "nice to have"s.  Like you said, this is no different than other commands like aae-status, etc.  The one thing I wold push for to do now, if you agree, is the output of the command when strong consistency is off.  I'm not getting the nice message you described, but that ugly stack trace.

engelsanchez:
I'm basically done with the review. A change to riak_ensemble_console:get_details/0 like this below would be enough to avoid the crash when strong consistency is disabled:

``` erlang
get_details() ->
      {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
      Enabled = riak_core_sup:ensembles_enabled(),
      Ensembles = case Enabled of
                      true -> ordered_ensembles();
                      _ -> []
                  end,
      #details{enabled     = Enabled,
               active      = riak_ensemble_manager:enabled(),
               paranoia    = medium,
               aae_enabled = riak_kv_entropy_manager:enabled(),
               ensembles   = Ensembles,
               quorums     = [],
               nodes       = riak_core_ring:ready_members(Ring),
               ring_ready  = riak_core_ring:ring_ready(Ring)}.
```

engelsanchez:
+1 b0cc282

I have pushed a fix to the most annoying crashes: When strong consistency is disabled and when an invalid non-numeric ensemble id is passed.

engelsanchez:
Notice that we'll have to retry and merge after the dependencies have merged.

jtuple:
@borshop merge

Issue #909, opened 2014-04-15
Make consistency trust configurable


jtuple:
There should be configurable trust levels for consistent K/V data, or as I've called them in previous discussion levels of paranoia. Before `2.0.0beta1`, the `riak_kv_ensemble_backend` always trusted backend data. After `beta1`, the backend never trusts backend data. This should be user configurable.

This level of trust directly affects the availability properties of strongly consistent Riak.

Currently, we never trust backend data after a node / vnode restarts. This means that a given peer must sync with a majority of peers (not including itself) before it can vote in a quorum. For `N=3`, this means that no nodes can ever be offline. If a replica is offline, all associated quorums will be unavailable. For `N=5`, we would need 4 replicas online at all times. For `N=7` we would need 5 replicas online at all times. This is more restrictive then pure quorum because we're assuming byzantine disk faults in the maximum paranoia setting -- eg. undetected lost of data by disk, eg. data that we previously wrote to disk but the disk has since forgotten about without Riak realizing, eg. data loss that CRC/etc never detects.

At the lower paranoia setting, we trust the disks. And normal quorum requirements apply. In other worlds, `N=3` can tolerate 1 node offline, `N=5` can tolerate 2, `N=7` can tolerate 3, etc.

Once we make this configurable, we need to determine the default. I highly suggest we ship with maximum paranoia (the current hardcoded setting as of `beta1`) and suggest users use `N=5` or `N=7` for consistent data.

Users that trust their disks (eg. ZFS mirroring or better), that are certain they will never accidentally delete Riak data by hand, that are certain they'll never accidentally partially restore from an old backup, etc can choose to go with the lower paranoia setting and have the lower quorum requirements.

A more concrete example of what these levels of paranoia are solving is discussed [here](https://gist.github.com/jtuple/8611559).

Remember that this is addressing a limitation that many consistent systems the world over gloss over. That just because you write data to disk doesn't mean that data is still there in the future, and that you may never realize it went missing. Since Riak is designed for maximum data protection, we should ship with the strong settings (IMO).

---

Making this configurable is easy. It's essentially a 1 or 2 line change to `riak_kv_ensemble_backend:trusted` to return `true` or `false` based on an `app.config` setting.

/cc basho/riak#536

reiddraper:
@jtuple is this still a 2.0-RC task?

jtuple:
Yes. Everything on [this list](https://github.com/basho/riak/issues/536) is still a 2.0 RC task.

[...]

PR #980, opened 2014-06-13
Make ensemble trust/paranoia configurable

jtuple:
This commit makes K/V ensemble trust user configurable, allowing for two settings: `low` and `medium`.

The `low` setting assumes that vnodes are durable -- that once data has been committed to a vnode, that data will never be silently lost. In this mode, K/V ensembles are not required to perform an AAE exchange before being trusted. This mode is similar to how most consistent databases operate, and is suitable when users trust that their disks will not lose data undetected. Using a filesystem design to detect bit rot (such as ZFS) on a redundant RAID configuration is essential when using `low` paranoia.

The `medium` setting assumes that vnodes are not durable, and that they may silently lose data without detection. This mode requires all K/V peers that restart/crash to first perform an AAE exchange with a majority of other peers before being trusted. Since this exchange must occur with a majority of peers (not counting the untrusted peer), this mode requires more than a usual quorum majority to guarantee availability.

The `medium` mode assumes that the AAE trees on remote peers are not corrupted such that they are missing the exact same keys which the corrupted node is missing from its K/V data. This should be true in practice, since AAE trees are rebuilt from scratch once a week across the entire cluster.

Note: the paranoia setting is designed to deal with undetected data loss -- an issue that many systems have no protection against. Normal data corruption that results in a detected CRC error can be handled without AAE and works in either paranoia mode.

Addresses issues #909 

This pull-request does not yet expose Cuttlefish versions of this new setting. Input on if this should be configurable in `riak.conf` as well as suggestions for the user-facing setting is would be appreciated.


engelsanchez:
I'm taking this review

engelsanchez:
I'm curious about why we have "low" and "medium" but no "high" :)

I don't see a reason not to have a straight cuttlefish mapping for this: strong_consistency.paranoia = low | medium.   It sounds very user friendly to me.  I'll attach a commit with that to this PR and we can keep it if nobody has objections.  /cc @seancribbs @michellep

seancribbs:
Yes, please add this to the configuration schema if it is configurable before startup.

jtuple:
"high" was going to be a paranoia setting where we didn't trust AAE trees either, and required AAE trees to be rebuilt from scratch before syncing. I never got around to implementing "high", and I think it would be terrible given how both AAE and AAE-based ensemble syncing are currently implemented. Ensembles could end up being unavailable for _days_ while you check integrity/etc. So, punted on "high" for 2.0, but plan to investigate "high" in a future release.

engelsanchez:
While reviewing this, I ran across a nasty problem that prevented ensembles from starting up at all.  As far as I can tell, it's not directly related to this simple config (although it may have made a race easier to hit).  So I'm going to open a separate issue to track it. Unfortunately I have not been able to reproduce it.

engelsanchez:
+1 a7ad6b9
Added cuttlefish support for strong_consistency.paranoia and did all my testing with it.  The code looks good and is very simple.  Ensembles spend a while syncing if paranoia is medium when a node restarts, but become trusted immediately when set to low.

Unless @jtuple @andrewjstone or somebody else has objections on the cuttlefish name or related, this can be merged.

seancribbs:
Cuttlefish schemas have tests. Add to them!

Also, config changes should get PM/CSE +1's.

engelsanchez:
@seancribbs Added a schema test for strong_consistency.paranoia.  Please illuminate us on how to go about getting those +1's. That's why I'm asking.

evanmcc:
Sorry, super late to the party with my two cents here, but for wishy washy
reasons, integrity or something world be my preference over paranoia.

seancribbs:
Config change +1's:
- [x] Eng
- [x] CSE
- [x] PM

/cc @angrycub @michellep

engelsanchez:
Ping @seancribbs @angrycub @michellep . What is the process to keep these things from stalling? If we can't get the required approval or feedback soon I might have to just drop my attempt to add cuttlefish support and leave it as an advanced thing. But we need to keep this moving.

bsparrow435:
looks good, CSE approved.

seancribbs:
@engelsanchez Looks like this will need rebase to merge cleanly.

engelsanchez:
Yes @seancribbs I'm doing a round of rebases/merges across my SC issues right now.  Got ship this shit!

michellep-basho:
Sorry for the delay. +1 from PM on the concept. I agree with Evan that we should call it something other than paranoia.  Discussed with Engel and recommending that we use <i>trust</i>, reversing the order of the setting levels, as there are additional components to data integrity.  So the current low would become high, medium would be unchanged.

engelsanchez:
Thanks @michellep !  @bsparrow435 I'm currently implementing the change as proposed by Michelle. Since you were the CSE who approved the previous version I'm checking with you if you are OK with the change from paranoia -> trust here.

andrewjstone:
Tests run fine as well. Should be :+1: for at least the last commit.

engelsanchez:
+1 d9fc05e

engelsanchez:
@borshop merge

PR #987, opened 2014-06-17
Fix MOD_MAP when datatype update request includes return_body=true

seancribbs:
This was an oversight in the previous review. /cc @russelldb


russelldb:
+1 fd260df

seancribbs:
@borshop merge

PR #992, opened 2014-06-17
Correct discrepancy between do_get_term and case clause in do_delete

seancribbs:
This was causing node shutdown when running the Java client
integration tests. On a small scale, this would be fine (one vnode
crashing would start back up), but in massive-delete scenarios entire
clusters could go down. We should heed the TODO message in do_get_term
and handle this holistically, but for now this should prevent the
vnode from crashing.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1536,7 +1536,7 @@ do_get(_Sender, BKey, ReqID,
 %% @private
 -spec do_get_term({binary(), binary()}, atom(), tuple()) ->
                          {{ok, riak_object:riak_object()}, tuple()} |
-                         {{error, not_found}, tuple()} |
+                         {{error, notfound}, tuple()} |
                          {{error, any()}, tuple()}.
 do_get_term({Bucket, Key}, Mod, ModState) ->
     case do_get_object(Bucket, Key, Mod, ModState) of
@@ -1762,7 +1762,7 @@ do_delete(BKey, State) ->
                     %% not a tombstone or not all siblings are tombstones
                     {reply, {fail, Idx, not_tombstone}, State#state{modstate=UpdModState}}
             end;
-        {{error, not_found}, UpdModState} ->
+        {{error, notfound}, UpdModState} ->
             %% does not exist in the backend
             {reply, {fail, Idx, not_found}, State#state{modstate=UpdModState}}
     end.


evanmcc:
Sorry, that's probably my typo;  Is there a way to get dialyzer to help here?

engelsanchez:
Oops. So this is in 1.4.9 too, right?  I remember seeing an old TODO in the code to make these not_found/notfound atoms consistent :(

seancribbs:
@engelsanchez This is a first step, I think I need to look at all areas for consistency. Dialyzer didn't find this because the spec of `do_get_term` encompasses both possible spellings, even if I change it. Also, putting a spec on that private function is probably less useful because it could hide interface errors like this.

seancribbs:
Here's a quick scan of riak_kv source: https://gist.github.com/seancribbs/9a439ba7d7b3fb869ab9

seancribbs:
In my cursory look, it seems only the backends use `not_found` while all the upper layers use `notfound`, which means we have two options AFAICT:
1. Keep this PR as-is so that `not_found` is converted into `notfound` in the vnode.
2. Convert all the backend code and tests to return/expect `notfound` instead of `not_found`.

The one edge-case may be AAE. It is unclear to me which form it expects just from the `ack` output linked above.

engelsanchez:
@seancribbs we should probably separate the overarching issue of standardizing the responses (and probably add it to 2.1) and this, which was introduced by a recent change to the vnode code (https://github.com/basho/riak_kv/commit/6cd0376ac97569355b31959be481e948827cc323). Fix this, then make sure cleanup happens at a better time. Sounds good?

seancribbs:
@engelsanchez Ok. I ran out of time yesterday, so I will make sure that this solves the issue with the Java client.

reiddraper:
Assigned 2.0-RC milestone. Feel free to change.

seancribbs:
Riak node survived the Java integration tests on multiple passes. :smile_cat:

seancribbs:
Except it's crashing on something else unrelated :frowning: 

```
beam.smp(27904,0xb190e000) malloc: *** error for object 0x7fe85213c4b0: pointer being freed was not allocated
                                                                                                                                *** set a breakpoint in malloc_error_break to debug
                               Abort trap: 6
```

seancribbs:
@borshop merge

PR #993, opened 2014-06-17
Fix AAE on/off detection

### LINKED ISSUE #988: Strong consistency AAE check broken
The current check to see if AAE is enabled in riak_kv_ensembles is expecting riak_kv/anti_entropy to be on, but this parameter is funky and is actually {on, []} or {on, [debug]}.  So the warning in the logs happens even if all things are good:

```
2014-06-17 12:00:05.829 [warning] <0.384.0>@riak_kv_ensembles:maybe_aae_configuration_valid:168 Strong consistency has been enabled without AAE -- all consistent operations will timeout until AAE is enabled.
```

/cc @jburwell

engelsanchez:
riak_kv/anti_entropy takes a value of the form {on|off, Opts}, not on
off.

This fixes basho/riak_kv#988

Without this fix, you would see messages like this:

```
2014-06-17 12:00:05.829 [warning] <0.384.0>@riak_kv_ensembles:maybe_aae_configuration_valid:168 Strong consistency has been enabled without AAE -- all consistent operations will timeout until AAE is enabled.
```

Even with active anti-entropy enabled. Now, these messages are only shown if anti_entropy is set to passive.  When it is active or active-debug, they will not show up

### DIFF
--- a/src/riak_kv_ensembles.erl
+++ b/src/riak_kv_ensembles.erl
@@ -161,10 +161,10 @@ maybe_bootstrap_ensembles() ->
             end
     end.
 
--spec maybe_aae_configuration_valid(atom() | any()) -> boolean().
-maybe_aae_configuration_valid(on) ->
+-spec maybe_aae_configuration_valid({on|off,_}) -> boolean().
+maybe_aae_configuration_valid({on, _}) ->
     true;
-maybe_aae_configuration_valid(_) ->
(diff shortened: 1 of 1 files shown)


coderoshi:
+1 c37ca2a
yup

engelsanchez:
@borshop merge

PR #997, opened 2014-06-24
Ignore all warnings in generated riak_core_pb

engelsanchez:
Because tools.mk performs a simple grep filter using the patterns in
dialyzer.ignore-warnings, simply using the module name (with the extra
colon just to be a bit safer) is enough to make it ignore all warnings
in that module. This should be a time saver

/cc @reiddraper   I'm preparing a separate tools.mk PR to deal with the line number issue.


engelsanchez:
I've amended the one commit by removing the extra colon. That makes it compatible with the latest version of https://github.com/basho/riak_kv/pull/998 while still ignore all warnings for the riak_core_pb module.

engelsanchez:
Testing this should be easy: with the changes, run make dialyzer and notice that the warnings for riak_core_pb are still all ignored.

lordnull:
:+1: e55adad

There is no difference on output, so that would be a successful change.

engelsanchez:
@borshop merge and thanks @lordnull !

PR #998, opened 2014-06-24
Ignore line numbers in dialyzer ignore file

engelsanchez:
This change uses sed to remove the line numbers from the ignore file and
the warnings file. That way the comparison doesn't care if line numbers
change, which happens very often.

Notice that the final output does not have the current line number for
the warning. I think that is OK for now.  With some trickery involving
sort and sed the original line numbers could be re-added to the output,
but that will have to wait for another day.

/cc @reiddraper I was going to open the PR on the tools.mk repo first, but I thought it would be easier to test here, then you/I can apply the change to tools.mk

### DIFF
--- a/tools.mk
+++ b/tools.mk
@@ -36,7 +36,18 @@ ${LOCAL_PLT}: compile
 
 dialyzer-run:
 	@echo "==> $(shell basename $(shell pwd)) (dialyzer)"
-	@if [ -f $(LOCAL_PLT) ]; then \
+# The bulk of the code below deals with the dialyzer.ignore-warnings file
+# which contains strings to ignore if output by dialyzer.
+# Typically the strings include line numbers. Using them exactly is hard
+# to maintain as the code changes. This approach instead ignores the line
+# numbers, but takes into account the number of times a string is listed
+# for a given file (See the sort + uniq -c commands, which add the count
+# as a prefix). So if one string is listed once, for example, and it
+# appears twice in the warnings, the user is alerted. It is possible but
+# unlikely that this approach could mask a warning if one ignored warning
+# is removed and two warnings of the same kind appear in the file, for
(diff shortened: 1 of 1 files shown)


engelsanchez:
I have added a commit to improve on the previous approach. Instead of completely ignoring the line number and ignoring all instances of a given warning in a module, the latest approach takes into account the number of times a duplicate warning is listed in dialyzer.ignore-warnings.  So for example, if you ignore two instances of a warning in a module, but the warning appears 3 times in the output of dialyzer, we know something is wrong.  It would require a lot more crazy luck now to ignore a warning that should not be ignored (i.e. a warning suddenly disappears and is replaced by another one with exactly the same output in another line), but I think that is very unlikely and a good trade off. The current situation of having most PRs have merge conflicts because of the ignore warnings file is leading us to many last minute tweaks to PRs is a massive time waster.

reiddraper:
Haven't looked at this in detail yet, but thanks! Can you make the PR against the basho/tools.mk repo? And then we'll update all repos once review is done there?

reiddraper:
Derp, sorry, just read your comment. Yeah, happy to do review here first.

reiddraper:
@engelsanchez this looks great. We could also later use `tee` to keep the original line numbers on the actual dialyzer output. Do you want to make a PR against tools.mk, and then I'll roll out the new version to all repos? Can you keep these nice commit messages too in the tool.mk PR?

reiddraper:
Pushed in basho/tools.mk@45130f34aac90034b058bd9e00d12c2ef5376419 and basho/tools.mk@0ab19a529cb7b0306e5f08c502decc6ae7f6d70a.

reiddraper:
Next, I'll tag tools.mk and begin rolling it out into our repositories.

PR #999, opened 2014-07-03
Update tools.mk to v0.5.4

reiddraper:
Removes the need to specify line numbers in the dialyzer ignore file.

### DIFF
--- a/tools.mk
+++ b/tools.mk
@@ -36,7 +36,18 @@ ${LOCAL_PLT}: compile
 
 dialyzer-run:
 	@echo "==> $(shell basename $(shell pwd)) (dialyzer)"
-	@if [ -f $(LOCAL_PLT) ]; then \
+# The bulk of the code below deals with the dialyzer.ignore-warnings file
+# which contains strings to ignore if output by dialyzer.
+# Typically the strings include line numbers. Using them exactly is hard
+# to maintain as the code changes. This approach instead ignores the line
+# numbers, but takes into account the number of times a string is listed
+# for a given file (See the sort + uniq -c commands, which add the count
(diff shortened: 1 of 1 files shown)


engelsanchez:
I'm sanity checking/reviewing this

tools.mk
```diff
@@ -36,7 +36,18 @@ ${LOCAL_PLT}: compile
...
+# Typically the strings include line numbers. Using them exactly is hard
+# to maintain as the code changes. This approach instead ignores the line
+# numbers, but takes into account the number of times a string is listed
+# for a given file (See the sort + uniq -c commands, which add the count
+# as a prefix). So if one string is listed once, for example, and it
+# appears twice in the warnings, the user is alerted. It is possible but
+# unlikely that this approach could mask a warning if one ignored warning
+# is removed and two warnings of the same kind appear in the file, for
+# example. But it is a trade-off that seems worth it.
+	@-if [ -f $(LOCAL_PLT) ]; then \
```
engelsanchez, line 50: I just thought it worth to mention here that the @if was changed to @-if to avoid echoing all these shell commands. It might surprise some people used to seeing them when running make dialyzer

reiddraper, line 50: Thanks, sounds good to me.

engelsanchez:
:+1: 65c6ae9

Verified by making sure no warnings are reported now, then re-introduced some warnings to make sure it failed.

But since I've experienced the sadness that is not having line numbers in the output, I'm going to try to add this on the side and submit another PR at some point soon. For now people will have to be reminded to look at the dialyzer_warnings raw file if they are confused.

reiddraper:
@borshop merge

PR #1000, opened 2014-07-08
Update tools.mk to v0.5.5

reiddraper:


reiddraper:
See #999, which was closed to make a minor fix. And #998, where @engelsanchez originally introduced the tools.mk change.

russelldb:
+1 48ec574

Verified that the `tools.mk` is the same as in `basho/tools.mk` tagged `0.5.5`. Given the review and testing that got theese changes merged there, that seems to be enough to pass this.

reiddraper:
@borshop merge

PR #1002, opened 2014-07-11
Update to latest riak_ensemble integrity approach

jtuple:
This is the sibling pull-request for basho/riak_ensemble#37. That pull-request completely changes how `riak_ensemble` performs data validation and integrity checking, completely removing the need for backend based trust/syncing.

This pull-request updates `riak_kv` to work with this new approach. For the most part, that means deleting a bunch of code that is no longer necessary. As well as updating the `ensemble_status` command to no longer reference AAE or trust -- since neither of those concepts apply anymore.

For more details, look at the individual commits.

**_Note:**_ a discussion with product management needs to happen concerning the new output of `ensemble_status` as well as how we should expose the new integrity checking settings to `riak.conf`. The old discussion about `trust` no longer applies since that setting has been removed. I think the approach as implemented in this pull-request is good enough for 2.0 RC But, we need to have that discussion / change things before 2.0 final.


jtuple:
Concerning the builder failure, this pull-request will fail until the sibling `riak_ensemble` pull-request is merged.

src/riak_kv_ensemble_console.erl
```diff
@@ -386,17 +399,11 @@ print_variations() ->
     ensemble_overview(Details2),
     io:format("~n~n"),
 
-    %% Enabled, enough nodes, but AAE not enabled
-    io:format("(*** AAE disabled when required ***)~n~n"),
-    Details3 = Details2#details{nodes=['dev1@127.0.0.1',
```
engelsanchez, line 391: Poor Details3 got lost in the shuffle.

src/riak_kv_exchange_fsm.erl
```diff
@@ -304,31 +296,11 @@ read_repair_keydiff(RC, LocalVN, RemoteVN, {Bucket, Key, Reason}) ->
...
-                    ok
-            end
-    after 10000 ->
-            %% TODO: Timeout should be configurable
-            %% lager:info("repair_consistent timeout"),
-            exit(self(), kill)
-    end.
+repair_consistent(BKey) ->
+    Ensemble = riak_client:ensemble(BKey),
+    Timeout = 60000,
```
engelsanchez, line 301: Maybe just byte the bullet and make this configurable in advanced config?

src/riak_kv_ensemble_console.erl
jtuple, line 391: Aye, easier to just kill it then renumber all the other lines. Keeping diff minimal/etc.

src/riak_kv_exchange_fsm.erl
jtuple, line 301: We could, but maybe later? Normal Riak read repair triggered by AAE is just a `riak_client:get` which uses a `60000` hardcoded timeout in `riak_client`. Figured copying existing behavior was good enough for now.

engelsanchez, line 301: Yes, we can do config cleanup later.

andrewjstone:
:+1: 1519686
Code looks good and riak tests pass with the following changes:
- https://github.com/basho/riak_test/pull/657
- https://github.com/basho/riak_test/pull/658
- https://github.com/basho/riak_ensemble/pull/37

jtuple:
@borshop merge

PR #1005, opened 2014-07-23
Add RangeReg as a new supported datatype

lenary:
RangeReg is a register for integers that tracks the largest,
the smallest, the first (by timestamp) and the last (by timestamp)
integers assigned into it.

The new type is called "rangereg" inside the codebase, and can be
activated just like any other bucket-type datatype. It exposes
the exact same statistics as the existing 2.0 datatypes.

This requires the work in basho/riak_dt#59 to be merged before it can
be merged. I'm also about to open a lot of sub-pull-requests for riak_pb, 
riak_test and several other repos, all of which go along with this PR.

It's worth saying this is 2.x (where x > 0) work, so not immediate.


src/riak_kv_crdt.erl
```diff
@@ -107,26 +107,34 @@ value(RObj) ->
...
 
 %% @doc convenience for (e.g.) MapReduce functions. Pass an object,
 %% get a 2.0+ Map type value, or `[]' if no Map is present.
 -spec map_value(riak_object:riak_object()) -> proplist:proplist().
 map_value(RObj) ->
-    {{_Ctx, Map}, _Stats}  = value(RObj, ?MAP_TYPE),
-    Map.
+    simple_value(RObj, ?MAP_TYPE).
+
+
```
lenary: This could do with some docs. Oops

reiddraper:
Has there been any input from product management on this?

russelldb:
Nope. What kind of input do you think they need?

randysecrist:
+1.
I like this addition, complements counters nicely.

lenary:
@reiddraper I think extra crdts that complement the gaming vertical nicely has been discussed, even if I don't remember the outcome. This one is for things like max scores (but also maybe financial info), I also want to work on an ordered set at some point for league tables etc.

lenary:
So, from the eng mumble today: this is on hold for the moment, until we do more thinking about product management.

cmeiklejohn:
For what it's worth, I had a discussion with a customer yesterday about this feature.

russelldb:
Equally, we've had a few asks in IRC for it. I guess awaiting @michellep or someone else (is there anyone else??) from Product Management to join the discussion.

yorkshire-steve:

PR #1008, opened 2014-08-01
Use SC bucket types and buckets to know ensembles

### LINKED ISSUE #1007: Ensembles not created for strong bucket type with n_val different from default
When we start with default bucket props containing `n_val = 5` and then create a strongly consistent bucket_type with `n_val = 3`, the corresponding riak_ensembles for that new n_val will not be created. The problem seems to stem from the fact that `riak_kv_ensembles:required_ensembles/1` checks all current bucket n_vals  for ensembles to create, but doesn't check bucket types that don't yet have buckets created. Since all buckets are created on demand with a put, and it's not possible to put if the ensemble doesn't exist, the buckets and ensembles never get created.

andrewjstone:
If we use a default bucket property with a different n_val from the
strongly consistent bucket type, ensembles for that bucket type will
never be created. The problem seems to stem from the fact that
riak_kv_ensembles:required_ensembles/1 checks all current bucket n_vals
for ensembles to create, but doesn't check bucket types that don't yet
have buckets created. Since all buckets are created on demand with a
put, and it's not possible to put if the ensemble doesn't exist, the
buckets and ensembles never get created.

To fix this we now iterate through all bucket types and collect all
n_vals from them, as well as the n_vals from the existing buckets in
order to figure out the set of all n_vals required to create ensembles.

Requires a corresponding riak_core PR and riak_test to show the issue. Those are coming shortly.

Fixes #1007

### DIFF
--- a/src/riak_kv_ensembles.erl
+++ b/src/riak_kv_ensembles.erl
@@ -28,8 +28,7 @@
 -behaviour(gen_server).
 
 %% API
--export([start_link/0,
-         possible_change/0]).
+-export([start_link/0]).
 
 %% Support API
 -export([ensembles/0,
@@ -47,7 +46,7 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
--record(state, {last_ring_id :: term()}).
+-record(state, {}).
 
 %%%===================================================================
 %%% API
@@ -55,9 +54,6 @@
 
 start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
-
-possible_change() ->
-    ?MODULE ! reset_ring_id.
 
 local_ensembles() ->
     Node = node(),
@@ -102,7 +98,7 @@ check_membership(Ensemble, CHBin) ->
 
 init([]) ->
     schedule_tick(),
-    {ok, #state{last_ring_id = undefined}}.
+    {ok, #state{}}.
 
 handle_call(_Request, _From, State) ->
     {reply, ok, State}.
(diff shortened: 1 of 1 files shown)


jtuple:
+1 71016f20917afee93485ead0938fd64e7cd3f0f7

Design matches what we discussed and all code looks good. Tested against the new riak_test and all is well.

We should benchmark if there's any performance impact of rechecking all N every tick just to know how worse things are (or aren't), but this is the best approach for 2.0 so we're going to ship this way regardless.

Future work is to consider restating the smarter approach that also checks bucket type hashes.

jtuple:
+1 59883e44c1fe7882e8b6999ddaafcb6f46a109bb

Bah, +1 the wrong commit. This should do it.

andrewjstone:
@borshop merge

andrewjstone:
Thanks @jtuple !

jtuple:
It looks like this was merged to develop but not to the 2.0 branch, and thus isn't part of RC2. Without this change, users are unable to read/write data for consistent keys with a bucket N-val different than the default bucket N-val. This is a pretty major issue.

The changes shouldn't invalidate any RC2 performance testing, nor any correctness qualification outside of tests related to strong consistency. In fact, without this change merged, the related test that was added to test this issue is going to fail. So, merging this branch actually improves test results.

Is it possible to merge this into 2.0 for 2.0.0 final?
/cc @jaredmorrow @andrewjstone @jonmeredith @matthewvon

jtuple:
Looks like we're punting on this until 2.0.1 now since we've already started 2.0.0 final qualification tests and don't want to rerun.

So, going to clearly document things here in case someone hits this and people come looking for a solution.

If a user creates a consistent bucket type with an N value that is different then the default bucket N value, ensembles for that different N value may not be created. This will manifest as reads/writes to data under that bucket type always timing out.

If this occurs, a user should be able to fix this by running `riak_core_ring_manager:force_update()` from the Erlang console of the claimant node. If that doesn't work, doing a rolling restart of all nodes should also be sufficient.

jaredmorrow:
Added 2.0 Known Issue so it doesn't get lost and we make a note of it in the release notes.

PR #700, opened 2013-10-11
Issue/665 retry put coordinator failure

slfritchie:
Addresses #665 

I strongly disagree that the default should be 'off'.  This is Riak 2.0 and is a good time to introduce a behavior change that should've happened many releases ago.

There's a tentative cuttlefish thingie in one of the commit log messages.

PR #1009, opened 2014-08-06
Backport put fsm forwarding.

lordnull:
Takes #627, #700, and #913 and puts them in one pull request to take the put fsm forwarding feature from 2.0 and place it in 1.4.

Tasks remaining:
- [ ] Create an r_t to test the feature. If that cannot be done in the next 3 days (EOD Friday) at least be able to create a step-by-step manual verification process.
- [ ] If commit c927c7faa8b (or its descendants) is correct & is well-reviewed, then it should also be applied to `develop`.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -180,6 +180,10 @@ start(_Type, _StartArgs) ->
                                           [[pncounter],[]],
                                           []),
 
+            riak_core_capability:register({riak_kv, put_fsm_ack_execute},
+                                          [enabled, disabled],
+                                          disabled),
+
             HealthCheckOn = app_helper:get_env(riak_kv, enable_health_checks, false),
             %% Go ahead and mark the riak_kv service as up in the node watcher.
             %% The riak_core_ring_handler blocks until all vnodes have been started
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -167,10 +167,14 @@ get_put_coordinator_failure_timeout() ->
     app_helper:get_env(riak_kv, put_coordinator_failure_timeout, 3000).
 
 make_ack_options(Options) ->
-    case riak_core_capability:get({riak_kv, put_fsm_ack_execute}, disabled) of
-        disabled ->
+    case (riak_core_capability:get(
+            {riak_kv, put_fsm_ack_execute}, disabled) == disabled
+          orelse
+          app_helper:get_env(
+            riak_kv, retry_put_coordinator_failure, on) == off) of
+        true ->
             {false, Options};
-        enabled ->
+        false ->
             case proplists:get_value(retry_put_coordinator_failure, Options, true) of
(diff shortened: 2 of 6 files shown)


slfritchie:
Sorry, -1, this patch set as-is doesn't work the way it does on `develop`.
1. Create a 5 node cluster with `vm.args` containing `+zdbbl 4`.
2. Suspend nodes 2 & 3, e.g. `kill -STOP $pid`.  Stopping them won't work.
3. Run the following Erlang code or equivalent on node1:
   
   {ok, C} = riak:local_client().  
   Big = list_to_binary(lists:duplicate(1024*1024, $x)).
   [begin timer:sleep(1000), spawn(fun() -> [{foo, Nd} ! Big || Nd <- nodes()] end), timer:sleep(100), Y = C:put(riak_object:new(<<"b">>, <<X:32>>, <<"val!">>)), io:format("Y was ~p at ~p\n", [Y, time()]) end || X <- lists:seq(0,10)].

Each put should be returning `ok` in less than 10 seconds.  Any delay beyond 10 seconds or any other return value is an error.  I see zero errors when using this procedure on `develop` but am seeing errors on this branch.

```
Y was ok at {18,30,4}
Y was ok at {18,30,5}
Y was ok at {18,30,6}
Y was {error,timeout} at {18,31,7}
Y was ok at {18,31,8}
[....]
```

slfritchie:
I found the same behavior in `develop` ... I just had to work harder at it.

jonmeredith:
Does that mean you've found an issue with put-foward-ackery?

On Thu, Aug 7, 2014 at 6:57 AM, Scott Lystig Fritchie <
notifications@github.com> wrote:

> I found the same behavior in develop ... I just had to work harder at it.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/basho/riak_kv/pull/1009#issuecomment-51467339.

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

slfritchie:
I've edited/updated my put loop recipe from 4 hours ago.

@jonmeredith Nope, it was a problem after the ackery.  See newly-pushed commit c927c7faa8b.

lordnull:
I've been able to test this patch manually using some information from the client that requested this patch set, and it does address the specific problem. Developing an r_t solution proved problematic as the test case we had used sysrq, which is not exactly portable (nor safe for a test considering it reset the machine). However, using the info @slfritchie has pointed out (use of kill -STOP and kill -CONT), an r_t solution seems entirely possible.

slfritchie:
Ping: comments on commit c927c7f?

jonmeredith:
@lordnull ping - would like to give customer an update by end of their day (which is soon as they're on BST)

lordnull:
r_t is not able to replicate exactly, but I was able to create a script with some notes on its use for a manual verification that this patch does solve the problem they described. I uploaded that script and notes to the ticket they opened.

PR #1013, opened 2014-08-08
Edoc fixes for readable HTML generation

macintux:
- Fix vertical whitespace and code formatting in module doc string (particularly noticeable in the wm modules)
- Some textual improvements to same
- Various fixes to `@see/@link`:
  - Separate `@see` from `@doc`
  - Replace `@see` with `@link` where appropriate
  - Fix cross-references to other Riak applications
  - Drop `@see` where it isn't actually part of edoc
- UTF-8 encoding for files where Nuno Preguiça is referenced in edoc strings so rendered properly
- Fix URLs in edoc strings (require [] for URL-ification)
- Various grammar fixes
- Replace "" and '' with `' where appropriate


macintux:
@reiddraper ^

reiddraper:
Back from PTO, reviewing asap.

src/riak_kv_ensemble_backend.erl
```diff
@@ -19,7 +19,7 @@
 %% -------------------------------------------------------------------
 
 %% @doc
-%% Implementation of {@link riak_ensemble_backend} behavior that
+%% Implementation of {@link //riak_ensemble/riak_ensemble_backend} behavior that
```
reiddraper, line 22: What's this `//`?

macintux, line 22: Global path, since this is a different project. See section 1.9 in http://www.erlang.org/doc/apps/edoc/chapter.html

reiddraper, line 22: Ah, found it in the [edoc manual](http://www.erlang.org/doc/apps/edoc/chapter.html). Probably just not linking correctly because we're building docs with `skip_deps=true`.

src/riak_kv_mrc_pipe.erl
```diff
@@ -20,7 +20,7 @@
 
 %% @doc Riak KV MapReduce / Riak Pipe Compatibility
 %%
-%% == About using `{modfun, Mod, Fun, Arg}' generator to a MapReduce job ==
+%% == About using {modfun, Mod, Fun, Arg} generator to a MapReduce job ==
```
reiddraper, line 23: Not allowed to use `code` blocks in titles, or something?

macintux, line 23: Doesn't render properly, whether it's allowed or not.

reiddraper, line 23: works for me

seancribbs, line 23: Is it possible to manually insert the heading instead of using the shortcut? e.g.

``` html
<h2>About using <code>{modfun, Mod, Fun, Arg}</code> generator to a MapReduce job</h2>
```

macintux, line 23: Didn't try.

seancribbs, line 23: No worries, doesn't seem critical.

macintux, line 23: Yeah, I'm thinking the cost of cluttering up the comments for future code readers probably higher than the value of having it be monospace on the webpage.

reiddraper:
@borshop merge

PR #1015, opened 2014-08-12
riak_kv_vnode:put() should use even-more-unreliable command

slfritchie:
Bring one patch, commit c927c7f, from PR #1009 forward to `develop`,
to fix a bug in put FSM processing when sending the put vnode op to
a node that isn't the coordinator blocks due to `busy_dist_port`
status.

cc: @jonmeredith @lordnull

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -219,7 +219,23 @@ put(Preflist, BKey, Obj, ReqId, StartTime, Options) when is_integer(StartTime) -
 
 put(Preflist, BKey, Obj, ReqId, StartTime, Options, Sender)
   when is_integer(StartTime) ->
-    riak_core_vnode_master:command(Preflist,
+    %% Reminder: command_unreliable() = "even more unreliable"
+    %%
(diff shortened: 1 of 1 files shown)


lordnull:
Using the process described in https://github.com/basho/riak_kv/pull/1009#issuecomment-51450810 shows no problems, even after upping the send loop to 100 interations. However, using the same process on develop also showed no problems. I do note the 'had to try harder' comment, though I'm not sure if that simply means more iterations to trigger it, or larger data sent to {foo, Node}.

src/riak_kv_vnode.erl
```diff
@@ -219,7 +219,23 @@ put(Preflist, BKey, Obj, ReqId, StartTime, Options) when is_integer(StartTime) -
 
 put(Preflist, BKey, Obj, ReqId, StartTime, Options, Sender)
   when is_integer(StartTime) ->
-    riak_core_vnode_master:command(Preflist,
+    %% Reminder: command_unreliable() = "even more unreliable"
```
russelldb, line 222: unless you have some non-idempotent command, like a counter increment (for example.)

yorkshire-steve:

PR #1016, opened 2014-08-15
Remove deprecated SC w/o AAE warning

engelsanchez:
The strong consistency system no longer requires AAE to be enabled to start up.  So this warning can be removed.  

Without this patch, running Riak with strong consistency enabled but with AAE disabled logged the warning in the logs.  It should be silent now.


andrewjstone:
:+1: ef5f797

engelsanchez:
@borshop merge

jtuple:
**_Note:**_ This should get merged into `2.0` branch after we build final 2.0.0 build so that it'll make it into 2.0.1.

PR #1022, opened 2014-09-16
Expose delete_mode to riak clients

krestenkrab:
This change exposes the `delete_mode` parameter to riak clients, so it can be controlled on individual DELETE operations.

This is most useful if the global `riak_kv.delete_mode` is set to `keep`, allowing a cleanup job (e.g. using
map/reduce) to later delete with mode `immediate` or a time interval.  Because we've been using EE w/fullsync, we need to keep tombstones around until we're certain that they have been propagated to connected clusters; and setting the delete_mode to a high timeout value is hard on the memory and the timer wheel.

The parameter `delete_mode` is also exposed to http and protocol buffer clients.  See https://github.com/basho/riak-erlang-client/pull/194 and https://github.com/basho/riak_pb/pull/102

This patch is against the 1.4 branch.

### DIFF
--- a/include/riak_kv_vnode.hrl
+++ b/include/riak_kv_vnode.hrl
@@ -45,6 +45,11 @@
           bkey :: {binary(), binary()},
           req_id :: non_neg_integer()}).
 
+-record(riak_kv_delete_req_v2, {
+          bkey :: {binary(), binary()},
+          req_id :: non_neg_integer(),
+          delete_mode :: undefined | keep | immediate | pos_integer() }).
+
 -record(riak_kv_map_req_v1, {
           bkey :: {binary(), binary()},
           qterm :: term(),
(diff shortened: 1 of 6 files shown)


krestenkrab:
In an effort to reduce the surface area of this pull request, I removed the exposure of this in the http / protobuf API's. (4311788)

Also, the new `#riak_kv_delete_req_v2` message was being generated even for request that did not specify a non-default delete_mode. The commit (ec989c9) makes it so that the new message is only used when the `delete_mode` option is actually passed into `riak_kv_delete` and/or `riak_kv_get_fsm`.  This should make deletes continue to work during a rolling upgrade; while the new functionality only works when the entire cluster is upgraded.

So this makes this PR local to only this repo. basho/riak-erlang-client#194 and basho/riak_pb#102 are thus no longer relevant.

jonmeredith:
Thanks Kresten - will try and take a look at this in the next few days.

jonmeredith:
First scan of code looks good, working on a riak_test to prove it is good.

jonmeredith:
riak_test added https://github.com/basho/riak_test/pull/685

PR #1023, opened 2014-09-18
Update to use new breadth-first AAE exchange

jtuple:
This commit updates riak_kv to work with the new breath-first AAE exchange added to riak_core. This new level-by-level exchange is necessary in order to support the new pipelined AAE-based fullsync replication protocol in Riak Enterprise.

Note: this change makes local AAE also use the new BFS exchange. I plan to finish up running a performance test tomorrow and add numbers to this pull request to hopefully show that this new approach has no negative impact on normal AAE.

Sibling pull-requests: basho/riak_core#628, basho/riak_repl#616

### DIFF
--- a/src/riak_kv_exchange_fsm.erl
+++ b/src/riak_kv_exchange_fsm.erl
@@ -200,6 +200,10 @@ key_exchange(timeout, State=#state{local=LocalVN,
                      ok = disk_log:sync(Now),
                      ok = disk_log:close(Now),
                      ok;
+                (start_exchange_level, {_Level, _Buckets}) ->
+                     ok;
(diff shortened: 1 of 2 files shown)


jonmeredith:
Code inspection looks good - interesting what the value is if there are still other instances of compare/4 in use for other components. 

Is there value in merging this before we have resolved the other call sites to `compare`?

jonmeredith:
Understood why we need this merged.  Satisfied myself it is ok with rolling upgrades as changes to `hash tree:compare2` are node local.

jonmeredith:
Plus one merge - needs to be retargeted at 2.0

jonmeredith:
Merged as #1026

PR #1032, opened 2014-10-09
Backport 2.0.1 AAE improvements

engelsanchez:
This is a prerequisite for the backporting of a number of MDC
improvements done in the 2.0.X series to prepare a 1.4.11 release.
This version passes the AAE riak_test.

The original changes come from PRs against both riak_core and riak_kv.
But on the 1.4 branch, the hashtree module is still in riak_kv, so they are
both included here.

basho/riak_core#629
basho/riak_kv#1026

/cc @jonmeredith


src/hashtree.erl
```diff
@@ -515,7 +500,7 @@ hash(X) ->
     sha(term_to_binary(X)).
 
 sha(Bin) ->
-    Chunk = app_helper:get_env(riak_kv, anti_entropy_sha_chunk, 4096),
+    Chunk = get_env(anti_entropy_sha_chunk, 4096),
```
jonmeredith, line 503: This change moves anti_entropy_sha_chunk to come from the riak_core section rather than riak_kv.  Should stick with the original `app_helper:getenv` version, or have `get_env` pull from riak_kv instead.

jonmeredith:
The two versions of orddict_delta are interesting.  The 1.4 backport updates the code from riak_ensemble_util, however the 2.0 version uses the ensemble code from exchange_level/exchange_final but uses the original 1.4 code for compare_segments.

Do you remember the history/reason @jtuple ?

src/hashtree.erl
engelsanchez, line 503: it takes it from kv and default to an entry in core actually.  With the movement of hashtree between repos that should allow either.

engelsanchez:
@jonmeredith @jtuple I forgot to check with you about the orddict_delta change. The newer code in ensemble ensures that order of inputs is kept.   The code was using that version in ensemble (which doesn't exist in this parallel universe) so I thought the best would be to update the local code to have the fix from ensemble applied and use that.  Not 100% sure there aren't subtle consequences of doing that. It just looked good to me.

jonmeredith:
Provisional secondary reviewer, once we answer
- [x] which appenv anti_entropy_sha_chunk should come from
- [x] which version of orddict_delta should be used (and if we need to fix 2.0 branch)

engelsanchez:
for 1.4, I guess I should have kept the riak_kv app env really. Notice that I didn't make that up: https://github.com/basho/riak_core/commit/ded140d8948a71fb155d30617afe8992b84a7b78.

The dual "pick up from riak_kv, but maybe use from core if no kv" was added when the module was moved to core.  It will have no effect unless someone writes a value in the core section and kv. If only in riak_kv, it will behave just the same.  meh

jonmeredith:
Code review also complete by Joe, lots of local r_t by Engel, JoeB and JonM.
Merging.

PR #1034, opened 2014-10-15
Add new vnode command to get vnodeid(s) for a preflist

russelldb:
Take your pick, there is this way (with test here https://github.com/basho/riak_test/pull/689)


jonmeredith:
-1 

Prefer `vnode_status()` method.

PR #1037, opened 2014-10-24
Patch to prevent calculation of cpu stats.

jonmeredith:
For the Unices that aren't directly supported in the cpu_sup.c
(like FreeBSD) they make do multiple fork/execs to run system
utilities to gather stats.  If multiple instances of Riak
are running on a server (devrel or otherwise) it starts to stack up.

Add [{riak_kv, disable_cpu_stats, true}] to advanced.config to disable
reporting.  Will decide if this is a common enough setting to add to
cuttlefish at a future point.


HMcKelvey:
add to Jira

yorkshire-steve:

PR #1040, opened 2014-11-11
Add per key actor epochs

russelldb:
See riak_kv#679 and the associated platform_task RFC and summary.

Requires https://github.com/basho/riak_core/pull/646

This PR addresses the "doomstone", backup-restore, and some byzantine flavours of the kv679 bug.

The RFC explains the mechanism in detail but briefly:
1. add a persisted to disk vnode counter (persisted with leases, aysnc)
2. when ever a key is written to for the "first time" by a vnode, create an epoch actor for the key by concatenating vnodeid+counter (and increment the counter)

This ensures that a first time write for a key gets a new actor, this is the epoch for the key. It means we don't mix up deleted+re-created keys {a,1} event with the original {a, 1} event for some key, by ensuring an actor per epoch, without causing a keyspace wide actor explosion.

I hope that is enough details for the review to start. I'll start work on EQC for the status_mgr and the new vnode functions in the mean time.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -111,7 +111,10 @@
                 target :: pid()}).
 
 -record(counter_state, {
-          %% The number of writes co-ordinated by this vnode
+          %% The number of new epoch writes co-ordinated by this vnode
+          %% What even is a "key epoch?" It is any time a key is
+          %% (re)created. A new write, a write not yet coordinated by
+          %% this vnode, a write where local state is unreadable.
           cnt = 0 :: non_neg_integer(),
           %% Counter leased up-to. For totally new state/id
           %% this will be that flush threshold See config value
@@ -148,6 +151,7 @@
 -type index() :: non_neg_integer().
 -type state() :: #state{}.
 -type vnodeid() :: binary().
+-type counter_lease_error() :: {error, counter_lease_max_errors | counter_lease_timeout}.
 
 -define(MD_CACHE_BASE, "riak_kv_vnode_md_cache").
 -define(DEFAULT_HASHTREE_TOKENS, 90).
@@ -1409,7 +1413,6 @@ prepare_put(State=#state{mod=Mod,
                                      _ ->
                                          riak_object:prune_vclock(AMObj, PruneTime, BProps)
                                  end,
-                    %% @todo What ID for CRDTs? The Epoch ID? The bare vnode id?
                     case handle_crdt(Coord, CRDTOp, ActorId, ObjToStore) of
                         {error, E} ->
                             {{fail, Idx, E}, PutArgs, State2};
@@ -1907,7 +1910,6 @@ do_diffobj_put({Bucket, Key}=BKey, DiffObj,
             %% if out of date.  Ok to set VId/Starttime undefined as
             %% they are not used for non-coordinating puts.
             case put_merge(false, false, OldObj, DiffObj, undefined, undefined) of
-                %% Only coordinating puts mutate vnode state, so ignore returned state
                 {oldobj, _} ->
                     {ok, ModState};
                 {newobj, NewObj} ->
@@ -2312,7 +2314,7 @@ update_counter(State=#state{counter=CounterState}) ->
 %% the lease, do nothing.  If not yet blocking(equal) and we already
 %% asked for a new lease, do nothing.
 maybe_lease_counter(State=#state{counter=#counter_state{cnt=Lease, lease=Lease}}) ->
-    %% Block until we get a new lease, or crash
+    %% Block until we get a new lease, or crash the vnode
     {ok, NewState} = blocking_lease_counter(State),
     NewState;
 maybe_lease_counter(State=#state{counter=#counter_state{leasing=true}}) ->
@@ -2335,12 +2337,21 @@ maybe_lease_counter(State) ->
 %% from the mgr. Block until we receive a lease. If we get an exit,
 %% retry a number of times. If we get neither in a reasonable time,
 %% return an error.
+-spec blocking_lease_counter(#state{}) ->
+                                    {ok, #state{}} |
+                                    counter_lease_error().
 blocking_lease_counter(State) ->
     {MaxErrs, MaxTime} = get_counter_wait_values(),
     blocking_lease_counter(State, {0, MaxErrs, MaxTime}).
 
-blocking_lease_counter(State, {MaxErrs, MaxErrs, _}) ->
-    {error, counter_lease_max_errors, State};
+-spec blocking_lease_counter(#state{}, {Errors :: non_neg_integer(),
+                                        MaxErrors :: non_neg_integer(),
+                                        TimeRemainingMillis :: non_neg_integer()}
+                            ) ->
+                                    {ok, #state{}} |
+                                    counter_lease_error().
+blocking_lease_counter(_State, {MaxErrs, MaxErrs, _}) ->
+    {error, counter_lease_max_errors};
 blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
     #state{idx=Index, status_mgr_pid=Pid, counter=CounterState} = State,
     #counter_state{lease_size=LeaseSize} = CounterState,
@@ -2358,7 +2369,7 @@ blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
             {ok, State#state{counter=NewCS, vnodeid=VnodeId}}
     after
         MaxTime ->
-            {error, counter_lease_timeout, State}
+            {error, counter_lease_timeout}
     end.
 
 %% @private get the configured values for blocking waiting on
@@ -2368,7 +2379,7 @@ blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
                                       MaxTime :: non_neg_integer()}.
 get_counter_wait_values() ->
     MaxErrors = non_neg_env(riak_kv, counter_lease_errors, ?DEFAULT_CNTR_LEASE_ERRS),
-    MaxTime = non_neg_env(riak_kv, counter_lease_errors, ?DEFAULT_CNTR_LEASE_ERRS),
+    MaxTime = non_neg_env(riak_kv, counter_lease_timeout, ?DEFAULT_CNTR_LEASE_TO),
     {MaxErrors, MaxTime}.
 
 %% @private we don't want to crash riak because of a dodgy config
@@ -2384,7 +2395,9 @@ non_neg_env(App, EnvVar, Default) when is_integer(Default),
                N > 0 ->
             N;
         X ->
-            lager:warning("Non-integer/Negative integer ~p for vnode counter config ~p", [X, EnvVar]),
+            lager:warning("Non-integer/Negative integer ~p for vnode counter config ~p."
+                          " Using default ~p",
+                          [X, EnvVar, Default]),
             Default
     end.
 
@@ -2402,6 +2415,8 @@ non_neg_env(App, EnvVar, Default) when is_integer(Default),
                           riak_object:riak_object()) ->
                                  {binary(), #state{}}.
 maybe_new_key_epoch(false, State, _, _) ->
+    %% Never add a new key epoch when not coordinating
+    %% @TODO (rdb) need to mark actor as dirty though.
     {State#state.vnodeid, State};
 maybe_new_key_epoch(true, State, LocalObj, IncomingObj) ->
     #state{vnodeid=VId} = State,
@@ -2445,12 +2460,19 @@ key_epoch_actor(ActorBin, Cntr) ->
 %% @private highest actor is the latest/greatest epoch actor for a
 %% key. It is the actor we want to increment for the current event,
 %% given that we are not starting a new epoch for the key.  Must work
-%% with non-epochal and epochal actors.
+%% with non-epochal and epochal actors.  The return tuple is
+%% `{ActorId, Epoch, Counter}' where `ActorId' is the highest ID
+%% starting with `ActorBase' that has acted on this key, undefined if
+%% never acted before. `KeyEpoch' is the highest epoch for the
+%% `ActorBase'. `Counter' is the greatest event seen by the `VnodeId'.
 -spec highest_actor(binary(), riak_object:riak_object()) ->
-    {binary() | undefined, non_neg_integer(), non_neg_integer()}.
+    {ActorId :: binary() | undefined,
+     KeyEpoch :: non_neg_integer(),
+     Counter :: non_neg_integer()}.
 highest_actor(ActorBase, Obj) ->
     Actors = riak_object:all_actors(Obj),
     {Actor, Epoch} = highest_actor(ActorBase, Actors, undefined, 0),
+    %% get the greatest event for the highest/latest actor
     {Actor, Epoch, riak_object:actor_counter(Actor, Obj)}.
 
 %% @private find the highest actor and epoch for the `ActorBase'. If
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -718,9 +718,7 @@ handle_command(?KV_VNODE_STATUS_REQ{},
     #counter_state{cnt=Cnt, lease=Lease, lease_size=LeaseSize, leasing=Leasing} = CS,
     CounterStatus = [{counter, Cnt}, {counter_lease, Lease},
                      {counter_lease_size, LeaseSize}, {counter_leasing, Leasing}],
-    %% @TODO remove the counter/vnodeid, or at least make it _work_
-    %% with the stats code
-    VNodeStatus = [{backend, BackendStatus}, {vnodeid, VId} | CounterStatus],
+    VNodeStatus = [BackendStatus, {vnodeid, VId} | CounterStatus],
     {reply, {vnode_status, Index, VNodeStatus}, State};
 handle_command({reformat_object, BKey}, _Sender, State) ->
     {Reply, UpdState} = do_reformat(BKey, State),
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1910,9 +1910,9 @@ do_diffobj_put({Bucket, Key}=BKey, DiffObj,
             %% they are not used for non-coordinating puts.
             case put_merge(false, false, OldObj, DiffObj, undefined, undefined) of
                 %% Only coordinating puts mutate vnode state, so ignore returned state
-                {oldobj, _, StateData} ->
+                {oldobj, _} ->
                     {ok, ModState};
-                {newobj, NewObj, StateData} ->
+                {newobj, NewObj} ->
                     AMObj = enforce_allow_mult(NewObj, riak_core_bucket:get_bucket(Bucket)),
                     case IndexBackend of
                         true ->
@@ -2443,7 +2443,7 @@ key_epoch_actor(ActorBin, Cntr) ->
 %% given that we are not starting a new epoch for the key.  Must work
 %% with non-epochal and epochal actors.
 -spec highest_actor(binary(), riak_object:riak_object()) ->
-    binary().
+    {binary(), non_neg_integer()}.
 highest_actor(ActorBase, Obj) ->
     Actors = riak_object:all_actors(Obj),
     highest_actor(ActorBase, Actors, ActorBase, 0).
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -112,12 +112,15 @@
 
 -record(counter_state, {
           %% The number of writes co-ordinated by this vnode
-          counter :: non_neg_integer(),
+          cnt = 0 :: non_neg_integer(),
           %% Counter leased up-to. For totally new state/id
           %% this will be that flush threshold See config value
           %% `{riak_kv, counter_lease_size}'
-          counter_lease :: non_neg_integer(),
-          counter_lease_size :: non_neg_integer()}).
+          lease = 0 :: non_neg_integer(),
+          lease_size = 0 :: non_neg_integer(),
+          %% Has a lease been requested but not granted yet
+          leasing = false :: boolean()
+         }).
 
 -record(state, {idx :: partition(),
                 mod :: module(),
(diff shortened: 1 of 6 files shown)


src/riak_kv_vnode.erl
```diff
@@ -2286,102 +2297,202 @@ new_md_cache(VId) ->
...
+maybe_lease_counter(State=#state{counter=#counter_state{cnt=Lease, lease=Lease}}) ->
+    %% Block until we get a new lease, or crash the vnode
+    {ok, NewState} = blocking_lease_counter(State),
+    NewState;
+maybe_lease_counter(State=#state{counter=#counter_state{leasing=true}}) ->
+    %% not yet at the blocking stage, waiting on a lease
+    State;
+maybe_lease_counter(State) ->
+    #state{status_mgr_pid=MgrPid, counter=CS=#counter_state{cnt=Cnt, lease=Lease, lease_size=LeaseSize}} = State,
+    %% @TODO configurable??
```
seancribbs, line 2325: Yes, this should probably be configurable/tunable.

src/riak_kv_vnode.erl
```diff
@@ -2286,102 +2297,202 @@ new_md_cache(VId) ->
...
+                            ) ->
+                                    {ok, #state{}} |
+                                    counter_lease_error().
+blocking_lease_counter(_State, {MaxErrs, MaxErrs, _MaxTime}) ->
+    {error, counter_lease_max_errors};
+blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
+    #state{idx=Index, status_mgr_pid=Pid, counter=CounterState} = State,
+    #counter_state{lease_size=LeaseSize} = CounterState,
+    Start = os:timestamp(),
+    receive
```
seancribbs, line 2359: Could this whole blocking paradigm be replaced with `gen_server:call` with a timeout? Also isn't it possible to enter this receive without having sent a lease request yet? Where is the original cast?

src/riak_kv_vnode.erl
```diff
@@ -2286,102 +2297,202 @@ new_md_cache(VId) ->
...
+    {ok, NewState} = blocking_lease_counter(State),
+    NewState;
+maybe_lease_counter(State=#state{counter=#counter_state{leasing=true}}) ->
+    %% not yet at the blocking stage, waiting on a lease
+    State;
+maybe_lease_counter(State) ->
+    #state{status_mgr_pid=MgrPid, counter=CS=#counter_state{cnt=Cnt, lease=Lease, lease_size=LeaseSize}} = State,
+    %% @TODO configurable??
+    %% has more than 80% of the lease been used?
+    CS2 = case (Lease - Cnt) =< 0.2 * LeaseSize of
```
seancribbs, line 2327: All of the expressions in this are guard expressions, an `if` would be clearer in this case.

src/riak_kv_vnode.erl
```diff
@@ -1484,11 +1553,19 @@ select_newest_content(Mult) ->
 
 %% @private
 put_merge(false, true, _CurObj, UpdObj, _VId, _StartTime) -> % coord=false, LWW=true
+    %% @TODO Do we need to mark the clock dirty here? I think so
```
seancribbs, line 1556: Why do you need to mark the clock dirty here? LWW should ignore the clock.

src/riak_kv_vnode.erl
```diff
@@ -1368,6 +1435,8 @@ prepare_new_put(true, RObj, VId, StartTime, CRDTOp) ->
     %% Make a new crdt, stuff it in the riak_object
     do_crdt_update(VClockUp, VId, CRDTOp);
 prepare_new_put(false, RObj, _VId, _StartTime, _CounterOp) ->
+    %% @TODO Not coordindating, not found local, is there an entry for
```
seancribbs, line 1438: Is the goal of marking the entry/clock dirty to ensure that it gets a new epoch next time this actor coordinates? If so, then :+1: to that. That seems like an early indicator that data was lost.

src/riak_kv_vnode.erl
```diff
@@ -1279,12 +1346,12 @@ prepare_put(#state{vnodeid=VId,
...
-                        false;
-                    _ ->
-                        true
-                end
+                %% We need to perform a local get, to merge contents,
+                %% if the local object has events unseen by the
+                %% incoming object. If the incoming object descends
+                %% the cache (i.e. has seen all its events) no need to
+                %% do a local get and merge, just overwrite.
+                not riak_object:vclock_descends(RObj, Clock)
```
seancribbs, line 1354: I love this clarification.

src/riak_kv_vnode.erl
```diff
@@ -1174,18 +1239,18 @@ raw_put({Idx, Node}, Key, Obj) ->
...
-        BProps ->
-            BProps
-    end,
-    case proplists:get_value(rr, Options, false) of
-        true ->
-            PruneTime = undefined;
-        false ->
-            PruneTime = StartTime
-    end,
+    BProps =  case proplists:get_value(bucket_props, Options) of
```
seancribbs, line 1242: Thank you for removing the case-export warnings.

src/riak_kv_vnode.erl
```diff
@@ -1134,8 +1176,31 @@ handle_info({final_delete, BKey, RObjHash}, State = #state{mod=Mod, modstate=Mod
...
+    State2 = State#state{counter=CS1, vnodeid=VnodeId},
+    {ok, State2}.
+
+handle_exit(Pid, Reason, State=#state{status_mgr_pid=Pid, idx=Index, counter=CntrState}) ->
+    lager:error("Vnode status manager exit ~p", [Reason]),
+    %% The status manager died, start a new one
+    #counter_state{lease_size=LeaseSize, leasing=Leasing} = CntrState,
+    {ok, NewPid} = riak_kv_vnode_status_mgr:start_link(self(), Index),
+
+    case Leasing of
```
seancribbs, line 1196: This seems like a clear spot for `if` instead of `case`.

src/riak_kv_vnode.erl
```diff
@@ -1134,8 +1176,31 @@ handle_info({final_delete, BKey, RObjHash}, State = #state{mod=Mod, modstate=Mod
                    {{error, _}, ModState1} ->
                        State#state{modstate=ModState1}
                end,
-    {ok, UpdState}.
-
+    {ok, UpdState};
+handle_info({counter_lease, {FromPid, VnodeId, NewLease}}, State=#state{status_mgr_pid=FromPid}) ->
+    #state{counter=CounterState} = State,
+    #counter_state{lease=OldLease} = CounterState,
+    %% @TODO (rdb) Paranoia? If the NewLease < OldLease something is very broken,
```
seancribbs, line 1183: We should discuss the implications of the NewLease being smaller than the OldLease. Seems like reordered messages could cause it (but are nigh-impossible between two local processes), but could dropped messages? Perhaps this needs a PULSE model.

src/riak_kv_vnode.erl
```diff
@@ -2286,102 +2297,202 @@ new_md_cache(VId) ->
...
+    MaxTime = non_neg_env(riak_kv, counter_lease_timeout, ?DEFAULT_CNTR_LEASE_TO),
+    {MaxErrors, MaxTime}.
+
+%% @private we don't want to crash riak because of a dodgy config
+%% value, and by this point, cuttlefish be praised, we should have
+%% sane values. However, if not, ignore negative values for timeout
+%% and max errors, use the defaults, and log the insanity. Note this
+%% expects the macro configured defaults to be positive integers!
+-spec non_neg_env(atom(), atom(), pos_integer()) -> pos_integer().
+non_neg_env(App, EnvVar, Default) when is_integer(Default),
```
seancribbs, line 2391: For the future, this seems like something that could be in `app_helper` in core.

src/riak_kv_vnode.erl
```diff
@@ -2286,102 +2297,202 @@ new_md_cache(VId) ->
...
+%% @private generate an epoch actor, and update the vnode state.
+-spec new_key_epoch(#state{}) -> {EpochActor :: binary(), #state{}}.
+new_key_epoch(State) ->
+    NewState=#state{counter=#counter_state{cnt=Cntr}, vnodeid=VId} = update_counter(State),
+    EpochId = key_epoch_actor(VId, Cntr),
+    {EpochId, NewState}.
+
+%% @private generate a new epoch ID for a key
+-spec key_epoch_actor(vnodeid(), pos_integer()) -> binary().
+key_epoch_actor(ActorBin, Cntr) ->
```
seancribbs, line 2457: It seems you only call this from the function above. Do we need a separate abstraction for the construction of this ID? If we do, shouldn't we have a deconstructor too (probably used by `highest_actor()`)?

src/riak_kv_vnode.erl
```diff
@@ -2286,102 +2297,202 @@ new_md_cache(VId) ->
...
+%% starting with `ActorBase' that has acted on this key, undefined if
+%% never acted before. `KeyEpoch' is the highest epoch for the
+%% `ActorBase'. `Counter' is the greatest event seen by the `VnodeId'.
+-spec highest_actor(binary(), riak_object:riak_object()) ->
+    {ActorId :: binary() | undefined,
+     KeyEpoch :: non_neg_integer(),
+     Counter :: non_neg_integer()}.
+highest_actor(ActorBase, Obj) ->
+    Actors = riak_object:all_actors(Obj),
+    {Actor, Epoch} = highest_actor(ActorBase, Actors, undefined, 0),
```
seancribbs, line 2474: Like I mentioned on old commits, this feels like a fold since you aren't bailing early from the processing.  How about this refactor:

``` erlang
{Actor, Epoch} = lists:foldl(find_highest_actor(ActorBase), {undefined, 0}, Actors),
% ...

find_highest_actor(ActorBase) -> 
    fun(ActorID, Acc) -> 
        find_highest_actor(ActorBase, ActorID, Acc)
    end.

find_highest_actor(ActorBase, <<ActorBase:8/binary, EpochCntr:32/integer>>=A, {_HighestActor, MaxEpoch}) when
      EpochCntr > MaxEpoch ->
    {A, EpochCntr};
find_highest_actor(ActorBase, <<ActorBase:8/binary, EpochCntr:32/integer>>, {_HighestActor, MaxEpoch}=Acc) when
      EpochCntr =< MaxEpoch -> 
    %% TODO: (sdc) This clause may not even be necessary! When the
    %% EpochCntr isn't greater, we just pass along the existing
    %% highest actor.
    Acc;
%% Since an actor without an epoch is lower than an actor with one,
%% this means in the unmatched case, the ActorBase passes through as
%% the highest actor, and the epoch (of zero) passes through too.
find_highest_actor(_ActorBase, _, Acc) ->
    Acc.
```

Or even simpler:

``` erlang
find_highest_actor(ActorBase) -> 
    fun(<<MatchingBase:8/binary, EpochCntr:32/integer>>=A, {_HighestActor, MaxEpoch}) 
        when MatchingBase == ActorBase andalso 
             EpochCntr > MaxEpoch -> 
               {A, EpochCntr};
        (_, Acc) ->
               %% Since an actor without an epoch is lower than an
               %% actor with one, this means in the unmatched case,
               %% the ActorBase passes through as the highest actor,
               %% and the epoch (of zero) passes through too.
               %%
               %% NB (sdc): this also covers where the ActorBase
               %% doesn't match, i.e. it is some other actor entirely.
               Acc
    end.
```

I guess the reason I like to try to identify things that look like folds an reimplement them as such is it forces me to find ways to make things more succinct. In this instance, I found some of those clauses aren't even needed -- two are handled by the implementation of `lists:foldl/3`, and one is a subset of a catch-all case.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+-record(state, {
+          %% vnode status directory
+          status_file :: undefined | file:filename(),
+          %% vnode index
+          index :: undefined | non_neg_integer(),
+          %% The vnode pid this mgr belongs to
+          vnode_pid :: undefined | pid()
+         }).
+
+-type status() :: [proplists:property()] | [].
```
seancribbs, line 52: The empty list is not strictly necessary in this type. Also, is there any case where you will have bare atoms in the list? If not, `orddict:orddict()` might be better all around.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+-type init_args() :: [init_arg()].
```
seancribbs, line 53: Might be safer to use a tuple instead of a list, then the ordering is explicit and dialyzer can find incorrect orderings easier.

``` erlang
-type init_args() :: {pid(), non_neg_integer()}.
```

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+%% @end
+%%--------------------------------------------------------------------
+-spec get_vnodeid_and_counter(pid(), non_neg_integer()) ->
+                                     {ok, {VnodeId :: binary(),
+                                           Counter :: non_neg_integer(),
+                                           LeaseSize :: non_neg_integer()}}.
+get_vnodeid_and_counter(Pid, LeaseSize) when is_integer(LeaseSize),
+                                          LeaseSize > 0,
+                                          LeaseSize =< ?MAX_CNTR ->
+    %% @TODO decide on a sane timeout for getting a vnodeid/cntr pair
```
seancribbs, line 90: Yes, you should probably use `infinity` timeout here, or otherwise have a safety mechanism that ensures you don't have false negatives. There could be a race with the disk and the vnode deciding to give up on the counter process.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+
+%%--------------------------------------------------------------------
+%% @doc Blocking call to remove the vnode id and counter and from
+%% disk. Used when a vnode has finished and will not act again.
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec clear_vnodeid(pid()) -> {ok, cleared}.
+clear_vnodeid(Pid) ->
+    gen_server:call(Pid, clear).
```
seancribbs, line 121: In `clear_vnodeid`, `status`, and `stop`, consider whether the five-second default timeout makes sense at all. It may be better as `infinity`.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+                           {non_neg_integer(), non_neg_integer(), binary(), status()}.
+get_counter_lease(LeaseSize, Status) ->
+    PrevLease = proplists:get_value(counter, Status, ?MAX_CNTR),
+    VnodeId0 = proplists:get_value(vnodeid, Status, undefined),
+    if
+        (VnodeId0 == undefined) or (PrevLease == ?MAX_CNTR) ->
+            {VnodeId, Status2} = assign_vnodeid(erlang:now(),
+                                                riak_core_nodeid:get(),
+                                                Status),
+            {0, LeaseSize, VnodeId, replace(counter, LeaseSize, Status2)};
```
seancribbs, line 219: Like I mentioned in the type specs above, if no duplicates or bare keys are allowed or used in the status lists, it might as well be an `orddict`. Then you can just use `orddict:store/3`.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+
+%% @private read the vnode status from `File'. Returns `{ok,
+%% status()}' or `{error, Reason}'. If the file does not exist, and
+%% empty status is returned.
+-spec read_vnode_status(file:filename()) -> {ok, status()} |
+                                       {error, term()}.
+read_vnode_status(File) ->
+    case file:consult(File) of
+        {ok, [Status]} when is_list(Status) ->
+            {ok, proplists:delete(version, Status)};
```
seancribbs, line 278: Should you be checking the version read from disk to ensure it's the same?

seancribbs, line 278: Also, if you switch to `orddict` for the status, you'll want to use `from_list` and `erase` to accomplish the same thing as `proplists:delete`.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+-ifdef(TEST).
+
+-ifdef(EQC).
+%% prop_monotonic_counter() ->
+%%     ok.
+
+-endif.
+
+
+replace_test() ->
```
seancribbs, line 304: This test will become unnecessary if you use `orddict`.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,394 @@
...
+%% not be written at any other place/time in the system.
+-spec write_vnode_status(status(), file:filename()) -> ok.
+write_vnode_status(Status, File) ->
+    VersionedStatus = replace(version, ?VNODE_STATUS_VERSION, Status),
+    riak_core_util:replace_file(File, io_lib:format("~p.", [VersionedStatus])).
+
+-ifdef(TEST).
+
+-ifdef(EQC).
+%% prop_monotonic_counter() ->
```
seancribbs, line 298: Did this get moved to the separate EQC module? If so, no need to keep it here.

src/riak_object.erl
```diff
@@ -75,7 +75,8 @@
 -define(EMPTY_VTAG_BIN, <<"e">>).
 
 -export([new/3, new/4, ensure_robject/1, ancestors/1, reconcile/2, equal/2]).
--export([increment_vclock/2, increment_vclock/3]).
+-export([increment_vclock/2, increment_vclock/3, prune_vclock/3, vclock_descends/2, all_actors/1]).
```
seancribbs, line 78: I love that you increased the encapsulation here. We should look to move more things from the vnode into the riak_object.

src/riak_kv_vnode.erl
russelldb, line 2359: re gen_server call, no, I think that breaks the encapsulation and serialization of file access, and lays us open to races.
re entering receive without requesting a lease: how? The state for counter must be "leasing=true"

src/riak_kv_vnode.erl
russelldb, line 2327: If you say so.

src/riak_kv_vnode.erl
russelldb, line 1556: But we allow people to change from LWW->OTHER, so a detected dirty clock should still be marked for future reads.

src/riak_kv_vnode.erl
russelldb, line 1438: But this is future work, of course, as it requires further riak_object metadata.

src/riak_kv_vnode.erl
russelldb, line 1196: Personally _HATE_ erlang `if` especially in this case, as it mentally comes out as

   if true -> thing;
     true -> other_thing
    end

src/riak_kv_vnode.erl
russelldb, line 1183: I agree, let's discuss.

Does erlang  drop messages between local processes?

src/riak_kv_vnode.erl
cmeiklejohn, line 1196: I'm with @russelldb.

src/riak_kv_vnode.erl
russelldb, line 2391: Noted, happy to move

src/riak_kv_vnode.erl
russelldb, line 2457: Hmmm. Ok.

src/riak_kv_vnode.erl
russelldb, line 2474: Ok

src/riak_kv_vnode_status_mgr.erl
russelldb, line 52: I don't know, it is inherited.

src/riak_kv_vnode_status_mgr.erl
russelldb, line 53: I can do that? This is cargo-culted from all other Riak gen_servers

src/riak_kv_vnode_status_mgr.erl
russelldb, line 90: Can you describe this race? Seems that the time out causes the vnode to crash, which causes the manager to crash. Then the vnode comes back up, and we start over.

seancribbs, line 90: So, if the vnode does a strict match, that makes sense. Thanks for clarifying.

src/riak_kv_vnode.erl
russelldb, line 2325: Yeah. It is hard to decide if it should be. Trying to figure out a scenario where you'd want to change this. And adding it adds more chance to shoot yourself in the foot. And more complex validation of the config item. And can it change in the vnode's lifetime? Do we need to add it to state, or check the env var every iteration?

I'm not against making it configurable, I'd just like a better understanding of the benefit vs. risk/complexity.

src/riak_kv_vnode.erl
russelldb, line 1183: Actually this is very broken. The old lease can legitimately be smaller than the old lease. If we've gone over the `?MAX_INT` boundary and got a new vnodeid, then the new lease is expected to be smaller, and using `max` is wrong and broken, we'll be handing out counters that are not on disk.

src/riak_kv_vnode.erl
russelldb, line 2457: So, we'd be tying ourselves to a set length for the `ActorBin` part, no? And we'd need versioning in case that changed in future. Maybe `highest_actor()` should use the length of `VnodeId` as the base to match from? Or can we trust the `VnodeId` is 8 bytes for ever and add the deconstructor as you suggest?

russelldb, line 2457: I changed this so that we don't know, nor need to know, that `vnodeid` is currently 8 bytes. This makes changing the `vnodeid` scheme down the road simpler, and keeps the matching backwards/forwards compatible.

src/riak_kv_vnode.erl
russelldb, line 2474: Just re-wrote it as you suggest, indeed it is more concise and there were redundant clauses. Thank you!

src/riak_kv_vnode_status_mgr.erl
russelldb, line 278: WRT to checking version, maybe should check that it is 1 or 2? Since older vnode statuses can be read by the new manager.

russelldb:
I rebased against 2.0 before it moves too far away. I addressed your comments. Sadly I can't push to this branch now. Instead I opened a new PR and ref'd this one from riak_kv#1053

Issue #1036, opened 2014-10-18
Enable more complex backend filtering for multi backend folds [JIRA: RIAK-1750]


bsparrow435:
Currently the multi backend fold only has a backend filter for 2i reindexing[1]. It would be very helpful in support situations if the {backend, [LIST_OF_BACKENDS]} option was supported so we only scan a subset of backends.

[1] https://github.com/basho/riak_kv/blob/develop/src/riak_kv_multi_backend.erl#L517-L525

DSomogyi:
Comment for Jira.

jrwest:
Is it known that these "comment for JIRA" comments spam everyone who is watching the repo not just basho employees? It's nice to be able to watch the project for progress but this makes it rather a pain in the ass. It also reflects poorly on the open source project. I understand that you need to integrate JURA and GitHub but is there not a better way?

PR #1041, opened 2014-11-12
Add backend filter fold option to multi_backend [JIRA: RIAK-1594]

nickelization:
Adds an option that can be passed to multi_backend bucket folds. The
option {backend, [BACKEND_LIST]} will cause the fold to skip over any
backends not explicitly named in the backend list.

This enhancement addresses issue #1036.

This is my first Basho PR, so apologies in advance if I've neglected any
important details! Fingers crossed :)

### DIFF
--- a/src/riak_kv_multi_backend.erl
+++ b/src/riak_kv_multi_backend.erl
@@ -85,7 +85,7 @@
 -define(CAPABILITIES, [async_fold, index_reformat]).
 -define(ANY_CAPABILITIES, [indexes, iterator_refresh]).
 
--record (state, {backends :: [{atom(), atom(), term()}],
+-record (state, {backends :: [{atom(), atom(), term()}], % [{BackendName, BackendModule, SubState}]
                  default_backend :: atom()}).
 
 -type state() :: #state{}.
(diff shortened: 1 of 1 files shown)


bsparrow435:
Thank you so much for the PR! We'll get it assigned to a milestone and reviewed as soon as possible.

src/riak_kv_multi_backend.erl
```diff
@@ -85,7 +85,7 @@
 -define(CAPABILITIES, [async_fold, index_reformat]).
 -define(ANY_CAPABILITIES, [indexes, iterator_refresh]).
 
--record (state, {backends :: [{atom(), atom(), term()}],
+-record (state, {backends :: [{atom(), atom(), term()}], % [{BackendName, BackendModule, SubState}]
```
seancribbs, line 88: Wow, I didn't realize this assumed atom for the BackendName. In the [configs](https://github.com/basho/riak_kv/blob/develop/priv/multi_backend.schema#L59) they are all binaries so they can be set via bucket props.

nickelization, line 88: Huh, weird. I didn't realize they could be anything other than atoms, but I guess other types should work just as well. I'm still learning the basics with Riak, but are bucket props all supposed to be binaries? From the type specs in riak_core_bucket_props, it looks like anything should be valid, and in the unit test I wrote it seems like atoms worked fine. Or, then again maybe I'm totally misinterpreting something....

nickelization:
Just realized this never got approved. Any changes needed?

andrewjstone:
+1 5558738

Code looks good, test passes

andrewjstone:
@borshop merge

PR #1006, opened 2014-07-29
use exometer metrics

uwiger:
This is a new pull request for the Exometer adaptation, with a cleaner commit history (all changes collected into one diff).

See riak [PR #576](https://github.com/basho/riak/pull/576)

The old riak_kv PR was #748 (for comment history)

PR #1044, opened 2014-11-14
Introduce exometer metrics into 2.0

uwiger:
This PR is an adaptation of PR #1006 ("Use exometer metrics"), which was relative to the 'develop' branch.

The current PR is relative to the 2.0 branch. Note that rebar.config is not affected.


jburwell:
:+1: cf2d199

jburwell:
@borshop merge

PR #1050, opened 2014-11-20
Add skipped_repairs stat.

uwiger:
lager message on stat_update_error changed from error to debug.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -88,7 +88,7 @@ maybe_dispatch_to_sidejob(false, Arg) ->
     ok.
 
 stat_update_error(Arg, Class, Error) ->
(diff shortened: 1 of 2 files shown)


jburwell:
@ulf Does skipped_read_repair stat need to be added the legacy stat map in both riak_kv and riak_core?

uwiger:
@jburwell, yes if you want to report it via both HTTP (riak_kv) and riak-admin status (riak_core).

jburwell:
Previously, we have collected the skipped_read_repairs stat, but not provided in `riak-admin status` or via the HTTP endpoint.  In discussing with @bsparrow435, it seems reasonable to add it as part of this bug fix.

/cc @uwiger

jburwell:
:+1: d23478e

jburwell:
@borshop merge

PR #1053, opened 2014-11-24
per-key actor epoch (kv679) rebased on 2.0

russelldb:
See original PR for discussion, rebased for easier merge (before squashing.)

This PR addresses most of the comments from the original PR at basho/riak_kv#1040

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -111,7 +111,10 @@
                 target :: pid()}).
 
 -record(counter_state, {
-          %% The number of writes co-ordinated by this vnode
+          %% The number of new epoch writes co-ordinated by this vnode
+          %% What even is a "key epoch?" It is any time a key is
+          %% (re)created. A new write, a write not yet coordinated by
+          %% this vnode, a write where local state is unreadable.
           cnt = 0 :: non_neg_integer(),
           %% Counter leased up-to. For totally new state/id
           %% this will be that flush threshold See config value
@@ -148,6 +151,7 @@
 -type index() :: non_neg_integer().
 -type state() :: #state{}.
 -type vnodeid() :: binary().
+-type counter_lease_error() :: {error, counter_lease_max_errors | counter_lease_timeout}.
 
 -define(MD_CACHE_BASE, "riak_kv_vnode_md_cache").
 -define(DEFAULT_HASHTREE_TOKENS, 90).
@@ -1409,7 +1413,6 @@ prepare_put(State=#state{mod=Mod,
                                      _ ->
                                          riak_object:prune_vclock(AMObj, PruneTime, BProps)
                                  end,
-                    %% @todo What ID for CRDTs? The Epoch ID? The bare vnode id?
                     case handle_crdt(Coord, CRDTOp, ActorId, ObjToStore) of
                         {error, E} ->
                             {{fail, Idx, E}, PutArgs, State2};
@@ -1907,7 +1910,6 @@ do_diffobj_put({Bucket, Key}=BKey, DiffObj,
             %% if out of date.  Ok to set VId/Starttime undefined as
             %% they are not used for non-coordinating puts.
             case put_merge(false, false, OldObj, DiffObj, undefined, undefined) of
-                %% Only coordinating puts mutate vnode state, so ignore returned state
                 {oldobj, _} ->
                     {ok, ModState};
                 {newobj, NewObj} ->
@@ -2312,7 +2314,7 @@ update_counter(State=#state{counter=CounterState}) ->
 %% the lease, do nothing.  If not yet blocking(equal) and we already
 %% asked for a new lease, do nothing.
 maybe_lease_counter(State=#state{counter=#counter_state{cnt=Lease, lease=Lease}}) ->
-    %% Block until we get a new lease, or crash
+    %% Block until we get a new lease, or crash the vnode
     {ok, NewState} = blocking_lease_counter(State),
     NewState;
 maybe_lease_counter(State=#state{counter=#counter_state{leasing=true}}) ->
@@ -2335,12 +2337,21 @@ maybe_lease_counter(State) ->
 %% from the mgr. Block until we receive a lease. If we get an exit,
 %% retry a number of times. If we get neither in a reasonable time,
 %% return an error.
+-spec blocking_lease_counter(#state{}) ->
+                                    {ok, #state{}} |
+                                    counter_lease_error().
 blocking_lease_counter(State) ->
     {MaxErrs, MaxTime} = get_counter_wait_values(),
     blocking_lease_counter(State, {0, MaxErrs, MaxTime}).
 
-blocking_lease_counter(State, {MaxErrs, MaxErrs, _}) ->
-    {error, counter_lease_max_errors, State};
+-spec blocking_lease_counter(#state{}, {Errors :: non_neg_integer(),
+                                        MaxErrors :: non_neg_integer(),
+                                        TimeRemainingMillis :: non_neg_integer()}
+                            ) ->
+                                    {ok, #state{}} |
+                                    counter_lease_error().
+blocking_lease_counter(_State, {MaxErrs, MaxErrs, _}) ->
+    {error, counter_lease_max_errors};
 blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
     #state{idx=Index, status_mgr_pid=Pid, counter=CounterState} = State,
     #counter_state{lease_size=LeaseSize} = CounterState,
@@ -2358,7 +2369,7 @@ blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
             {ok, State#state{counter=NewCS, vnodeid=VnodeId}}
     after
         MaxTime ->
-            {error, counter_lease_timeout, State}
+            {error, counter_lease_timeout}
     end.
 
 %% @private get the configured values for blocking waiting on
@@ -2368,7 +2379,7 @@ blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
                                       MaxTime :: non_neg_integer()}.
 get_counter_wait_values() ->
     MaxErrors = non_neg_env(riak_kv, counter_lease_errors, ?DEFAULT_CNTR_LEASE_ERRS),
-    MaxTime = non_neg_env(riak_kv, counter_lease_errors, ?DEFAULT_CNTR_LEASE_ERRS),
+    MaxTime = non_neg_env(riak_kv, counter_lease_timeout, ?DEFAULT_CNTR_LEASE_TO),
     {MaxErrors, MaxTime}.
 
 %% @private we don't want to crash riak because of a dodgy config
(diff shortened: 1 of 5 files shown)


src/riak_kv_vnode.erl
```diff
@@ -398,7 +435,11 @@ init([Index]) ->
     IndexBufSize = app_helper:get_env(riak_kv, index_buffer_size, 100),
     KeyBufSize = app_helper:get_env(riak_kv, key_buffer_size, 100),
     WorkerPoolSize = app_helper:get_env(riak_kv, worker_pool_size, 10),
-    {ok, VId} = get_vnodeid(Index),
+    %%  This _has_ to be a non_neg_integer(), and really, if it is
+    %%  zero, you are fysncing every.single.key epoch.
```
jplock, line 439: spelling mistake - `fysncing` -> `fsyncing`

russelldb, line 439: Good one, thanks!

kellymclaughlin:
There are a lot of outstanding TODOs around marking the clock as dirty. Are those intended to be resolved before this is merged or just markers for future improvements?

russelldb:
Markers for future work that requires an update to the riak_object format.

kellymclaughlin:
Regarding the test failures, I got the same error output running locally:

```
12:36:23.153 [info] Application exometer started on node nonode@nohost
...........................12:36:23.619 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
.........12:36:23.628 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
......12:36:23.658 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
........12:36:23.731 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
...12:36:23.765 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
.......12:36:23.819 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
........12:36:23.924 [error] Failed to update stat skipped_read_repairs due to (error) function_clause.
```

Rebasing from the latest `2.0` made that output go away. The `make test` run is still not clean, but I can dig into that more today.

test/kv_vnode_status_mgr_eqc.erl
```diff
@@ -0,0 +1,108 @@
+%%% @author Russell Brown <russelldb@basho.com>
```
kellymclaughlin, line 1: Probably need to wrap this with `ifdef(EQC)`. Perhaps add some eunit accountrements to facilitate running it with CI as well (and an `ifdef(TEST)`wrapper also if so)?

test/kv_vnode_status_mgr_eqc.erl
```diff
@@ -0,0 +1,108 @@
...
+    {postcondition_failed, "Ets and Disk don't match", MoCnt, Cnt}.
+
+%% @doc weight/2 - Distribution of calls
+-spec weight(S :: eqc_statem:symbolic_state(), Command :: atom()) -> integer().
+weight(_S, lease_counter) -> 1;
+weight(_S, _Cmd) -> 1.
+
+%% @doc Default generated property
+-spec prop_monotonic() -> eqc:property().
+prop_monotonic() ->
```
kellymclaughlin, line 78: I'm curious about the choice to use `eqc_statem` when there is only one actual state? Might it be simpler to use a normal eqc property and have one of the generated inputs be a number of times to call into the `riak_kv_vnode_status` server per test iteration?

test/kv_vnode_status_mgr_eqc.erl
```diff
@@ -0,0 +1,108 @@
...
+    [
+
+     frequency([{5, ?SUCHTHAT(Lease, ?LET(I, largeint(), abs(I)), Lease =< ?MAX_INT andalso Lease > 0)},
+                {5, ?SUCHTHAT(Lease, ?LET(I, int(), abs(I)), Lease =< ?MAX_INT andalso Lease > 0)}])
+    ].
+
+lease_counter(Lease) ->
+    %% @TODO (rdb) handle the 32 bit threshold, roll over to new
+    %% id. Basics first.
+    [{status, LastId, MoCnt, Pid}] = ets:lookup(vnode_status, status),
```
kellymclaughlin, line 38: Why use ets to track the expected test state versus using the `state` record? It seems like in the this function you strictly want to call functions from the module under test and not have the ets information influence the behavior in the test if that is tracking the expected state you are using to determine the efficacy of the MUT.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,388 @@
...
+%%--------------------------------------------------------------------
+%% @private
+%%--------------------------------------------------------------------
+-spec handle_cast({lease, non_neg_integer()}, #state{}) ->
+                         {noreply, #state{}}.
+handle_cast({lease, LeaseSize}, State) ->
+    #state{status_file=File, vnode_pid=Pid} = State,
+    {ok, Status} = read_vnode_status(File),
+    {_Counter, LeaseTo, VnodeId, UpdStatus} = get_counter_lease(LeaseSize, Status),
+    ok = write_vnode_status(UpdStatus, File),
```
kellymclaughlin, line 189: What are your thoughts on swapping this line and the next so that the vnode can be on about its business without waiting for a bunch of file access function calls to flow through the file server?

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,388 @@
...
+    {stop, normal, ok, State}.
+
+%%--------------------------------------------------------------------
+%% @private
+%%--------------------------------------------------------------------
+-spec handle_cast({lease, non_neg_integer()}, #state{}) ->
+                         {noreply, #state{}}.
+handle_cast({lease, LeaseSize}, State) ->
+    #state{status_file=File, vnode_pid=Pid} = State,
+    {ok, Status} = read_vnode_status(File),
```
kellymclaughlin, line 187: If all access to the file is serialized through this server and there is one per vnode, could you optimize this by tracking the latest status in the server state and avoid reading from disk?

test/kv_vnode_status_mgr_eqc.erl
```diff
@@ -0,0 +1,108 @@
...
+weight(_S, _Cmd) -> 1.
+
+%% @doc Default generated property
+-spec prop_monotonic() -> eqc:property().
+prop_monotonic() ->
+    ?FORALL(Cmds, non_empty(commands(?MODULE)),
+            begin
+                ets:new(vnode_status, [named_table, set]),
+                ets:new(vnodeids, [named_table, set]),
+                {ok, Pid} = riak_kv_vnode_status_mgr:start_link(self(), 1),
```
kellymclaughlin, line 83: Might be interesting to add some intentional crashing of status manager process to the test to verify the disk persistence is really effective and the expectations will still hold.

kellymclaughlin:
I have all tests that run with `make test` passing locally though it concludes with a strange crash:

```
Finished in 1066.860 seconds
313 tests, 0 failures
Cover analysis: /home/kelly/basho/repos/riak_kv/.eunit/index.html
{"init terminating in do_boot",{terminated,[{io,format,[<0.24.0>,"escript: Internal error: ~p\n",[badarg]],[]},{escript,start,1,[{file,"escript.erl"},{line,287}]},{init,start_it,1,[]},{init,start_em,1,[]}]}}
```

test/kv_vnode_status_mgr_eqc.erl
russelldb, line 1: I'm sorry, I should have said, the eqc test is not finished. I plan on adding a load more commands (crash, delete file, etc)

test/kv_vnode_status_mgr_eqc.erl
russelldb, line 78: It's easier, to my mind, to generate a sequence of commands than a state that tests the property.

test/kv_vnode_status_mgr_eqc.erl
russelldb, line 38: Because I don't want to deal with symbolic vs dynamic and mutating the state representation at generation time. What is MUT?

src/riak_kv_vnode_status_mgr.erl
russelldb, line 189: Isn't the vnode on about it's business as soon as gen_server:cast is called above in the API call? The whole point is not to allow a vnode to use a counter that is not on disk.

src/riak_kv_vnode_status_mgr.erl
russelldb, line 187: Maybe.  I did think of it, but I'm not sure on the benefit.

test/kv_vnode_status_mgr_eqc.erl
russelldb, line 83: See comment above, I'm sorry that I should have said, this test is not complete. My plan was to work on the test while the code was reviewed. I'm still kind of hoping someone else will pitch in and write some unit tests for the riak_kv_vnode portions of the code. And I still don't know how best to test the interaction/blocking portions of the protocol.

src/riak_kv_vnode_status_mgr.erl
kellymclaughlin, line 189: Yes, you're right. My eyes tricked me into seeing `handle_call` instead of `handle_cast`.

test/kv_vnode_status_mgr_eqc.erl
kellymclaughlin, line 38: Module Under Test :)

test/kv_vnode_status_mgr_eqc.erl
kellymclaughlin, line 83: I can look at some further testing on the vnode changes. 

For the testing of the blocking portions of the protocol, this is perhaps relevant or might present some ideas. I wrote [this](https://github.com/basho/riak_cs/blob/develop/test/riak_cs_gc_d_eqc.erl#L105) property to exercise the riak cs gc process api. Also happy to rubber duck more about it if you need.

kellymclaughlin, line 83: > I can look at some further testing on the vnode changes. 

This seems pretty tricky just due to the dependencies involved with the vnode. What do you think about using a stripped down imitation vnode (similar to `fsm_eqc_vnode`) to get some assurances about the interactions between the vnode and the status manager?

russelldb, line 83: Yes, I think maybe fsm_eqc_vnode is too complex even. But we need a process that starts the vnode_status_mgr and makes casts, and waits for replies, and that calls the blocking functions in vnode. So maybe just an eunit test inside the `riak_kv_vnode` module file, so it can call private functions from within a test process?

src/riak_kv_vnode.erl
```diff
@@ -1904,91 +1983,23 @@ max_hashtree_tokens() ->
...
+%% to the previous function, as it will now _always_ store a new
+%% status to disk as it will grab a new lease.
+%%
+%%  @TODO document the need for, and invariants of the counter
+-spec get_vnodeid_and_counter(pid(), non_neg_integer()) ->
+                                     {ok, {vnodeid(), #counter_state{}}} |
+                                     {error, Reason::term()}.
+get_vnodeid_and_counter(StatusMgr, CounterLeaseSize) ->
+    {ok, {VId, Counter, Lease}} = riak_kv_vnode_status_mgr:get_vnodeid_and_counter(StatusMgr, CounterLeaseSize),
+    {ok, {VId, #counter_state{cnt=Counter, lease=Lease, lease_size=CounterLeaseSize}}}.
```
kellymclaughlin, line 1998: Shouldn't this also set `leasing` to `true`?

kellymclaughlin, line 1998: Nvm, I interpretted the comment about the purpose of this field incorrectly. I see now that is just is to indicate that the request for a lease has gone out, but not been fulfilled yet.

russelldb:
Closing, re-opened re-rebased at basho/riak_kv#1070

PR #1057, opened 2014-12-04
Update MapReduce JSON Parsing

hazen:
While adding client support for M/R jobs with bucket types, I found the code comments a little behind the times.  So the `parse_inputs/1` error string now shows alternate usage with optional bucket types.  This will hopefully make maintenance easier in the future.  I also found that if the `default` bucket type is specified, Riak does not like this.  So now if `<<"default">>` is specified, it will be parsed as if no bucket type had been specified giving the desired behavior.  Tests for the new parsing behavior were also added.

### DIFF
--- a/src/riak_kv_mapred_json.erl
+++ b/src/riak_kv_mapred_json.erl
@@ -72,6 +72,8 @@ parse_request(Req) ->
             {error, not_json}
     end.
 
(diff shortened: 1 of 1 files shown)


yorkshire-steve:

PR #1063, opened 2014-12-16
Bugfix for precondition_context test failures.

seancribbs:
Previously every riak_dt type simply called riak_dt:to_binary/1 for 
serialization, and before 2.0.0 optimizations, was an entire copy of the 
datatype. This meant that the precondition context was handled through the
type's to_binary/1 function. The 2.0.0 optimizations use only a VV as the
context, however, so we don't need to serialize via the type's to_binary/1
call, we can call riak_dt_vclock:to_binary/1 directly.

This bug was uncovered by basho/riak_dt#110, which changes how the sets and maps are
manipulated before serialization.

### DIFF
--- a/src/riak_kv_crdt_json.erl
+++ b/src/riak_kv_crdt_json.erl
@@ -265,10 +265,11 @@ encode_fetch_response_test_() ->
                                      {<<"value">>,
                                       {struct,
                                        [% NB map sorts its keys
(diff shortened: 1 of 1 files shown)


russelldb:
Looks good +1 from me

russelldb:
+1 53bf5eb

russelldb:
@borshop merge

PR #1065, opened 2014-12-17
Fix bug related to riak_dt#110 serialization changes

seancribbs:
Previously, Mod:from_binary/1 would just return the Datatype, but now it
returns {ok, Datatype}. Fixed this in the one place it was called here
and added a quickcheck property for it.

This should fix the [related test failures in yokozuna](http://buildbot.bos1/builders/test-yokozuna/builds/915/steps/shell_4/logs/stdio).


russelldb:
Thanks @seancribbs, sorry I missed these, running all of riak_test is prohibitively expensive locally

russelldb:
+1 a00e0f9

russelldb:
@borshop merge

seancribbs:
@russelldb It was no problem, I should have caught them in review.

PR #1067, opened 2014-12-23
Make read-repair stats be 'proplist' instead of 'value'

uwiger:
'value' is actually something that should practically never be
used in exometer function metrics. In this case, it causes the
return value to be wrong if the exometer:aggregate() function doesn't
find any detailed stats to aggregate.


jburwell:
:+1: f91bf21

jburwell:
@borshop merge

Issue #679, opened 2013-10-02
Possible data loss on corruption of coordinating node replica


engelsanchez:
Russell brought this up today, and perhaps he can elaborate on this: 
- A replica becomes unavailable in a put coordinating node due to corruption or user error.
- The incoming object may or may not contain a vector clock.
- We get a not_found in the local put.  We have lost the information corresponding to this vnode id.
- The vector clock created could have a stale number for this vnode actor id. Either 1 if no vclock or < than the real one if more writes have happened since.
- The vector clock is subsumed by the other replicas when sent out, so this write will lose and will eventually be read repaired completely out of existence.

/cc @jtuple @jrwest @evanmcc @Vagabond @russelldb

russelldb:
Yeah. That captures it @engelsanchez. I think this is a problem we introduced recently by treating errors on local read as `notfound`. And I think the answer is as simple as creating a new `vnodeid` as soon as local corruption is detected.

If for whatever reason a coordinating PUT to a vnode fails to read an existing K/V out of the backend (we have a catch for it here:- https://github.com/basho/riak_kv/blob/develop/src/riak_kv_vnode.erl#L1258) then we have this problem.

@engelsanchez pointed out we have the same problem if a user removes the backend data and _not_ the vnodeid, or removes the backend data while the vnode is running. At this point I'd say all bets are off and there is nothing we can do.

In the case of corruption where we _do_ get an error attempting to read the data, we can do something (get the data from another vnode, generate a new vnodeid.) Ignoring the error and  returning a `notfound` is not correct though.

Imagine putting to Key K with an empty vclock, the item has been written by vnode A before. Vnode A coordinates the write for this PUT. Vnode A's backend is corrupt due to cosmic rays. Vnode A attempts to read key K to get a local vector clock. Vnode A gets an error but we silently parlay that to a `notfound`. Vnode A creates a vclock entry of {A, 1} for the PUT and stores the item (even though the item on disk (that we can't read) has entry {A, 3}).

K's data is returned as a frontier object (not!) to the FSM which sends it downstream to Vnode B and C for local merge. Vnode B and C see that their local vclocks for K dominate the one they just received and drop the write and ack the FSM. The FSMs ack the user.

The user tries to read K. A's replica is dominated by B's and C's, read repair kicks in and A's data is replaced. We lost the data.

[...]

PR #1070, opened 2014-12-30
Rebase per-key actor epoch (kv679) on 2.0 [JIRA: RIAK-1331]

russelldb:
See basho/riak_kv#679 and the associated platform_task RFC and summary.

This PR addresses the "doomstone", backup-restore, and some byzantine flavours of the kv679 bug.

The RFC explains the mechanism in detail but briefly:

add a persisted to disk vnode counter (persisted with leases, aysnc)
when ever a key is written to for the "first time" by a vnode, create an epoch actor for the key by concatenating vnodeid+counter (and increment the counter)
This ensures that a first time write for a key gets a new actor, this is the epoch for the key. It means we don't mix up deleted+re-created keys {a,1} event with the original {a, 1} event for some key, by ensuring an actor per epoch, without causing a keyspace wide actor explosion.

There are riak_tests at https://github.com/basho/riak_test/tree/rdb/gh-kv679

---

This has been long running work, there is a PR here basho/riak_kv#1040 and here basho/riak_kv/pull/1053 that are closed in favour of this.

---
- [x] review
- [x] benchmarking
- [x] riak_test review
- [x] add Kelly's EQC test


russelldb:
Mr @slfritchie gave me some review over email. Reproduced below are his comments and my replies. As usual he was right and insightful and I made the changes he suggested.

On 19 Dec 2014, at 07:22, Scott Lystig Fritchie scott@basho.com wrote:

> Howdy!
> 
> I can put these comments into the PR, but there are enough
> comments/conversation fragments there that I don't want things to get
> lost right now.
> 
> Question: blocking_lease_counter()
> 
> I don't believe that 5 seconds is long enough.  I've seen Linux
> systems block all writes for 15+ seconds when disk has been
> overwhelmed by random dirty pages and/or a disk device is stuttering
> (i.e., in an early hardware failure state).

I have set to 20seconds, you think more?

> Also, see that the 5 seconds is what's in the 1.4 branch.  We might
> want to patch that, depending on results of the discussion here.

This is all new code, where in 1.4 do we have such a timeout?

> Question: put_merge() and prepare_new_put()
> 
> I'm not knowlegeable to know if the new @TODO things are must-do or
> would-be-nice-to-do-in-unicorn-land or somewhere in between.

These are things that cannot be done without object format change as =
they require some persistent per-key-metadat about the clock. Leaving as =
TODOs, if that is OK?

They are nice-to-have as they solve the byzantine case of dataloss.

> Question: -define(DEFAULT_CNTR_LEASE, 1000).
> 
> Increase by a factor of 5 or 10?

Made it 10k

> Question: update_counter() and key_epoch_actor()
> 
> key_epoch_actor() says that the counter is 32 bits.  update_counter()
> doesn't appear to have any overflow protection?

The overflow protection is in the status manager, you can't hand out
above the Lease, and the lease is protected from overflowing. Makes 
sense?

> riak_kv_vnode_status_manager has checks for ?MAX_CNTR for
> get_vnodeid_and_counter() and lease_counter(), but those are checking
> the lease size instead of checking the lease's upper bound.
> 
> HRM, ok, so there _is_ an enforcement, but it feels weird.
> get_counter_lease() has this:
> -            LeaseTo =3D min(PrevLease + LeaseSize, ?MAX_CNTR),
> 
> So, if we ask for a lease range of 100 and a PrevLease of ?MAX_CNTR-2,
> then LeaseTo will only be ?MAX_COUNTER, which means a lease size of 2;
> the vnode will (nearly) immediately ask for a new lease,
> get_counter_lease() will see PrevLease =3D=3D ?MAX_CNTR, and then=
> assign a
> new vnodeID and a new lease range starting at zero.

OK, we discussed this, and the answer is smaller Max Lease size, and =
always give out a lease of the size requested, even if it means a new =
vnodeid =93prematurely"

> I'm still puzzled by the "LeaseSize =3D< ?MAX_CNTR" checks, though, =
> they
> feel very distracting.

Will remove.

> Question: assign_vnodeid()
> 
> Looks like you just moved this function and didn't edit it, but ... it
> seems possible for VnodeEpoch to overflow 32 bits.  I don't think that
> matters in The Big Scheme of Things(tm), but perhaps I'm wrong?

I hope it doesn=92t matter, but I=92ll give it a poke.

> Suggestion: riak_kv_vnode_status_mgr_eqc:lease_counter()
> 
> Who knows how flakey the CI system might be.  I'd change it to 30 or
> 60 seconds and then wait to be exasperated about CI when something
> that ludicrous fails.  I've seen enough eunit tests with CI fail at 5
> second when they should take 2 milliseconds, so =85

Will do, thanks.

> Hrm, there isn't an EUnit wrapper for the QuickCheck property in this
> module, is that intentional?

Ooops.

slfritchie:
> I have set to 20seconds, you think more?

If there's a buffer flush storm, or if the disk hardware is stuttering, then 20 seconds might or might not be enough.  Other reviewers, do you have opinions?

slfritchie:
I've looked through the other diffs, +1 from me.

src/riak_kv_vnode.erl
```diff
@@ -110,6 +110,21 @@
                 reqid :: term(),
                 target :: pid()}).
 
+-record(counter_state, {
```
macintux, line 113: Minor nit: suggest renaming this record to epoch_counter (and giving the state record field that same name) to be more explicit

russelldb, line 113: Not sure I agree. It's just a counter: the _keys_ get epochs.

macintux, line 113: Ok, but it's not a purposeless counter, it has a specific role. Is there a good name for its role?

russelldb, line 113: Whatever man. Rename it if you like. Thanks for the contribution.

seancribbs, line 113: Naming is hard. I think the inline comments make up for any possible confusion about its purpose.

russelldb, line 113: Hey @macintux, I'm sorry, that wasn't cool on my part. This is at least the 3rd review, and I appreciate more eyes. It's late and I shouldn't have responded until morning. What I mean is, if you want to rename it, please be my guest, maybe it would help a future maintainer understand the purpose of the counter.

macintux, line 113: Bike and sleep deprived. No sweat. I'll keep reading and thinking.

andrewjstone:
I'm seeing the following failure when running `make test` for riak_kv on this branch.

```
Failures:

  1) riak_kv_schema_tests:commit_hooks_test/0
     Failure/Error: ?assertEqual({error,apply_translations,{error,[{error,[84,114,97,110,115,108,97,116,105,111,110,32,102,111,114,32,39,114,105,97,107,95,99,111,114,101,46,100,101,102,97,117,108,116,95,98,117,99,107,101,116,95,112,114,111,112,115,46,112,111,115,116,99,111,109,109,105,116,39,32,102,111,117,110,100,32,105,110,118,97,108,105,100,32,99,111,110,102,105,103,117,114,97,116,105,111,110,58,32,105,110,99,111,114,114,101,99,116,32,104,111,111,107,32,102,111,114,109,97,116,32,39,106,115,76,79,76,39]},{error,[84,114,97,110,115,108,97,116,105,111,110,32,102,111,114,32,39,114,105,97,107,95,99,111,114,101,46,100,101,102,97,117,108,116,95,98,117,99,107,101,116,95,112,114,111,112,115,46,112,114,101,99,111,109,109,105,116,39,32,102,111,117,110,100,32,105,110,118,97,108,105,100,32,99,111,110,102,105,103,117,114,97,116,105,111,110,58,32,105,110,99,111,114,114,101,99,116,32,104,111,111,107,32,102,111,114,109,97,116,32,39,98,97,100,58,109,111,100,58,102,117,110,39]}]}}, Config)
       expected: {error,apply_translations,
                     {error,
                         [{error,
                              "Translation for 'riak_core.default_bucket_props.postcommit' found invalid configuration: incorrect hook format 'jsLOL'"},
                          {error,
                              "Translation for 'riak_core.default_bucket_props.precommit' found invalid configuration: incorrect hook format 'bad:mod:fun'"}]}}
            got: {error,apply_translations,
                     {errorlist,
                         [{error,
                              {translation_invalid_configuration,
                                  {"riak_core.default_bucket_props.postcommit",
                                   "incorrect hook format 'jsLOL'"}}},
                          {error,
                              {translation_invalid_configuration,
                                  {"riak_core.default_bucket_props.precommit",
                                   "incorrect hook format 'bad:mod:fun'"}}}]}}
     %% test/riak_kv_schema_tests.erl:278:in `riak_kv_schema_tests:-commit_hooks_test/0-fun-0-/2`


  2) riak_kv_schema_tests:correct_error_handling_by_multibackend_test/0
     Failure/Error: ?assertMatch({ error , apply_translations , { error , _ } }, Config)
       expected: = { error , apply_translations , { error , _ } }
            got: {error,apply_translations,
                     {errorlist,
                         [{error,
                              {translation_invalid_configuration,
                                  {"riak_kv.multi_backend",
                                   "Error processing multi_backend configuration for backend default"}}}]}}
     %% test/riak_kv_schema_tests.erl:305:in `riak_kv_schema_tests:correct_error_handling_by_multibackend_test/0`

```

seancribbs:
@andrewjstone Those are fixed by #1071.

andrewjstone:
Thanks @seancribbs

src/riak_kv_vnode.erl
```diff
@@ -2322,102 +2346,286 @@ new_md_cache(VId) ->
...
+                                    {ok, #state{}} |
+                                    counter_lease_error().
+blocking_lease_counter(_State, {MaxErrs, MaxErrs, _MaxTime}) ->
+    {error, counter_lease_max_errors};
+blocking_lease_counter(State, {ErrCnt, MaxErrors, MaxTime}) ->
+    #state{idx=Index, vnodeid=VId, status_mgr_pid=Pid, counter=CounterState} = State,
+    #counter_state{lease_size=LeaseSize} = CounterState,
+    Start = os:timestamp(),
+    receive
+        {'EXIT', Pid, Reason} ->
```
andrewjstone, line 2410: We would need to call `process_flag(trap_exit, true)` in init/1 to get an `EXIT` message.
See "Receiving Exit Signals" [here](http://www.erlang.org/doc/reference_manual/processes.html)

Are we sure we want to actually trap exits in the vnode though? @jonmeredith @jtuple

src/riak_kv_vnode.erl
```diff
@@ -2322,102 +2346,286 @@ new_md_cache(VId) ->
...
+            lager:error("Failed to lease counter ~p", [Reason]),
+            {ok, NewPid} = riak_kv_vnode_status_mgr:start_link(self(), Index),
+            ok = riak_kv_vnode_status_mgr:lease_counter(NewPid, LeaseSize),
+            NewState = State#state{status_mgr_pid=NewPid},
+            Elapsed = timer:now_diff(os:timestamp(), Start),
+            blocking_lease_counter(NewState, {ErrCnt+1, MaxErrors, MaxTime - Elapsed});
+        {counter_lease, {Pid, VId, NewLease}} ->
```
andrewjstone, line 2417: What happens if this message comes in late? It seems like next call to `blocking_lease_counter/3` will pull the old message of the mailbox. Is that harmful?

src/riak_kv_vnode.erl
seancribbs, line 2410: We already trap exits in [`riak_core_vnode`](https://github.com/basho/riak_core/blob/develop/src/riak_core_vnode.erl#L186).

andrewjstone, line 2410: Ahhh, Thank you @seancribbs

src/riak_kv_vnode.erl
```diff
@@ -2322,102 +2346,286 @@ new_md_cache(VId) ->
...
+            %% Never acted on this object before, new epoch.
+            new_key_epoch(State);
+        {LocalId, LocalEpoch, LocalCntr} -> %% Present locally
+            case highest_actor(VId, IncomingObj) of
+                {_InId, InEpoch, InCntr} when InEpoch > LocalEpoch;
+                                              InCntr > LocalCntr ->
+                    %% In coming actor-epoch or counter greater than
+                    %% local, some byzantine failure, new epoch.
+                    %%
+                    %% @TODO (rdb) do we log/stat here?
```
andrewjstone: I would add at least a warning log here. Either there is a bug in the backend code or the disk is losing things.

russelldb: I’m not sure about “Late" and "Old message"?

The vnode has sent a lease message to the manager. It has yet to receive an answer so it enters this blocking receive. What is the old message? Do you mean if we timeout and crash the vnode? There is no old message as the message queue dies with the crashed vnode.

Is this the sequence you’re thinking of that leads to there being an “old message” for the lease
1. vnode asks for a lease
2. manager sends a lease message
3. manager crashes
4. vnode in blocking lease receive gets the exit message first (‘cos of the order of receives in statement)
5. vnode starts a new manager and requests a lease
6. vnode pulls the original message (pre crash)

Is that step 6 the “old message?”

I don’t think it can happen. The `Pid` in the receive is the new managers `Pid`, the old one will never match. Even if it would I think it is safe. The vnode is getting a new lease! Yay! It can keep working. The lease it gets is smaller than that on disk, yay! Invariant preserved. I assume then there is a step 7
1. vnode gets new lease message

All is still OK. We just have a larger lease/longer until we re-lease. If the vnode crashes it just wasted some integers.

Does that seem OK to you? (This is the first time I have thought it through, so I think it was a valid question, if I answered the question you asked, that is.)

I don’t think there is any way for the vnode to request a new lease before it receives the old lease, and Erlang delivers messages between processes on the same node in order, so they’ll be received in order, meaning the lease always increases. Happy to mumble about it in the morning to be sure, though.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,408 @@
...
+-spec get_counter_lease(non_neg_integer(), status()) ->
+                           {non_neg_integer(), non_neg_integer(), binary(), status()}.
+get_counter_lease(LeaseSize, Status) ->
+    PrevLease = get_status_item(counter, Status, undefined),
+    VnodeId0 = get_status_item(vnodeid, Status, undefined),
+    Version = get_status_item(version, Status, 1),
+
+    case {Version, PrevLease, VnodeId0} of
+        {_, _, undefined} ->
+            new_id_and_counter(Status, min(LeaseSize, ?MAX_CNTR));
```
andrewjstone: I would extract the use of `min(LeaseSize, ?MAX_CNTR)` into it's own variable and use that in each clause for clarity.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,408 @@
...
+get_status_item(Item, Status, Default) ->
+    case orddict:find(Item, Status) of
+        {ok, Val} ->
+            Val;
+        error ->
+            Default
+    end.
+
+%% @private generate a file name for the vnode status, and ensure the
+%% path to exixts.
```
andrewjstone: typo here.

src/riak_kv_vnode_status_mgr.erl
```diff
@@ -0,0 +1,408 @@
...
+    ?assertEqual(<<1, 2, 3, 4, 70,116,143,150>>, Vid1),
+    %% Simulate clear
+    Status2 = orddict:erase(vnodeid, Status1),
+    %% Reassign
+    {Vid2, _Status3} = assign_vnodeid(Now2, NodeId, Status2),
+    ?assertEqual(<<1, 2, 3, 4, 70,116,143,250>>, Vid2).
+
+%% Check assigning a vnodeid with a earlier date - just in case of clock skew
+assign_vnodeid_restart_earlier_ts_test() ->
+    Now1 = {1000,224520,343546}, %% <<70,116,143,150>>
```
andrewjstone: The commented binaries are backwards here

src/riak_kv_vnode.erl
andrewjstone: @russelldb That all makes sense to me. Originally I was thinking that we returned an error to the client after the timeout and continued with the same valid manager/vnode. I just missed the
 `{ok, NewState} = blocking_lease_counter(State)` in `maybe_lease_counter/1`. In this case you are correct that the crashing vnode will clear the mailbox so there is no problem.

In the latter case that you so thoughtfully put together, since the manager crashes  there will indeed be a new Pid and the match will not succeed. Thanks for taking the time to walk me through it.

andrewjstone:
@russelldb This PR looks very good. I've gone over all the code and ran the eqc test a few times. Tomorrow I will go over the corresponding riak_test PR and run that against this code.

src/riak_kv_vnode_status_mgr.erl
russelldb: Ooooh, good one, because actually the `LeaseSize` isn't normalised in the final case clause's body.

russelldb:
Arrggh. I messed up. Need to re-do that change, hold off re-review for now

russelldb:
Urgh, I went and rebased again. And I don't know how to "unrebase" and I don't want to force push. Give me a bit of time to wrestle with the tools here.

russelldb:
there we go

andrewjstone:
:+1:  d77f65f
riak_tests specific to this pass. So do dt tests and sibling explosion tests. Code looks good as well.

seancribbs:
I was checking this over to see if you had already inserted a metric for when the epoch increments (or leases granted, etc), and found you have a stray file in the PR, `test/vnode_driver.erl~`. I'll make a separate task for the metrics but removing that stray file would be good, then we can re-approve and merge.

Issue #1054, opened 2014-11-26
Quick question on riak_kv_pb_mapred


kuenishi:
Looks like [this head](https://github.com/basho/riak_kv/blob/2.0/src/riak_kv_pb_mapred.erl#L146-L162)

``` erlang
process_stream({'DOWN', Ref, process, Pid, Reason}, Ref,
               State=#state{req=#rpbmapredreq{},
                            req_ctx=#pipe_ctx{sender={Pid, Ref}}=PipeCtx}) ->
```

is never called as the second argument is actually `ReqId` in riak_api sense, as the `process` function is [returning PipeRef](https://github.com/basho/riak_kv/blob/2.0/src/riak_kv_pb_mapred.erl#L209) and then never matches with the monitor reference, even when the pipe input sender crashes. And then sender error or normal finish all fall through to the final ignored head of `process_stream/3`. So maybe this head should be like this in theory?:

``` erlang
process_stream({'DOWN', Ref, process, Pid, Reason}, PipeRef,
               State=#state{req=#rpbmapredreq{},
                            req_ctx=#pipe_ctx{ref=PipeRef, sender={Pid, Ref}}=PipeCtx}) ->
```

This is likely not the cause of the problem I'm currently investigating, but I just found when reviewing codes.

krestenkrab:
It looks like you're right. The parameter you call `PipeRef` is called `ReqId` in the previous head - https://github.com/basho/riak_kv/blob/2.0/src/riak_kv_pb_mapred.erl#L106-L113.  

Would this be catchable by adding some extra type specs?  Looks a little tricky.

kuenishi:
Maybe changing PipeRef to something other than `reference()`, like `{pipe, reference}` makes dialyzer happy, but I just don't know the code impact. If `ReqId` stuff for riak_pb_api_service is opaque, then that might work.

Oh, I just found riak_api_pb_service is using old-style behaviour_info definition. Maybe changing to this style (not sure this works..) ?

``` erlang
-callback init() -> {ok, term()} | {error, term()}.
-callback decode(integer(), binary()) -> {ok, msg_record()} | {error, term()}.
-callback encode(msg_record()) -> {ok, binary()} | {error, term()}.
-callback process(msg_record()) -> {reply, {stream, req_id(T)}} | ... | {error, term(), state()}.
-callback process_stream(msg_record(), req_id(T), state()) -> ....
```

seancribbs:
[...]

PR #1073, opened 2015-01-22
Bugfix/ku/add probe logmsg

kuenishi:
This patch adds a log message line when pipe fitting died and MapReduce failed with ambiguous error. This is also a probe when a customer issue (zd://9291) reproduced. At the same time this adds fix for #1054 . CC: https://github.com/basho/riak_pipe/pull/92

### DIFF
--- a/src/riak_kv_mrc_sink.erl
+++ b/src/riak_kv_mrc_sink.erl
@@ -293,7 +293,7 @@ handle_info({'DOWN', _, process, Pid, Reason}, _,
     %% don't stop when the builder exits 'normal', because that's
     %% probably just the pipe shutting down normally - wait for the
     %% owner to ask for the last outputs
-    _ = lager:error("Pipe builder down. Reason: ~p", [Reason]),
+    _ = lager:warning("Pipe builder down. Reason: ~p", [Reason]),
     {stop, normal, State};
 handle_info(_, StateName, State) ->
     {next_state, StateName, State}.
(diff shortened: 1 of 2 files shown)


kuenishi:
The original work for 1.4 branch is here, just linking: https://github.com/basho/riak_kv/tree/ku-tmp-mapred-probe

kuenishi:
Fm. Maybe error level doesn't match with the comments, but it rarely happens so it's worth logging. Maybe warning would be sufficient?

seancribbs:
I agree that the `error` level is too high. Perhaps warning or notice? The style guide might be helpful here.

jadeallenx:
I would ask - what is the impact of this failure?  How much does a human need to care about it? What operational impact will it have?  

Without understanding the context for this change better, I think Sean's instinct for a warning or notice is right on.

kuenishi:
@mrallen1 That's the problem. The impact of the failure is MapReduce fail -  storage calculation. Operators can't know how much amount each bucket consuming as Riak CS logical space, and to the result, operators can't create a bill to the bucket owner. We have to catch the root cause of this fail and be keen at this. Riak client often returns useless information for this kind of failure. Operators usually don't have to take care of this immediately. - CS operators should, but the log will be output in CS.
That's why I'm thinking warning is suitable.

jadeallenx:
:+1:

cmeiklejohn:
:+1: f704d08

cmeiklejohn:
This merge is blocked on failure on riak_kv, that are being investigated in a different thread.

cmeiklejohn:
@borshop retry

cmeiklejohn:
@borshop retry

cmeiklejohn:
@borshop: retry

seancribbs:
https://github.com/basho/internal_wiki/wiki/Bors-automated-integration-CI-bot#quick-reference

> How to retry: `@borshop: retry` on the commit comments

cmeiklejohn:
@borshop merge

cmeiklejohn:
This should also be manually merged back to develop.  I'll take care of that once this PR merges.

cmeiklejohn:
Manually merged this to develop.

PR #1074, opened 2015-01-22
Enhance/miklix/auto create hash tree

lixen:
Introduce build and wait for lock. Needed by https://github.com/basho/riak_repl/pull/655
Wait for lock to give the trees time to finish building.

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -75,7 +75,7 @@
                 vnode_pid,
                 built,
                 expired :: boolean(),
-                waiting_for_lock :: undefined | pid(),
+                waiting_for_lock :: undefined | {any(), pid()},
                 lock :: undefined | reference(),
                 path,
                 build_time,
@@ -197,7 +197,7 @@ get_lock(Tree, Type, Pid) ->
 %% @doc Acquire the lock for the specified index_hashtree if not already
 %%      locked, and associate the lock with the calling process.
 %%      When not_built pid are saved and given the lock when build finish.
--spec wait_for_lock(pid(), any()) -> ok | not_built | already_locked.
+-spec wait_for_lock(pid(), any()) -> ok | building | not_built | already_locked.
 wait_for_lock(Tree, Type) ->
     wait_for_lock(Tree, Type, self()).
 
@@ -598,7 +598,7 @@ do_new_tree(Id, State=#state{trees=Trees, path=Path}) ->
     Trees2 = orddict:store(Id, NewTree, Trees),
     State#state{trees=Trees2}.
 
--spec do_get_lock(any(), pid(), state()) -> {building, not_built | ok | already_locked, state()}.
+-spec do_get_lock(any(), pid(), state()) -> {building | not_built | ok | already_locked, state()}.
 do_get_lock(_, _, State) when is_pid(State#state.built) ->
     lager:debug("Building: ~p :: ~p", [State#state.index, State#state.built]),
     {building, State};
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -203,7 +203,7 @@ wait_for_lock(Tree, Type) ->
 
 %% @doc Acquire the lock for the specified index_hashtree if not already
 %%      locked, and associate the lock with the provided pid.
--spec wait_for_lock(pid(), any(), pid()) -> ok | not_built | already_locked.
+-spec wait_for_lock(pid(), any(), pid()) -> ok | building | not_built | already_locked.
 wait_for_lock(Tree, Type, Pid) ->
     gen_server:call(Tree, {wait_for_lock, Type, Pid}, 60000).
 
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -579,7 +579,12 @@ handle_command({hashtree_pid, Node}, _, State=#state{hashtrees=HT}) ->
             %% after the vnode was already running
             case HT of
                 undefined ->
-                    State2 = maybe_create_hashtrees(State),
+                    case riak_repl2_fssource:get_strategy() == aae of
```
engelsanchez: @rsltrifork, we can not have riak_kv reference repl code, which is not open source and is not a dependency of riak_kv

src/riak_kv_index_hashtree.erl
```diff
@@ -272,6 +293,14 @@ handle_call({get_lock, Type, Pid}, _From, State) ->
     {Reply, State2} = do_get_lock(Type, Pid, State),
     {reply, Reply, State2};
 
+handle_call({wait_for_lock, Type, Pid}, _From, State) ->
+    case do_get_lock(Type, Pid, State) of
+        {building, State2} ->
+            {reply, building, State2#state{waiting_for_lock = {Type, Pid}}};
```
engelsanchez: So it just forgets any previous Pid? Should this be a list?

src/riak_kv_index_hashtree.erl
```diff
@@ -190,6 +193,21 @@ get_lock(Tree, Type) ->
...
+%%      When not_built pid are saved and given the lock when build finish.
+-spec wait_for_lock(pid(), any()) -> ok | building | not_built | already_locked.
+wait_for_lock(Tree, Type) ->
+    wait_for_lock(Tree, Type, self()).
+
+%% @doc Acquire the lock for the specified index_hashtree if not already
+%%      locked, and associate the lock with the provided pid.
+-spec wait_for_lock(pid(), any(), pid()) -> ok | not_built | already_locked.
+wait_for_lock(Tree, Type, Pid) ->
+    gen_server:call(Tree, {wait_for_lock, Type, Pid}, 60000).
```
engelsanchez, line 211: Should we also set up a monitor to detect if the tree process dies and therefore forgets that this process is to be notified when the trees are built?  I guess it could be done by the caller

src/riak_kv_index_hashtree.erl
lixen: Now it will return locked if someone is waiting for the lock.

src/riak_kv_index_hashtree.erl
lixen, line 211: All callers already have monitoring.

src/riak_kv_index_hashtree.erl
```diff
@@ -564,7 +598,13 @@ do_new_tree(Id, State=#state{trees=Trees, path=Path}) ->
...
     State#state{trees=Trees2}.
 
--spec do_get_lock(any(), pid(), state()) -> {not_built | ok | already_locked, state()}.
+-spec do_get_lock(any(), pid(), state()) -> {building | not_built | ok | already_locked, state()}.
+ do_get_lock(_, _, State) when State#state.waiting_for_lock /= undefined ->
+    lager:debug("Already waiting for lock: ~p ~p", [State#state.index, State#state.waiting_for_lock]),
+    {already_locked, State};
+do_get_lock(_, _, State) when is_pid(State#state.built) ->
+    lager:debug("Building: ~p :: ~p", [State#state.index, State#state.built]),
+    {building, State};
```
engelsanchez: Is it really necessary to introduce the `building` reply instead of returning not_built? Notice that this makes the get_lock call also start returning this, which propagates down many paths in the Riak code.

src/riak_kv_index_hashtree.erl
```diff
@@ -370,7 +403,8 @@ handle_cast(build_failed, State) ->
     {noreply, State2};
 handle_cast(build_finished, State) ->
     State2 = do_build_finished(State),
-    {noreply, State2};
+    State3 = do_inform_waiting(State2),
```
engelsanchez: We cast a message if the tree gets built, but remain silent if the build failed (function clause above this).

src/riak_kv_index_hashtree.erl
```diff
@@ -619,6 +659,14 @@ do_build_finished(State=#state{index=Index, built=_Pid}) ->
     riak_kv_entropy_info:tree_built(Index, BuildTime),
     State#state{built=true, build_time=BuildTime, expired=false}.
 
+do_inform_waiting(#state{waiting_for_lock = {Type, Waiting}} = State) when is_pid(Waiting) ->
+    {ok, State2} = do_get_lock(Type, Waiting, State#state{waiting_for_lock = undefined}),
+    Waiting ! {hashtree_lock, State2#state.index},
+    State;
```
engelsanchez, line 678: We are returning State without removing the waiting_for_lock field (that was done to State2).

src/riak_kv_index_hashtree.erl
engelsanchez: Also, we should probably remove the wait for lock Pid after notifying that process that the build failed.

engelsanchez:
What happens when a process is waiting for a lock and the cluster changes making the riak_kv_index_hashtree go away? Or for that matter, what happens when that process dies or goes away for other reasons?  This is related to my comment about maybe adding monitoring to the call.  The caller that assumes it will get an ack when the tree it is waiting for needs to be notified when that will never happen.

engelsanchez:
Since the AAE FS code does monitor the tree process and will notice if it goes away, my note above doesn't apply to that use. But should at least warn in the comments for that function that any other future users should take care to do the same.

PR #1076, opened 2015-01-23
Add capability and env_var to control binary format of map/set (riak#667)

russelldb:
See basho/riak#667 (RIAK-1436) (RIAK-1482) for more details.

This PR adds a capability for the binary output of Maps and Sets in riak. It also uses and app_env `{riak_kv, crdts_mixed_versions}` to override the capability so that an operator may force the riak cluster to use v1 binary format until all clusters in a multi-datacenter set up are upgraded to riak-2.0.x>4.

There is a cost associated with running in mixed cluster mode. See basho/riak_dt#111 for details. A `v2` map written to `v1` must fully recurse the structure to convert to a format readable by `v1`.

### DIFF
--- a/include/riak_kv_types.hrl
+++ b/include/riak_kv_types.hrl
@@ -34,5 +34,8 @@
 
 -define(DATATYPE_STATS_DEFAULTS, [actor_count]).
 
+-define(R1_DATATYPE_VERSIONS, [{?COUNTER_TYPE, 2}]).
+-define(R2_DATATYPE_VERSIONS, [{?MAP_TYPE, 2}, {?SET_TYPE, 2}, {?COUNTER_TYPE, 2}]).
+
 -type crdt() :: ?CRDT{}.
 -type crdt_op() :: ?CRDT_OP{}.
--- a/include/riak_kv_types.hrl
+++ b/include/riak_kv_types.hrl
@@ -34,8 +34,12 @@
 
 -define(DATATYPE_STATS_DEFAULTS, [actor_count]).
 
--define(R1_DATATYPE_VERSIONS, [{?COUNTER_TYPE, 2}]).
--define(R2_DATATYPE_VERSIONS, [{?MAP_TYPE, 2}, {?SET_TYPE, 2}, {?COUNTER_TYPE, 2}]).
+%% These proplists represent the current versions of supported
+%% datatypes. The naming `EN_DATATYPE_VERSIONS' means `Epoch' and
+%% number. `N' is incremented when any new version of any datatype is
+%% introduced, thus bumping the data type `Epoch'.
+-define(E1_DATATYPE_VERSIONS, [{?COUNTER_TYPE, 2}]).
+-define(E2_DATATYPE_VERSIONS, [{?MAP_TYPE, 2}, {?SET_TYPE, 2}, {?COUNTER_TYPE, 2}]).
 
 -type crdt() :: ?CRDT{}.
 -type crdt_op() :: ?CRDT_OP{}.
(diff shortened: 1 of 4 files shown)


src/riak_kv_crdt.erl
```diff
@@ -415,6 +420,24 @@ to_record(?SET_TYPE, Val) ->
...
     lists:member(Mod, riak_core_capability:get({riak_kv, crdt}, [])).
 
+%% @private get the binary version for a crdt mod, default to `1' for
+%% pre-versioned.
+-spec crdt_version(module()) -> pos_integer().
+crdt_version(Mod) ->
+    %% due to the riak-2.0.4 disaster where mixed format maps were
+    %% written to disk (see riak#667 for more) override the cluster
+    %% negotiated capability with an env var, this is to ensure that
+    %% in a multi-cluser environment, v1 binary format is used until
```
macintux: "cluster"

include/riak_kv_types.hrl
```diff
@@ -34,5 +34,8 @@
 
 -define(DATATYPE_STATS_DEFAULTS, [actor_count]).
 
+-define(R1_DATATYPE_VERSIONS, [{?COUNTER_TYPE, 2}]).
```
macintux: Could we add some comments above this? Even R1/R2 isn't immediately obvious: is it 1.4 vs 2.0 or 2.0.<4 vs 2.0.>4?

macintux:
Would like to see a high-level summary of the history of data type changes, perhaps tied to the R1/R2 macro comments I suggested

macintux:
I do think `crdt_mixed_versions` needs to be an integer (still thinking about proplist) rather than a true/false value for future-proofing.

Alternatively, name this environment variable specific to this transition. That may be safer than betting we'll know what we need next time.

**UPDATE**: see later proposal

macintux:
This comment could use some revision/expansion/removal/whatever, neh?

```
%% @doc Check cluster capability for crdt support
%% @TODO what does this mean for Maps?
supported(Mod) ->
    lists:member(Mod, riak_core_capability:get({riak_kv, crdt}, [])).
```

macintux:
After more consideration, I'd vote for naming `R1/2_*` as `EPOCH1/2_*`, giving the capability the name `crdt_epoch`, and the environmental override `mdc_crdt_epoch`.

russelldb:
I agree mainly, except the capability name: the value is a proplist map of mod->versions.

russelldb:
But yes, I like the env override name, since it will make sense _if_ we need it come the next epoch change.

src/riak_kv_crdt.erl
```diff
@@ -411,10 +416,27 @@ to_record(?SET_TYPE, Val) ->
...
+    %% due to the riak-2.0.4 disaster where mixed format maps were
+    %% written to disk (see riak#667 for more) override the cluster
+    %% negotiated capability with an env var, this is to ensure that
+    %% in a multi-cluster environment, Epoch 1 binary format is used until
+    %% all clusters are Epoch 2 capable.
+    case app_helper:get_env(riak_kv, mdc_crdt_epoch) of
+        1 ->
+            proplists:get_value(Mod, ?E1_DATATYPE_VERSIONS, 1);
+        _ ->
+            %% use any term except true to unset app env. Default to
```
seancribbs: With the name and semantic changes, this comment is now minorly misleading. Any value other than `1` triggers this clause.

russelldb:
@macintux and @seancribbs pushed a change to that last point. What do you think?

russelldb:
@borshop merge

PR #1081, opened 2015-02-15
add is_crdt convenience fn

zeeshanlakhani:
Convenience fun related to @seancribbs's comment on https://github.com/basho/yokozuna/pull/452#discussion-diff-24717256. If there's a better name for this fun, I'll update it accordingly :).

### DIFF
--- a/src/riak_kv_crdt.erl
+++ b/src/riak_kv_crdt.erl
@@ -28,6 +28,8 @@
 -export([log_merge_errors/4, meta/2, merge_value/2]).
 %% MR helper funs
 -export([value/1, counter_value/1, set_value/1, map_value/1]).
+%% Other helper funs
+-export([is_crdt/1]).
 
 -include("riak_kv_wm_raw.hrl").
 -include("riak_object.hrl").
@@ -131,6 +133,19 @@ set_value(RObj) ->
 map_value(RObj) ->
     {{_Ctx, Map}, _Stats}  = value(RObj, ?MAP_TYPE),
     Map.
+
+%% @doc convenience function for (e.g.) Yokozuna. Checks the bucket props for
+%% the object, if it has a supported datatype entry, returns true; otherwise
+%% false if not a 2.0 CRDT.
+-spec is_crdt(riak_object:riak_object()) -> boolean().
+is_crdt(RObj) ->
+    Bucket = riak_object:bucket(RObj),
+    case riak_core_bucket:get_bucket(Bucket) of
+        BProps when is_list(BProps) ->
+            Type = proplists:get_value(datatype, BProps),
+            Mod = riak_kv_crdt:to_mod(Type),
+            supported(Mod)
+    end.
 
 %% @TODO in riak_dt change value to query allow query to take an
 %% argument, (so as to query subfields of map, or set membership etc)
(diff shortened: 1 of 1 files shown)


src/riak_kv_crdt.erl
```diff
@@ -132,6 +134,19 @@ map_value(RObj) ->
     {{_Ctx, Map}, _Stats}  = value(RObj, ?MAP_TYPE),
     Map.
 
+%% @doc convenience function for (e.g.) Yokozuna. Checks the bucket props for
+%% the object, if it has a supported datatype entry, returns true; otherwise
+%% false if not a 2.0 CRDT.
+-spec is_crdt(riak_object:riak_object()) -> boolean().
+is_crdt(RObj) ->
+    Bucket = riak_object:bucket(RObj),
+    case riak_core_bucket:get_bucket(Bucket) of
```
russelldb, line 143: Why the `case` statement and not just:

```
BProps = riak_core_bucket:get_bucket(Bucket),
```

?

src/riak_kv_crdt.erl
```diff
@@ -507,13 +522,13 @@ eqc_test_() ->
      ?_test(?TIMED_QC(prop_binary_roundtrip()))}.
 
 prop_binary_roundtrip() ->
-    ?FORALL({_Type, Mod}, oneof(?MOD_MAP), 
+    ?FORALL({_Type, Mod}, oneof(?MOD_MAP),
```
russelldb, line 578: Is your emacs set to automatically remove trailing whitspace on save?

russelldb:
Looks good to me, although a unit test wouldn't hurt.

src/riak_kv_crdt.erl
zeeshanlakhani, line 578: Yeah. I know it can cause confusion. Should I turn it off?

src/riak_kv_crdt.erl
zeeshanlakhani, line 143: I kinda stole most of this from https://github.com/basho/riak_kv/blob/feature/zl/helper-fn-check-if-object-bucket-props-supports-crdt/src/riak_kv_crdt.erl#L103 haha, which probably has it just for the guard. But, you're right, the case is unnecessary.

zeeshanlakhani:
Yeah... I'm basically testing it for the use-case here, https://github.com/basho/yokozuna/pull/452/files#diff-a765b250cdc4d78ac1eef6cfddbbb280R508, but I'll add a unit test.

src/riak_kv_crdt.erl
russelldb, line 578: I turned mine off after working on some old files and making every line into the diff, despite a small change. I guess I'm indifferent about it in general (I argued that having the autosave feature was a good thing since trailing whitespace is "bad"), though it can make review harder for more complex changes since it draws the reviewers attention.

russelldb:
Thanks, I think a unit test here is needed. Just something simple, meck out riak_core_bucket and return buckets for all the datatypes, some datatype=nonsense, and some no datatype is exhaustive.

src/riak_kv_crdt.erl
zeeshanlakhani, line 578: Yeah... for larger diffs w/ older files, I'll make sure to turn it off. I also believe trailing whitespace is "bad."

src/riak_kv_crdt.erl
zeeshanlakhani, line 143: Looking at other spots, e.g. https://github.com/basho/riak_kv/blob/ec7e525fd3af7e6324474e8ea8949626c5c8b061/src/riak_kv_util.erl#L163, I decided to keep the case, but actually have an error case if BProps is not a list for some reason.

seancribbs:
:+1: 9f17bcc

zeeshanlakhani:
@borshop merge

PR #1082, opened 2015-02-16
Introduce `riak_kv_multi_prefix_backend`

kuenishi:
`riak_cs_kv_multi_backend` has been separated from riak_kv repo, as it was rather independent and supposed to be only for Riak CS. After a long time, it has been well tested in many production systems, it would be good time to bundle this into KV repository.
Now that backend is reliable enough so that the cost of configuration complexity for this multibackend is now much bigger. So now it is time to bundle in Riak release package.

This pull request copies `riak_cs_kv_multi_backend` to this repository with its name changed to `riak_kv_multi_prefix_backend`. This pull request also introduces a new config item to simplify backend configuration for CS without harming current KV clusters. After this pull request, `riak.conf` to run KV as CS backend will be like this:

```
storage_backend = prefix_multi
cs_version = 20000
```

You'll still need `add_paths` for storage calculation in `advanced.config`, but no other configuration will required. If you still unlike this, old configurations can be put out to `advanced.config`, too.

### DIFF
--- a/src/riak_kv_multi_prefix_backend.erl
+++ b/src/riak_kv_multi_prefix_backend.erl
@@ -1,0 +1,831 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_multi_prefix_backend: switching between multiple storage engines
+%%
+%% Copyright (c) 2007-2015 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc riak_kv_multi_backend allows you to run multiple backends within a
+%% single Riak instance. The 'backend' property of a bucket specifies
+%% the backend in which the object should be stored. If no 'backend'
+%% is specified, then the 'multi_backend_default' setting is used.
+%% If this is unset, then the first defined backend is used.
+%%
+%% If the 'multi_backend_prefix_list' list is defined and is non-empty,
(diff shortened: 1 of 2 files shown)


kuenishi:
~~Note: `{allow_mult, true}` for riak_core is still needed after this pull request.~~ it is defined in riak_kv.schema.

src/riak_kv_multi_prefix_backend.erl
```diff
@@ -0,0 +1,854 @@
...
+    {ok, Caps} = Mod:capabilities(ModState),
+    lists:member(index_reformat, Caps).
+
+%% ===================================================================
+%% EUnit tests
+%% ===================================================================
+-ifdef(TEST).
+-ifdef(TEST_IN_RIAK_KV).
+
+%% %% @private
```
cmeiklejohn, line 668: Why is all of this code commented out?

cmeiklejohn:
I'm a bit concerned about the commented out tests, given it looks like we're about to merge a new backend without any coverage.  Is there a reason for the removal of the tests?

cmeiklejohn:
@slfritchie Now that Kota's on leave, can you comment on the included, commented out, tests here?

slfritchie:
Hrm, my guess is that those test functions are from the original generic backend test suite + the original multi backend's invocation of them ... and that this is a prerequisite PR for later 2.0-related merge work.  @shino would know better than me about the prerequisites, though.  Shino, your opinion?

Given that this backend is only used by CS, and CS isn't even ready for it quite yet, I'd be OK with making a new ticket to uncomment the tests and get them passing _prior_ to the 2.0 CS release.

shino:
@slfritchie it seems ancient history before i know of it, i.e. I have no idea :smile:

cmeiklejohn:
OK, well, I'm :+1: to merge, given it's only used by Riak CS.

cmeiklejohn:
@borshop merge

cmeiklejohn:
:+1: 7270a0e

cmeiklejohn:
@borshop merge

kuenishi:
Thank you for review and merge! The problem being just only for CS is riak.conf, where cuttlefish automatically shows a choice of `prefix_multi` in the comment. So I'm +1 to Scott:

> Given that this backend is only used by CS, and CS isn't even ready for it quite yet, I'd be OK with making a new ticket to uncomment the tests and get them passing _prior_ to the 2.0 CS release.

kuenishi:
But that depends on _when_ this code is shipped. Unlikely to be in 2.0 series... then just old style backend configuration will be used and described in the release doc.

PR #1083, opened 2015-02-16
add wm code for getting active preflist

zeeshanlakhani:
**Note**: **Bors will fail** until **https://github.com/basho/riak_pb/pull/105** and **https://github.com/basho/riak_api/pull/79** is reviewed/merged.

Update for webmachine/pb interface.
### Description

Part of [RIAK-1481](https://bashoeng.atlassian.net/browse/RIAK-1481).

The active preflist will return primary/fallback partitions and nodes for the available nodes at the time of query. Primary/fallback will be annotated.
This involves updates to our RiakKV WB code, RiakPB, Riak-Erlang-Http-Client, Riak-Erlang-Client.

*The impetus for adding this to the API started with a mailing list question answered by Charlie Voiselle, http://lists.basho.com/pipermail/riak-users_lists.basho.com/2015-January/016527.html, which involved a snippet of code that we've given clients multiple times. This was then discussed with Russell Brown on HipChat, https://basho.hipchat.com/history/room/867200/2015/01/13?q=enterprising&t=rid-867200#12:23:22.

Other PRs in the series:
- https://github.com/basho/riak_pb/pull/105
- https://github.com/basho/riak-erlang-http-client/pull/50
- https://github.com/basho/riak-erlang-client/pull/204

### DIFF
--- a/src/riak_kv_pb_bucket.erl
+++ b/src/riak_kv_pb_bucket.erl
@@ -50,7 +50,9 @@
          decode/2,
          encode/1,
          process/2,
-         process_stream/3]).
+         process_stream/3,
+         bucket_type/2,
+         maybe_create_bucket_type/2]).
 
 -record(state, {client,    % local client
                 req,       % current request (for multi-message requests like list keys)
@@ -186,8 +188,17 @@ check_bucket_type(Type) ->
             error
     end.
 
--spec maybe_create_bucket_type(binary(), binary()) -> binary() | {binary(), binary()}.
+-spec maybe_create_bucket_type(binary()|undefined, binary()) -> riak_core_bucket:bucket().
 maybe_create_bucket_type(<<"default">>, Bucket) ->
+    Bucket;
+maybe_create_bucket_type(undefined, Bucket) ->
     Bucket;
 maybe_create_bucket_type(Type, Bucket) when is_binary(Type) ->
     {Type, Bucket}.
+
+%% always construct {Type, Bucket} tuple, filling in default type if needed
+-spec bucket_type(binary()|undefined, binary()) -> riak_core_bucket:bucket().
+bucket_type(undefined, B) ->
+    {<<"default">>, B};
+bucket_type(T, B) ->
+    {T, B}.
--- a/src/riak_kv_pb_bucket_key_apl.erl
+++ b/src/riak_kv_pb_bucket_key_apl.erl
@@ -1,0 +1,78 @@
+%% --------------------------------------------------------------------------
+%%
+%% riak_kv_pb_bucket_key_apl: Expose Core active preflist functionality to
+%%                            Protocol Buffers
+%%
+%% Copyright (c) 2015 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% --------------------------------------------------------------------------
+
+%% @doc <p>The Bucket-Key Preflist (Primaries & Fallbacks) PB service
+%% for Riak Core. This service covers the following request messages in the
+%% original protocol:</p>
+%%
+%% <pre>
+%% 33 - RpbGetBucketKeyPreflistReq
+%% </pre>
+%%
+%% <p>This service produces the following responses:</p>
+%%
+%% <pre>
+%% 34 - RpbGetBucketKeyPreflistResp
+%% </pre>
+%%
+%% @end
+
+-module(riak_kv_pb_bucket_key_apl).
+
+-export([init/0,
+         decode/2,
+         encode/1,
+         process/2,
+         process_stream/3]).
+
+-include_lib("riak_pb/include/riak_kv_pb.hrl").
+
+init() ->
+    undefined.
+
+%% @doc decode/2 callback. Decodes an incoming message.
+decode(Code, Bin) when Code == 33 ->
+    Msg = #rpbgetbucketkeypreflistreq{type =T, bucket =B, key =Key} =
+        riak_pb_codec:decode(Code, Bin),
+    Bucket = riak_kv_pb_bucket:bucket_type(T, B),
+    {ok, Msg, {"riak_kv.get_preflist", {Bucket, Key}}}.
+
+%% @doc encode/1 callback. Encodes an outgoing response message.
+encode(Message) ->
+    {ok, riak_pb_codec:encode(Message)}.
+
+%% Get bucket-key preflist primaries
+process(#rpbgetbucketkeypreflistreq{bucket = <<>>}, State) ->
+    {error, "Bucket cannot be zero-length", State};
+process(#rpbgetbucketkeypreflistreq{key = <<>>}, State) ->
+    {error, "Key cannot be zero-length", State};
+process(#rpbgetbucketkeypreflistreq{type = <<>>}, State) ->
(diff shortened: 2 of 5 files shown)


src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,226 @@
...
+         allowed_methods/2,
+         resource_exists/2,
+         content_types_provided/2,
+         encodings_provided/2,
+         produce_preflist_body/2,
+         malformed_request/2
+        ]).
+
+-record(ctx, {
+          api_version :: integer(),                   %% version of the API to use.
```
seancribbs: Not sure copying this state record wholesale is necessary. E.g., you don't need:
1. `api_version` (always 3 or whatever, this is a new api)
2. `client` (you're only getting bucket props)
3. `prefix` (again related to ancient API)
4. `riak` (this is just legacy cruft, we only "connect" locally)
5. `method` (are you supporting anything other than GET/HEAD?)

zeeshanlakhani: Yeah, I was gonna ask. Will update.

src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,226 @@
...
+                    {false, RD, Ctx}
+            end
+    end.
+
+-spec allowed_methods(#wm_reqdata{}, context()) ->
+    {[atom()], #wm_reqdata{}, context()}.
+%% @doc Get the list of methods this resource supports.
+%%      Properties allows, GET.
+allowed_methods(RD, Ctx) ->
+    {['GET'], RD, Ctx}.
```
seancribbs: `HEAD` might be good to support, and requires no extra work.

src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,226 @@
...
+                                    #wm_reqdata{},
+                                    context()}.
+%% @doc Produce bucket/key annotated preflist as JSON.
+produce_preflist_body(RD, #ctx{bucket=Bucket0,
+                               bucket_type=BType,
+                               key=Key}=Ctx) ->
+    Bucket = riak_kv_wm_utils:maybe_bucket_type(BType, Bucket0),
+    Preflist = riak_core_apl:get_apl_ann_with_pnum({Bucket, Key}),
+    case Preflist of
+        [] ->
```
seancribbs: In what case can the preflist calculation return an empty list? N=0?

src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,226 @@
...
+        {{PartitionNumber, Node}, T} <- Preflist].
+
+-spec jsonify_results({non_neg_integer(), node()}, primary|fallback) ->
+                            [{struct, list()}].
+
+jsonify_results({PartitionNumber, Node}, primary) ->
+    [{struct,
+      [{<<"partition">>, PartitionNumber},
+       {<<"node">>, atom_to_binary(Node, utf8)},
+       {<<"primary">>, true}]}]
```
seancribbs: You could collapse this function to a single clause if you replace `primary` in the function head with `Ann` and use `Ann == primary` in the field. I would also call this function `jsonify_result` or `jsonify_preflist_entry` or something to distinguish it from the 1-arity list-taking function above.

src/riak_kv_wm_preflist.erl
zeeshanlakhani: @seancribbs good point... I guess this was in the case that no preflist was found. But, all the ann preflist specs expect values within the list. But, yes, my thought was at N=0, which means the riak_core fun show allow such a type. What are you're thoughts?

src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,226 @@
...
+                                        lists:flatten(jsonify_results(P))}]}),
+            {Json, RD, Ctx}
+    end.
+
+%% Private
+
+-spec jsonify_results(riak_core_apl:preflist_with_pnum_ann()) -> list().
+%% @doc Jsonify active preflist to json.
+jsonify_results(Preflist) ->
+    [jsonify_results({PartitionNumber, Node}, T) ||
```
seancribbs: Nitpick: I'd either completely destructure these tuples (make the called function 3-arity), or just pass the entire list entry to the called function.

src/riak_kv_wm_preflist.erl
seancribbs: Pretty sure N=0 is a nonsense value.

src/riak_kv_pb_bucket.erl
```diff
@@ -186,8 +188,17 @@ check_bucket_type(Type) ->
             error
     end.
 
--spec maybe_create_bucket_type(binary(), binary()) -> binary() | {binary(), binary()}.
+-spec maybe_create_bucket_type(binary(), binary()) -> riak_core_bucket:bucket().
```
seancribbs: This spec looks clearly violated if one can pass `undefined`.

zeeshanlakhani: hmm, yeah, odd. fixing and will rerun.

src/riak_kv_pb_bucket_key_apl.erl
```diff
@@ -0,0 +1,88 @@
...
+-include_lib("riak_pb/include/riak_pb_kv_codec.hrl").
+
+init() ->
+    undefined.
+
+%% @doc decode/2 callback. Decodes an incoming message.
+decode(Code, Bin) when Code == 33 ->
+    Msg =  riak_pb_codec:decode(Code, Bin),
+    case Msg of
+        #rpbgetbucketkeypreflistreq{type =T, bucket =B, key =Key} ->
```
seancribbs: Don't make a case expression with only one clause, simply pattern match on the result, or skip the match altogether.

``` erlang
Msg = #rpbgetbucketkeypreflistreq{type=T, bucket=B, key=Key} = riak_pb_codec:decode(Code, Bin),
Bucket = riak_kv_pb_bucket:bucket_type(T, B),
{ok, Msg, {"riak_core.get_preflist", {Bucket, Key}}
```

zeeshanlakhani: yep... I tend to start w/ multiple cases and then cleanup lamely... fixing :).

src/riak_kv_pb_bucket_key_apl.erl
```diff
@@ -0,0 +1,88 @@
...
+    {error, "Bucket cannot be zero-length", State};
+process(#rpbgetbucketkeypreflistreq{key = <<>>}, State) ->
+    {error, "Key cannot be zero-length", State};
+process(#rpbgetbucketkeypreflistreq{type = <<>>}, State) ->
+    {error, "Type cannot be zero-length", State};
+process(#rpbgetbucketkeypreflistreq{type=T, bucket=B0, key =K}, State) ->
+    B = riak_kv_pb_bucket:maybe_create_bucket_type(T, B0),
+    Preflist = riak_core_apl:get_apl_ann_with_pnum({B, K}),
+    case Preflist of
+        [] ->
```
seancribbs: Again, not likely for a preflist to be empty. Even if it is, it seems suitable to send the empty preflist to the client.

src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,165 @@
...
+-spec encodings_provided(#wm_reqdata{}, context()) ->
+    {[{Encoding::string(), Producer::function()}], #wm_reqdata{}, context()}.
+%% @doc List the encodings available for representing this resource.
+%%      "identity" and "gzip" are available for a preflist request.
+encodings_provided(RD, Ctx) ->
+    %% identity and gzip for top-level and bucket-level requests
+    {riak_kv_wm_utils:default_encodings(), RD, Ctx}.
+
+resource_exists(RD, #ctx{bucket_type=BType}=Ctx) ->
+    {riak_kv_wm_utils:bucket_type_exists(BType), RD, Ctx}.
```
seancribbs, line 128: Does this function handle bucket types of 'undefined'?

seancribbs, line 128: Strike that, I see that it sets the bucket_type to `<<"default">>` in riak_kv_web.

seancribbs:
:+1: 9fede91

zeeshanlakhani:
@borshop merge

PR #1084, opened 2015-02-19
Feature/az/api entrypoints coverage

hmmr:
This PR depends on https://github.com/basho/riak_pb/pull/106, in turn depending on the bunch of PRs by @zeeshanlakhani.

### DIFF
--- a/src/riak_kv_apiep.erl
+++ b/src/riak_kv_apiep.erl
@@ -1,0 +1,89 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_kv_apiep: Common functions for ring/coverage protobuff
+%%                callback and Webmachine resource
+%%
+%% Copyright (c) 2015 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Supporting functions shared between riak_kv_{wm,pb}_apiep.
+
+-module(riak_kv_apiep).
+
+-export([get_entrypoints/1, get_entrypoints_json/1,
+         get_entrypoints/2, get_entrypoints_json/2]).
+
+-type proto() :: http|pbc.
+-type ep() :: {string(), non_neg_integer()}.
+-type bkey() :: {Bucket::binary(), Key::binary()}.
+
+-define(RPC_TIMEOUT, 10000).
+
+
+-spec get_entrypoints_json(proto(), bkey()) -> iolist().
+%% @doc Produce API entry points for a given Bucket and Key, in a JSON form.
+get_entrypoints_json(Proto, BKey) ->
+    mochijson2:encode(
+       {struct, get_entrypoints(Proto, BKey)}).
+
+-spec get_entrypoints_json(proto()) -> iolist().
+%% @doc Produce all API entry points in a JSON form.
+get_entrypoints_json(Proto) ->
+    mochijson2:encode(
+       {struct, get_entrypoints(Proto)}).
+
+
+-spec get_entrypoints(proto(), bkey()) -> [ep()].
+%% @doc For a given protocol, determine ip:port entry points of hosts
+%%      running riak nodes containing requested bucket and key.
+get_entrypoints(Proto, BKey) ->
+    UpNodes = riak_core_node_watcher:nodes(riak_kv),
+    Preflist = riak_core_apl:get_apl_ann(BKey, UpNodes),
+    Nodes =
+        lists:usort(
+          [N || {{_Index, N}, _Type} <- Preflist]),  %% filter on type?
+    get_proto_entrypoints(Nodes, Proto).
+
+-spec get_entrypoints(proto()) -> [ep()].
+%% @doc Returns all API entry points for a given protocol.
+get_entrypoints(Proto) ->
+    Nodes = riak_core_node_watcher:nodes(riak_kv),
+    %% is it possible that a vnode running riak_kv has no listeners?
+    get_proto_entrypoints(Nodes, Proto).
+
+
+%% ===================================================================
+%% Local functions
+%% ===================================================================
+
+-spec get_proto_entrypoints([node()], proto()) -> [ep()].
+%% @private
+get_proto_entrypoints(Nodes, Proto) ->
+    %% 0. check that listeners are up, and also get the port numbers
+    {ResL, FailedNodes} =
+        rpc:multicall(
+          Nodes, riak_api_lib, get_active_listener_entrypoints, [Proto], ?RPC_TIMEOUT),
+    case FailedNodes of
+        [] ->
+            fine;
+        FailedNodes ->
+            lager:log(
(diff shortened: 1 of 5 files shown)


src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,229 @@
...
+          client,       %% riak_client() - the store client
+          prefix,       %% string() - prefix for resource uris
+          riak,         %% local | {node(), atom()} - params for riak client
+          method,       %% atom() - HTTP method for the request
+          security      %% security context
+         }).
+
+-type context() :: #ctx{}.
+%% @doc Annotated preflist type
+-type preflist_with_number_ann() :: [ {{non_neg_integer(), node()},
```
zeeshanlakhani: @hmmr take a look at https://github.com/basho/riak_core/pull/705/files#diff-b5b475eb9ebcaac228237ca394f6aaacR49 to keep consistency, as we now export partition_id.

src/riak_kv_wm_preflist.erl
```diff
@@ -0,0 +1,229 @@
...
+         allowed_methods/2,
+         resource_exists/2,
+         content_types_provided/2,
+         encodings_provided/2,
+         produce_preflist_body/2,
+         malformed_request/2
+        ]).
+
+-record(ctx, {
+          api_version,  %% integer() - Determine which version of the API to use.
```
cmeiklejohn: Is there a reason you are not specifying the types in the signature instead of just in a comment?  For example: `api_version :: integer()`.

zeeshanlakhani: That's actually from my PR... but was keeping consistency w/ https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_object.erl#L157 and other files. The better question is why we've been doing that I reckon.

zeeshanlakhani: @cmeiklejohn @hmmr I've made an update to https://github.com/basho/riak_kv/blob/develop/src/riak_kv_wm_object.erl#L157 that updates types for wm_preflist. Still relies on riak_core pr -> https://github.com/basho/riak_core/pull/705/files (for funs and types).

hmmr: Actually, I have no function that takes a preflist as an argument or returns one, so I should be fine. It's good to know where those types are, however.

An unrelated question: in newly written code, there is only one api_version I have to deal with (the current one). Do I still need to future-proof it by introducing api_version into my ctx record?

hmmr:
It still merges cleanly, so no changes; just a rebase onto HEAD of develop. Failed due to required bits (https://github.com/basho/riak_pb/pull/106) still awaiting approval.

seancribbs:
I disagree generally with the approach. We already have preflist calculation as an API request, why not simply provide the nodename + host/port/protocol for every interface? 

The client can cache the network information and simply request the preflist for the desired bucket-type/bucket/key, and then match up the nodenames to existing connections or new connections to be opened. Obviously, the client should also periodically re-request for the live nodes' interfaces, or by be triggered some signal from the server.

Also, I'm doubful about how you're deriving the host interface information, because it is totally valid for a listener to start up with `0.0.0.0` or `::0` as the address, which binds all interfaces but is not routable. Also, you don't have adequate information about how the client can (or cannot) be routed to the bound interfaces.

hmmr:
@seancribbs The issue you point out concerning listening ip where the external ip is meant, is now quite embarrassingly obvious to me. It's a blunder I've just committed (no, I have sat on it feeling good for a week!). What's needed here is some code path eventually calling `getifaddrs()` and picking the `addr` of the interface we assume to be the one externally accessible.

I will review the whole thing tomorrow and report back.

hmmr:
Reworked, now the functions return the proper external ip instead of 127.0.0.1.

@seancribbs concerning your suggestion to return the nodename and host/port, much as I'd like to just go with that, the java guys seem to only want the external ip and the port(s), not the nodenames proper or the part after '@' in those. I've been scratching my head hard, but other than calling getifaddrs on location, I just don't see how one can figure out the ip to which external clients can connect. I'll be happy to stand corrected, but at any rate, the scope of DP-6 is precisely about that.

I grepped the entire riak/deps/ and found no matches on getifaddr (except an odd one in erlang_js).

I am provisionally force-pushing the amended commit, for @srgg to continue his work on riak java client for the Thursday demo. There are a couple of issues to resolve, notably what 'heuristics' to apply in the case of multihomed hosts (such as those with infiniband in addition to plain eth0).

seancribbs:
I suppose this is fine for a demo, but before this merges, I'd like to see the following changes:
1. The host interface information is at a fundamentally different granularity than the preflist information. Since we already have requests for fetching the preflist, there's no reason to duplicate that effort here; all we need is the host/port/protocol information.
2. The code in this pull request RPCs to _every node in the preflist_ on _every request_. The host interface information doesn't change often enough to warrant this, and under load RPC will become a bottleneck. Instead, each host should publish its listener information in cluster metadata, which will be gossiped around automatically. Fetching the data is an inexpensive, local operation.

hmmr:
To go the whole hog, I now see I'll have to:
1. Provide a function to enumerate the interfaces and export the bound addresses to the preflist (still have to figure out how to do the latter).
2. For the first time, call that function at listener app start, then call it once every minute (or not, with a big fat warning to whoever will be moving hosts around, to take care).
3. If a listener is configured to listen on an address other than 0.0.0.0 that ends up being different from what has been discovered, reconfigure that listener.
4. I will scrap my bucket+key coverage call. The other call, which returns the full coverage map, would remain.

Is that a plan?

seancribbs:
1. I don't know what the preflist has to do with the bound addresses, but yes, you will need to export them to cluster metadata.
2. I don't think you need to check every minute. Why would a listener address be changing that often? Or are you talking about from the client? If the latter, then perhaps that could be configurable. Polling might be ok, or triggered on a failed connection to a previously discovered host.
3. Not sure what you mean here. Can you elaborate?
4. When you say "coverage map", what do you mean exactly? "Coverage" usually means "a list of partition/node pairs covering the entire keyspace", and doesn't require a bucket/key to seed it. The "preflist (preference list)" usually refers to the replicas calculated for an individual key, based on consistent hashing.

hmmr:
@seancribbs Thanks for commenting and dropping precious hints.

I. My line of reasoning was about this call: https://github.com/basho/riak_kv/blob/develop/src/riak_kv_web.erl#L115. It returns (on my devrel setup):

`$ curl  localhost:10018//buckets/B/keys/K/preflist`
`{"preflist":[{"partition":29,"node":"dev5@127.0.0.1","primary":true},...]}`

where I expect these new properties to appear: `"http_ip":"192.168.33.44", "http_port":10018` (plus same for pbc). I see I used the word 'preflist' meaning 'cluster metadata' in my previous comment, inaccurate me. But, are these new entry point details now in the proper place as returned by this call?

II. Indeed, let it be triggered by a dropped connection (and failure to establish one initially, too). Then, when a client times out or sees it gone, it will re-request the entire entry points map, possibly with &force_refresh=1 (or some such) to cause it to be regenerated. And then no periodic autorefresh is necessary.

III. Suppose riak.conf initially has, e.g, `listener.protobuf.internal = 10.10.10.10:10017`, and then the host migrates from 10.10.10.10 to 10.10.10.11. If the configuration had 0.0.0.0:10017, the listener would just keep going. Otherwise, the listener would listen on an address which is no longer routed (it would even fail to restart, correct?).

My proposition is to check whether the configured listening address, unless it is 0.0.0.0, does match the bound address on some ethX in the system, and take steps to correct the situation.

IV. It's simply the complete list of all API entry points into the cluster. The idea is to extend this list with details on what key ranges are quickest accessed via each entry point -- and thus attain the meaning of "coverage". Properly naming things goes a long way, indeed.

hmmr:
Another reworking effort to address @seancribbs latest comments. Due to massive changes over what went into this branch, I pushed the new work into https://github.com/basho/riak_kv/tree/feature/az/api-entrypoints-coverage-rebased.

hmmr:
Work and further discussions moved to https://github.com/basho/riak_kv/pull/1087; closing this one then.

PR #1087, opened 2015-03-11
DP-6 API entry points coverage, http and pbc API, cached

hmmr:
There are accompanying changes in newly created branches:
- riak_pb/feature/az/api-entrypoints-coverage, and
- riak_api/feature/az/api-entrypoints-coverage

that will have to be merged first.  Python client changes are in
- riak-python-client/feature/az/api-entrypoints-coverage.

The following HTTP requests are accepted (values for parameter P are
"http", "pbc", assuming the latter by default; for F, values are "1"
and "0", the latter by default):

1) /ring/coverage?proto=P&force_update=F

  returning a JSON like this:

   {<del>"struct":</del>{"addr":"192.168.1.10","port":10017,"last_checked":<del>1426041629</del>"2015-03-15T11:22:33"},
    ...}

providing a complete set of API entry points into the cluster via
protocol P. The value of 'last_checked' is the time of the last
actual call to getifaddrs/0 used to get the bound address.

2) /ring/coverage/bucket/B/key/K?proto=P&force_update=F

returning a similar JSON with details as above.  There will be a
single element in the list, in which addr:port is where an entry point
is for shortest data access to the key K in bucket B, via the given
protocol API.

On the first call, or with force_update=1, results are cached in
cluster metadata.  There is no expiry period; clients finding their
connections dropped should re-request interesting entry points with
force_update=1.

HTTP responses are further cached on their own, until the next
multiple of (a default of) 15 secs, configurable via parameter
'ring_coverage_cache_expiry_time'.

For pb, two new messages are provided:

 RpbApiEpReq (code 90),
 RpbApiEpResp(code 91).

declared in riak_pb/src/riak_kv.proto.

The coverage information is obtained in two steps:
1. using riak_core_apl:get_apl_ann(), get active preflist for given
   Bucket and Key, and extract riak nodes from it;
2. make an rpc call on those nodes to:
   
   (a) determine which of the relevant listener apps (i.e., riak_api_pb_listener,
      riak_api_web) are running on those nodes,
      and find the ports these are listening on.
   
   (b) get the bound ip address of the first non-loopback interface on the
      underlying os (all code is in riak_api/src/riak_api_lib.erl).
   
      If a listener was configured with a specific address rather than a wildcard,
      then when the bound ip does not match it, a warning is logged and 'not_routed'
      is returned as a value of 'addr' in the JSON blob.

The corresponding new riak python client method is:
- client.get_api_entry_points(bucket, key, proto = 'pbc', force_update = False)

where bucket and key can be None to request the complete map.


hmmr:
@seancribbs At this point, it looks to me good enough for the next round of review. The two PRs dependent on this one are https://github.com/basho/riak_pb/pull/106 and https://github.com/basho/riak_api/pull/84. Then, there is a PR in https://github.com/basho/riak_api/pull/84.

mdigan:
Closing this pull request. We're taking a different approach. John and Mark will be enhancing the coverage plan API, and we'll be adding functionality in the java client to use the new API features, and using all of those in the spark connector.

PR #1094, opened 2015-03-19
Fix dataloss bug caused by reusing causal histories

russelldb:
See http://github.com/riak_kv/issues/679 for a full history and
discussion. In summary: certain scenarios may lead to riak breaking a
strict invariant of version vectors, namely that each actor issue an
increasing, total order of events. When this invariant is broken some
new data item may be lost when some older item appears to be from a
later causal time. This fix adds per-key-epochs, so causal histories
are not re-used.
## Examples:

Tombstone->doomstone
- Key is deleted
- Tombstone is written to 2 primaries and a fallback
- fallback is unavailable
- tombstone read-repaired over 3 primaries and reaped
- new key written
- fallback hands off
- old tombstone subsumes new key, data lost

Dataloss
- A replica becomes unavailable in a put coordinating node due to corruption or user error.
- The incoming object may or may not contain a vector clock.
- We get a not_found in the local put. We have lost the information corresponding to this vnode id.
- The vector clock created could have a stale number for this vnode
  actor id. Either 1 if no vclock or < than the real one if more
  writes have happened since.
- The vector clock is subsumed by the other replicas when sent out, so
  this write will lose and will eventually be read repaired completely
  out of existence.

There are others, but these are the two that this commit addresses.
See the kv679\* tests in riak_test for more details.
## How it Does it:

When a coordinating vnode updates a key it has no local clock entry
for, it creates a new vnode id for that key only, using its base
vnodeid and a monotonic per vnode epoch counter, this ensures that any
version vector from the past/future would conflict with the new one,
not overwrite it. We use a counter lease scheme, and async leasing to
keep the counter update durable and quick. The actor is per-key, so no
 general actor explosion occurrs.

There is an off switch {riak_kv, [{use_epoch_counter, false}]}. The
default is `true`.
## Not done:

There is another flavour of the issue, see private issue tracker, that
is not yet addressed, but can be in future with some extra clock
metadata.

### DIFF
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -73,22 +73,12 @@ callback_after(Time, Ref, Msg) when is_integer(Time), is_reference(Ref) ->
 -ifdef(TEST).
 
 standard_test(BackendMod, Config) ->
-    {spawn,
-     [
-      {setup,
-       local,
-       fun() -> setup({BackendMod, Config}) end,
-       fun cleanup/1,
-       fun(X) ->
-               [basic_store_and_fetch(X),
-                fold_buckets(X),
-                fold_keys(X),
-                delete_object(X),
-                fold_objects(X),
-                empty_check(X)
-               ]
-       end
-      }]}.
+    {"Basic Backend",
+     fun() ->
+             {Mod, State} = setup({BackendMod, Config}),
+             State2 = basic_store_and_fetch(Mod, State),
+             cleanup({Mod, State2})
+     end}.
 
 make_test_bucket(Backend, Suffix) ->
     try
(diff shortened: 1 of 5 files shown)


russelldb:
This is the rebased-onto-develop, squashed, and with-added-off-switch version of the kv679 work It's been +1's by @slfritchie and @andrewjstone before, but I would really appreciate a final set, or two, of eyes on it before it is merged for the next release.

Differences to last reviewed code at https://github.com/basho/riak_kv/pull/1070 is
1. off switch
2. some document/comment typos fixes

src/riak_kv_backend.erl
```diff
@@ -231,62 +206,55 @@ fold_objects({Backend, State}) ->
...
+                  {{B3,K3},<<"v3">>}],
+                 begin
+                     {ok, Objects} =
+                         Backend:fold_objects(FoldObjectsFun,
+                                              [],
+                                              [],
+                                              State2),
+                     lists:sort(ObjFilter(Objects))
+                 end),
+    empty_check(Backend, State).
```
slfritchie: Pass `State2` in 2nd arg?

src/riak_kv_backend.erl
```diff
@@ -231,62 +206,55 @@ fold_objects({Backend, State}) ->
...
-             ?assertEqual(false, Backend:is_empty(State)),
-             ?assertMatch({ok, _}, Backend:delete(B1,K1, [], State)),
-             ?assertMatch({ok, _}, Backend:delete(B3,K3, [], State)),
-             ?assertEqual(true, Backend:is_empty(State))
-     end
-    }.
+    ?assertEqual(false, Backend:is_empty(State)),
+    {ok, State2} = Backend:delete(B1,K1, [], State),
+    {ok, State3} = Backend:delete(B3,K3, [], State2),
+    ?assertEqual(true, Backend:is_empty(State3)),
```
slfritchie, line 256: Wow, so we rely on the B1 and B3 keys to already exist (and no other keys to exist) as inherited from the previous test. I guess that's OK, it's just a bit odd to see lines 250-252 doing gyrations to make the exact two keys that we're expecting to already be there.  I guess I'd expected (perhaps to insert some keys into a known empty state), then a fold to enumerate all keys, the delete them all.  {shrug}

russelldb, line 256: I don't know @slfritchie, I just made a basic functional backend (`dict`) and it failed all these tests as they each expect side effects in the backend and ignore the `State` var. I simply attempted to re-write them exactly as is, but use that `State` var so they work for purely functional backends.

russelldb:
Addressed review comment.

seancribbs:
:+1: 8f54cc7

seancribbs:
@russelldb This looks broken on the builder because of meck but everything passes cleanly on my machine. I'll look into it in the morning.

seancribbs:
@borshop merge

seancribbs:
@russelldb Cleared up the builder, cheers. :tada: :balloon:

PR #1097, opened 2015-03-23
Initial Implementation of data validation module

gordonguthrie:
This commit introduces:

the first cut of the intermediate output of the DDL language as a set of records
a library module that works on examples of those records to generate validation functions in Erlang and load them into the VM
a comprehensive test suite
The types supported for this are:

integer
float
binary
timeseries
booleans
maps
sets
any (the values in this field)
Maps, and sets are expositional - work needs to be done to make them
correspond to typed CRDTs - as do other specific CRDT types

The validation generator contains a debug setting which will dump the
corresponding Erlang source and AST to a file


src/riak_kv_ddl.hrl
```diff
@@ -0,0 +1,38 @@
...
+
+-record(riak_field, {
+	  name = throw("field 'name' in #riak_field{} is required") :: binary(),
+	  type = throw("field 'type' in #riak_field{} is required"):: field_type()
+	 }).
+
+-record(ddl, {
+	  bucket     = throw("field 'bucket' in #ddl{} is required") :: binary(),
+	  fields     = []                                            :: [riak_field()],
+	  colocation = throw("field 'bucket' in #ddl{} is required") :: colocation()
```
andrewjstone: should be 'colocation' instead of 'bucket' in the throw()

src/riak_kv_ddl_compiler.erl
```diff
@@ -0,0 +1,586 @@
...
+    ModName = make_module_name(Bucket),
+    {ModAttr, N1} = make_module_attr(ModName, N),
+    {ExpAttr, N2} = make_export_attr(N1),
+    {ModName, [ModAttr, ExpAttr], N2}.
+
+make_funs(Fields, N, M) ->
+    %% we will reverse the field list at the start
+    %% becuz we can reverse an AST but keeping fields in order
+    %% makes it nicer for the next person to look at this
+    %% I <3 Me <- so helpful, such love
```
andrewjstone: lol

andrewjstone:
As discussed, we need to allow the DDL to handle optional fields.

andrewjstone:
Most of this code should have type specs. Additionally there are a bunch of dialyzer errors that require fixing.

andrewjstone:
I'm curious how generating a module will work with DDL updates. I suppose if it is only used for filtering at the edge it doesn't matter. We just reject old in flight messages as necessary.

src/riak_kv_ddl_compiler.erl
```diff
@@ -0,0 +1,878 @@
...
+    recurse_over_fns(T, N2, M2, Acc1a, [Funs | Acc2]).
+
+make_attrs(Bucket, N) when is_binary(Bucket)    ->
+    ModName = make_module_name(Bucket),
+    {ModAttr, N1} = make_module_attr(ModName, N),
+    {ExpAttr, N2} = make_export_attr(N1),
+    {ModName, [ModAttr, ExpAttr], N2}.
+
+make_funs(Fields, N, M) ->
+    {Args, Fields2} = case Fields of
```
andrewjstone, line 236: This case clause was kind of confusing to me. It took me a bit to realize that List is just Fields. I would change the name of that variable to make it more explicit. Also, if there is a way to break the whole case clause into a separate function, that may make the code read a bit better. It's not clear to me as a reader why the fields get put in tuples. 

Basically I guess I'm arguing for higher levels of abstraction here and elsewhere for those not familiar with the Erlang AST. If each set of make_functions was wrapped in a logical unit like `create_field_forms` where tuples with names were created that would allow the user to better understand how the AST was being built up and learn it in smaller doses.

src/riak_kv_ddl_compiler.erl
```diff
@@ -0,0 +1,878 @@
...
+    SrcFileName = filename:join([OutputDir, ModName]) ++ ".erl",
+    write_file(SrcFileName, Contents).
+
+write_file(FileName, Contents) ->
+    ok = filelib:ensure_dir(FileName),
+    {ok, Fd} = file:open(FileName, [write]),
+    io:fwrite(Fd, "~s~n", [Contents]),
+    file:close(Fd).
+
+recurse_over_fns([], N, _, [], Acc2) ->
```
andrewjstone: I think a comment explaining what N and M represent would make this code more understandable. As far as I can tell each AST form needs a unique number, but some numbers seem to be shared among different forms. Also, I'm not exactly sure what M represents in this case. Also, I think the function name could probably be named in a manner describing it's intent and not what it's doing. Maybe something like `build_validation_functions`.

src/riak_kv_index.hrl
```diff
@@ -0,0 +1,69 @@
...
+%% TODO these types will be improved over the duration of the time series project
+-type selection()  :: term().
+-type filter()     :: term().
+-type operator()   :: term().
+-type sorter()     :: term().
+-type combinator() :: term().
+-type colocation() :: any().
+-type limit()      :: any().
+
+-record(riak_kv_li_index_v1, {
```
andrewjstone: I realize we have other index structures in here, but it's a bad pattern to expose records in .hrl files. I think we should consider abstracting this inside a module instead of directly exposing the records, similar to how riak_object works.

include/riak_kv_index.hrl
```diff
@@ -45,4 +45,30 @@
           max_results :: integer() | undefined
          }).
 
--define(KV_INDEX_Q, #riak_kv_index_v3).
+-record(colocation, {
```
andrewjstone, line 48: Kind of a nitpick, but I don't particularly like the term colocation. We've used the term partition/hashkey/hashgroup before. I'm really just bikeshedding here, but maybe we can open the bikeshed up in timeseries room.

gordonguthrie, line 48: I agree - I couldn't think of the right name and Russell and I chewed it over but didn't come to any conclusion - I think its wrong but would like a consensus on what right is...

src/riak_kv_ddl_compiler.erl
gordonguthrie, line 236: This clause is messy because it is where the Map type (which contains its own field set) is pulled out from a list of functions and then rewrapped for a second function further down - so it is the very application specific bit of the module.

Make funs has to do 2 separate jobs.

andrewjstone:
:+1: a8e4493

Code looks better, tests pass, optional values added. Nice work @gordonguthrie.

I just want to mention here for others that this still requires parsing incoming puts and checking the ddl positions so that they can be appropriately ordered for validation.

andrewjstone:
@borshop merge

PR #1098, opened 2015-03-25
Small refactor for fast_path special buckets

russelldb:
This file needs a full refactor for a more extensible, readable, hell
_sane_ scheme. Suggesting erlog and prolog defined rules for validation
but really anything will do.

The main work here is to try and simplify enough that the tests still
pass and we can introduce validation for `fast_path` buckets. Please
review as such.

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -396,9 +396,9 @@ validate_update_default_props(New) ->
     %% Only called if not already a consistent, datatype, fast_path
     %% bucket. Check that none of those are being set to `true'/valid
     %% datatypes.
-    ensure_not_present(New, [], [], [{datatype, ""},
-                                     {consistent, false, ""},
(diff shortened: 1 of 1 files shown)


src/riak_kv_bucket.erl
```diff
@@ -325,6 +371,35 @@ validate_create_dt_props(Unvalidated0, Valid, Invalid) ->
...
+%% is not attempting to become one.
+-spec validate_update_default_props(New :: props()) ->
+                                           {Unvalidated :: props(),
+                                            Valid :: props(),
+                                            Error :: props()}.
+validate_update_default_props(New) ->
+    %% Only called if not already a consistent, datatype, fast_path
+    %% bucket. Check that none of those are being set to `true'/valid
+    %% datatypes.
+    ensure_not_present(New, [], [], [{datatype, ""},
```
russelldb: oops, these need error messages!

PR #1100, opened 2015-03-30
Feature/fd/riak 1534

fadushin:
This branch contains a special code path for "write_once" buckets.  Bucket types that have this property set are intended to be such that entries for a given key are written exactly once, in other words, no updates are expected.  Instead of the ordinary coordinated put, requiring a read before write, this code path casts write operations directly to vnodes, and aggregates replies from primaries or replicas in the pref list.  Nval, w, and pw parameters are supported.  If dw is specified and greater than w, then the w value takes dw (i.e., w = max{w, dw}).

In place of the put_fsm used in the normal path, we introduce a collection of new intermediate worker processes (implementing gen_server behavior).  The role of these intermediate processes is to dispatch put requests to vnode(proxie)s in the preflist, and to aggregate replies.  Unlike the put_fsm, the write-once workers are long-lived for the lifecyle of the raik_kv application.  They are therefore stateful, and store request state in a state-local dictionary.

The put path employs the vnode command framework support for handoff and overload.  There is, however, no overload protection on the write once workers themselves, so it is possibly to over-drive the workers if vnodes are unresponsive.

The get path remains largely unmodied, except that merges of siblings in write_once buckets will select the least object (in lexical order) of object SHA-1 hashes of siblings.  This behavior exhibits a deterministically "random" sibling selection from the point of view of the application client.

The write_once property may only be specified at bucket type creation time, and once set, cannot be modified.  This bucket property is incompatible with data types and the (strong) consistent proeprty, so attempts to set this property on buckets with data types or the conistnet property set to true will fail.


src/riak_kv_bucket.erl
```diff
@@ -263,7 +304,13 @@ coerce_bool(_) ->
 %% value to true? Well, the user maybe type "fals" so lets be careful.
 -spec validate_create_consistent_props(any(), props()) -> {props(), props(), errors()}.
 validate_create_consistent_props(true, New) ->
-    {lists:keydelete(consistent, 1, New), [{consistent, true}], []};
+    % write_once and consistent can't both be true
+    case get_boolean(write_once, New) of
+        true ->
+            {lists:keydelete(consistent, 1, New), [], [{consistent, "Fast path buckets must be not be consistent=true"}]};
```
jonmeredith: `Fast Path` should be `Write once`

src/riak_kv_bucket.erl
```diff
@@ -324,6 +371,35 @@ validate_create_dt_props(Unvalidated0, Valid, Invalid) ->
...
+-spec validate_update_default_props(New :: props()) ->
+                                           {Unvalidated :: props(),
+                                            Valid :: props(),
+                                            Error :: props()}.
+validate_update_default_props(New) ->
+    %% Only called if not already a consistent, datatype, write_once
+    %% bucket. Check that none of those are being set to `true'/valid
+    %% datatypes.
+    ensure_not_present(New, [], [], [{datatype, "`datatype` must not be defined."},
+                                     {consistent, true, "Fast Path buckets must not be consistent=true"},
```
jonmeredith: `Fast Path` should be `Write once`

src/riak_kv_bucket.erl
```diff
@@ -385,6 +455,64 @@ validate_update_dt_props(New, Valid, Invalid) ->
...
+%% @private validate the boolean property, if `write_once' was present.
+%% precondition: write_once is not an entry in Unvalidated
+%%               write_once is an entry in Valid
+%% The following rules apply when write_once is true:
+%%   - datatype may not be defined
+%%   - consistent may not be true
+-spec validate_w1c_props(props(), props(), errors()) -> {props(), props(), errors()}.
+validate_w1c_props(Unvalidated, Valid, Errors) ->
+    ensure_not_present(Unvalidated, Valid, Errors,
+                       [{consistent, true, "Fast Path buckets must not be consistent=true"},
```
jonmeredith: `Fast Path` should be `Write once`

src/riak_kv_bucket.erl
```diff
@@ -385,6 +455,64 @@ validate_update_dt_props(New, Valid, Invalid) ->
...
+                        {datatype, "Fast Path buckets must not have datatype defined"}
```
jonmeredith: `Fast Path` should be `Write once`

include/riak_kv_vnode.hrl
```diff
@@ -7,6 +7,18 @@
           start_time :: non_neg_integer(),
           options :: list()}).
 
+-record(riak_kv_w1c_put_req_v1, {
+    bkey :: {binary(),binary()},
+    encoded_val :: binary(),
```
jonmeredith: Would recommend `encoded_obj` instead to make sure people don't think it's the value that should be wrapped in a riak object at the vnode.

src/riak_kv_bucket.erl
```diff
@@ -498,103 +638,76 @@ prop_create_valid() ->
...
-                                   merge(New, Existing)
-                           end,
-
+                Result={Good, Bad} = validate(update, Bucket, Existing, New),
+
+                %% All we really want to check is that every key in
+                %% Good replaces the same key in Existing, right?
+                %% Remove `Bad' from the inputs to validate.
+                {NoBadExisting, OnlyGoodNew} = lists:foldl(fun({Name, _Err}, {Old, Neu}) ->
+                                                                   {value, V, Neu2} = lists:keytake(Name, 1, Neu),
```
seancribbs: The line wrapping here is ridiculous. Could you put the anonymous function above the fold call, or make it a module function? It doesn't look like it needs any bindings from the surrounding code.

src/riak_kv_vnode.erl
```diff
@@ -813,7 +817,33 @@ handle_command({get_index_entries, Opts},
         false ->
             lager:error("Backend ~p does not support incorrect index query", [Mod]),
             {reply, ignore, State}
-    end.
+    end;
+
+handle_command(?KV_W1C_PUT_REQ{bkey={Bucket, Key}, encoded_obj=EncodedVal, type=Type},
```
seancribbs, line 823: Some general comments, and you may leave them for later refactoring/TODO:
1. I really dislike using `case` where a function clause is possible. The nested cases _look_ similar but aren't completely, and the nesting makes that harder to see.
2. You use manual `reply` when that could instead be included in the return value of each clause. I think this obfuscates the code, especially when you discard the value of `Reply` in the first branch.
3. The return values of the branches have different types and you don't match on them. Dialyzer will complain about this. Furthermore, you export `ModState2` from the case when you could just return it.

src/riak_kv_w1c_sup.erl
```diff
@@ -0,0 +1,55 @@
...
+        MaxR :: non_neg_integer(), MaxT :: non_neg_integer()},
+        [ChildSpec :: supervisor:child_spec()]
+    }}).
+init([]) ->
+    Workers = riak_kv_w1c_worker:workers(),
+    Children = [
+        {
+            Name, {riak_kv_w1c_worker, start_link, [Name]},
+            permanent, 5000, worker, [riak_kv_w1c_worker]
+        } || Name <- tuple_to_list(Workers)],
```
seancribbs, line 50: Is the use of a tuple really necessary? Just use a list. I also really dislike having a static list... this should be a tuning knob that operators can adjust for the capacity of their nodes/clusters.

jonmeredith, line 50: I suspect the tuple is copied over from jtuples code - it's a performance hack that lets you use element(N, T) vs lists:nth(N, L) - makes a difference for long lists, but shorter ones are more dubious.

src/riak_kv_vnode.erl
fadushin, line 823: See commit 9ea0e3d630c77af181e362cfbd8524a654ed49be

src/riak_kv_bucket.erl
fadushin: Please see commit 8722eb7580503f328dbe17e316f53736aa127d8a

src/riak_kv_w1c_sup.erl
seancribbs, line 50: Sure, I figured that out in the other code. I think we need to punt on this nit but make a todo for the number of workers to be configurable.

seancribbs:
:+1: 8722eb7

fadushin:
+1 f0b3975

fadushin:
+1 208f70b

fadushin:
@borshop merge

PR #1102, opened 2015-04-09
Warn if max ports less than open file limit

joecaswell:
also attempt to guess the path to sysctl on linux if it is not already in the path.

### DIFF
--- a/src/riak_kv_env.erl
+++ b/src/riak_kv_env.erl
@@ -119,17 +119,16 @@ test_file_limit(FileLimit0) ->
             case FileLimit < 4096 of
                 true ->
                     {warn, "Open file limit of ~p is low, at least "
-                     ++ "4096 is recommended", [FileLimit]};
+                        ++ "4096 is recommended", [FileLimit]};
                 false ->
-                    {info, "Open file limit: ~p", [FileLimit]}
-            end,
-            case FileLimit >= erlang:system_info(port_limit) of
-                true ->
-                    {warn, "Erlang ports limit should be greater than open "
-                     ++ "file limit",[]};
-                false -> ok
+                    case FileLimit >= erlang:system_info(port_limit) of
+                        true ->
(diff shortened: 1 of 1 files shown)


bsparrow435:
:+1: a82d67e705aee592d025a7b111da8185f55a14a8

matthewvon:
This seems erroneous.  What is the problem being solved?  Leveldb wants 112,000 or larger for nofiles.  That has nothing to do with ports.

bsparrow435:
@matthewvon in Erlang, ports represent any communication that leaves the vm(port driver) which includes opening files and network communication. So, the open_files limit should always at least be equal to max_ports and vice versa.  More detail on this can be found in the developers guide: http://erlang.org/doc/reference_manual/ports.html and http://erlang.org/doc/man/erl.html#max_ports

bsparrow435:
@joecaswell this should have been merged a year ago, apologies. Will get it manually tested again today to make sure nothing has changed and merge it in.

matthewvon:
Still not convinced.  What user problem lead to this PR?

bsparrow435:
I'm sorry for the confusion, this has been made much simpler in newer version of the Erlang runtime as there is no longer a max_ports or max_open_file setting. They were combined for clarity into a single +Q setting which is the same as max_ports(which includes open file handles as well). However, since we havent moved to these newer erlangs yet, this change will warn users if they have a config which could lead to a system_limit crash because their settings do not make sense. The max_ports configuration includes both open files as well as open network ports so having max_ports set lower than the open file limit will effectively lower max_open_files to the value of max_ports.

This change will warn users on startup and once a day(by default) if max_ports is less than open_files.

bsparrow435:
@joecaswell tested on OSX and it successfully warned when I had a high ulimit with a manually set low max_port.  Going to fire up an ubuntu vm and quick test there as well.

bsparrow435:
Success, all unit and integration tests pass on riak_kv_env. Merged.

Issue #1086, opened 2015-02-26
Metrics doesn't work after consistent put or get. [JIRA: RIAK-1575]


ksauzz:
`riak-admin status` and `riak-admin stat` fails after consistent put or get. According to crash.log it seems exometer's process died.

crash.log

```
2015-02-27 04:31:44 =CRASH REPORT====
  crasher:
    initial call: exometer_proc:-spawn_process/2-fun-0-/0
    pid: <0.684.0>
    registered_name: []
    exception error: {function_clause,[{exometer_shallowtree,insert,[undefined,166979104625,{t,0,500,[]}],[{file,"src/exometer_shallowtree.erl"},{line,53}]},{exometer_histogram,into_heap,4,[{file,"src/exometer_histogram.erl"},{line,314}]},{exometer_histogram,update_int,3,[{file,"src/exometer_histogram.erl"},{line,309}]},{exometer_histogram,probe_update,2,[{file,"src/exometer_histogram.erl"},{line,303}]},{exometer_probe,handle_msg,2,[{file,"src/exometer_probe.erl"},{line,545}]},{exometer_probe,loop,1,[{file,"src/exometer_probe.erl"},{line,514}]},{proc_lib,init_p,3,[{file,"proc_lib.erl"},{line,224}]}]}
    ancestors: [exometer_admin,exometer_core_sup,<0.154.0>]
    messages: []
    links: []
    dictionary: [{{exometer_proc,info},{info,<0.156.0>,[],{sys,undefined,exometer_probe,[riak,riak_kv,consistent,puts,objsize]}}}]
    trap_exit: false
    status: running
    heap_size: 46422
    stack_size: 27
    reductions: 289
  neighbours:
```
#### To reproduce (riak-2.0.5)

setup strong consistency

```
% sed -i.back 's/## strong_consistency = on/strong_consistency = on/g' riak.config
% riak-admin bucket-type create strongly_consistent '{"props":{"consistent":true}}'
% riak-admin bucket-type activate strongly_consistent
```

put object

```
% curl -XPUT localhost:8098/types/strongly_consistent/buckets/demo/keys/key1 -d 'hi' -i
HTTP/1.1 204 No Content
Vary: Accept-Encoding
Server: MochiWeb/1.1 WebMachine/1.10.5 (jokes are better explained)
Date: Thu, 26 Feb 2015 19:59:23 GMT
Content-Type: application/x-www-form-urlencoded
Content-Length: 0
```

attempt to get metrics

```
[...]

PR #1105, opened 2015-04-22
Handling undefined object size on kv stats during consistent put/get.

ksauzz:
To avoid stat's crash, riak_kv_stat should handle undefined object size during consistent put/get. 

riak_test: https://github.com/basho/riak_test/pull/788
original issue: #1086


JeetKunDoug:
Closing as this was done in #1167 (RIAK-2180)

PR #1106, opened 2015-04-24
riak_kv_vnode:hashtree_pid should return {error, wrong_node} appropriately

JeetKunDoug:
riak_kv_vnode:hashtree_pid should be returning {error, wrong_node} when it's no longer the primary (but at some point was). It used to return {ok, undefined} in this case.

Related to deadlock issues in riak_repl:

https://github.com/basho/riak_repl/pull/671
https://github.com/basho/riak_test/pull/789


seancribbs:
:+1: cd00fc7

Please make sure there's a backport to 2.0 after this merges.

jonmeredith:
lgtm too - Sean already approved.

JeetKunDoug:
@borshop merge

PR #1109, opened 2015-04-28
Bugfix for allow_mult=true leaking into the default/legacy bucket type

seancribbs:
See basho/riak#727 (RIAK-1662) (RIAK-1681 (#727 (RIAK-1662)))

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -103,7 +103,8 @@ start(_Type, _StartArgs) ->
        {rw, quorum},
        {basic_quorum, false},
        {notfound_ok, true},
-       {write_once, false}
+       {write_once, false},
+       {allow_mult, false}
    ]),
 
     %% Check the storage backend


seancribbs:
An riak_test is already available thanks to @russelldb: https://github.com/basho/riak_test/pull/782

jonmeredith:
@seancribbs should this be targeted at 2.1 instead of develop?

seancribbs:
@jonmeredith My plan was develop, 2.1 and 2.0.

jonmeredith:
Tested and does not work. Will have to revisit post-2.0.6.

(Dragons beware - the order of the new / old merge in riak_core_bucket_props looks wrong, but is explained in the riak_core_bucket:append_bucket_defaults edoc comment

``````
%% @doc Add a list of defaults to global list of defaults for new
%%      buckets.  If any item is in Items is already set in the
%%      current defaults list, the new setting is omitted, and the old
%%      setting is kept.  Omitting the new setting is intended
%%      behavior, to allow settings from app.config to override any
%%      hard-coded values.
```)
``````

JeetKunDoug:
This sadly didn't work, and @macintux resolved it in a different way in https://github.com/basho/riak_core/pull/765, so closing this.

PR #1110, opened 2015-04-30
Do not crash riak node if vnode status file is unreadable

russelldb:
`file:consult/1` the vnode would start, fail to consult, crash until
the restart limit was hit and the node would fail to start/crash.

This change makes that more robust by simply throwing away a status file
that cannot be read, and using a new, empty vnode status. This has the
cost of a new vnodeid+epoch counter for the vnode, but is more robust.

Some vnode files are un-consultable because they were written with
`io_lib:format("~p.", [Term]).` `io:format` docs say:

```
"Writes the data with standard syntax in the same way as ~w, but breaks
terms whose printed representation is longer than one line into many
lines and indents each line sensibly. It also tries to detect lists of
 printable characters and to output these as strings."
```

Which is why this change also use `io_lib:format("~w.", [Term])` to
write vnode status files. This is backwards compatible so these status
files can be read on r16 and previous riak in the case of downgrade.


russelldb:
Please hang fire on this, I'm going to re-work it without the core dependancy.

russelldb:
Stupidly conflated the issue of vnode status file corruption, and r17 ness. This should just be for the former

russelldb:
OK, Have at it. Nothing to do with erlang-17 now, just handles "corrupt" files for whatever reason. TODO address reading files from < 17 in 17+ nodes (see https://github.com/basho/riak_core/pull/733)

russelldb:
Needs re-target on 2.1 branch

Issue #1093, opened 2015-03-18
The riak_kv schema tests assume riak_kv is the top level [JIRA: RIAK-1625]


russelldb:
https://github.com/basho/riak_kv/blob/develop/test/riak_kv_schema_tests.erl#L298-L304

However, if you want to test riak_kv with deps that riak builds/ships with you usually need to test it from riak/deps/riak_kv with

`./rebar eunit skip_deps=true deps_dir=../`

which will cause these tests to fail.

It would be _ace_ if they would run in a location independent manner. Please.

Basho-JIRA:
PR: https://github.com/basho/riak_kv/pull/1111 

**_[posted via JIRA by Derek Somogyi]**_

PR #1111, opened 2015-04-30
Make schema tests less path-sensitive

macintux:
Addresses #1093 (RIAK-1630)


yorkshire-steve:

Issue #1112, opened 2015-04-30
Failing tests on develop [JIRA: RIAK-1791]


macintux:
Seeing these frequently but not every time:

```
  1) riak_kv_pb_object:empty_bucket_key_test_/0:398: RpbPutReq with empty key is disallowed
     Failure/Error: ?assertMatch({ ok , _ }, Result)
       expected: = { ok , _ }
            got: {error,closed}
     %% src/riak_kv_pb_object.erl:398:in `riak_kv_pb_object:-empty_bucket_key_test_/0-fun-1-/0`


  2) riak_kv_pb_object:empty_bucket_key_test_/0:402: RpbPutReq with empty bucket is disallowed
     Failure/Error: ?assertMatch({ ok , _ }, Result)
       expected: = { ok , _ }
            got: {error,closed}
     %% src/riak_kv_pb_object.erl:402:in `riak_kv_pb_object:-empty_bucket_key_test_/0-fun-3-/0`


  3) riak_kv_pb_object:empty_bucket_key_test_/0:406: RpbGetReq with empty key is disallowed
     Failure/Error: ?assertMatch({ ok , _ }, Result)
       expected: = { ok , _ }
            got: {error,closed}
     %% src/riak_kv_pb_object.erl:406:in `riak_kv_pb_object:-empty_bucket_key_test_/0-fun-5-/0`


  4) riak_kv_pb_object:empty_bucket_key_test_/0:409: RpbGetReq with empty bucket is disallowed
     Failure/Error: ?assertMatch({ ok , _ }, Result)
       expected: = { ok , _ }
            got: {error,closed}
     %% src/riak_kv_pb_object.erl:409:in `riak_kv_pb_object:-empty_bucket_key_test_/0-fun-7-/0`
```

macintux:
This was due to a lack of an `epmd` daemon. Will update the kv test suite to launch it automatically.

PR #1119, opened 2015-05-20
Addresses test failures when epmd is not running

macintux:
Addresses #1112

### DIFF
--- a/src/riak_kv_test_util.erl
+++ b/src/riak_kv_test_util.erl
@@ -175,6 +175,9 @@ setup(TestName, SetupFun) ->
     Deps = dep_apps(TestName, SetupFun),
     do_dep_apps(load, Deps),
 
+    %% Start epmd
+    os:cmd("epmd -daemon"),
+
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
Are you getting a clean run with this, or is it just removing a subset of the errors?

macintux:
Double checking. I know the test in question passes with `epmd` running, and doesn't pass with it not running, and I know `make test` seems to work with this change, but I haven't compared the # of tests run with and without.

macintux:
Yep, something's off. 296 successes either way, 4 failures on develop. Thanks for pointing this out; do you know what's going on?

macintux:
Scratch that: I was misreading eunit's output. Derp. 296 _tests_, not 296 successes. Is there something you specifically were concerned about, @cmeiklejohn?

cmeiklejohn:
:+1: 8eb9c1a

macintux:
@borshop merge

PR #1115, opened 2015-05-12
Do not crash riak node if vnode status file is unreadable

russelldb:
If for any reason the vnode status file is unreadable with
`file:consult/1` the vnode would start, fail to consult, crash until
the restart limit was hit and the node would fail to start/crash.

This change makes that more robust by simply throwing away a status file
that cannot be read, and using a new, empty vnode status. This has the
cost of a new vnodeid+epoch counter for the vnode, but is more robust.

Some vnode files are un-consultable because they were written with
io_lib:format("~p.", [Term]). `io:format' docs say:

```
"Writes the data with standard syntax in the same way as ~w, but breaks
terms whose printed representation is longer than one line into many
lines and indents each line sensibly. It also tries to detect lists of
printable characters and to output these as strings."
```

Which is why this change also use `io_lib:format("~w.", [Term])` to
write vnode status files. This is backwards compatible so these status
files can be read on r16 and previous riaks in the case of downgrade.

----- AH BUT don't conflate erlang 17 handling with corrupt vnode status file!

Issue #1118, opened 2015-05-20
Occasional test failure: riak_kv_stats [JIRA: RIAK-1820]


macintux:
Seeing this on the builders during some runs. Sometimes this failure is much noisier, with hundreds/thousands of lines of meck status dump.

/cc @uwiger

```
  riak_kv_stat:create_or_update_histogram_test/0
    %% Unknown error: timeout
  undefined
    %% Unknown error: {blame,[1,6]}


8 tests, 0 failures, 3 cancelled

=ERROR REPORT==== 20-May-2015::15:42:48 ===
** Generic server riak_core_ring_manager_meck terminating
** Last message in was {'EXIT',<0.262.0>,killed}
** When Server state == {state,riak_core_ring_manager,
                         [{start_link,0},
                          {refresh_my_ring,0},
                          {refresh_ring,2},
                          {set_my_ring,1},
                          {get_ring_id,0},
                          {get_bucket_meta,1},
                          {write_ringfile,0},
                          {set_cluster_name,1},
                          {is_stable_ring,0},
                          {force_update,0},
                          {prune_ringfiles,0},
                          {stop,0},
                          {init,1},
                          {handle_call,3},
                          {handle_cast,2},
                          {handle_info,2},
                          {terminate,2},
                          {code_change,3},
                          {run_fixups,3},
                          {do_write_ringfile,1},
                          {read_ringfile,1},
                          {find_latest_ringfile,0},
                          {get_raw_ring,0},
                          {get_raw_ring_chashbin,0},
                          {get_chash_bin,0},
                          {cleanup_ets,1},
                          {setup_ets,1},
                          {set_ring_global,1},
                          {ring_trans,2},
                          {get_my_ring,0},
                          {start_link,1}],
                         {dict,1,16,16,8,80,48,
                          {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                          {{[],[],[],[],[],[],[],[],[],
                            [[{get_my_ring,0}|
                              {{get_my_ring,0},
                               [{{args_matcher,[],<<>>,false},
                                 {meck_exec,
                                  #Fun<riak_kv_stat.15.83696121>}}]}]],
                            [],[],[],[],[],[]}}},
                         true,[],
                         {false,no_passthrough_cover},
                         false,undefined,[]}
** Reason for termination ==
** killed
```

macintux:
[...]

PR #1121, opened 2015-05-21
Fix test that fails due to inadequate cleanup from earlier tests

macintux:
Perhaps `riak_kv_test_util` would remove all environment variables it sets, but that seems like a heavy-handed change. Instead, override the platform_data_dir variable so we're sure to catch the correct vnode status file with our delete operations.

Addresses test failures highlighted by the build for #1115

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -2742,7 +2742,7 @@ list_buckets_test_() ->
              clean_test_dirs(),
              application:start(sasl),
              Env = application:get_all_env(riak_kv),
-         exometer:start(),
+             exometer:start(),
              riak_kv_stat:register_stats(),
              {ok, _} = riak_core_bg_manager:start(),
              riak_core_metadata_manager:start_link([{data_dir, "kv_vnode_test_meta"}]),
@@ -2752,7 +2752,7 @@ list_buckets_test_() ->
              riak_core_ring_manager:cleanup_ets(test),
              riak_kv_test_util:stop_process(riak_core_metadata_manager),
              riak_kv_test_util:stop_process(riak_core_bg_manager),
-         exometer:stop(),
+             exometer:stop(),
              application:stop(sasl),
              [application:unset_env(riak_kv, K) ||
                  {K, _V} <- application:get_all_env(riak_kv)],
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -2621,12 +2621,16 @@ blocking_setup() ->
     application:set_env(riak_core, platform_data_dir, ?DATA_DIR),
     (catch file:delete(?DATA_DIR ++ "/kv_vnode/0")).
 
+blocking_teardown() ->
+    application:unset_env(riak_core, platform_data_dir),
+    (catch file:delete(?DATA_DIR ++ "/kv_vnode/0")).
+
 %% @private test the vnode and vnode mgr interaction NOTE: sets up and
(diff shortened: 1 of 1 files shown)


macintux:
/cc @russelldb @jonmeredith 

Not sure whether this should be applied against Russell's branch directly.

jonmeredith:
Nope. Should merge first then we ask bors to retry and it should force-push the remerged branch.

jonmeredith:
Buildbot chokes with

```
=INFO REPORT==== 21-May-2015::15:08:49 ===
    application: lager
    exited: stopped
    type: permanent

Pending:
  riak_kv_stat:create_or_update_histogram_test/0
    %% Unknown error: timeout
  undefined
    %% Unknown error: {blame,[29,6]}

```

Very strange.

jonmeredith:
This is coming from eunit...  From the comments in eunit_proc.erl

```
% where Id identifies the item that the message pertains to, and the
%% Info part can be one of:

%%           {blame, Id}        forced to terminate because of item `Id'
```

jonmeredith:
seems like eunit not eating it's own messages properly

```
deps/eunit_formatters/src/eunit_progress.erl
327:    io_lib:format("Unknown error: ~p", [Reason]).
```

macintux:
This is the problem I referenced Ulf in: 
https://github.com/basho/riak_kv/issues/1118

jonmeredith:
@borshop retry

src/riak_kv_vnode.erl
```diff
@@ -2615,13 +2615,18 @@ highest_actor(ActorBase, Obj) ->
...
+    (catch file:delete(?DATA_DIR ++ "/kv_vnode/0")).
 
 %% @private test the vnode and vnode mgr interaction NOTE: sets up and
 %% tearsdown inside the test, the mgr needs the pid of the test
 %% process to send messages. @TODO(rdb) find a better way
 blocking_test_() ->
-    {setup, fun() -> (catch file:delete("undefined/kv_vnode/0")) end,
-     fun(_) -> file:delete("undefined/kv_vnode/0") end,
+    {setup, fun() -> blocking_setup() end,
+     fun(_) -> file:delete(?DATA_DIR ++ "/kv_vnode/0") end,
```
engelsanchez: If you are setting an env var in the setup function, you should unset it as part of the cleanup.

macintux: Good point (and how we got into this mess in the first place).

Unfortunately this patch has already gone into 2.1; mind if I file a separate PR to clean it up rather than have different patches for develop and 2.1?

engelsanchez: I think I do mind. I would prefer to fix it properly here before merging, then port to 2.1

macintux:
@engelsanchez Done. Will file a PR against 2.1 shortly

jonmeredith:
+1 b595ab5735d490a20706f2fe5a8999becacf3c82

macintux:
@borshop merge

src/riak_kv_vnode.erl
```diff
@@ -2733,7 +2742,7 @@ list_buckets_test_() ->
              clean_test_dirs(),
              application:start(sasl),
              Env = application:get_all_env(riak_kv),
-	     exometer:start(),
+             exometer:start(),
```
engelsanchez, line 2745: Whitespace.

Sorry, I've never tunced anyone before!

engelsanchez:
Good man @macintux

PR #1130, opened 2015-05-27
Removing cover has exposed a problem: riak_kv_backend in rebar's

macintux:
erl_first_files is necessary to make the behavior work, but causes
problems with rebar and eunit. The non-test version of riak_kv_backend
is being invoked when other backends run riak_kv_backend:standard_test,
and that function is not exposed without the TEST macro definition.

Refactor the test code out of riak_kv_backend into a new file
test/backend_test_util.erl.

Also eliminated some superfluous whitespace along the way.


macintux:
See also #1127

cmeiklejohn:
:+1: db9e2ab

cmeiklejohn:
@borshop merge

cmeiklejohn:
@borshop merge

PR #1131, opened 2015-05-27
2.0 fix for rebar/eunit problems with backend tests

macintux:
Move test code from riak_kv_backend into its own module to avoid rebar/eunit complications. See #1127


cmeiklejohn:
:+1: 6d7507a

cmeiklejohn:
@borshop merge

PR #1141, opened 2015-06-10
Do not crash riak node if vnode status file is unreadable.

cmeiklejohn:
file:consult/1 the vnode would start, fail to consult, crash until the
restart limit was hit and the node would fail to start/crash.


jonmeredith:
+1 d32be75

jsvisa:
Thanks  @mrallen1, but I want to know why the vnode status file becomes unreadable?

jadeallenx:
It interprets the vnode status file as improper Unicode since that's the default file encoding in 17+

russelldb:
Note, there is another PR needed for erlang 17+ that attempts to consult a pre 17 file as pre 17 encoding. @tburghart worked on it, but it's not needed until we support 17.

PR #1143, opened 2015-06-15
Do not crash riak node if vnode status file is unreadable.

cmeiklejohn:
file:consult/1 the vnode would start, fail to consult, crash until the
restart limit was hit and the node would fail to start/crash.

Squashed and rebased to include test fix.  

Confirmed that in the event of {error, eacces}, write will cause 
error to be returned and will not be swallowed.


jonmeredith:
+1  9972c1b

Nice backport man.

cmeiklejohn:
@borshop merge

Issue #1142, opened 2015-06-11
Vnode won't start with crash error for mixed vnode_status file [JIRA: RIAK-1857]


russelldb:
From Dams in #riak on IRC

```
CRASH REPORT Process <0.15403.0> with 0 neighbours exited with reason: no function clause matching riak_kv_vnode:terminate({{{case_clause,{1,10000,<<"Lý<äUxªÏ">>}},[{riak_kv_vnode_status_mgr,get_counter_lease,3,[{file,...},...]},...]},...}, undefined) line 1155 in gen_fsm:terminate/7 line 600
```

Looks like it is possible for the vnode status file to be written as v1, and have a counter. Have asked for more information in IRC.

https://github.com/basho/riak_kv/blob/2.1/src/riak_kv_vnode_status_mgr.erl#L259

Maybe the scenario was run with vnode_counters, then turn off vnode_counters, then turn them on.

russelldb:
As a workaround until a patch, removing the vnode_status file allows the vnode to start again safely

russelldb:
See basho/riak_kv/pull/1144

JeetKunDoug:
Now that basho/riak_kv#1144 is merged into 2.1, can we close this?

Basho-JIRA:
[~russelldb] Is this the same as PR1144 mentioned above (JIRA Riak-1857)?  Should this be closed and done?  I can see that PR1144 was merged but don't see PR1142 as merged in the comparison list I have from Jon M.   

**_[posted via JIRA by Patricia Brewer]**_

PR #1144, opened 2015-06-18
Fix crash on vnode start if status file is mixed 1 and 2

russelldb:
Reported by booking.com in basho/riak_kv#1142: an upgrade to 2.1, followed by
a downgrade to 2.0, followed by an upgrade to 2.1 caused  them to
have mixed vnode_status files, where `counter` was present, but
the version was `1`. A none-exhaustive case statement in vnode
status manager did not handle the case, and the vnode could not
start, leading to a node that would not start.

A simpler way to get into this mess is simply to turn on epochs,
turn them off, and then turn them on  again.

This fix deals with mixed vnode_status files

---

Was unable to repro, so wrote a simple unit test instead.

### DIFF
--- a/src/riak_kv_vnode_status_mgr.erl
+++ b/src/riak_kv_vnode_status_mgr.erl
@@ -268,12 +268,12 @@ get_counter_lease(LeaseSize0, Status, ?VNODE_STATUS_VERSION) ->
         {?VNODE_STATUS_VERSION, undefined, _ID} ->
             %% Lost counter? Wha? New ID
             new_id_and_counter(Status, LeaseSize);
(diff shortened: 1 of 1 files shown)


jonmeredith:
+1 ec3e0a4

Tests pass, logically correct to use the counter if it's present and you're allowed (as otherwise it wouldn't have been updated).

jonmeredith:
Verified that without the patch, if you tweak the data/kv_status/\* files by hand to change version back from 2 to 1 vnodes go into a nasty crashy loop.

russelldb:
@borshop merge

PR #1145, opened 2015-06-23
Prevent duplication of bucket properties in state of get/put FSMs

JeetKunDoug:
When using non-typed buckets, the bucket properties from the bucket are merged with the default bucket properties. However, they were merged using lists:keymerge rather than lists:ukeymerge, which kept all of the duplicated propertied in the merged list. This PR changes riak_kv_get_fsm and riak_kv_put_fsm to use lists:ukeymerge instead, which takes the value from the custom properties list if it exists, and discards the matching key from the default list.

### DIFF
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -191,7 +191,7 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
     Props = 
         case is_tuple(Bucket) of
             false ->
-                lists:keymerge(1, lists:keysort(1, BucketProps), 
+                lists:ukeymerge(1, lists:keysort(1, BucketProps),
                                lists:keysort(1, DefaultProps));
             true ->
                 BucketProps
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -289,7 +289,7 @@ prepare(timeout, StateData0 = #state{from = From, robj = RObj,
     Props = 
         case is_tuple(Bucket) of
             false -> 
-                lists:keymerge(1, lists:keysort(1, BucketProps), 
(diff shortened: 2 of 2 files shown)


russelldb:
Looks good to me, silly and trivial as it is, a test would have stopped the original issue ever getting out, so I'd really recommend adding a unit test that shows this does what is expected (no dups, favours the custom props over the default props)

src/riak_kv_get_fsm.erl
```diff
@@ -191,7 +191,7 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
     Props = 
         case is_tuple(Bucket) of
             false ->
-                lists:keymerge(1, lists:keysort(1, BucketProps), 
+                lists:ukeymerge(1, lists:keysort(1, BucketProps),
```
russelldb, line 194: Could probably use https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_props.erl#L33 instead?

src/riak_kv_put_fsm.erl
```diff
@@ -289,7 +289,7 @@ prepare(timeout, StateData0 = #state{from = From, robj = RObj,
     Props = 
         case is_tuple(Bucket) of
             false -> 
-                lists:keymerge(1, lists:keysort(1, BucketProps), 
+                lists:ukeymerge(1, lists:keysort(1, BucketProps),
```
russelldb, line 292: Could use https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_props.erl#L33?

russelldb:
I guess it is a pain to use the `riak_core_bucket_props` fun since this PR will touch 2 repos. Is up to you. Is such a simple fun.

yorkshire-steve:

Issue #1138, opened 2015-06-10
Bad match in PUT or transfer path [JIRA: RIAK-1850]


nerophon:
The line:

https://github.com/basho/riak_kv/blob/develop/src/riak_kv_vnode.erl#L1044

... suffers a bad match causing process failure. Encountered by a customer running dual memory/leveldb backend. Triggers during handoff send. Crash causes transfer counts to balloon (on the sender) and eventual node crash.

Example error output:

```
2015-06-10 13:35:54.390 [info] <0.17404.21>@riak_core_handoff_sender:start_fold:153 Starting ownership transfer of riak_kv_vnode from 'riak@172.16.3.6' 17126972312471518572699431633393941636592959488 to 'riak@172.16.3.8' 17126972312471518572699431633393941636592959488
2015-06-10 13:35:56.089 [error] <0.176.0>@riak_core_handoff_manager:handle_info:340 An outbound handoff of partition riak_kv_vnode 17126972312471518572699431633393941636592959488 was terminated because the vnode died
2015-06-10 13:35:56.089 [error] <0.17276.21> gen_fsm <0.17276.21> in state active terminated with reason: no match of right hand value {reply,{riak_kv_w1c_put_reply_v1,ok,primary},{state,17126972312471518572699431633393941636592959488,riak_kv_multi_backend,false,{state,[{<<"memory">>,riak_kv_memory_backend,{state,624165094,623902949,undefined,undefined,0,undefined}},{<<"leveldb">>,riak_kv_eleveldb_backend,{state,<<>>,"/data/leveldb/leveldb/17126972312471518572699431633393941636592959488",[{block_cache_threshold,33554432},{block_restart_interval,16},{block_size_steps,16},{compression,true},{create_if_missing,true},{delete_threshold,...},...],...}}],...},...}} in riak_kv_vnode:handle_handoff_command/3 line 1044
```

fadushin:
Is this a mixed-mode cluster?  I.e., are not all nodes in the cluster on 2.1.1?

sacreman:
$ ansible -m shell -a "riak version" riak -K -s

```
riakn2 | success | rc=0 >>
2.1.1

riakn5 | success | rc=0 >>
2.1.1

riakn3 | success | rc=0 >>
2.1.1

riakn4| success | rc=0 >>
2.1.1

riakn1 | success | rc=0 >>
2.1.1

riakn6 | success | rc=0 >>
2.1.1

riakn9 | success | rc=0 >>
[...]

PR #1153, opened 2015-07-16
Fix bad pattern match during handoff with write_once buckets

fadushin:
This PR addresses a bug with write_once buckets that may arise during handoff.  If data is coming into a write_once bucket while handoff is taking place, the riak_core_vnode will crash with a bad_match.

C.f., https://github.com/basho/riak_kv/issues/1138

The fix is to pattern match cleanly against possible return values from handle_command with a KV_W1C_PUT_REQ as input.

A corresponding fix to the verify_handoff_write_once riak_test is on the bugfix/fd/RIAK-1850 branch in the riak_test repo, under PR https://github.com/basho/riak_test/pull/812

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1047,8 +1047,8 @@ handle_handoff_command(Req=?KV_PUT_REQ{}, Sender, State) ->
 
 handle_handoff_command(?KV_W1C_PUT_REQ{}=Request, Sender, State) ->
     NewState0 = case handle_command(Request, Sender, State) of
-        {noreply, NewState} -> lager:info("FDUSHIN> w1c handle_handoff_command noreply", []), NewState;
-        {reply, Reply, NewState} -> lager:info("FDUSHIN> w1c handle_handoff_command reply: ~p", [Reply]), NewState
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1046,8 +1046,11 @@ handle_handoff_command(Req=?KV_PUT_REQ{}, Sender, State) ->
     end;
 
 handle_handoff_command(?KV_W1C_PUT_REQ{}=Request, Sender, State) ->
-    {noreply, NewState} = handle_command(Request, Sender, State),
-    {forward, NewState};
+    NewState0 = case handle_command(Request, Sender, State) of
+        {noreply, NewState} -> NewState;
+        {reply, _Reply, NewState} -> NewState
```
JeetKunDoug: After talking with Jon, in the {reply, ...} case we should send the reply ourselves with riak_core_vnode:reply/? and then continue to forward the original put request to the new vnode. Otherwise, the current owner will not respond in the non-async case.

JeetKunDoug:
:+1: e56e471

fadushin:
@borshop merge

PR #1155, opened 2015-07-24
Converting stop_fold exception breaks vnode worker

engelsanchez:
This fixes a problem where, occasionally, paginated 2i queries would be missing values.  The fix is very short, but the explanation is very long:

2i queries eventually execute folds, which run in one of the the riak_kv vnode pool workers.  The request to run work in a pool worker has a fold function with an accumulator, and a finish function that runs on the result of the fold function ([riak_kv_worker:handle_work](https://github.com/basho/riak_kv/blob/ec9b41ece5e5fe2ce6d6b30cfe43c2cbc93db917/src/riak_kv_worker.erl#L44)). See how a stop_fold exception is handled by skipping the finish function.  In the case of paginated 2i queries, the fold sends back results to the index FSM using [riak_kv_vnode:result_fun_ack](https://github.com/basho/riak_kv/blob/ec9b41ece5e5fe2ce6d6b30cfe43c2cbc93db917/src/riak_kv_vnode.erl#L1714) (see [2i handler](https://github.com/basho/riak_kv/blob/ec9b41ece5e5fe2ce6d6b30cfe43c2cbc93db917/src/riak_kv_vnode.erl#L821)), which waits for an ack from the index FSM and may throw stop_fold if the FSM signals it does not need any more results.

Results are collected in batches before being sent in a [riak_kv_fold_buffer](https://github.com/basho/riak_kv/blob/ec9b41ece5e5fe2ce6d6b30cfe43c2cbc93db917/src/riak_kv_vnode.erl#L895), which is used as the accumulator in the fold.  Adding to this buffer and reaching the batch limit is what triggers the sending.  A [finish function](https://github.com/basho/riak_kv/blob/ec9b41ece5e5fe2ce6d6b30cfe43c2cbc93db917/src/riak_kv_vnode.erl#L1741-L1749) is provided so that any remaining items in the buffer by the end of the fold are also sent.  That's the key of the problem right there: the folder adds an item, the buffer gets full, the folder sends the batch to the FSM and receives a stop command, then throws stop_fold.  The intention is to stop the fold right there and not send any more items, so the finish function should be skipped. The stoppable_fold wrapper was catching that stop_fold and converting it to a {break, Acc} exception, where the accumulator was the value **before** the last items were sent, still containing most of those items.  The riak_kv_fold_worker never gets the exception, and simply processes that last buffer minus last item by calling the finish function on it, re-sending those items. An example:
- Last batch, after adding last item 9: [7,8,9]
- FSM receives [7,8,9]. Then, if it stays around long enough, it receives an extra [7,8] due to the finish function being called on the stale accumulator and ends up with [7,8,9,7,8].  **This is the bit that is timing sensitive and makes it non-deterministic**.
- The merge sort algorithm expects items in order. The result is undefined otherwise. So occasionally it will output items out of order. For example [7,8,7,9,8].  It also truncates to page size, so the [9,8] is dropped and the 7 dropped due to de-duplication.

/cc @jonmeredith @JeetKunDoug
The code around the fold functions is pretty hairy. Let me know if you need more clarification (you will).

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -704,12 +704,12 @@ fold_keys_fun(FoldKeysFun, {index, FilterBucket,
             fun(Bucket, Key, Acc) ->
                     case re:run(Key, TermRe) of
                         nomatch -> Acc;
-                        _ -> stoppable_fold(FoldKeysFun, Bucket, Key, Acc)
+                        _ -> FoldKeysFun(Bucket, Key, Acc)
                     end
             end;
         false ->
             fun(Bucket, Key, Acc) ->
-                    stoppable_fold(FoldKeysFun, Bucket, Key, Acc)
+                    FoldKeysFun(Bucket, Key, Acc)
             end
     end,
 
@@ -730,7 +730,7 @@ fold_keys_fun(FoldKeysFun, {index, FilterBucket, Q=?KV_INDEX_Q{return_terms=Term
     AccFun = case TermRe =:= undefined of
         true ->
             fun(Bucket, _Term, Val, Acc) ->
-                    stoppable_fold(FoldKeysFun, Bucket, Val, Acc)
+                    FoldKeysFun(Bucket, Val, Acc)
             end;
         false ->
             fun(Bucket, Term, Val, Acc) ->
@@ -738,7 +738,7 @@ fold_keys_fun(FoldKeysFun, {index, FilterBucket, Q=?KV_INDEX_Q{return_terms=Term
                         nomatch ->
                             Acc;
                         _ ->
-                            stoppable_fold(FoldKeysFun, Bucket, Val, Acc)
+                            FoldKeysFun(Bucket, Val, Acc)
                     end
             end
     end,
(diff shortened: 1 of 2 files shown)


JeetKunDoug:
I'm going to try to "TODONE" the todo to remove stoppable_fold - there are only 5 usages, all in riak_kv_eleveldb_backend, so unless I'm missing something it should be pretty painless to remove it completely and be done with it.

engelsanchez:
I incorporated Doug's cleanup and added a fix for the memory backend, then squashed the branch to a single commit.

engelsanchez:
This problem can be observed and verified with this new EQC based 2i test: https://github.com/basho/riak_test/pull/822

This is a pure EQC test, without PULSE. Ideally you could use PULSE in a cluster, but the tech is not there yet. Still, running that test for the eleveldb or memory backend with a single node using 32 vnodes shows the pagination problem immediately.

jonmeredith:
+1 9b82812

Mentally walked through the code.
Ran the EQC riak_test (at commit 2.0.0-234-gfa49385)
Ran the 'repro.erl' script that replicated the original problem and was resolved.

jonmeredith:
Forgot to mention also ran existing 2I r_t's

jonmeredith:
@borshop merge

PR #1157, opened 2015-07-27
Bump eper

kuenishi:


andytill:
eper has been bumped in the riak_ts-develop branch of riak_kv used as develop for Riak Time Series, it will be used in KV once the two products have been mergd cc @javajolt

PR #1158, opened 2015-07-27
Remove stoppable_fold as it's no longer used.

JeetKunDoug:
TODONE the TODO from incorrect-2i-pagination. Any reason we shouldn't do this @engelsanchez ?


engelsanchez:
I was trying to reduce the line count since we will merge the change to several branches. It's good, I'm including this change in the original PR, plus a fix to the memory backend.

PR #1159, opened 2015-07-27
Bugfix/fd/riak 1937

fadushin:
This PR assumes a modification to the yz_kv:index function, to support passing both the bucket and key, which it did not support, before.  Passing a bucket is needed for the write_once put path, where there is no (decoded) Riak Object.  We need to decode the object, when indexing (in order to get at the value), but we want to do so in such a 

Cf. https://github.com/basho/yokozuna/pull/529


zeeshanlakhani:
:+1: 4d34e30fff64d6e192a03dd6e857334bc0f180c8

zeeshanlakhani:
@borshop merge

Issue #1069, opened 2014-12-29
Never call  `hd/1` in `riak_object:most_recent_content/1` with the empty list [JIRA: RIAK-1743]


russelldb:
Reported here http://lists.basho.com/pipermail/riak-users_lists.basho.com/2014-December/016459.html

Some merge/reconcile bug is causing the empty list to be passed to `hd`, probably passed to `most_recent_content` in fact. Investigate the cause. Add a guard earlier to catch the error sooner.

Settings are LWW=true, allow_mult=false.

ghost:
We're actually seeing this same error.. One key appears to be the culprit... It's crashing in initialcall: riak_kv_Get_fsm:init/1 

Which is awesome, in that the request actually then times out... 

Any ideas as to a clean way to patch this in the meantime, either delete the key by hand, or something else.

russelldb:
@boardom any chance I can see this data item? You can sanitize the content/key as you see fit, but if I could see the vclock it would be really valuable for me to diagnose the root cause.

You'd need to get each key straight from it's vnode, rather than via the API. If you're willing to do this let me know, and I'll link a snippet you can run on an attached riak node to get this data for me.

ghost:
Sure man, no problems. Shoot me the snippet

russelldb:
@boardom if you attach to a node with

```
riak attach
```

then run https://gist.github.com/russelldb/0b11fbcea1c6f618d2f5

you well get 3 copies of the object output on the console. If you don't want to paste them here, feel free to email me russelldb@basho.com

DSomogyi:
Comment for Jira.

russelldb:
Current research shows this to be an issue with conflicting bucket properties.

```
{dvv_enabled, true}
{last_write_wins, true}
```

and an indexing backend (`leveldb` or `memory`.)

Work around is just to set `{dvv_enabled, false}` if you are using `lww`. We're working on code that means you can't shoot yourself in the foot with this. We have manual repro, working now on automating it ,then the fix.

Basho-JIRA:
Advisory issued. Need to add bucket validation and fixup to make sure users can't do this by accident in the future. 

**_[posted via JIRA by Douglas Rohrer]**_

Basho-JIRA:
Assigned Fred per the Riak Social / Mumble 

**_[posted via JIRA by Patricia Brewer]**_

Basho-JIRA:
[...]

PR #1162, opened 2015-07-30
Add bucket property validation for last_write_wins xor dvv_enabled

jvoegele:
https://bashoeng.atlassian.net/browse/RIAK-1743 (#1069)

If `last_write_wins` is true then `dvv_enabled` cannot also be true. Add
bucket property validation to ensure that this is the case. The bucket
property validation applies to create and update on any buckets or
bucket types. Since this validation applies to a combination of two
separate properties in both the create and update scenarios, a new
`validate_post_merge` function was added, which allows for validation on
properties after they have been fully resolved and merged.

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -119,7 +119,8 @@ validate_create_bucket_type(BucketProps) ->
         Consistent ->
             {Unvalidated, Valid, Errors} = validate_create_consistent_props(Consistent, BucketProps)
     end,
-    validate(Unvalidated, Valid, Errors).
+    {Good, Bad} = validate(Unvalidated, Valid, Errors),
+    validate_post_merge(Good, Bad).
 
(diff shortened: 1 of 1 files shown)


jonmeredith:
@jvoegele do you know if Fred handed off review to anybody else?

jvoegele:
@jonmeredith If Fred handed off review to anybody else, I am not aware of it.

jvoegele:
@jonmeredith: Russell would be the logical choice to look at this in Fred's absence.

russelldb:
I can look at it on UK Friday morning @jvoegele @jonmeredith

jonmeredith:
+1  33178e0

Going to merge but would still be good if @russelldb could double-check.  Noticed it was possible to do this (but you could on previous versions).

```
riak-admin bucket-type create scmap '{"props":{"consistent":true,"datatype":"map"}}'
```

jonmeredith:
@borshop merge

PR #1104, opened 2015-04-18
fix some of the easier dialyzer issues

hmmr:
Also, dialyzer.ignore-warnings has been regenerated from the current
dialyzer output, in which entries are now sorted.

PR #1164, opened 2015-08-03
Fix/az/assorted dialyzer issues

hmmr:
A couple of months ago I created this janitor branch, off then-`develop`, and pushed some commits (#1104) , which got +1'ed, and which I was remiss to merge in time. Instead, I piled more commits onto it, and then got preoccupied by other stuff. This resulted in it collecting dust since.

Per @macintux suggestion, I have rebased this branch onto the current branch 2.1, reset it to the last commit that received a +1 (from @jonmeredith) and pushed with force. The idea is to merge it this far, and add the rest of commits later for eventual inspection.

### DIFF
--- a/src/riak_kv_pncounter.erl
+++ b/src/riak_kv_pncounter.erl
@@ -50,6 +50,15 @@
 -export_type([pncounter/0, pncounter_op/0]).
 
 -opaque pncounter() :: {riak_kv_gcounter:gcounter(), riak_kv_gcounter:gcounter()}.
+%% Redeclaring pncounter() with -type to please dialyzer, which would
+%% otherwise issue the following warning:
+%%  riak_kv_pncounter.erl:100: Invalid type specification for
+%%    function riak_kv_pncounter:merge/2. The success typing is
+%%    ({[{_,pos_integer()}],[{_,pos_integer()}]},{[{_,pos_integer()}],[{_,pos_integer()}]})
+%%    -> {[{_,pos_integer()}],[{_,pos_integer()}]}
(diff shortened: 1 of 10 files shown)


macintux:
This does show a reduction of dialyzer messages. Next step is to look at each change and make sure I'm comfortable with them.

src/riak_kv_pncounter.erl
```diff
@@ -50,6 +50,15 @@
 -export_type([pncounter/0, pncounter_op/0]).
 
 -opaque pncounter() :: {riak_kv_gcounter:gcounter(), riak_kv_gcounter:gcounter()}.
+%% Redeclaring pncounter() with -type to please dialyzer, which would
+%% otherwise issue the following warning:
+%%  riak_kv_pncounter.erl:100: Invalid type specification for
+%%    function riak_kv_pncounter:merge/2. The success typing is
+%%    ({[{_,pos_integer()}],[{_,pos_integer()}]},{[{_,pos_integer()}],[{_,pos_integer()}]})
+%%    -> {[{_,pos_integer()}],[{_,pos_integer()}]}
```
macintux, line 58: Curious about this: when I run dialyzer (r16) I only see a no local return error for `merge`, not the detailed message here.

hmmr, line 58: I noticed dialyzer tends to issue "summary" warnings, such as "no local return", which are consequential to some other issue with the function in question, and which disappear when that underlying issue is fixed.

Anyway, I think I should just trim the hairiness of this comment (in a follow-up commit).

PR #1165, opened 2015-08-05
Converting stop_fold exception breaks vnode worker - BDP patch [JIRA: RIAK-2184]

engelsanchez:
If the riak_kv_worker sees a stop_fold exception, it will stop and do no
further work. The code was blocking that exception and handling it in
the backend by return the value of the accumulator _before_ the
exception. This is wrong. The kv fold logic has already processed that
accumulator and sent the results back. This causes some extra values to
be appended to the results. In the case of paginated 2i queries that use
list:merge to do a merge sort of the results, the extra values made the
partial lists not be in sorted order, and list:merge in turn returns
results that are in the incorrect order. This causes some results to be
moved to the end and be left out of the final merged result.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -710,12 +710,12 @@ fold_keys_fun(FoldKeysFun, {index, FilterBucket,
             fun(Bucket, Key, Acc) ->
                     case re:run(Key, TermRe) of
                         nomatch -> Acc;
-                        _ -> stoppable_fold(FoldKeysFun, Bucket, Key, Acc)
+                        _ -> FoldKeysFun(Bucket, Key, Acc)
                     end
             end;
         false ->
             fun(Bucket, Key, Acc) ->
-                    stoppable_fold(FoldKeysFun, Bucket, Key, Acc)
+                    FoldKeysFun(Bucket, Key, Acc)
             end
     end,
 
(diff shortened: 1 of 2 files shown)


DSomogyi:
create jira issue

Basho-JIRA:
Created at Heather's request via email (Subject = Data Platform Riak Deltas, sent 10/9/2015 @ 3:23 PST). 

**_[posted via JIRA by Derek Somogyi]**_

Basho-JIRA:
John, can you get this in by Wednesday of this week? 

**_[posted via JIRA by Derek Somogyi]**_

macintux:
It’s unlikely. My early attempts to trace the behavior of this code were unsuccessful because of the instability of the integration efforts, and since then I’ve been alternating working on the 2i code both for BDP and for the BDP/TS merge, and the 0.8 tests.

I can drop 2i in favor of this, but tomorrow will also involve time spent discussing our 3-week target, so I don’t know how much time I’ll have to code this.

I’ll try to have a better feel for the scope of the fix by tomorrow’s daily call.

-John

macintux:
Oops, replied to this thinking it was a different issue. Will see what I can do.

macintux:
Welp, after digging and poking I was satisfied with the PR, but a wee bit more digging and poking revealed this was already merged at some previous date (possibly as @jonmeredith was merging 2.0 bits into 2.1). Closing.

PR #1169, opened 2015-08-12
Interpret new node/join error

macintux:
The default error message is perfectly fine, but this is a bit more explicit.


yorkshire-steve:

PR #1175, opened 2015-08-21
Remove erroneous value from error tuple

jvoegele:
When validation of a bucket property fails, the key of the erroneous
property is added to an errors list, along with an error message. For
the dvv_enabled property, however, we were also adding the value of the
erroneous property to the tuple that is added to the errors list. To be
consistent with other validation failures, do not include the value in
the error tuple.


yorkshire-steve:

PR #1180, opened 2015-09-07
DDL Distribution (WIP, Do not merge)

andytill:
DDL distribution across the cluster.
- [x] Use `riak-admin` to create a bucket type with a `table_def` prop that is a valid riak_ql create table statement. This will compile the DDL into a beam file and store it to disk on each node in the cluster. 
- [ ] Use `riak-admin` to activate the bucket. If the DDL was not compiled successfully (because the op is not complete or failed) then activation fails.


andytill:
Meant to be merged against integration/timeseries.

PR #1182, opened 2015-09-14
Time Series DDL Distribution

andytill:
Changes in riak_kv to distribute the DDL and compile the module for the DDL across the cluster.  
- Support the `table_def` JSON property in the riak-admin create bucket  command which takes a riak_ql `CREATE TABLE` statement.
- The statement is parsed into a DDL and compiled to byte code before the bucket type is created to verify it is valid before being propogated around the cluster. The beam code is thrown away at this point, it is only for validation.
- The `table_def` bucket type property is replaced with a `ddl` property, the value is the `#ddl_v1` record. The bucket properties are then stored into the bucket type metadata as usual.

This PR relies on the riak_core changes in branch [type_ready_checks_ddl](https://github.com/basho/riak_core/tree/type_ready_checks_ddl) that has a gen_event manager for notifying listeners when metadata is stored. This is used in DDL distribution for receiving a callback when bucket type metadata is stored.
- `riak_kv_metadata_store_listener` is added to the gen_event manager for bucket type metadata store events.
- When bucket type metadata is stored `riak_kv_metadata_store_listener` calls `riak_kv_ts_newtype` to start compiling the DDL into a beam module on that node. A new process is spawned for the compilation, mostly for fault isolation because of the number of possible errors that could occur during compilation and writing the beam to disk.
- `riak_kv_compile_tab` is an API for a dets table that stores the compilation state for a bucket type.
- When compilation is complete and successful the beam is written to the `data/ddl_ebin` directory, which is added to the erlang VM beam path so modules can be loaded as requested.

A bucket type that has a `table_def` property must have compiled the DDL module and stored it to disk before it can be activated.
- As well as checking if a bucket type exists across the cluster. To verify that the DDL is compiled to a module the metadata listeners are called, with the `is_type_compiled` message. `riak_kv_metadata_store_listener` handles this message and checks in the dets table if it is in the compiled state and returns true if it is.

The `is_type_compiled` can optionally not be handled. If any of the listeners return true then the type can be activated. It is done in this way to give loose coupling between riak_core and riak_kv, riak_core needs to check that the type is compiled but cannot make direct calls to riak_kv since it is not a required dependency.

In this way riak_core has knowledge of the ddl attribute and the listeners but is otherwise ignorant of Time Series and DDL specifics.

To test, cd to the riak deps and checkout `riak_kv/feature/jrd/ts-ddl-dist` and `riak_core/type_ready_checks_ddl`. Build, start and join the cluster. Create a bucket type with a table def:

```
export TAB=at45; dev/dev1/bin/riak-admin bucket-type create $TAB "{\"props\": {\"table_def\":\"create table $TAB (v varchar not null, temperature_k timestamp, primary key (v))\"}}"
```

The `TAB` variable makes it easier to create new bucket types.  The module should now be compiled on each node in data/ddl_ebin.

Activate works as normal.

```
dev/dev1/bin/riak-admin bucket-type activate at45
```


src/riak_kv_pb_timeseries.erl
```diff
@@ -1,65 +1,65 @@
 -module(riak_kv_pb_timeseries).
 
--include_lib("riak_pb/include/riak_kv_pb.hrl").
--include_lib("riak_ql/include/riak_ql_sql.hrl").
--include_lib("riak_ql/include/riak_ql_ddl.hrl").
--behaviour(riak_api_pb_service).
+% -include_lib("riak_pb/include/riak_kv_pb.hrl").
+% -include_lib("riak_ql/include/riak_ql_sql.hrl").
```
hmmr: riak_ql_sql.hrl has been removed, its contents is now in riak_ql_ddl.hrl (https://github.com/basho/riak_ql/pull/22 merged today).

src/riak_kv_ts_sup.erl
```diff
@@ -0,0 +1,84 @@
...
+%%                     ignore |
+%%                     {error, Reason}
+%% @end
+%%--------------------------------------------------------------------
+init([]) ->
+    riak_kv_compile_tab:new(app_helper:get_env(riak_core, platform_data_dir)),
+
+    RestartStrategy = one_for_one,
+    MaxRestarts = 1000,
+    MaxSecondsBetweenRestarts = 3600,
```
hmmr, line 71: These numbers seem to be too large for MaxR and MaxT. Is there a reason to wait, say, 59 min before there are 1000 restarts of wayward children failing at a rate of one over slightly under a second?

src/riak_kv_ts_newtype.erl
```diff
@@ -0,0 +1,200 @@
...
+-export([handle_call/3]).
+-export([handle_cast/2]).
+-export([handle_info/2]).
+-export([terminate/2]).
+-export([code_change/3]).
+
+-record(state, {
+}).
+
+-define(new_metadata_appeared,new_metadata_appeared).
```
macintux: - Macro should be all caps
- Suggest renaming the atom to make more ops-friendly, perhaps  `bucket_type_changed_mid_compile`

src/riak_kv_pb_timeseries.erl
andytill: Merged the header file changes from feature/az/fetch-qry-results.

PR #1183, opened 2015-09-14
Change the default value of `vclock_data_encoding` to `encode_raw`

hazen:
@engelsanchez discovered this one weird trick which seems to greatly improve Time Series write performance and slightly improves standard Riak write performance.


jonmeredith:
Have we had vlock_data_encoding as optional for at least two majors back before we flip this?

hazen:
It would appear so: https://github.com/basho/riak_kv/blob/1.4/src/riak_object.erl#L770-L771

yorkshire-steve:

PR #1187, opened 2015-09-16
RTS-254 Time Series uses the write once path 

andytill:
Add the `"write_once":true` property for bucket types with the `table_def` property since time series always uses write once.  An error is thrown if `write_once` is specifically set to `false`.

Key rewriting has been removed from the normal path in riak_kv_vnode and put into the riak_kv_w1c_worker.  If the riak_object has the `<<"timeseries-local-key">>` key in its meta data then this is the key that should be used to write the value and use the msgpack encoding.


hmmr:
+1

PR #1191, opened 2015-09-18
hashtree:flush_buffer before finishing build of hashtree

zeeshanlakhani:
simple fix, but make sure to call hashtree:flush_buffer before buildtime, meta for build_finished

^^ @jonmeredith

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -622,12 +622,7 @@ apply_tree(Id, Fun, State=#state{trees=Trees}) ->
 do_build_finished(State=#state{index=Index, built=_Pid, trees=Trees0}) ->
     lager:debug("Finished build: ~p", [Index]),
     Trees = orddict:map(fun(_Id, Tree) ->
-                            try
-                                hashtree:flush_buffer(Tree)
-                            catch _:_ ->
-                                lager:warning("Failed to flush trees during build_finish"),
-                                Tree
-                            end
+                            hashtree:flush_buffer(Tree)
                         end, Trees0),
     {_, Tree0} = hd(Trees),
     BuildTime = get_build_time(Tree0),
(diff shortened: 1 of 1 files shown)


jonmeredith:
-1

Similar issues to basho/yokozuna#557 you need to store the state after flushing the buffer.  It's probably still correct, you'd just double-write those last entries which should still be correct.

zeeshanlakhani:
@jonmeredith updated as per https://github.com/basho/yokozuna/pull/557. Derp on me... for not tracing through the state record.

src/riak_kv_index_hashtree.erl
```diff
@@ -619,14 +619,22 @@ apply_tree(Id, Fun, State=#state{trees=Trees}) ->
...
 -spec do_build_finished(state()) -> state().
-do_build_finished(State=#state{index=Index, built=_Pid}) ->
+do_build_finished(State=#state{index=Index, built=_Pid, trees=Trees0}) ->
     lager:debug("Finished build: ~p", [Index]),
-    {_,Tree0} = hd(State#state.trees),
+    Trees = orddict:map(fun(_Id, Tree) ->
+                            try
+                                hashtree:flush_buffer(Tree)
+                            catch _:_ ->
+                                lager:warning("Failed to flush trees during build_finish"),
```
jonmeredith: In the dangerous territory of heavily reviewed small PRs :)  The only reason I can see for this to happen is if leveldb:put fails which could also happen on `hashtree:insert` or `hashtree:delete` and would throw an exception killing the process.

If there is a failure like that, we do not want to mark the tree as built.  We'd probably like to crash and start over.

The interesting question is usability.  The most likely reasons I can think for for a failure are eperm or enospc which a user could do something about, but if we want to fix that we should wrap the other hashtree update calls with nicer logging too.  If we hit either of those conditions, it's unlikely it'll be here first.  Perhaps handle as a separate issue?

The other alternative is to move the write_meta/tree_build calls inside the try/catch and probably add which index/treeid caused problems to the log message for somebody trying to work out which index had problems from the log.

jonmeredith:
-1 does not rebuild on disk error flushing.

PR #1195, opened 2015-09-29
RTS-348 Fix console/HTTP errors creating bucket types

andytill:
Fixes a badarity error thrown when no bucket type properties are given via riak-admin.

Fixes list_to_existing_atom error thrown on the HTTP interface when a table_def is specified.

riak-admin and the HTTP bucket props interface share the `table_def` to `ddl` conversion code. Error messages should be prettier, no stack traces for anticipated problems i.e. write_once:false and BucketType =/= TableName.

There are commands on the JIRA for testing the HTTP interface.


andytill:
@borshop merge

PR #1199, opened 2015-09-30
Define a type/table does not exist error code

andytill:
When calls to the riak_ql helper module throws an undef then return a specific bucket type does not exist error. It is likely to be a common error and not catching it returns a horrible (and mostly uninteresting) stack trace.

This needs to be supported in the clients.


gordonguthrie:
+1 0f16b61

andytill:
@borshop merge

PR #1200, opened 2015-09-30
Fix 2i requests with limits on number of results

macintux:
@javajolt has tested this; hopefully he can plus it too.


hazen:
:+1: f397c8c

macintux:
@borshop merge

PR #1204, opened 2015-10-05
Error messages for invalid where messages

andytill:
Better error messages for select statement where clauses that do not properly cover the primary key.  For example if the where clause does not have a greater than and less than clauses over the time then these messages are returned:

```
Where clause has no upper bound.
```

or:

```
Where clause has no lower bound.
```

If the where clause is missing a primary key field then this message is returned when the missing field is `user`:

```
Missing parameter user in where clase.
```


hmmr:
I am fine with your changes in riak_kv_qry.erl (particularly, with removal of potentially useful but not currently used `submit/1`). I can see how and where subqueries are created and how they become what used to be 'side effects'. I can confirm `ts_A_select_pass_{1,2}` pass.

andytill:
@borshop merge

PR #1206, opened 2015-10-06
Better messages for parse errors performed when parsing riak_ql

andytill:
Instead of returning a badmatch exception and stack trace, return an error message containing the invalid error message.

yecc does not give any usable information except for the token and token type e.g.

```
{select, <<"select">>}
```

Where the token type is the `select` atom. If the token is unknown then the type is just `chars`.


andytill:
@borshop merge

PR #1212, opened 2015-10-09
Beware LWW on bucket type siblings

macintux:
The claimant should be responsible for handling any conflicts during the creation of bucket types, but the default `riak_core_metadata:get` invocation will impose last write wins on siblings and push the results back to cluster metadata.

For now, LWW is a reasonable strategy for individual nodes to decide what to compile wrt DDLs, but we don't want the nodes to push that result back to CMD, so add the appropriate flag to `get`.

/cc @andytill


hazen:
@macintux If you want to explain this me, I'm happy to review it.

andytill:
+1 e6b4009

macintux:
@borshop merge

PR #1213, opened 2015-10-09
Reset status and query state when the query is complete RTS-365

andytill:
To test, assuming you have the GeoCheckin example TS bucket type created, open a socket run the following query:

```
{ok, P} = riakc_pb_socket:start_link("127.0.0.1", 10017).

riakc_ts:query(P, "SELECT * FROM GeoCheckin6 WHERE time > 0 AND time < 20 AND myfamily = 'till' AND myseries = 'moderfamily'").
```

Run it consecutively a bunch of times, it should not throw an error.


andytill:
Building now in e2638ad

andytill:
@borshop merge

PR #1217, opened 2015-10-13
Fix filters for non-primary key fields (RTS-380)

andytill:
Where clauses in queries that filtered on fields not in the primary key would be ignored. This is because the filter was not tagged correctly as a proplists property and so was not picked up eleveldb.


gordonguthrie:
+1 c9f905d

andytill:
@borshop merge

PR #1218, opened 2015-10-13
Update the code path in the VM outside init the canonical way

macintux:
Instead of sending a message to itself, which can result in race problems, use a 0ms timeout as the result from `riak_kv_ts_newtype:init`.

Checked the proper behavior of the system by running `ts_A_select_pass_1`.


macintux:
Ha. Turns out the workshop instructors fed me a lie. Dropping this and using `gen_server:cast` instead.

PR #1221, opened 2015-10-14
Reset the state record between queries, not just fields (RTS-395)

andytill:
A previous change reset some fields before queries but errors could still occur because some fields were not reset. This change creates a new state record after query results have been returned to ensure there is no shared state between queries


gordonguthrie:
+1 2c4f6d3

andytill:
@borshop merge

PR #1222, opened 2015-10-14
Return the service state with TS PB errors

andytill:
Some error replies were not returning the PB service state with the error reply so the message was not properly formatted.

Returned a friendly reply for missing bucket types in queries.


gordonguthrie:
+1 a23ec2c

gordonguthrie:
@borshop merge

PR #1223, opened 2015-10-14
single-key deletes for TS data

hmmr:
[RTS-201](https://bashoeng.atlassian.net/browse/RTS-201): Single key deletes for TS, via standard delete path. A new `tsdelreq` pb message is allocated, see the riak_pb PR below.

Depends on  https://github.com/basho/riak_pb/pull/136. There is a simple (not comprehensive) check of `riakc_ts:delete/4` (which function appears in https://github.com/basho/riak-erlang-client/pull/240) in https://github.com/basho/riak_test/pull/889.

### DIFF
--- a/src/riak_kv_delete.erl
+++ b/src/riak_kv_delete.erl
@@ -102,11 +102,11 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
             end
     end.
 
-new_tombstone_object(Table, {PK, LK}) ->
+new_tombstone_object(Bucket, {PK, LK}) ->
     %% TS objects have this peculiar way to use the two keys: time
     %% quanta-based partition key, for calculating coverage; and local
     %% key for actual lookups:
-    RO = new_tombstone_object(Table, PK),
+    RO = new_tombstone_object(Bucket, PK),
     MD1 = riak_object:get_update_metadata(RO),
     MD2  = dict:store(?MD_TS_LOCAL_KEY, LK, MD1),
     riak_object:update_metadata(RO, MD2);
--- a/src/riak_kv_delete.erl
+++ b/src/riak_kv_delete.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_delete: two-step object deletion
 %%
-%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2015 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -60,7 +60,8 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,undefined) ->
         {R, PR, PassThruOpts} ->
             RealStartTime = riak_core_util:moment(),
             {ok, C} = riak:local_client(),
-            case C:get(Bucket,Key,[{r,R},{pr,PR},{timeout,Timeout}]++PassThruOpts) of
+            case C:get(Bucket, Key,
+                       [{r,R}, {pr,PR}, {timeout,Timeout}] ++ PassThruOpts) of
                 {ok, OrigObj} ->
                     RemainingTime = Timeout - (riak_core_util:moment() - RealStartTime),
                     delete(ReqId,Bucket,Key,Options,RemainingTime,Client,ClientId,riak_object:vclock(OrigObj));
@@ -80,9 +81,9 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
             ?DTRACE(?C_DELETE_INIT2, [-1], []),
             Client ! {ReqId, {error, Reason}};
         {W, PW, DW, PassThruOptions} ->
-            Obj0 = riak_object:new(Bucket, Key, <<>>, dict:store(?MD_DELETED,
-                                                                 "true", dict:new())),
-            Tombstone = riak_object:set_vclock(Obj0, VClock),
+            Tombstone =
+                riak_object:set_vclock(
+                  new_tombstone_object(Bucket, Key), VClock),
             {ok,C} = riak:local_client(ClientId),
             Reply = C:put(Tombstone, [{w,W},{pw,PW},{dw, DW},{timeout,Timeout}]++PassThruOptions),
             Client ! {ReqId, Reply},
@@ -100,6 +101,20 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
                     nop
             end
     end.
+
+new_tombstone_object(Table, {PK, LK}) ->
+    %% TS objects have this peculiar way to use the two keys: time
+    %% quanta-based partition key, for calculating coverage; and local
+    %% key for actual lookups:
+    RO = new_tombstone_object(Table, PK),
+    MD1 = riak_object:get_update_metadata(RO),
+    MD2  = dict:store(?MD_TS_LOCAL_KEY, LK, MD1),
+    riak_object:update_metadata(RO, MD2);
+new_tombstone_object(Bucket, Key) ->
+    riak_object:new(
+      Bucket, Key, <<>>,
+      dict:store(?MD_DELETED, "true", dict:new())).
+
 
 get_r_options(Bucket, Options) ->
     BucketProps = riak_core_bucket:get_bucket(Bucket),
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -180,7 +180,7 @@ init({test, Args, StateProps}) ->
     {ok, validate, TestStateData, 0}.
 
 %% @private
-prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
+prepare(timeout, StateData=#state{bkey = {Bucket, Key},
                                   options=Options,
                                   trace=Trace}) ->
     ?DTRACE(Trace, ?C_GET_FSM_PREPARE, [], ["prepare"]),
@@ -196,7 +196,9 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
             true ->
                 BucketProps
         end,
-    DocIdx = riak_core_util:chash_key(BKey, BucketProps),
+
+    DocIdx = riak_core_util:chash_key({Bucket, riak_kv_pb_timeseries:pk(Key)}, BucketProps),
+
     Bucket_N = get_option(n_val, BucketProps),
     CrdtOp = get_option(crdt_op, Options),
     N = case get_option(n_val, Options) of
@@ -223,12 +225,14 @@ prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
                         riak_core_apl:get_apl_ann(DocIdx, N, UpNodes)
                 end,
             new_state_timeout(validate, StateData#state{starttime=riak_core_util:moment(),
-                                                n = N,
-                                                bucket_props=Props,
-                                                preflist2 = Preflist2,
-                                                tracked_bucket = StatTracked,
-                                                crdt_op = CrdtOp})
+                                                        bkey = {Bucket, riak_kv_pb_timeseries:lk(Key)},
+                                                        n = N,
+                                                        bucket_props=Props,
+                                                        preflist2 = Preflist2,
+                                                        tracked_bucket = StatTracked,
+                                                        crdt_op = CrdtOp})
     end.
+
 
 %% @private
 validate(timeout, StateData=#state{from = {raw, ReqId, _Pid}, options = Options,
--- a/src/riak_kv_get_fsm.erl
+++ b/src/riak_kv_get_fsm.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_get_fsm: coordination of Riak GET requests
 %%
-%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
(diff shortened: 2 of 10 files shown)


src/riak_kv_pb_timeseries.erl
```diff
@@ -126,6 +129,54 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
...
+                {error, notfound} ->
+                    {reply, tsdelresp, State};
+                {error, Reason} ->
+                    {reply, make_rpberrresp(
+                              ?E_DELETE, io_lib:format("Failed to delete record: ~p", [Reason])),
+                     State}
+            end
+    catch error:{undef, _} ->
+            BucketProps = riak_core_bucket:get_bucket(Bucket),
+            {reply, missing_helper_module(Bucket, BucketProps)}
```
andytill: This should be a three-element tuple, with the service state at the end. I fixed this for other cases in riak_kv_pb_timeseries.erl this morning, so don't worry about those.

src/riak_kv_delete.erl
```diff
@@ -101,6 +100,19 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
             end
     end.
 
+new_maybe_ts_object(Bucket, {PK, LK}) ->
+    %% TS objects have this peculiar way to use the two keys: time
+    %% quanta-based partition key, for calculating coverage; and local
+    %% key for actual lookups:
+    RO_ = riak_object:new({Bucket, Bucket}, PK, <<>>),
```
andytill: I got confused reading this with `RO_` as it is and `_RO` as I mistakenly read it meaning it is going to be an unused binding.  Should be `RO` and `MD1`, `MD2`.

src/riak_kv_pb_timeseries.erl
hmmr: Okay.. Let's let git do the right thing when it comes to conflict resolution.

src/riak_kv_w1c_worker.erl
```diff
@@ -142,6 +142,7 @@ put(RObj, Options) ->
             Error
     end.
 
+
```
hmmr: These are the traces of an earlier effort to add a TS `delete` here, which proved to be a false lead.

hmmr:
Rebased onto current e2e/ts and addressed @andytill's comments.

src/riak_kv_delete.erl
```diff
@@ -101,6 +100,19 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
             end
     end.
 
+new_maybe_ts_object(Bucket, {PK, LK}) ->
```
macintux: Suggest naming this `create_tombstone_object` or similar and combining it with the next Tombstone binding line.

hmmr:
@nickelization Because the last two commits are fixups to the previous commits in this PR, they are to be mentally squashed. In isolation, the fixup appears quite out of context, and it's hard to give it an adequate commit message -- it will appear either too short or too long. (I reworded that commit message, however, per your suggestion,)

But your remark is still very pertinent in riak_ts, our excuse being the insane code churn with several concurrent feature and bugfix branches at any point in time. We are workin' on that.

src/riak_kv_pb_timeseries.erl
```diff
@@ -127,6 +130,53 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
...
+    CompoundKey = riak_pb_ts_codec:decode_cells(CompoundKey_),
+
+    Mod = riak_ql_ddl:make_module_name(Table),
+    try Mod:get_ddl() of
+        DDL ->
+            PKLK = make_ts_keys(CompoundKey, DDL),
+            %% = {PK, LK} = Key to pass to riak_client:delete
+            DeleteF =
+                fun() -> riak_client:delete_vclock(
+                           Table, PKLK, VClock, Options,
```
macintux: Suspect this is where {Table, Table} should be passed per @nickelization's comment about delete

src/riak_kv_delete.erl
```diff
@@ -60,7 +60,10 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,undefined) ->
         {R, PR, PassThruOpts} ->
             RealStartTime = riak_core_util:moment(),
             {ok, C} = riak:local_client(),
-            case C:get(Bucket,Key,[{r,R},{pr,PR},{timeout,Timeout}]++PassThruOpts) of
+            %% a TS bucket is a {Table, Table}
+            DupBucketForTS = if is_binary(Key) -> Bucket; el/=se -> {Bucket, Bucket} end,
```
nickelization: Instead of having a TS-specific clause in the `riak_kv_delete` code, I'm thinking we could do a better job encapsulating the TS code if we converted the table to a bucket in `riak_kv_pb_timeseries`. That way the existing `riak_kv_delete:delete` function doesn't need to be modified with a special case.

src/riak_kv_get_fsm.erl
```diff
@@ -180,23 +180,35 @@ init({test, Args, StateProps}) ->
...
+                lists:keymerge(1, lists:keysort(1, BucketProps),
                                lists:keysort(1, DefaultProps));
             true ->
                 BucketProps
         end,
-    DocIdx = riak_core_util:chash_key(BKey, BucketProps),
+
+    %% TS data decompose Key into {PK, LK} and replace the Key with LK
+    %% for lookup, while PK is used to calculate the coverage plan
+    {PK, LK} =
```
nickelization: This breaks normal, non-TS gets: we can't match `StandardKey` to a 2-tuple.

src/riak_kv_get_fsm.erl
```diff
@@ -180,23 +180,35 @@ init({test, Args, StateProps}) ->
     {ok, validate, TestStateData, 0}.
 
 %% @private
-prepare(timeout, StateData=#state{bkey=BKey={Bucket,_Key},
+prepare(timeout, StateData=#state{bkey = {Bucket, PossiblyTSKey},
```
macintux: Instead of doing TS key handling internal to this function, I would like to see two function heads: 

```
prepare(timeout, StateData=#state{bkey={Bucket, Key}}...) when is_binary(Key) ->
    prepare(timeout, {Bucket, {Key, Key}}); %% pseudocode
prepare(timeout, StateData=#state{bkey={Bucket, {PK, LK}}} ->
```

macintux: Actually, scratch that for a better idea...

Match on Key, and anywhere we need a partition key, we call a `?PARTITION_KEY` macro, also a `?LOCAL_KEY` macro, and the logic of divvying up the tuple goes into those macros.

This way the get/put/delete paths can remain virtually untouched but we're being very explicit whenever we call into another function about which semantic purpose is relevant.

nickelization:
For all modules that aren't specific to TS, I'd like to see names like `LocalKey` and `PartitionKey` instead of `LK` and `PK`. For someone not familiar with TS, it's going to be really confusing seeing the abbreviated variable names and not knowing what they stand for.

src/riak_kv_pb_timeseries.erl
```diff
@@ -47,8 +47,9 @@
 -define(E_PUT,      4).
 -define(E_NOCREATE, 5).
 -define(E_NOT_TS_TYPE, 6).
--define(E_MISSING_TYPE, 7).
+-define(E_MISSING_TYPE,   7).
```
nickelization: Do we want extra spaces here (and for E_DELETE, below)? Not sure if you were trying to align the different macros in this block of code, but it doesn't quite look right here.

src/riak_kv_pb_timeseries.erl
```diff
@@ -127,6 +130,53 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
             {reply, missing_helper_module(Bucket, BucketProps), State}
     end;
 
+%% @ignore DELETE (as if)
```
nickelization: Could you elaborate a little more in this comment? Or, maybe I'm missing something, I don't know if `@ignore` has some special meaning, but I don't understand what this line is trying to say.

src/riak_kv_pb_timeseries.erl
```diff
@@ -127,6 +130,53 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
...
+process(#tsdelreq{table = Table, key = CompoundKey_,
```
nickelization: This is just a stylistic note, but typically in other Riak code we don't use variable names with trailing underscores like `CompoundKey_`. I do agree that it's helpful to mark something that you're going to update, but typical Riak style would be to name it `CompoundKey0` instead.

src/riak_kv_pb_timeseries.erl
```diff
@@ -127,6 +130,53 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
...
+                  vclock = VClock_, timeout = Timeout},
+        State) ->
+    Options =
+        lists:flatmap(
+          fun({_, undefined}) -> [];
+             ({_, default})   -> [];
+             ({K, V})         -> [{K,V}] end,
+          [{timeout, Timeout}]),
+    VClock = if VClock_ =:= undefined -> undefined;
```
nickelization: Another style nitpick, but we don't normally use `if` blocks unless it's a poor fit for a `case` statement. In this situation I think it would blend in better with the rest of the code and be slightly clearer to do something like this:

```
VClock = case VClock0 of
    undefined -> undefined;
    _ -> riak_object:decode_vclock(VClock0)
end,
```

src/riak_kv_pb_timeseries.erl
```diff
@@ -127,6 +130,53 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
             {reply, missing_helper_module(Bucket, BucketProps), State}
     end;
 
+%% @ignore DELETE (as if)
+process(#tsdelreq{table = Table, key = CompoundKey_,
+                  vclock = VClock_, timeout = Timeout},
+        State) ->
+    Options =
```
nickelization, line 145: I think `lists:flatmap` is overkill when you only have a singleton list you're processing :). This could be made much simpler and clearer by just using a case statement.

src/riak_kv_pb_timeseries.erl
```diff
@@ -127,6 +130,53 @@ process(#tsputreq{table = Bucket, columns = _Columns, rows = Rows}, State) ->
...
+             end,
+
+    CompoundKey = riak_pb_ts_codec:decode_cells(CompoundKey_),
+
+    Mod = riak_ql_ddl:make_module_name(Table),
+    try Mod:get_ddl() of
+        DDL ->
+            PKLK = make_ts_keys(CompoundKey, DDL),
+            %% = {PK, LK} = Key to pass to riak_client:delete
+            DeleteF =
```
nickelization: Instead of creating an anonymous function here, and then immediately calling it in the case statement, it would seem simpler to just say something like `Result = riak_client:delete_vclock(...` followed by `case Result of`.

src/riak_kv_get_fsm.erl
hmmr: Good catch. I knew I will not regret inviting you for a review!

src/riak_kv_pb_timeseries.erl
hmmr: `@ignore` will prevent edoc (or what's used for generating documentation for riak) from looking at this comment as pertaining to the function that follows. `DELETE` (in capitals) is to be understood in SQL-ish sense (there are `CREATE TABLE` and `SELECT` at other heads of `process/2`. And, `as if` refers to the fact that these are separate API calls, not string arguments to `query/2`.

Eventually, I removed all those cryptic comments, because the code is self-explanatory as to what messages `process` handles (from the #tsdothing arg).

src/riak_kv_pb_timeseries.erl
hmmr, line 145: It's a remnant of a comprehensive set of options. Replaced with an `if` statement.

hmmr:
Addressed reviewers' comments. Also, rebased again onto current e2e/ts, and squashed all fixups together into a single commit.

macintux:
Looking at the new code.

src/riak_kv_delete.erl
```diff
@@ -101,6 +102,20 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
             end
     end.
 
+new_tombstone_object(Table, {PK, LK}) ->
```
macintux: This binding should be named `Bucket` not `Table` to avoid confusion.

src/riak_kv_delete.erl
```diff
@@ -101,6 +102,20 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
...
+    %% TS objects have this peculiar way to use the two keys: time
+    %% quanta-based partition key, for calculating coverage; and local
+    %% key for actual lookups:
+    RO = new_tombstone_object(Table, PK),
+    MD1 = riak_object:get_update_metadata(RO),
+    MD2  = dict:store(?MD_TS_LOCAL_KEY, LK, MD1),
+    riak_object:update_metadata(RO, MD2);
+new_tombstone_object(Bucket, Key) ->
```
macintux: Ditto here.

nickelization:
Everything looks good, thanks @hmmr! I'll +1 this and the other PRs as soon as John's comment about Bucket/Table naming is addressed.

nickelization:
:+1: dbb1f4d

hmmr:
@borshop retry

hazen:
@borshop merge

PR #1224, opened 2015-10-14
Feature/new encoding

erikleitch:
This is related to eleveldb PR: https://github.com/basho/eleveldb/pull/157

Per group discussion last Friday it was decided to change the riak_object content encoding format for TS, from:

<< EncBin/binary >>

to

<<?ENC_MAGIC/integer, EncBin/binary>>

The change to riak_object.erl are to implement this new encoding format.  Changes are:

1) Add MSGPACK_MAGIC macro (it was decided this should be 2) 

2) Modified from_binary() to remove the now obsolete encoding specification (since it is contained in the encoded contents as the magic number)

However I left a stub with the encoding argument around just in case someone still tries to call the old from_binary/4

3) Removed the obsolete encoding argument from the call chain initiated by from_binary/3

4) changed encode_msgpack to use the new format,ie:

 encode_msgpack(Bin) ->
-    msgpack:pack(Bin, [{format, jsx}]).
-    <<?MSGPACK_MAGIC:8/integer, (msgpack:pack(Bin, [{format, jsx}]))/binary>>.

5) Added a pattern match to decode_maybe_binary to match on the new magic number:

decode_maybe_binary(<<?MSGPACK_MAGIC:8/integer, Bin/binary>>) ->
   decode_msgpack(Bin);

6) And one questionable change that warrants some thought:  I added a default for no valid starting byte (either 0, 1 or ?MSGPACK_MAGIC), the idea being that if anyone tries to read back data written with the previous msgpack encoding format, the code would quietly read it and not try to decode it -- ie treat it like a user-supplied binary of unknown content, same as if the initial byte were a 1:

decode_maybe_binary(<<Bin/binary>>) ->
  Bin.


src/riak_object.erl
```diff
@@ -80,6 +80,7 @@
 -define(MAGIC, 53).      %% Magic number, as opposed to 131 for Erlang term-to-binary magic
                          %% Shanley's(11) + Joe's(42)
 -define(EMPTY_VTAG_BIN, <<"e">>).
+-define(MSGPACK_MAGIC, 2). %% Magic number for msgpack encoding
```
lordnull: Minor quibble: Why are we using 'magic' in a macro? It's improved by saying it's part of msgpack, but wouldn't 'msgpack_format_flag' be more informative? (I do see 'magic' on line 80, but that's a legacy, and so I see it as outside scope of this pr).

src/riak_object.erl
```diff
@@ -936,34 +937,38 @@ binary_version(<<?MAGIC:8/integer, 1:8/integer, _/binary>>) -> v1.
 -spec from_binary(bucket(),key(),binary(),encoding()) ->
     riak_object() | {error, 'bad_object_format'}.
 
-from_binary(B,K,Obj) ->
-    from_binary(B,K,Obj,erlang).
-from_binary(_B,_K,<<131, _Rest/binary>>=ObjTerm, _) ->
+%% Keep deprecated interface around, because I have a feeling someone
+%% will still try to use it
+
+from_binary(B,K,ObjTerm, _Enc) ->
```
lordnull, line 943: You can add an "deprecated" edoc tag to this function to make it appear in generated docs.

lordnull:
Running a few r_t's that are meant to read/write data, as well as some poking manually on the shell seems to indicate no issues. However, I would like to see an automated test added for this. "test/riak_object_eqc.erl" seems a good candidate for this.

erikleitch:
Thanks for the comments!  I changed the macro to MSGPACK_ENCODING_FLAG, and added an automated unit test to riak_object.erl (packObj_test/0), to test encoding/decoding in both msgpack and normal term_to_binary() encoding.   (This was previously tested in my sandbox, but not committed as a unit test)

lordnull:
+1 23535f7

Cover indicates the test covers the code, and is easier than doing it on my own.

PR #1226, opened 2015-10-15
Time Series key prefix to group sorted keys

andytill:
Currently all keys are prefixed with `o`. Time Series keys should be prefixed with a different prefix so that they are stored together on disk when sorted by leveldb. This will make key scans more efficient because they will not have to skip over normal object keys in the case where they may be stored on the same node.

`riak_kv_eleveldb_backend` does not know which keys are Time Series and which are not, This change created a new type for Key, `{colocated, LocalKey::binary()}` which is part of the write once messages to the vnodes but not part of the object.  If the key is tuple tagged with the `colocated` atom then it will use the prefix `c`.

Further considerations:
- AAE is currently disabled for Time Series but will most likely need to differentiate between partition and local keys in a similar way.
- There is no write once path for merges, the riak_kv_vnode will need to check `riak_object:get_ts_locat_key` in the object metadata to construct the `colocated` key.
- Single key deletes will also need to use the `colocated` key.


hmmr:
It appears single-key deletes PR (#1223) has the chances to get merged first. Will you then take care to rebase this PR on the new e2e/ts and massage the `LK` accordingly around here: https://github.com/basho/riak_kv/pull/1223/files#diff-021a303d7b16715d543d2b7a6239d15fR113?

gordonguthrie:
I am taking this PR over @hmmr so will do

erikleitch:
+1 after git://github.com/basho/riak_ql.git is changed to git@github.com:basho/riak_ql.git in rebar.config, so that get deps doesn't fail

macintux:
Two changes: fixed rebar.config, fixed merge conflict due to the bucket/table overlap fix (and which negates my `{Bucket, Bucket}` concern).

erikleitch:
+1

PR #1228, opened 2015-10-15
Ignore unknown encodings for RIAK-2221

JeetKunDoug:
In case of downgrade from a version of Riak that uses new binary types other than term-to-binary or "whatever the user gave us" binaries, make sure at minimum we don't crash when trying to read those keys back. Also, if a user provides an "X-Riak-Val-Encoding" metadata tag, store the binary with that value as its magic byte to provide round-trip support.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -24,7 +24,7 @@
 {deps, [
         {sidejob, ".*", {git, "git://github.com/basho/sidejob.git", {branch, "2.0"}}},
         {erlang_js, ".*", {git, "git://github.com/basho/erlang_js.git", {tag, "1.3.0"}}},
-        {bitcask, ".*", {git, "git://github.com/basho/bitcask.git", {branch, "1.7"}}},
+        {bitcask, ".*", {git, "git://github.com/basho/bitcask.git", {tag, "1.7.2"}}},
         {eper, ".*", {git, "git://github.com/basho/eper.git", {tag, "0.78"}}},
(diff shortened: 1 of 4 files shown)


JeetKunDoug:
Will update to fix Lager issue later (was using deps from riak 2.0 locally - make deps is broken on 2.0 branch right now). For review/comment on implementation.

JeetKunDoug:
@gordonguthrie @jonmeredith this should now build/test cleanly (hopefully BORS agrees). This is the 2.0(.7) version of the "handle binary types other than 0 (erlang binary_to_term) or 1 (anything else) PR series. Once this is merged, I'll forward-merge to 2.1 and develop. Please review/comment.

JeetKunDoug:
So reading over this again this morning, and I'm not happy with what I did for encode_maybe_binary - going to rework it a bit real quick this morning.

JeetKunDoug:
@erikleitch I believe you're correct - I think I changed that early in the process and hadn't realized it's a different 1:8 than the one I actually care about - will take a look and see if I can revert the change at  https://github.com/basho/riak_kv/pull/1228/files#diff-23f6ea5339469f5d292dca9ba27d2914L874. Thanks for the review... will let you know.

JeetKunDoug:
@erikleitch thanks for the review - you're right - that one change was an overzealous removal of "1:8" in pattern matches. Fixed in ed465fd.

erikleitch:
+1

JeetKunDoug:
:+1: ed465fd

hazen:
@borshop merge

PR #1229, opened 2015-10-16
Make DDL distribution more robust

macintux:
- Do not allow DDL to be updated post-activation
- Include a DDL comparison as part of the claimant-driven activation process so that if messages are delayed the claimant doesn't accidentally validate a node who has compiled an earlier version of the DDL but has yet to receive the later version

Requires basho/riak_core#790

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -20,12 +20,11 @@
 %%
 %% -------------------------------------------------------------------
 
-%% TODO use dets
-
 -module(riak_kv_compile_tab).
 
 -export([is_compiling/1]).
 -export([get_state/1]).
+-export([get_ddl/1]).
 -export([new/1]).
 -export([insert/4]).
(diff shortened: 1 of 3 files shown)


macintux:
Manual test that I'd like to turn into a formal one:
1. Define a bucket type via the cli
2. Wait until all nodes have had time to see/compile the beam
3. Retrieve the DDL from cluster metadata
   - `DDL = riak_kv_compile_tab:get_ddl(BucketType)`
4. Run the command sequence below via `riak attach`
5. Tweak the DDL term
6. Run the `rpc:multicall` again with the new DDL

```
Nodes = riak_core_ring:all_members(Ring),
rpc:multicall(Nodes, riak_core_metadata_evt_sup,  is_type_compiled, [BucketType, DDL]).
```

JeetKunDoug:
:+1: 4224b26

macintux:
@borshop merge

PR #1233, opened 2015-10-19
fix for bug

gordonguthrie:
Previously the last key of a sub-query would be included and the first key of the next one would be too and they would be the same.

Now the sub-query goes up to the first record in the next quanta - but that record is then discarded from the range scan


hmmr:
These two commits are begging to be squashed.

macintux:
@borshop merge

macintux:
@javajolt and I have also verified this independently using the new EQC test `ts_B_random_query_pass_eqc`

PR #1235, opened 2015-10-21
bitcask dependency fix

fadushin:
Made bitcask dependency a tagged dependency on the same bit cask version that ships with Riak 2.0.6 (fe8cb6e8962ed8d29814d8f5de49147a439415d8).  This bitcask SHA corresponds with the bit cask 1.7.2 tag (June 25, 2015).

This fix addresses a build failure (make deps) because of some recent changes in the bit cask branch that is (recursively) pulling in an incompatible version of lager from that which is pulled in from riak_core.


JeetKunDoug:
:+1: b5e8382

fadushin:
@borshop merge

PR #1241, opened 2015-10-27
Merge/2.0 to 2.1 for riak_object encoding updates for TD

JeetKunDoug:
This change allows riak_object to not crash if "other" object encodings than 0 (erlang:term_to_binary) or 1 ("Arbitrary binary - used by CRDTs?") are found - in those cases, we add the encoding value to metadata (new behavior), and return the binary as-is like we do for type 1.


fadushin:
+1 17526cc

fadushin:
+1 17526cc

JeetKunDoug:
@borshop merge

PR #1246, opened 2015-10-29
Removed riak_ql dep from riak_kv

cuyler:
This removes the riak_ql dependency from riak_kv, in favor of a modification to load riak_ql via reltool.config.


hazen:
:+1: b7d9d05

hazen:
@borshop merge

PR #1248, opened 2015-10-30
Fix for reversed rows in make_tsqueryresp output.

alexmoore:
The cell/row order is correct going into assemble_records, but the row order is reversed on the way out.


alexmoore:
@borshop merge

PR #1249, opened 2015-10-30
Two fixes for put/2 interactions with its worker process

macintux:
- If a DOWN message arrives after the timeout, the calling process could
  hang forever
- If the timeout cancel succeeds and a response is returned, the monitor
  is` not removed (h/t @jonmeredith)

### DIFF
--- a/src/riak_kv_w1c_worker.erl
+++ b/src/riak_kv_w1c_worker.erl
@@ -116,7 +116,10 @@ put(RObj, Options) ->
             after Timeout ->
                 gen_server:cast(Worker, {cancel, ReqId}),
                 receive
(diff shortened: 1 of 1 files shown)


hmmr:
Wait.. Should it not be against end-to-end/timeseries?

macintux:
I assumed we'd have another reconciliation against 2.1/develop, but if not, yes, we should do it there also

hazen:
Nice catch, @macintux.

Yes, with the 2.0->2.1 merges we just saw, merging soon with 2.1 again makes good sense, too.

macintux:
@borshop merge

PR #1251, opened 2015-10-30
Two fixes to put/2 interactions with its worker process (for develop)

macintux:
- If a DOWN message arrives after the timeout, the calling process could
  hang forever
- If the timeout cancel succeeds and a response is returned, the monitor
  is` not removed (h/t Jon Meredith)


macintux:
@borshop merge

PR #1253, opened 2015-11-03
RTS-523: Update KV Unit Tests to Current DDL

hazen:
- Apply @macintux's bugfix for badarith (https://github.com/basho/riak_kv/compare/end-to-end/timeseries...bugfix/jrd/badarith)
- Fix unit tests to use updated DDL (quantum at end)
- Replace all tabs with spaces

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -39,7 +39,7 @@ compile(#ddl_v1{}, #riak_sql_v1{is_executable = true}) ->
 compile(#ddl_v1{}, #riak_sql_v1{'SELECT' = []}) ->
     {error, 'full table scan not implmented'};
 compile(#ddl_v1{} = DDL, #riak_sql_v1{is_executable = false,
-                      type          = sql} = Q) ->
+                                      type          = sql} = Q) ->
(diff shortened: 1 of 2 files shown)


alexmoore:
Works from the front end, will inspect code next. 
itesttimeseries_java_-_riak-client_-____src_basho_riak-java-client_

hazen:
@alexmoore Which GUI tool are you using?

erikleitch:
+1

The substantive change here is the change to expand_where(), to extract time from the last element of the start and end key lists, instead of the first.  The use of lists:last() is appropriate, since it guards against primary keys that (maybe now?  in future?)  are allowed to have != 3 components.  

The rest are formatting changes, and changes to unit tests to get them to run with the new primary key order (with the exception of the badarith_regression_test, which is designed to fail).   The proof there is in the pudding -- all unit tests appear to pass, so I think we are good to go.

erikleitch:
+1 c6f335d

hazen:
@borshop merge

alexmoore:
@javajolt That's just the Java Client tests in IntelliJ.  

@erikleitch Thanks for the in-depth code review! I owe you a beer.

PR #1254, opened 2015-11-04
Batching put support using w1c workers

macintux:
Per discussions over the last couple of weeks, performance on large batch puts is dramatically better if those batches are parallelized in Riak. @jonmeredith pointed out that the obvious way to do so without introducing more infrastructure is to leverage the write-once workers already present on the coordinating node, hence this PR.

This changes **all** puts handled by the write-once logic, so it will need to be reviewed carefully.

Before this is approved I/we need to come up with a test to verify the behavior on worker failure and timeouts.

### DIFF
--- a/src/riak_kv_w1c_worker.erl
+++ b/src/riak_kv_w1c_worker.erl
@@ -72,14 +72,14 @@ workers() ->
 start_link(Name) ->
     gen_server:start_link({local, Name}, ?MODULE, [], []).
 
-%% @spec put(RObj :: riak_object:riak_object(), riak_object:options()) ->
+%% @spec put(RObj :: riak_object:riak_object(), proplists:proplist()) ->
 %%        ok |
 %%       {error, timeout} |
 %%       {error, term()}
 put(RObj, Options) ->
     synchronize_put(async_put(RObj, Options), Options).
 
--spec async_put(RObj :: riak_object:riak_object(), riak_object:options()) ->
+-spec async_put(RObj :: riak_object:riak_object(), proplists:proplist()) ->
                        {ok, {reference(), pid()}} |
                        {error, term()}.
 async_put(RObj, Options) ->
@@ -131,7 +131,7 @@ async_put(RObj, Options) ->
             Error
     end.
 
--spec async_put_replies(ReqIdTuples :: list({reference(), pid()}), riak_object:options()) ->
+-spec async_put_replies(ReqIdTuples :: list({reference(), pid()}), proplists:proplist()) ->
                                        list(term()).
 async_put_replies(ReqIdTuples, Options) ->
     async_put_reply_loop(ReqIdTuples, [], os:timestamp(),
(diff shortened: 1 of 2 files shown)


macintux:
I performed manual testing by inserting crashes and long sleeps into the write once worker; the flow seems correct (after fixing and then removing the effective timeout logic).

src/riak_kv_w1c_worker.erl
```diff
@@ -354,3 +349,82 @@ erase_request_record(ReqId, #state{entries=Entries} = State) ->
...
+    end;
+synchronize_put(Error, _Options) ->
+    Error.
+
+%% Invoked by async_put_reply/2 to wait for all responses
+async_put_reply_loop([], Responses, _Timeout) ->
+    Responses;
+async_put_reply_loop(ReqIds, Responses, Timeout) ->
+    receive
+        {'DOWN', ReqId, process, _Pid, _Reason} when is_reference(ReqId) ->
```
lordnull: This concerns me a bit. If the calling process is monitoring anything other than the put worker, and that other goes down after the async_put_reply_loop starts (or even before it's able to receive that down message), it gets swallowed here, enduing up with an incorrect 'w1c_server crashed' error. This is repeated in the aysnc_put_cleanup function.

macintux: Notice that it's pattern matching on the reference given back to us by monitor. Does that alleviate your concerns?

lordnull: Except it doesn't seem to here (It does in the async_put function, so no worries there). We have a list of ReqIds in the function head, the receive clause just checks that ReqId is a reference (which all monitor's are), and blindly deletes from the ReqIds list, which will work even if the key doesn't exist. Unless I've completely mis-read the code...

macintux: Oh, bloody ‘ell, you’re right. This was easier when we were expecting one specific ID.

Good catch, thanks.

-John

macintux:
Unclear why dialyzer is griping. Even a local test reveals nothing other than an error exists somewhere.

lordnull:
+1 526880d

Ran this through some of the r_t ts's, and they pass.

Local dialyzer passes here as well; looks like bors is running a second one for whatever reason? Weird, but seems unrelated to the change. The tests that are failing appear to be query parsing and compilation, something this does't touch at all.

macintux:
Unit tests pass locally, dialyzer still failing locally with no useful information even with `--verbose` added.

macintux:
Introduced some sleeps and crashes to the worker processes, things behave as expected.

Pushed a dialyzer fix, but the unit tests are probably doomed to fail until all of the renaming PRs are merged (at which point this branch might need rebasing, but I doubt it).

macintux:
Ah, phooey. @lordnull I can't retry the build until you +1 the latest commit

hazen:
@macintux Hope this helps

hazen:
:+1:  e0af5b1

macintux:
@borshop merge

PR #1255, opened 2015-11-04
Remove sleep and poll querying RTS-415

andytill:
This PR removes the poll for results used for queries. Now the client process pid is given to the worker which messages the results back to the client process when it is complete.

The workers make a gen_server:call to the riak_kv_qry_queue, the result is the query to execute along with the client pid to send the results to.  The call will not return until there is a query ready. When the worker is complete it will ask another and reset its state.


hazen:
Unit tests all pass

src/riak_kv_pb_timeseries.erl
```diff
@@ -380,21 +374,6 @@ make_ts_keys(CompoundKey, DDL = #ddl_v1{local_key = #key_v1{ast = LKParams},
...
--spec fetch_with_patience(query_id(), non_neg_integer()) ->
-                                 {ok, [{Key::binary(), riak_pb_ts_codec:ldbvalue()}]} |
-                                 {error, atom()}.
-fetch_with_patience(QId, 0) ->
-    lager:info("Query results on qid ~p not available after ~b secs", [QId, ?FETCH_RETRIES]),
-    {ok, []};
-fetch_with_patience(QId, N) ->
-    case riak_kv_qry_queue:fetch(QId) of
-        {error, in_progress} ->
-            timer:sleep(1000),
```
hazen, line 422: Love the removal of this `sleep/1`

include/riak_kv_qry_queue.hrl
```diff
@@ -1,39 +0,0 @@
...
--define(RIAK_KV_QRY_QUEUE_HRL, included).
-
--include("riak_kv_index.hrl").
--include_lib("riak_ql/include/riak_ql_ddl.hrl").
-
--record(queue_query, {
-          qry = #riak_sql_v1{}
-         }).
-
--type qry_fsm_name() :: atom().
```
hazen, line 33: Why did you want to kill all of these types?

andytill, line 33: They still exist, just in riak_kv_qry_queue. Types should be exported from a module rather than a header file.

andytill:
@borshop merge

andytill:
+1 f2f1a78

andytill:
@borshop merge

PR #1264, opened 2015-11-08
return_terms, return_body can be undefined instead of false

macintux:
Yet again I made the same mistake: assuming that `return_body` is either `true` or `false`. Due to carelessness in defining the value in protobufs, it can be (and almost always is) `undefined`.

Addresses #1263 (RIAK-2269)

### DIFF
--- a/src/riak_kv_pb_index.erl
+++ b/src/riak_kv_pb_index.erl
@@ -241,13 +241,13 @@ make_continuation(_, _, _)  ->
 %% @doc process_stream/3 callback. Handle streamed responses
 process_stream({ReqId, done}, ReqId, State=#state{req_id=ReqId,
                                                   continuation=Continuation,
-                                                  req=#rpbindexreq{return_body=false}=Req,
+                                                  req=#rpbindexreq{return_body=true}=Req,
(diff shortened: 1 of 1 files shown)


hazen:
:+1: 510ac38

macintux:
@borshop merge

lukebakken:
@macintux is there a way to audit or otherwise catch these `undefined` issues in this and other PB messages?

macintux:
Ideally they would be defined with default values in the .proto files. What I haven’t yet researched is whether changing those after the fact is a backwards compatibility problem for protobufs.

I considered setting the default value for return_body to false since it’s effectively new to Riak, but it seems better to be behaviorally symmetric with return_terms, so I’ll wait to figure out whether I can change return_terms at the same time.

-John

lukebakken:
You should be safe to set those both to `false` as the default. This indicates that `false` should be, by definition, the default for booleans that don't explicitly set the default to `true`:

https://developers.google.com/protocol-buffers/docs/proto#optional

I suppose this means that the erlang protobuf lib we use isn't 100% compliant with the standard.

PR #1271, opened 2015-11-19
Queries must use full ranges and other error messages RTS-578

andytill:
Queries must specify a full range for the quantum: `time > 1000 AND time <= 2000`.

Queries not using the greater/less than [or equals to] will return an error message.

Queries using a valid range but add an additional filter return an error:  `time > 1000 AND time <= 2000 AND time != 1101`. No technical reason, it is just not implemented yet. I have left a commented test which I used to investigate whether this was already supported.

Allow multiple timestamps for the partition and local key, prior to this change the quantum was not extracted from the key correctly.

Check that the query range can actually match something: `time > 100 AND time < 50` returns an error. As does `time > 100 AND time < 100`.

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -29,6 +29,7 @@
 -include_lib("riak_ql/include/riak_ql_ddl.hrl").
 -include("riak_kv_index.hrl").
 -include("riak_kv_ts_error_msgs.hrl").
+-include_lib("eunit/include/eunit.hrl").
 
 -define(MAXSUBQ, 5).
 
@@ -172,6 +173,8 @@ check_if_timeseries(#ddl_v1{table = T, partition_key = PK, local_key = LK},
     catch
         error:{incomplete_where_clause, _} = E ->
             {error, E};
+        error:{where_clause_is_not_a_range, _} = E ->
+            {error, E};
         error:Reason ->
             % if it is not a known error then return the stack trace for
             % debugging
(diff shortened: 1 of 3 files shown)


macintux:
A bit concerned about consistency/clarity. e.g.:

```
6> riakc_ts:query(Pid, "select * from test1 where time > 5 and time < 5").
{error,{1001,
        <<"lower_and_upper_bounds_are_equal_when_no_equals_operator: The upper and lower boundaries are equal b"...>>}}
7> riakc_ts:query(Pid, "select * from test1 where time < 5").
{error,{1001,
        <<"{where_not_supported,[{'<',<<\"time\">>,{integer,5}}]}">>}}
```

I'm generally a fan of the second type of error message for clients to be able to customize/localize, but in this case there's not really enough information to understand why `where_not_supported`.

andytill:
@macintux I have added error messages for cases when the time bounds do not use an AND or are not specified.

I have removed two tests because they modified the query DDL after it was parsed from a string query, the DDL is internal to riak so this is not a useful test.

macintux:
This error message is inaccurate:

```
> riakc_ts:query(Pid, "select * from testone where time > 10").
{error,{1001,
        <<"incomplete_where_clause: Neither upper or lower time bounds were specified in the query.">>}}
```

macintux:
This is also inaccurate (field1 and field2 are timestamp fields but not the range field):

```
> riakc_ts:query(Pid, "select * from testtwo where time > 1 and time < 6 and field1 = 2 or field2 = 4").
{error,{1001,
        <<"time_bounds_must_use_and_op: The time bounds used OR but must use AND.">>}}
```

andytill:
I have fixed all but your last example, which is covered by RTS-634 and scheduled for 1.1. This query will not give a correct error message because of the way that the WHERE tree is being built.

```
riakc_ts:query(Pid, "select * from testtwo where time > 1 and time < 6 and field1 = 2 or field2 = 4").
```

The other queries tested, unit tests exist also.

``` erlang
riakc_ts:query(Pid, "select * from GeoCheckin8 where time > 5 and time < 5").
{error,{1001,
        <<"lower_and_upper_bounds_are_equal_when_no_equals_operator: The upper and lower boundaries are equal but the query uses the greater and less than operators.  Change the bounds time or use the greater/less than or equals to on either side.">>}}

riakc_ts:query(Pid, "select * from GeoCheckin8 where time < 5").
{error,{1001,
        <<"incomplete_where_clause: Where clause has no lower bound.">>}}

riakc_ts:query(Pid, "select * from GeoCheckin8 where time > 10").
{error,{1001,
        <<"incomplete_where_clause: Where clause has no upper bound.">>}}
```

andytill:
@borshop merge

PR #1274, opened 2015-11-20
Change the flat_format string in decoder_parse_error_resp/1

hazen:
If the token is not a string, this error message blows up.


hmmr:
If unsure, use "~p". Lesson One in Erlang for Beginners. As well as in Erlang for Uberprofessionals.

hazen:
@borshop merge

PR #1275, opened 2015-11-21
Catch and return nice errors when decoding bad tscells.

alexmoore:
Companion to basho/riak_pb#166 (RIAK-1234) (RIAK-2079) (RIAK-2237).

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -111,37 +111,44 @@ encode(Message) ->
                      {reply, #tsqueryresp{} | #rpberrorresp{}, #state{}}.
 process(#tsputreq{rows = []}, State) ->
     {reply, #tsputresp{}, State};
-process(#tsputreq{table = Table, columns = _Columns, rows = Rows}, State) ->
+process(#tsputreq{table = Table, columns = _Columns, rows = EncodedRows}, State) ->
     Mod = riak_ql_ddl:make_module_name(Table),
-    Data = riak_pb_ts_codec:decode_rows(Rows),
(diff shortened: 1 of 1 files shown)


src/riak_kv_pb_timeseries.erl
```diff
@@ -111,37 +111,44 @@ encode(Message) ->
...
-                    {reply, Error, State}
-            end;
-        false ->
-            {reply, make_rpberrresp(?E_IRREG, "Invalid data"), State};
-        {_, {undef, _}} ->
-            BucketProps = riak_core_bucket:get_bucket(table_to_bucket(Table)),
-            {reply, missing_helper_module(Table, BucketProps), State}
+
+    case riak_pb_ts_codec:decode_rows(EncodedRows) of
+        {error, {bad_ts_cell, BadCell=#tscell{}}} ->
```
macintux, line 118: If we do change the last `decode_cell` function head to match any value, not just `tscell` records, this line (and similar) should also be changed to drop the match.

src/riak_kv_pb_timeseries.erl
```diff
@@ -111,37 +111,44 @@ encode(Message) ->
...
+            make_decode_error_response("Error decoding row cell", BadCell, State);
+
+        Rows when is_list(Rows) ->
+            %% validate only the first row as we trust the client to send us
+            %% perfectly uniform data wrt types and order
+            case (catch Mod:validate_obj(hd(Rows))) of
```
macintux, line 124: Perhaps pattern match line 121 so you don't have to invoke `hd` here:

```
[Row1|_T]=Rows ->
    case (catch Mod:validate_obj(Row1)) of
```

PR #1276, opened 2015-11-23
unbreak normal list_keys (regression caused by adding TS list_keys)

hmmr:
The commit that triggers the regression is 77cd66e3 (part of #1268).

With it, tests/verify_listkeys passes again.


hazen:
:+1: 605c455

PR #1186, opened 2015-09-15
Apply Engel's Vector Clock Compression Optimization [JIRA: RTS-440]

hazen:
This is the same thing as #1183 except it's explicitly for the TS branch

PR #1277, opened 2015-11-24
Hard-code vclock_encoding_method/0 to return encode_raw

hazen:
This takes the work of https://github.com/basho/riak_kv/pull/1186 and further optimizes it by removing the call to `riak_core_capability:get/2`. The only downside is now the vclock encoding cannot be changed.


DSomogyi:
RTS-641

hmmr:
This function is shared between Riak TS and Riak proper. I am concerned that this optimisation, while making sense in TS, might seep through unnoticed when a grand merge happens between Riak TS and Riak proper. And it surely would become a regression then.

hazen:
@hmmr At the moment we are not planning to merge it since it does not make as huge a performance improvement as first thought

PR #1280, opened 2015-11-30
Minor optimizations for TS put requests

macintux:
- Don't embed the local key in the Riak object when it will be
  immediately extracted and removed
- Directly pass the bucket, keys, and encoding function from PB put
  rather than intuiting them from the Riak object

### DIFF
--- a/src/riak_kv_w1c_worker.erl
+++ b/src/riak_kv_w1c_worker.erl
@@ -21,7 +21,7 @@
 -behaviour(gen_server).
 
 %% API
--export([start_link/1, put/2, async_put/2, async_put_replies/2, workers/0]).
+-export([start_link/1, put/2, async_put/5, async_put_replies/2, workers/0]).
 -export([init/1,
     handle_call/3,
     handle_cast/2,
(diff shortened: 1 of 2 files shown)


src/riak_kv_w1c_worker.erl
```diff
@@ -337,6 +333,22 @@ get_request_record(ReqId, #state{entries=Entries} = _State) ->
...
+%% Utility function for put/2: is this a TS object with its special
+%% requirements or a more traditional KV object?
+%%
+%% When riak_kv_pb_timeseries is driving a put request, it can provide
+%% all of these details directly to async_put/5, but when a tombstone
+%% is being put via riak_kv_delete, these must be extracted from the
+%% object.
+kv_or_ts_details(RObj, {ok, LocalKey}) ->
+    MD  = riak_object:get_metadata(RObj),
+    MD1 = dict:erase(?MD_TS_LOCAL_KEY, MD),
```
hmmr, line 345: There will be no local key to erase here, as you are no longer adding it in the metadata in the first plase (per changes in `riak_kv_timeseries:put_data`). Or am I missing something?

macintux, line 345: This is for scenarios where `riak_kv_w1c_worker:put` is invoked by an alternate path (specifically, putting a tombstone during a delete request).

hmmr, line 345: Ok. I was going to argue that we wouldn't need local key in metadata for deletes either, but then remembered that we in fact will.

I'll give my +1 as long as ts_basic passes on it (which I understand it does).

hazen:
```
Test Results:
ts_basic-riak_kv_bitcask_backend,riak_kv_eleveldb_backend: pass
---------------------------------------------
0 Tests Failed
1 Tests Passed
That's 100.0% for those keeping score
```

hazen:
:+1:  b5a42f9

macintux:
@borshop merge

PR #1282, opened 2015-12-01
Unregister per-vnode stats when cleanly shutting down.

jonmeredith:
This removes the stats on a happy path terminate, however if the vnode crashes the terminate callback will not be executed.

The vnode manager currently monitors vnodes for exits and cleans up per-vnode msgq stats, but we need to extend the vnode manager to clean up non-core stats as well.

RIAK-2312


JeetKunDoug:
+1 9af448c

JeetKunDoug:
@borshop merge

PR #1283, opened 2015-12-01
handle backend and qry worker timeouts

hmmr:
RTS-653
1. The riak_kv_qry_worker gen_server previously did not handle the timeout messages from the backend process. Under load, these messages would crash the query worker, with messages in the logs such as this:
   `{{function_clause,[{riak_kv_qry_worker,handle_info,[{{1, 'riak@10.0.24.42',#Ref<0.0.30.90883>}},{error,timeout}}`
   
   This condition is fixed in https://github.com/basho/riak_kv/commit/85d000bf40, whereby all backend error messages will be passed through to riak_kv_pb_timeseries and eventuate in corresponding `#rpberrorresp`.
2. The timeout messages will now result in a special error code 1013 and message "backend timeout" (the generic code 1001 did not distinguish between timeouts and other error conditions such as malformed requests).
3. If the backend process fails to send us any messages in 10 sec, the error code will be the same (1013) but the message now will be "no response from backend".

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -65,6 +65,7 @@
 -define(E_GET,               1010).
 -define(E_BAD_KEY_LENGTH,    1011).
 -define(E_LISTKEYS,          1012).
+-define(E_TIMEOUT,           1013).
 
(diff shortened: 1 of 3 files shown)


andytill:
+1 96acc25

hmmr:
@borshop merge

PR #1284, opened 2015-12-01
Organize data by destination partition for better performance

macintux:
Determining the target partition for each row of a batch upload and then determining most `async_put` parameters by bucket or partition rather than rerunning the same functions for each row is significantly more efficient.


macintux:
@BrianMMcClain reports 10% boost in batch performance, and all relevant riak_test tests pass, including `verify_write_once` since its put path is impacted.

hazen:
Independent testing is underway

hazen:
As @macintux already said, all of the `riak_test` tests pass.
:+1:

PR #1286, opened 2015-12-01
Mv cpp pb timeseries1

erikleitch:


hazen:
@erikleitch I assume that means +1?

erikleitch:
sorry -- I was merging two experimental branches via github.  Just ignore

PR #1290, opened 2015-12-03
Not equals are treated as equals for key fields in where expression RTS-633

andytill:
This change checks that the equals operator is used for family and series fields in the key, other operators are not allowed, although they can be used for filters.

Also, a change to catch riak_ql lexer errors.


hazen:
@borshop merge

PR #1292, opened 2015-12-06
add a unit test for ranges spanning exact multiples of quanta

hmmr:
RTS-637.

This PR includes a eunit test for the case where a query has SELECT range spanning an exact multiple quanta. A fix for this case is in https://github.com/basho/riak_ql/pull/62.


macintux:
:+1:

macintux:
@borshop merge

PR #1294, opened 2015-12-07
Do not include fallbacks in the coverage plan to prevent missing data RTS-698

andytill:
Do not include fallbacks in the coverage plan to prevent missing data otherwise a race condition can occur where the fallbacks return a result first which might be partial or empty. The riak_kv_qry_worker accepts the first sub-query result so the correct result from the primary could be discarded.

I tested manually using a modified eleveldb backend that does a `timer:sleep` when it folds results, so vnodes with less or zero results will return more quickly.  The diff is here:

``` diff
diff --git a/src/riak_kv_eleveldb_backend.erl b/src/riak_kv_eleveldb_backend.erl
index f1fe366..ab31afb 100644
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -470,9 +470,11 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
     EndK3 = riak_ql_ddl:make_key(Mod, LK, EndK2),
     EndKey = eleveldb_ts:encode_key(EndK3),
     EndKey2 = to_object_key(Bucket, EndKey),
-    FoldFun = fun({K, V}, Acc) ->
-                    [{K, V} | Acc]
-            end,
+    FoldFun =
+        fun({K, V}, Acc) ->
+            timer:sleep(40),
+                   [{K, V} | Acc]
+           end,
     Options = [
                {start_key,    StartKey2},
                {end_key,      EndKey2},
```

With the diff in place the error can be provoked with the following riak test:

``` erlang
-module(ts_C_select_pass_2).

-behavior(riak_test).

-include_lib("eunit/include/eunit.hrl").

-export([confirm/0]).

confirm() ->
    DDL  = ts_util:get_ddl(),
    Data = ts_util:get_valid_select_data_spanning_quanta(),
    Qry  = ts_util:get_valid_qry_spanning_quanta(),
    Expected = {
        ts_util:get_cols(),
        ts_util:exclusive_result_from_data(Data, 2, 9)},
    {[_Node, N2, N3], Conn} = ClusterConn = ts_util:cluster_and_connect(multiple),
    ?assertEqual(
        Expected,
        ts_util:ts_query(ClusterConn, normal, DDL, Data, Qry)),
    % Stop Node 2 after bucket type has been activated
    rt:stop(N2),
    rt:stop(N3),
    ?assertEqual(
        Expected,
        ts_util:single_query(Conn, Qry)),
    pass.
```

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -470,9 +470,11 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
     EndK3 = riak_ql_ddl:make_key(Mod, LK, EndK2),
     EndKey = eleveldb_ts:encode_key(EndK3),
     EndKey2 = to_object_key(Bucket, EndKey),
-    FoldFun = fun({K, V}, Acc) ->
-		     [{K, V} | Acc]
-	     end,
+    FoldFun =
+        fun({K, V}, Acc) ->
+            timer:sleep(40),
+		    [{K, V} | Acc]
+	    end,
     Options = [
                {start_key,    StartKey2},
                {end_key,      EndKey2},
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -472,7 +472,6 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
     EndKey2 = to_object_key(Bucket, EndKey),
     FoldFun =
         fun({K, V}, Acc) ->
-            timer:sleep(40),
 		    [{K, V} | Acc]
 	    end,
     Options = [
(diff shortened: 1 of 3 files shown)


src/riak_kv_eleveldb_backend.erl
```diff
@@ -470,9 +470,11 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
     EndK3 = riak_ql_ddl:make_key(Mod, LK, EndK2),
     EndKey = eleveldb_ts:encode_key(EndK3),
     EndKey2 = to_object_key(Bucket, EndKey),
-    FoldFun = fun({K, V}, Acc) ->
-		     [{K, V} | Acc]
-	     end,
+    FoldFun =
+        fun({K, V}, Acc) ->
+            timer:sleep(40),
```
macintux: Pretty sure you didn't mean to leave this sleep in the production code

macintux:
@borshop merge

PR #1296, opened 2015-12-07
Change time series key encoding to use sext format encoded tuples.

jonmeredith:
The previous encoding from leveldb_ts:encode_key stored the length first as a variable int, then the binary which meant that all 1-byte strings appeared before all 2-byte strings ( "A", "B", "AA", "AB", "BA" ) rather than in lexicographical order ("A", "AA", "AB", "B", "BA") which could cause future problems for ranges of family keys.

Instead of using the custom encoding in eleveldb_ts:encode_keys, the time series keys become plain tuples that are sext-encoded as part of the object.  On encode this is transparent, however on decode the riak_kv_eleveldb:from_object_key has been modified to keep the key portion as a binary regardless of whether it is a standard object or a time series object.  The key tuple can be recovered at any time useing `sext:decode` and means that any Riak code or support scripts still see decoded keys as binaries.

The time series list keys code has been modified to use the normal list keys code path and just decode the keys at the end (I think this would have been possible with the old encoding if we had written a DDL-aware decoder, but there is no type information information included in the format).

A consequence of the list keys change is that deleted objects will be included in key listing until their tombstones are removed from the underlying eleveldb database.

There are still some eunit failures that need to be resolved.  Opening to give a chance for reviewers to comment while they are fixed up.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -895,12 +895,15 @@ fold_objects_fun(FoldObjectsFun, undefined) ->
 %% that represents the starting key for the scope. For example, since
 %% we store objects under {o, Bucket, Key}, the first key for the
 %% bucket "foo" would be `sext:encode({o, <<"foo">>, <<>>}).`
+%%
+%% For starting ranges, use key undefined.  Keys are either binaries
+%% for tuples for time series.  Either of sort *after* a bare atom.
 to_first_key(undefined) ->
     %% Start at the first object in LevelDB...
-    to_object_key({<<>>, <<>>}, <<>>);
+    to_object_key({<<>>, <<>>}, undefined); 
 to_first_key({bucket, Bucket}) ->
     %% Start at the first object for a given bucket...
-    to_object_key(Bucket, <<>>);
+    to_object_key(Bucket, undefined);
 to_first_key({index, incorrect_format, ForUpgrade}) when is_boolean(ForUpgrade) ->
     %% Start at first index entry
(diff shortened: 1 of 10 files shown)


src/riak_kv_eleveldb_backend.erl
```diff
@@ -464,18 +465,18 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
         end,
     StartK2 = [{Field, Val} || {Field, _Type, Val} <- StartK],
     StartK3 = riak_ql_ddl:make_key(Mod, LK, StartK2),
-    StartKey = eleveldb_ts:encode_key(StartK3),
-    StartKey2 = to_object_key(Bucket, StartKey),
+    StartK4 = list_to_tuple([Val || {_Type, Val} <- StartK3]), %% TODO: Avoid adding/removing type info
```
JeetKunDoug: this `list_to_tuple([V || {_T, V} <- Key])` should really be wrapped in a function somewhere that's well named and has some edoc about what the heck it's doing and why... it's pretty cryptic trying to understand it reading through the code.

jonmeredith:
Pushed fixes for things that came up in the walkthrough. Will address any review feedback next.

JeetKunDoug:
:+1:

JeetKunDoug:
:+1:    66f22a9

JeetKunDoug:
@borshop merge

jonmeredith:
@borshop merge

jonmeredith:
@borshop merge

PR #1297, opened 2015-12-08
Adjust the timestamps provided for quanta calculation based on where clause

jonmeredith:
The quanta calculations are based on a default inclusive start range, exclusive
end range.  Now that the quanta calculations are correct in riak_ql commit
12118e the values being passed in need to be adjusted.

Found by ts_B_random_query_pass_eqc with counterexample {1,1,{1,d},1}.

Really needs to have some kind of local tests added that check for subqueries around quanta boundaries.


macintux:
Was made necessary by basho/riak_ql#62 (RIAK-2137)

macintux:
@borshop merge

macintux:
(I'm working on a unit test as a separate PR, no reason to hold up this fix)

PR #1299, opened 2015-12-08
Add a new test for our failure yesterday

macintux:
Make sure a test that terminates at the end of a quanta (inclusive) generates two separate SQL records.

Not particularly sophisticated, but should be sufficient to avoid a basic regression.

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -863,6 +863,18 @@ simple_spanning_boundary_test() ->
            ],
     ?assertEqual(Expected, Got).
 
+%% Values right at quanta edges are tricky. Make sure we're not
+%% missing them: we should be generating two queries instead of just
+%% one.
+boundary_quanta_test() ->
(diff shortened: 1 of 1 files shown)


macintux:
Heh, the merge for #1297 happened just after this started its testing

jonmeredith:
+1 e16c9987f0ca5b8b94576145166a230a119ed060

hmmr:
It's a duplicate of https://github.com/basho/riak_kv/pull/1292 (approved by @macintux, the author of the current ticket), but I would prefer yours because the 'sophistication' in mine is simply a result of copypasting the bigger test (it's right above the diff chunk), and it doesn't need to be. Just checking that there are 2 subqueries and not 3 is enough.

Letting both be is just as fine, though.

macintux:
There are similarities but I didn't feel these were tackling quite the same problem. @hmmr please feel free to +1 this one, I'd like to see both in there.

macintux:
@borshop merge

PR #1300, opened 2015-12-09
Regression handling of the new syntax for TS 1.1

gordonguthrie:


andytill:
Now might be a good time to move these erl files into riak_ql, there will be a lot of PRs that need corresponding ql and kv merges to happen otherwise.

gordonguthrie:
This PR is redundant. Closed.

andytill:
Closed.

PR #1302, opened 2015-12-09
Print helpful error messages for lexer errors.

andytill:
Return readable messages for known lexer exceptions instead of printed stack traces.

Some TS specific error handling has been moved from kv_console to ts_util.  Tested provoking lexer errors in queries, which already return friendly errors.

Relates to RTS-615.

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -524,8 +524,10 @@ bucket_type_create(CreateTypeFn, Type, {struct, Fields}) ->
                 {ok, Props2} ->
                     Props3 = [riak_kv_wm_utils:erlify_bucket_prop(P) || P <- Props2],
                     CreateTypeFn(Props3);
-                {error, ErrorMessage} when is_list(ErrorMessage) ->
-                    bucket_type_print_create_result(Type, {error, ErrorMessage});
+                {error, ErrorMessage} when is_list(ErrorMessage) orelse is_binary(ErrorMessage) ->
+                    bucket_type_print_create_result_error_header(Type),
+                    io:format("~ts~n", [ErrorMessage]),
+                    error;
                 {error, Error} ->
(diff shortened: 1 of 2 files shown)


src/riak_kv_console.erl
```diff
@@ -524,8 +524,10 @@ bucket_type_create(CreateTypeFn, Type, {struct, Fields}) ->
                 {ok, Props2} ->
                     Props3 = [riak_kv_wm_utils:erlify_bucket_prop(P) || P <- Props2],
                     CreateTypeFn(Props3);
-                {error, ErrorMessage} when is_list(ErrorMessage) ->
-                    bucket_type_print_create_result(Type, {error, ErrorMessage});
+                {error, ErrorMessage} when is_list(ErrorMessage) orelse is_binary(ErrorMessage) ->
+                    bucket_type_print_create_result_error_header(Type),
+                    io:format("~ts~n", [ErrorMessage]),
```
hazen, line 529: Should the be `lager` instead of `io:format`?

src/riak_kv_console.erl
```diff
@@ -548,15 +550,15 @@ bucket_type_print_create_result(Type, ok) ->
         false ->
             ok
     end;
-bucket_type_print_create_result(Type, {error, {_LineNo, riak_ql_parser, Reason}}) ->
-    io:format("Error validating table definition for bucket type ~ts:\n~s\n", [Type, Reason]),
-    error;
```
hazen, line 553: Interesting.  Looks like `io:format` is used quite a bit.

andytill, line 553: In this case it really does need to be printed to the console, maybe to console.log wouldn't hurt as well.

hazen:
:+1: cacc988

hazen:
@borshop merge

PR #1303, opened 2015-12-10
Gracefully handle queries that do not have a WHERE clause RTS-746

andytill:
Previously, the pb client process would crash and the local client would be disconnected, now it returns an error message.


macintux:
@borshop merge

PR #1305, opened 2015-12-14
First refactoring of query worker

gordonguthrie:


lehoff:
It is a bit confusing that `handle_call/3` is being used as a regular function on 
https://github.com/basho/riak_kv/blob/610146dcbc71d903da1a1cd4d28dd5f6e347f088/src/riak_kv_qry_worker.erl#L118

Since `handle_call` at 
https://github.com/basho/riak_kv/blob/610146dcbc71d903da1a1cd4d28dd5f6e347f088/src/riak_kv_qry_worker.erl#L90
can never be invoked from the outside it would be better to move that code into a separate private function, say, `execute(Request, ReceiverPid, State)`.

The confusion arises when one expects `handle_call` to be a normal OTP callback, where the `From`
parameter is supposed to be a tuple with a pid and a tag. 

A minor thing would be to make `pop_next_query` use `gen_server:cast/2` - it is preferable if `handle_info` only has to deal with messages that cannot be delivered through the call/cast functions.

What is the benefit of carrying the field `run_sub_qs_fn` around in the process state? 
That function might as well be called directly from the new `execute` function.

The call to `blocking_pop/0` at
https://github.com/basho/riak_kv/blob/610146dcbc71d903da1a1cd4d28dd5f6e347f088/src/riak_kv_qry_worker.erl#L116
could - probably with benefit - be replaced by telling the `riak_kv_qry_queue` process that this worker is ready to serve and then wait for the queue to assign a new work item. And the queue should tell the worker through an API call that it has work for the worker, e.g.,

riak_kv_qry_worker:

``` erlang
new_query(Worker, QId, Qry, DDL, ReceiverPid) ->
    gen_server:cast(Worker, {new_query, {QId, Qry, DDL}, ReceiverPid}).

handle_cast({new_query, {QId, Qry, DDL}=Request}, State) ->
    % do the magic by calling the new execute function
    case execute(Request, ReceiverPid, State) of
        {ok, NewState} -> 
            {noreply, NewState};
        {{error,_}=Error, NewState} ->
            ReceiverPid ! Error,
           {noreply, NewState}.
    end.
```

In the `riak_kv_qry_queue` module there should be a function to report for duty.

``` erlang

request_work() ->
    gen_server:cast(?SERVER, {request_work, self()}).

handle_cast({request_work, WorkerPid}, State) ->
    % send some work to the WorkerPid using riak_kv_qry_worker:new_query
    % or put the WorkerPid into the list of idle pids
```

It "just" requires pushing the `WorkerPid` instead of the `From` argument from the `handle_call` to the `
waiting_workers` state field.

I assume that it is only the `riak_kv_qry_queue` that can assign work to the workers and then this would be sufficient. If that is not the case then the the worker is more of a FSM, which could be modelled as a state field in the state record or as a gen_fsm (a bit heavy). 

Having the worker just hanging in a synchronous call will make it a bit harder to debug what is going on as you cannot introspect the worker while it is hanging there. Would be nice to be able to use `sys:get_status` on the processes.

andytill:
@lehoff when the qry_worker is waiting for a query it has a blank state, it only accumulates state while it is processing a query.

lehoff:
@andytill true, but it still doesn't make the use of `handle_call/3` as a normal function a good thing. Nor does it address the problem of hanging indefinitely in the `blocking_pop/0` function call.

andytill:
I am tackling the direct use of `handle_call` right now.

What is the problem with blocking until a query is ready, given there is no state?  Spawning worker processes per query like we do for put and get FSMs but when this was a bigger change than I wanted to make at the time.

lehoff:
Blocking makes it impossible to ask the worker what state it is in or to get anything out of it. That might be something we want in the future.

I think it is a good idea to spawn the worker processes up front - no need to change that.
But changing the "report-for-duty" pattern to be asynchronous would make the design more flexible and easier to integrate with the resilience architecture that I suggested in the Jira ticket.

andytill:
@borshop merge

PR #1310, opened 2015-12-17
Support TS coverage requests

macintux:
WIP. @hmmr will be carrying this set of PRs forward.

Requires basho/riak_core#798, basho/riak_pb#173 (RIAK-1980), and basho/riak_pipe#103

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -46,6 +46,7 @@
 -export([for_dialyzer_only_ignore/3]).
 -export([ensemble/1]).
 -export([get_cover/3, get_cover/4, get_cover/5, replace_cover/6]).
+-export([vnode_target/1]).
 
 -include_lib("riak_core/include/riak_core_vnode.hrl").
 
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -491,11 +491,12 @@ consistent_delete_vclock(Bucket, Key, VClock, Options, _Timeout, {?MODULE, [Node
     end.
 
 %% @doc Retrieve a coverage plan
--spec get_cover(Module :: module(),
-                Bucket :: binary() | {binary(), binary()},
-                riak_client()) ->
-                       Plan :: list({atom(), term()}) |
-                               {error, Err :: term()}.
+%% -spec get_cover(Module :: module(),
+%%                 Bucket :: binary() | {binary(), binary()},
+%%                 riak_client()) ->
+%%                        Plan :: list({atom(), term()}) |
+%%                                {error, Err :: term()}.
+%% TODO for JohnD: fix the specs
 get_cover(Module, Bucket, Client) ->
     get_cover(Module, Bucket, undefined, Client).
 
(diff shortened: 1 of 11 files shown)


src/riak_kv_pb_timeseries.erl
```diff
@@ -87,8 +88,8 @@ init() ->
 decode(Code, Bin) ->
     Msg = riak_pb_codec:decode(Code, Bin),
     case Msg of
-        #tsqueryreq{query = Q}->
-            case catch decode_query(Q) of
+        #tsqueryreq{query = Q, cover_context = Cover}->
+            case catch decode_query(Q, Cover) of
```
lehoff, line 96: Best to use a try-catch like this:

``` erlang
try decode_query(Q, Cover) of
  {ok, DecodedQuery} -> ...
catch
   T:E -> ...
end
```

Allows for better reporting of the error.

src/riak_kv_pb_timeseries.erl
```diff
@@ -293,6 +296,156 @@ process(SQL = #riak_sql_v1{'FROM' = Table}, State) ->
...
             submit_query(DDL, Mod, SQL, State)
+    end;
+%% No support yet for replacing coverage components, so replace_cover
+%% must be undefined
+process(#tscoveragereq{query = Q,
+                       replace_cover=undefined,
+                       table = Table}, State) ->
+    Mod = riak_ql_ddl:make_module_name(Table),
+    Bucket = table_to_bucket(Table),
+    SQL = compile(Mod, catch decode_query(Q)),
```
lehoff, line 301: Better with try-catch as above.

src/riak_kv_pb_timeseries.erl
```diff
@@ -293,6 +296,156 @@ process(SQL = #riak_sql_v1{'FROM' = Table}, State) ->
...
+%% be nice to collapse them back together, probably with a closure,
+%% but time and effort.
+convert_cover_list({error, Error}, State) ->
+    {error, Error, State};
+convert_cover_list(Results, State) ->
+    %% Pull hostnames & ports
+    %% Wrap each element of this list into a rpbcoverageentry
+    Resp = #tscoverageresp{
+              entries=
+                  lists:map(fun({Cover, Range, SQLtext}) ->
```
lehoff: I would add a private helper function called `coverageentry_from_result/1` instead of the anonymous function.

src/riak_client.erl
```diff
@@ -541,23 +544,23 @@ get_cover_aux(MinPar, ReqId, N, RingSize) ->
                       {error, Reason}),
                     riak_client()) ->
                            list(list({atom(), term()})) | {error, term()}.
-replace_cover(_Bucket, _P, {error, Reason}, _OtherBroken, _Client) ->
+replace_cover(_Mod, _Bucket, _P, {error, Reason}, _OtherBroken, _Client) ->
     {error, Reason};
-replace_cover(Bucket, _P, {ok, Replace}, OtherBroken, _Client) ->
-    pick_cover_replacement(proplists:get_value(subpartition, Replace),
+replace_cover(Mod, Bucket, _P, {ok, Replace}, OtherBroken, _Client) ->
+    pick_cover_replacement(Mod, proplists:get_value(subpartition, Replace),
```
lehoff: Since the 2nd arg to `pick_cover_replacement` is only being used for picking traditional vs subpartition I suggest that the destiction is done here as

``` erlang
case proplists:get_value(subpartion, Replace) of
    undefined -> replace_traditional_cover(...);
    _  -> replace_subpartition_cover
end.
```

In my head it makes it easier to understand the branching.

src/riak_kv_qry_compiler.erl
```diff
@@ -408,6 +415,56 @@ rewrite2([#param_v1{name = [FieldName]} | T], Where1, Mod, Acc) ->
...
+    %%   Drop end_inclusive, start_inclusive properties
+    %%   Add new end_inclusive, start_inclusive properties based on the parameters
+    %%   Retain any other properties (currently only `filter')
+
+    NewStartKeyVal = modify_where_key(proplists:get_value(startkey, Props),
+                                     Field, StartVal),
+    NewEndKeyVal = modify_where_key(proplists:get_value(endkey, Props),
+                                   Field, EndVal),
+
+    SlimProps = proplists:delete(startkey,
```
lehoff: Do this with a fold instead - something like

``` erlang
lists:foldl( fun (Prop, Acc)) -> proplists:delete(Prop, Props) end, 
                Props
                [startkey, endkey, end_inclusive, start_inclusive])
```

lehoff:
+1

hmmr:
@borshop merge

PR #1311, opened 2015-12-18
Window aggregate functions

andytill:
**Do not merge**
Support window aggregate functions in riak_kv, such as the following:

``` sql
SELECT count(*) FROM mytable
WHERE time > 1234560 AND time < 1234570 AND myfamily = 'family1' AND myseries = 'series1'
```

Run in the erlang shell using the erlang riak client looks like the following:

``` erlang
120> riakc_ts:query(Pid, "select count(myseries) from GeoCheckin where time > 1234560 and time < 1234570 and myfamily = 'family1' and myseries = 'series1'").   
{[<<"aggregate">>],[{3}]}
```

One current limitation is that all return column names are `<<"aggregate">>` and all column return types are `sint64`.  Column names will eventually be the selection e.g. `count(*)` in this case and the correct return type will be returned for that function by checking the type signature in riak_ql.

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -142,16 +142,32 @@ compile_select_col(DDL, Select) ->
     % fun and ignore the state arg
     {rows, undefined, fun(Row, _) -> Fn(Row) end}.
 
-%%
+%% Returns a one arity fun which is stateless for example pulling a field from a
+%% row.
 compile_select_col_stateless(_, {identifier, [<<"*">>]}) ->
-    fun(Row) ->
-        Row
-    end;
+    fun(Row) -> Row end;
+compile_select_col_stateless(_, {Type, V})
+        when Type == integer; Type == varchar; Type == double; Type == boolean ->
+    fun(_) -> V end;
 compile_select_col_stateless(#ddl_v1{ fields = Fields }, {identifier, ColumnName}) ->
     Index = column_index_of(Fields, to_column_name_binary(ColumnName)),
     fun(Row) ->
         lists:nth(Index, Row)
-    end.
+    end;
+compile_select_col_stateless(DDL, {Op, A, B}) ->
+    Arg_a = compile_select_col_stateless(DDL, A),
+    Arg_b = compile_select_col_stateless(DDL, B),
+    compile_select_col_stateless2(Op, Arg_a, Arg_b).
+
+%%
+compile_select_col_stateless2('+', A, B) ->
+    fun(Row) -> A(Row) + B(Row) end;
+compile_select_col_stateless2('*', A, B) ->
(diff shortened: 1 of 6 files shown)


src/riak_kv_qry_compiler.erl
```diff
@@ -630,11 +698,11 @@ simplest_test() ->
     PK = get_standard_pk(),
     LK = get_standard_lk(),
     ?assertMatch(
-        [#riak_sql_v1{ is_executable = true,
-                       type          = timeseries,
-                       'WHERE'       = Where2,
-                       partition_key = PK,
-                       local_key     = LK }],
+        {ok, _, [#riak_sql_v1{ is_executable = true,
```
hmmr, line 717: The `InitialState` here is `[]`, so I suppose you should be matching against that, rather than `_`.

src/riak_kv_qry_worker.erl
```diff
@@ -239,13 +154,83 @@ extract_riak_object(SelectSpec, V) when is_binary(V) ->
...
+            State
+    end.
+
+%%
+prepare_final_results(rows, IndexedChunks) ->
+    %% sort by index, to reassemble according to coverage plan
+    {_, [R2]} = lists:unzip(lists:sort(IndexedChunks)),
+    lists:append(R2);
+prepare_final_results(aggregate, Aggregate) ->
+    [{<<"aggregate">>, A} || A <- Aggregate].
```
hmmr, line 232: And.. where do you call `riak_ql_window_agg_fn:finalise`?

andytill, line 232: Not got that far yet :)

src/riak_kv_qry_compiler.erl
```diff
@@ -56,45 +56,129 @@ comp2(#ddl_v1{} = DDL, #riak_sql_v1{is_executable = false,
...
+    Index = column_index_of(Fields, to_column_name_binary(ColumnName)),
     fun(Row) ->
         lists:nth(Index, Row)
-    end.
+    end;
+compile_select_col_stateless(DDL, {Op, A, B}) ->
+    Arg_a = compile_select_col_stateless(DDL, A),
+    Arg_b = compile_select_col_stateless(DDL, B),
+    compile_select_col_stateless2(Op, Arg_a, Arg_b).
+
```
bkerley, line 161: Can we also get a function body here to do `negate` c.f. https://bashoeng.atlassian.net/browse/RTS-753 ?

``` erlang
compile_select_col_stateless(DDL, {negate, A}) ->
  ToBeNegated = compile_select_col_stateless(DDL, A),
  fun(Row) -> -ToBeNegated(Row) end.
```

andytill:
Changes in this PR are part of https://github.com/basho/riak_kv/pull/1315, closing this PR.

PR #1313, opened 2015-12-23
Validate all rows, list which failed validation in error message.

alexmoore:
Fix for RTS-784.

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -381,7 +381,7 @@ validate_rows(Mod, Rows) ->
         end
     end,
 
-    {_, BadRowIdxs} = lists:foldl(ValidateFn, {0,[]}, Rows),
+    {_, BadRowIdxs} = lists:foldl(ValidateFn, {1, []}, Rows),
     BadRowIdxs.
 
 -spec make_rpberrresp(integer(), string()) -> #rpberrorresp{}.
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -381,7 +381,7 @@ validate_rows(Mod, Rows) ->
         end
     end,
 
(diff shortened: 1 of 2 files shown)


hazen:
Looks like we have some unit test failures:

```
Failed: 17.  Skipped: 0.  Passed: 315.
```

hazen:
Working on a set of more bad inputs: https://github.com/basho/riak_test/blob/feature/bch/more-bad-data-tests/tests/ts_A_put_invalid_data.erl

src/riak_kv_pb_timeseries.erl
```diff
@@ -140,14 +138,15 @@ process(#tsputreq{table = Table, columns = _Columns, rows = Rows}, State) ->
                     Error = make_rpberrresp(?E_IRREG, to_string({Class, Exception})),
                     {reply, Error, State}
             end;
-        false ->
-            {reply, make_rpberrresp(?E_IRREG, "Invalid data"), State};
+        BadRowIdxs when is_list(BadRowIdxs) ->
+            ValidationBadRowsString = string:join(lists:reverse(BadRowIdxs),", "),
+            EValidationMessage = flat_format("Invalid data found at row index(es) ~s", [ValidationBadRowsString]),
```
andytill: Should we specify it is base zero in the error message? I was surprised when I tested it but I am used to base one because of erlang.  Changing the base number doesn't fix it because some might be expecting base zero.

andytill:
I have made the batch index numbers base one and updated the warnings on row validations in the docs. I don't think base one or zero is definitively correct but at least it is documented. https://github.com/basho/private_basho_docs/tree/at/row_validation.

The riak test PR is here: https://github.com/basho/riak_test/pull/967

gordonguthrie:
+1 3f7555a

andytill:
@borshop merge

PR #1315, opened 2015-12-30
Major refactor of Query Worker for timeseries 1.1

gordonguthrie:
Support window aggregate functions in riak_kv, such as the following:

``` sql
SELECT count(*) FROM mytable
WHERE time > 1234560 AND time < 1234570 AND myfamily = 'family1' AND myseries = 'series1'
```

Run in the erlang shell using the erlang riak client looks like the following:

``` erlang
120> riakc_ts:query(Pid, "select count(myseries) from GeoCheckin where time > 1234560 and time < 1234570 and myfamily = 'family1' and myseries = 'series1'").   
{[<<"aggregate">>],[{3}]}
```

One current limitation is that all return column names are `<<"aggregate">>` and all column return types are `sint64`.  Column names will eventually be the selection e.g. `count(*)` in this case and the correct return type will be returned for that function by checking the type signature in riak_ql.

### DIFF
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -3,6 +3,8 @@ riak_core_pb.erl
 Callback info about the riak_core_vnode_worker behaviour is not available
 Callback info about the riak_core_coverage_fsm behaviour is not available
 Callback info about the riak_kv_backend behaviour is not available
+The created fun has no local return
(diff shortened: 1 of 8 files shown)


andytill:
@borshop: retry

gordonguthrie:
+1 1fd1886

andytill:
@borshop merge

PR #1321, opened 2016-01-06
Do not flatten NULL values in a row [JIRA: RTS-825]

andytill:
Replace the `lists:flatten/1` that flattens out NULL values which are represented as an empty list, empty lists are now retained.

It requires https://github.com/basho/riak_ql/pull/84 to be merged before it will build.

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -86,15 +86,21 @@ run_select(Select, Row, InitialState) ->
 
 %% @priv
 run_select2([], _, _, Acc) ->
-    lists:flatten(lists:reverse(Acc));
+    lists:reverse(Acc);
 run_select2([Fn | SelectTail], Row, [ColState1 | ColStateTail], Acc1) ->
-    ColState2 = Fn(Row, ColState1),
-    Acc2 = [ColState2 | Acc1],
+    Acc2 = prepend_select_columns(Fn(Row, ColState1), Acc1),
     run_select2(SelectTail, Row, ColStateTail, Acc2);
 run_select2([Fn | SelectTail], Row, RowState, Acc1) ->
-    Value = Fn(Row, RowState),
-    Acc2 = [Value | Acc1],
+    Acc2 = prepend_select_columns(Fn(Row, RowState), Acc1),
     run_select2(SelectTail, Row, RowState, Acc2).
+
+%% Check if the select column is actually multiple columns, as returned by
+%% SELECT * FROM, or the corner case SELECT *, my_col FROM. This cannot simply
+%% be flattened because nulls are represented as empty lists.
+prepend_select_columns([_|_] = MultiCols, Acc) ->
+    lists:reverse(MultiCols) ++ Acc;
+prepend_select_columns(V, Acc) ->
+    [V | Acc].
 
 %%
 compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause = Sel } } = Q) ->
@@ -1597,7 +1603,7 @@ function_1_initial_state_test() ->
                   "AND myseries = 'seriesX' AND time > 1 AND time < 2"),
     {ok, Select} = compile_select_clause(get_sel_ddl(), Rec),
     ?assertMatch(
-       #riak_sel_clause_v1{ initial_state = [0] },
+       #riak_sel_clause_v1{ initial_state = [[]] },
        Select
       ).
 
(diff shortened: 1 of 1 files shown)


DSomogyi:
create jira issue

gordonguthrie:
+1 db0adfe

gordonguthrie:
@borshop merge

gordonguthrie:
@borshop merge

src/riak_kv_qry_compiler.erl
```diff
@@ -97,16 +97,22 @@ run_select(Select, Row, InitialState) ->
 
 %% @priv
 run_select2([], _, _, Acc) ->
```
lehoff, line 99: This can be done with body recursion, which will be faster, especially for the ColState traversal, as it avoids the double reverse on the lists in the results.

``` erlang
-spec run_select(SelectionSpec::[compiled_select()], Row::[any()]) ->
                        [any()].
run_select(Select, Row) ->
    %% the second argument is the state, if we're return row query results then
    %% there is no long running state
    run_select2(Select, Row, undefined).

run_select(Select, Row, InitialState) ->
    %% the second argument is the state, if we're return row query results then
    %% there is no long running state
    run_select2(Select, Row, InitialState).


run_select2([], _Row, _RowState) ->
    [];
run_select2([Fn | SelectTail], Row, RowState) ->
    prepend_select_columns(Fn(Row, RowState), SelectTail, Row, RowState);
run_select2([Fn | SelectTail], Row, [ColState1 | ColStateTail] ) ->
    prepend_select_columns(Fn(Row, ColState1), SelectTail, Row, ColStateTail).

%% Check if the select column is actually multiple columns, as returned by
%% SELECT * FROM, or the corner case SELECT *, my_col FROM. This cannot simply
%% be flattened because nulls are represented as empty lists.
prepend_select_columns([], Select, Row, RowState) ->
    [ [] | run_select2(Select, Row, RowState) ];
prepend_select_columns(Cols, Select, Row, RowState) ->
    Cols ++ run_select2(Select, Row, RowState).
```

hazen, line 99: @lehoff So the `++` operation is more efficient than `lists:reverse`?

hazen:
@borshop merge

src/riak_kv_qry_compiler.erl
lehoff, line 99: @javajolt no, it has the same complexity, but running `lists:reverse` once while accumulating and at the end is slower.

If we used something else than [] for nulls then we could just do a

``` erlang
[ Fn(Row, RowState] | run_select2(SelectTail, Row, RowState) ]
```

in `run_select2`, but then we'd have to flatten the list at the end. I think that is as expensive as `lists:reverse` and `++`.

andytill, line 99: @lehoff how much slower than your version?  If we have numbers that show a different implementation improves throughput we need to put it in ASAP.

lehoff, line 99: Sorry, I missed that note in time.

I didn't have measurements at the time, but we are talking about a full list reverse times two vs an append of the whole list.
I recon that append and reverse have the same running time, so my version could be up to 50% faster.

Today I did an experiment using the code below and depending on the length of the list of lists and the lengths of the individual lists my version with body recursion is between 30-60% faster than the tail recursive version.

``` erlang
-module(append_reverse).

-compile(export_all).

list_of_lists() ->
    N =  3 + rand:uniform(6),
    [ random_list() || _ <- lists:seq(1,N) ].

random_list() ->
    Len = rand:uniform(5) - 1,
    lists:seq(1,Len).

my_append([]) ->
    [];
my_append([X|Xs]) ->
    prepend(X, Xs).

prepend([], Xs) ->
    [ [] | my_append(Xs) ];
prepend(Ys, Xs) ->
    Ys ++ my_append(Xs).

my_fold(Xs) ->
    my_fold2(Xs, []).

my_fold2([], Acc) ->
    lists:reverse(Acc);
my_fold2([X|Xs], Acc) ->
    Acc2 = prepend2(X, Acc),
    my_fold2(Xs, Acc2).

prepend2([_|_]=Xs, Acc) ->
    lists:reverse(Xs) ++ Acc;
prepend2(_, Acc) ->
    [ [] | Acc ].


test1() ->
    Data = list_of_lists(),
    {Time1, Res1} = timer:tc( fun my_append/1, [Data]),
    {Time2, Res2} = timer:tc( fun my_fold/1, [Data]),
    Res1 = Res2,
    {Time1, Time2}.

test() ->
    Samples = [ test1() || _ <- lists:seq(1,10000) ],
    {AppendTimes, FoldTimes} = lists:unzip(Samples),
    {stats(AppendTimes), stats(FoldTimes)}.

stats(Samples) ->
    Min = lists:min(Samples),
    Max = lists:max(Samples),
    Avg = lists:sum(Samples) / length(Samples),
    {Min, Avg, Max}.
```

PR #1323, opened 2016-01-08
Perf/ttb newmsg ts1.0 release candidate v2

erikleitch:
This is one of three related PRs (https://github.com/basho/riak_pb/pull/176 is the corresponding one for riak_pb, https://github.com/basho/riak-erlang-client/pull/259 is the corresponding one for riakc) to streamline put requests using TTB encoding instead of PB.
As with the current end-to-end/timeseries branch, the TTB encoded messages are still handled by the riak_kv_pb service; even though they are not PB-encoded messages, they reuse the erlang infrastructure around generating erlang messages from the .proto definitions.

Changes here are:

1) extend the riak_kv_pb_timeseries registration to handle MsgCodes 90-100, where 100 is the new streamlined message

2) add a case statement to decode(), switching on the new message type (tsputreqttb)

3) add new process() variants to handle the new message type

NB: This branch will fail to build unless rebar.config is modified to pick up the same branch of riak_pb.  I didn't want to commit that change, because as JD has pointed out, that risks merging into the mainline

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -122,70 +122,48 @@ encode(Message) ->
     {ok, riak_pb_codec:encode(Message)}.
 
 
+-spec process_tsreq(atom() | #tsputreq{} | #tsputreqttb{}, term(), #state{}) ->
+{reply, #tsqueryresp{} | #rpberrorresp{}, #state{}}.
+process_tsreq(#tsputreq{table = Table, columns = _Columns, rows = _Rows}, Data, State) ->
+  Mod = riak_ql_ddl:make_module_name(Table),
+
+  case (catch validate_rows(Mod, Data)) of
+    [] ->
+      try
+        case put_data(Data, Table, Mod) of
+          0 ->
+            {reply, #tsputresp{}, State};
+          ErrorCount ->
+            EPutMessage = flat_format("Failed to put ~b record(s)", [ErrorCount]),
+            {reply, make_rpberrresp(?E_PUT, EPutMessage), State}
+        end
+      catch
+        Class:Exception ->
+          lager:error("error: ~p:~p~n~p", [Class,Exception,erlang:get_stacktrace()]),
+          Error = make_rpberrresp(?E_IRREG, to_string({Class, Exception})),
+          {reply, Error, State}
+      end;
+    BadRowIdxs when is_list(BadRowIdxs) ->
+      {reply, validate_rows_error_response(BadRowIdxs), State};
+    {_, {undef, _}} ->
+      BucketProps = riak_core_bucket:get_bucket(table_to_bucket(Table)),
+      {reply, missing_helper_module(Table, BucketProps), State}
+  end.
+
 -spec process(atom() | #tsputreq{} | #tsputreqttb{} | #tsdelreq{} | #tsgetreq{} | #tslistkeysreq{}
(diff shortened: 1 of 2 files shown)


hazen:
@erikleitch My understanding was there was a process dictionary setting which JD added to use t2b instead of PB.  Did I remember that correctly? So now we are still using the PB transport (hence the new message), but are encoding with t2b instead of protobuf encoding in that message, right?

@javajolt That's correct -- still does work that way.  It's in riak_pb_codec:decode that the dict lookup happens.  Both the original implementation and this version use the PB transport, but just switch on how the message is encoded

src/riak_kv_pb_timeseries.erl
```diff
@@ -152,6 +154,38 @@ process(#tsputreq{table = Table, columns = _Columns, rows = Rows}, State) ->
...
+process(#tsputreqttb{rows = []}, State) ->
+    {reply, #tsputresp{}, State};
+process(#tsputreqttb{table = Table, columns = _Columns, rows = Rows}, State) ->
+
+    Mod = riak_ql_ddl:make_module_name(Table),
+    Data = Rows,
+
+    %% validate only the first row as we trust the client to send us
+    %% perfectly uniform data wrt types and order
+    case (catch Mod:validate_obj(hd(Data))) of
```
hazen: I'm assuming that `Rows` is a list. We are only validating the first row?

erikleitch: @javajolt This is carbon copy from the normal process statement -- we have only ever validated the first row, yes.  (Gordon decided to do that a long time ago to minimize perf impact)

hazen: @erikleitch So it looks like we started validating all rows back in https://github.com/basho/riak_kv/pull/1313.  **#tsputreqttb** differs in a few places from **#tsputeq** currently.

erikleitch: @javajolt thanks for catching -- I rebased this a few days ago, so this was the state of things as of then.

hazen: I added a new branch, `amendment/bch/ttb_newmsg_ts1.0_release_candidate_v2`, which consolidates processing **tsputreqttb** and **tsputreq**.

src/riak_kv_pb_timeseries.erl
```diff
@@ -120,7 +122,7 @@ encode(Message) ->
     {ok, riak_pb_codec:encode(Message)}.
 
 
--spec process(atom() | #tsputreq{} | #tsdelreq{} | #tsgetreq{} | #tslistkeysreq{}
+-spec process(atom() | #tsputreq{} | #tsputreqttb{} | #tsdelreq{} | #tsgetreq{} | #tslistkeysreq{}
```
hazen: So I did see your commit to change this name to avoid collision with previous versions, but the new name isn't terribly consistent with all of the others.  Do we actually have customers in the wild using the **TsTtbPutReq** message?

erikleitch:
Renamed tsputreqttb to tsttbputreq.

Had to modify process_tsreq because there was no pattern to match #tsttbputreq.  This now splits out Table and Rows from the messages in process(), then calls a unified process_tsreq with Table and Row as arg instead of Req

javajolt:
:+1: 7099685

hazen:
:+1: 8e0f837

hazen:
@borshop merge

PR #1324, opened 2016-01-13
check and report table is activated before calling any TS function on it

hmmr:
RTS-831

Checks for this error condition are in https://github.com/basho/riak_test/pull/974.

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -268,19 +268,11 @@ sub_tsttbputreq(Mod, _DDL, #tsttbputreq{table = Table, rows = Data},
 sub_putreq_common(Mod, Table, Data, State) ->
     case catch validate_rows(Mod, Data) of
         [] ->
-            try
-                case put_data(Data, Table, Mod) of
-                    0 ->
-                        {reply, #tsputresp{}, State};
-                    ErrorCount ->
-                        EPutMessage = flat_format("Failed to put ~b record(s)", [ErrorCount]),
-                        {reply, make_rpberrresp(?E_PUT, EPutMessage), State}
-                end
-            catch
-                Class:Exception ->
-                    lager:error("error: ~p:~p~n~p", [Class, Exception, erlang:get_stacktrace()]),
-                    Error = make_rpberrresp(?E_IRREG, to_string({Class, Exception})),
-                    {reply, Error, State}
+            case put_data(Data, Table, Mod) of
+                0 ->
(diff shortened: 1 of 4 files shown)


src/riak_kv_pb_timeseries.erl
```diff
@@ -123,32 +126,35 @@ encode(Message) ->
...
-      end;
-    BadRowIdxs when is_list(BadRowIdxs) ->
-      {reply, validate_rows_error_response(BadRowIdxs), State};
-    {_, {undef, _}} ->
-      BucketProps = riak_core_bucket:get_bucket(table_to_bucket(Table)),
-      {reply, missing_helper_module(Table, BucketProps), State}
-  end.
+    check_table_and_call(
+      Table,
+      fun(Mod, _DDL) ->
```
hazen: Maybe making this a named function might make the code more readable.  Same in the following cases.

hmmr: I did consider this, but then those functions would have to be moved pages of code down, past the last head of process/2, which would require much scrolling and detract from locality of code.

However, it's probably time to split the 37k-long riak_kv_pb_timeseries into smaller files. Let me see.

hmmr:
In 928a5d3 I undertake a more thought-through reshuffle towards better code locality. Fully admitting there be ways to do this differently, my reasoning has been to attempt to avert the state of, for example, https://github.com/basho/riak-erlang-client/blob/end-to-end/timeseries/src/riakc_pb_socket.erl, which has become a mire of interdependencies with its sibling (or is it subordinate?) module riak_ts.erl.

In detail, the diff against the previous commit does:
- move actual code from `riak_kv_pb_timeseries:process/2` into named functions;
- cluster those functions with supporting local functions, in anticipation of possible removal of those clusters into separate modules;
- move `make_ts_key/3` to ts_util, move `table_to_bucket/1`, `pk/1`, `lk/1` there as well;
- provide some specs, types and docstrings.

The only change in the substance is removal of heads of `process/2` matching on `#tsputreq{}` and `#tsttbputreq{}` with `rows = []`. Given that clients are not normally likely to be putting empty data, and given that putting `[]` will still work as expected after doing a couple of foldl's over empty lists downstream in `put_data`, an early shortcut at `process/3` does not seem to be worth the odd extra lines breaking the tidiness of `process/3`. But also, importantly, putting [] in a non-existing table will now return an error, as it ought to be.

andytill:
@borshop: retry

andytill:
+1 add9a4c

PR #1325, opened 2016-01-14
Obsolete: Support aggregation functions and arithmetic RTS-848

andytill:
Support aggregation functions and arithmetic. For example.

``` sql
COUNT(foo) + 1
```

And

``` sql
COUNT(foo) + COUNT(bar)
```

Clauses like.

``` sql
COUNT(COUNT(foo))
```

Will not be supported.  Depends on riak_ql PR https://github.com/basho/riak_ql/pull/90 && https://github.com/basho/riak_ql/pull/93

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -128,6 +128,14 @@ prepend_select_columns([_|_] = MultiCols, Acc) ->
 prepend_select_columns(V, Acc) ->
     [V | Acc].
 
+%% copy pasta mapfoldl from lists.erl, otherwise this reports
+%% a warning in dialyzer!
+my_mapfoldl(F, Accu0, [Hd|Tail]) ->
+    {R,Accu1} = F(Hd, Accu0),
+    {Rs,Accu2} = my_mapfoldl(F, Accu1, Tail),
+    {[R|Rs],Accu2};
+my_mapfoldl(F, Accu, []) when is_function(F, 2) -> {[],Accu}.
+
 %%
 compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause = Sel } } = Q) ->
     CompileColFn = 
@@ -141,7 +149,7 @@ compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause = 
     %% iterate from the right so we can append to the head of lists
     {ResultTypeSet, Q2} = lists:foldl(CompileColFn, Acc, Sel),
 
-    {ColTypes, Errors} = lists:mapfoldl(
+    {ColTypes, Errors} = my_mapfoldl(
         fun(ColASTX, Errors) ->
             infer_col_type(DDL, ColASTX, Errors)
         end, [], Sel),
@@ -302,7 +310,7 @@ compile_select_col_stateless(DDL, {Op, A, B}) ->
 
 %%
 -spec infer_col_type(#ddl_v1{}, selection(), Errors1::[any()]) ->
-        {Type::simple_field_type(), Errors2::[any()]}.
+        {Type::simple_field_type() | error, Errors2::[any()]}.
 infer_col_type(_, {Type, _}, Errors) when Type == sint64; Type == varchar;
                                           Type == boolean; Type == double ->
     {Type, Errors};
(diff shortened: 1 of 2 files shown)


src/riak_kv_qry_compiler.erl
```diff
@@ -207,52 +246,44 @@ compile_select_col(DDL, {{window_agg_fn, FnName}, [FnArg1]}) ->
...
             {ColTypes2, IsValid2, Compiled_arg1} =
                 compile_select_col_stateless(DDL, FnArg1),
-            %% all the windows agg fns so far are arity of 1
-            %% which we have forced in this clause by matching on a single argument in the
-            %% function head
             FnSig = riak_ql_window_agg_fns:get_arity_and_type_sig(FnName),
             {IsValid, ColRet} = check_types(FnName, FnSig, ColTypes2),
 
             IsValid3 = merge_validation(IsValid2, IsValid),
+            % all the windows agg fns so far are arity of 1
```
hmmr, line 268: On behalf of all cat^W emacs people, please keep this comment block marked with double `%`. Autoindenting this block in Emacs will move it far to the right. (Ditto for other blocks elsewhere.)

src/riak_kv_qry_compiler.erl
```diff
@@ -319,13 +374,13 @@ merge_validation({error, E1}, {error, E2}) -> {error, E1 ++ E2}.
 
 %%
 compile_select_col_stateless2('+', A, B) ->
-    fun(Row) -> A(Row) + B(Row) end;
+    fun(Row, State) -> A(Row, State) + B(Row, State) end;
```
hmmr: A `State` parameter in a function having '_stateless' in its name?

src/riak_kv_qry_compiler.erl
```diff
@@ -1640,6 +1672,39 @@ basic_select_arith_1_test() ->
        Sel
       ).
 
+varchar_literal_test() ->
+    {ok, Rec} = get_query("SELECT 'hello' from mytab"),
```
hmmr, line 1705: I note that under normal circumstances, validation of a SELECT query without WHERE clause will not succeed. Is there a plan to allow WHERE-less queries with a static expression not referencing any columns?

(I understand these tests only deal with parts of DDL structure, so the above question is asked in a broader context.)

src/riak_kv_qry_compiler.erl
andytill, line 268: I wondered why my comments kept getting double indented in other people's commits.

src/riak_kv_qry_compiler.erl
andytill, line 1705: This is just to minimise test data. I'm not sure about plans to do things like `SELECT 2 FROM dual`, maybe if there is ever a need for sub queries.

src/riak_kv_qry_compiler.erl
andytill: That variable is more like a `ColValue` than a state, or do you think there is a better name? IMHO the function is correctly named as it purposely does not handle aggregate functions.

src/riak_kv_qry_compiler.erl
```diff
@@ -22,11 +22,9 @@
...
 -module(riak_kv_qry_compiler).
 
--export([
-         compile/3,
-         run_select/2,
-         run_select/3  %% what is this for?
-        ]).
+-export([compile/3]).
+-export([finalise_aggregate/2]).
+-export([run_select/2, run_select/3]).
```
hazen, line 28: Are multiple exports just stylistic preference?

hazen:
All unit tests pass with the QL `at/aggregations_fns_and_arithmetic2` branch

src/riak_kv_qry_compiler.erl
```diff
@@ -113,6 +128,14 @@ prepend_select_columns([_|_] = MultiCols, Acc) ->
 prepend_select_columns(V, Acc) ->
     [V | Acc].
 
+%% copy pasta mapfoldl from lists.erl, otherwise this reports
+%% a warning in dialyzer!
+my_mapfoldl(F, Accu0, [Hd|Tail]) ->
```
hazen, line 133: Really?  You can't use `lists:mapfoldl`? This is the dialyzer bug mentioned the other day?

lehoff, line 133: It is the dialyzer potential bug - I have asked Andrei to create a Jira ticket for this or add to the one Andy created.

src/riak_kv_qry_compiler.erl
andytill, line 28: It makes it very easy to sort alphabetically and group functions with the same name but different arity.

PR #1327, opened 2016-01-21
Support aggregation functions and arithmetic RTS-848

hazen:
Support aggregation functions and arithmetic. For example.

``` sql
COUNT(foo) + 1
```

And

``` sql
COUNT(foo) + COUNT(bar)
```

Clauses the one below are not going to be supported.

``` sql
COUNT(COUNT(foo))
```

The select clause is now split into two distinct phases, **accumulation** and **finalisation**.  Previously all the select logic was happening in accumulation, for example on each row `+2` it or run the `AVG` function over it.  With clauses like `COUNT(foo) + 1` we need the final state of `COUNT` before `+1` can be added. A non-simple cases would be doing arithmetic with two stateful functions `COUNT(foo) + COUNT(bar)`.

This change pulls out aggregation functions from the select clause columns, if there are more than one function then a temporary column is created. So we may have some (pseduo) AST like the following:

``` erlang
{+, {window_agg_gfn, 'AVG', [{identifier, <<"mycol"">>}]}, {integer, 1}}
```

The accumulator AST would be the function `{window_agg_gfn, 'AVG', [{identifier, <<"mycol"">>}` and the finaliser AST would be `{+, {return_state, 1}, {integer, 1}}`.

The new `return_state` selection means the state of column one in the aggregate row once all the result rows for the query had been processed.  `return_state` takes temporary columns into consideration.

Now that we have temporary columns, the select clause columns and the finaliser column list may or may not be the same length.

This is a replacement for #1325 to merge against `riak_ts-develop-1.2`.  It is dependent upon https://github.com/basho/riak_ql/pull/95

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -128,6 +128,14 @@ prepend_select_columns([_|_] = MultiCols, Acc) ->
 prepend_select_columns(V, Acc) ->
     [V | Acc].
 
+%% copy pasta mapfoldl from lists.erl, otherwise this reports
+%% a warning in dialyzer!
+my_mapfoldl(F, Accu0, [Hd|Tail]) ->
+    {R,Accu1} = F(Hd, Accu0),
+    {Rs,Accu2} = my_mapfoldl(F, Accu1, Tail),
+    {[R|Rs],Accu2};
+my_mapfoldl(F, Accu, []) when is_function(F, 2) -> {[],Accu}.
+
 %%
 compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause = Sel } } = Q) ->
     CompileColFn = 
@@ -141,7 +149,7 @@ compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause = 
     %% iterate from the right so we can append to the head of lists
     {ResultTypeSet, Q2} = lists:foldl(CompileColFn, Acc, Sel),
 
-    {ColTypes, Errors} = lists:mapfoldl(
+    {ColTypes, Errors} = my_mapfoldl(
         fun(ColASTX, Errors) ->
             infer_col_type(DDL, ColASTX, Errors)
         end, [], Sel),
@@ -302,7 +310,7 @@ compile_select_col_stateless(DDL, {Op, A, B}) ->
 
 %%
 -spec infer_col_type(#ddl_v1{}, selection(), Errors1::[any()]) ->
-        {Type::simple_field_type(), Errors2::[any()]}.
+        {Type::simple_field_type() | error, Errors2::[any()]}.
 infer_col_type(_, {Type, _}, Errors) when Type == sint64; Type == varchar;
                                           Type == boolean; Type == double ->
     {Type, Errors};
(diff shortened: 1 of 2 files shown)


src/riak_kv_qry_compiler.erl
```diff
@@ -82,6 +83,20 @@ expand_query(#ddl_v1{local_key = LK, partition_key = PK},
             {ok, SubQueries}
     end.
 
+%% Calulate the final result for an aggregate.
+-spec finalise_aggregate(#riak_sel_clause_v1{}, [any()]) -> [any()].
```
hazen, line 87: @andytill Where is `finalise_aggregate` called?

src/riak_kv_qry_compiler.erl
```diff
@@ -145,197 +197,228 @@ compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause =
...
 compile_select_col_stateless2('/', A, B) ->
-    fun(Row) -> A(Row) / B(Row) end;
+    fun(Row, State) ->
+        riak_ql_window_agg_fns:divide(A(Row, State), B(Row, State))
+    end;
 compile_select_col_stateless2('-', A, B) ->
-    fun(Row) -> A(Row) - B(Row) end.
-
-check_types(_FnName, {_, RetType}, _Type)
-  %% functions such as COUNT taking an argument of any type have
```
hazen, line 331: @andytill This comment about COUNT functions is no longer true?

src/riak_kv_qry_compiler.erl
```diff
@@ -717,34 +800,12 @@ modify_where_key(TupleList, Field, NewVal) ->
     lists:keyreplace(Field, 1, TupleList, {Field, FieldType, NewVal}).
 
 -ifdef(TEST).
--compile(export_all).
```
hazen, line 720: Thank you, thank you, thank you!

src/riak_kv_qry_compiler.erl
andytill, line 87: @javajolt it is called in the `riak_kv_qry_worker`.

src/riak_kv_qry_compiler.erl
andytill, line 331: Yep that function head is no longer necessary.

javajolt:
Getting some test failures.  I'll investigate.
```
{riak_ql_window_agg_fns,
                                                 fn_type_signature,
                                                 ['COUNT',[varchar]],
                                                 []},
                                                {riak_kv_qry_compiler,
                                                 infer_col_function_type,3,
                                                 [{file,
                                                   \"src/riak_kv_qry_compiler.erl\"},
                                                  {line,347}]},
                                                {riak_kv_qry_compiler,
                                                 my_mapfoldl,3,
                                                 [{file,
                                                   \"src/riak_kv_qry_compiler.erl\"},
                                                  {line,134}]},
                                                {riak_kv_qry_compiler,
                                                 compile_select_clause,2,
                                                 [{file,
                                                   \"src/riak_kv_qry_compiler.erl\"},                                                                                     {line,152}]},
                                                {riak_kv_qry_compiler,
                                                 compile,3,
                                                 [{file,
                                                   \"src/riak_kv_qry_compiler.erl\"},
                                                  {line,46}]},
                                                {riak_kv_qry,
                                                 maybe_submit_to_queue,2,
                                                 [{file,\"src/riak_kv_qry.erl\"},
                                                  {line,98}]},
                                                {riak_kv_pb_timeseries,
                                                 sub_tsqueryreq,4,
                                                 [{file,
                                                   \"src/riak_kv_pb_timeseries.erl\"},
                                                  {line,644}]},
                                                {riak_api_pb_server,
                                                 process_message,4,
                                                 [{file,
                                                   \"src/riak_api_pb_server.erl\"}
```

hazen:
:+1: 1ead83b

hazen:
@borshop merge

PR #1328, opened 2016-01-22
support setting table properties in WITH clause in CREATE TABLE query

hmmr:
RTS-865

Relies on https://github.com/basho/riak_ql/pull/96.

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -900,18 +900,33 @@ bucket_type_create_with_timeseries_table_test() ->
           "user   varchar   not null, ",
           "time   timestamp not null, ",
           "PRIMARY KEY ((series, user, quantum(time, 15, m)), "
-          "series, user, time))">>,
-    JSON = json_props([{bucket_type, my_type}, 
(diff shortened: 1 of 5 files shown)


jonmeredith:
Very nice :)  Does the code path you used to create the buckets with properties goes through the checks in riak_kv_bucket and reports back any errors/problems?

hmmr:
@jonmeredith yes and no. Yes in the sense that this code path follows the one used in table creation via riak-admin (https://github.com/basho/riak_kv/blob/feature/az/create-with/src/riak_kv_pb_timeseries.erl#L217, same as in https://github.com/basho/riak_kv/blob/feature/az/create-with/src/riak_kv_console.erl#L526). And no, because it does not invoke https://github.com/basho/riak_kv/blob/feature/az/create-with/src/riak_kv_bucket.erl#L135. There appears to be room for improvement here.

hmmr:
There seems to be an agreement to go store properties in the new `#ddl_v2{}` record (https://github.com/basho/riak_kv/pull/1332). This PR is thus going to be closed.

PR #1330, opened 2016-01-22
Hotfix/2.0.6 memory backend ets leaks

JeetKunDoug:
Fix for a memory leak in the timer reference table used by the memory backend when attempting to limit to a certain amount of used memory.

This will resolve RIAK-1576 (#1064) for the 2.0 branch - will need to forward-merge for 2.1/2.2


JeetKunDoug:
Going to cherry-pick these onto a branch from 2.0's HEAD as the deps actually work there.

PR #1332, opened 2016-01-25
support setting table properties in WITH clause in CREATE TABLE query (#ddl_v2{})

hmmr:
RTS-865

Goes with https://github.com/basho/riak_ql/pull/99 and https://github.com/basho/riak_core/pull/805.

This is an alternative way to implement setting table properties with `WITH` clause in a CREATE TABLE query (compare https://github.com/basho/riak_kv/pull/1328). Here, the table properties are kept in the `properties` field in a new `#ddl_v2{}` record.


hmmr:
Will have to resubmit to riak_ts-develop-1.3. Closing this one.

PR #1333, opened 2016-01-27
☠️ WIP ☠️ Flexible keys RTS-621

andytill:
Previous to this change, partition and local keys had to be three fields long and end with the quanta field.

With this change the partition key must be at least one field long and end with the quanta function/field. The initial fields of local key must match the entire partition key but can include any number of fields afterwards.

This requires the riak_ql branch at/flexi_keys to build.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -454,12 +454,24 @@ fold_indexes_fun(FoldIndexFun) ->
             end
     end.
 
+build_list({_K, _V}=KV, Acc) ->
+    [KV | Acc].
+
 range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
                                               ref=Ref}) ->
     {_, Bucket, Qry} = proplists:lookup(index, Opts),
     ?SQL_SELECT{'WHERE'    = W,
                 helper_mod = Mod,
                 local_key  = LK} = Qry,
+    %% Work out what the elements of the local key are after the partitioning key.
+    %% Used below to add dummy fields to pad out the StartK2/EndK2 fields.
+    LKAST = LK#key_v1.ast,
+    ExtraLK = case length(LKAST) > 3 of
+		  true ->
+		      lists:nthtail(3, LKAST);
+		  false ->
+		      []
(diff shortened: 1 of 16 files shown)


src/riak_kv_eleveldb_backend.erl
```diff
@@ -478,17 +493,19 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
...
+    %% Pad with missing local keys after the timestamp - use minimum erlang term value of '0' to make
+    %% sure we start the search at the beginning.
+    StartK2 = [{Field, Val} || {Field, _Type, Val} <- StartK] ++ [{N, 0} || #param_v1{name = [N]} <- ExtraLK],
     StartK3 = riak_ql_ddl:make_key(Mod, LK, StartK2),
     StartK4 = riak_kv_ts_util:encode_typeval_key(StartK3), %% TODO: Avoid adding/removing type info
     StartKey = to_object_key(Bucket, StartK4),
-    EndK2 = [{Field, Val} || {Field, _Type, Val} <- EndK],
+    %% Pad the missing end - not perfect because there could be longer bitstrings, but will have to do for prototype
+    %% TODO: FIX END KEY
+    EndK2 = [{Field, Val} || {Field, _Type, Val} <- EndK] ++ [{N, <<16#ffffffff:64>>} || #param_v1{name = [N]} <- ExtraLK],
```
jonmeredith: There's still the issue of the end key to be addressed - if the end key had 65 set bits it would be missed.

The problem is Erlang Term ordering goes like this
number < atom < reference < fun < port < pid < tuple < map < nil < list < bit string

So if you're trying to find a term that will be after all of the other terms, doing it this naive way you would need an infinitely long stream of bits.  I think you could do it by converting the binary back to an integer, incrementing by 1 and converting back to a binary, but you would have to handle overflow cases.

lehoff: I know it is going to cost some extra space, but we could tag the keys `{'a', Key}` and use `{'e', <<>>}` as the end key.

That way we create a type that allows for detecting the end key.

andytill:
Closing and opening a new PR against 1.3.

PR #1335, opened 2016-01-28
fix of schema changes

gordonguthrie:


priv/riak_kv.schema
```diff
@@ -678,26 +678,77 @@ end
...
-  {datatype, integer}
+  {commented, 10000},
+  {datatype, integer},
+  hidden
+]}.
+
+%% @doc Timeout for Time Series queries, after which riak
+%% will return a timeout error - default is 10,000 miliseconds
+{mapping, "riak_kv.query.timeseries.timeout", "riak_kv.timeseries_query_timeout_ms", [
+  {default, "10000ms"},
```
angrycub, line 694: you could use 10s as well.  It will be converted to ms automatically because you specified that it was a duration in ms.

hazen:
:+1: 5400e8a

hazen:
@borshop merge

PR #1337, opened 2016-01-29
Remove ~p formatting from Time Series error messages.

andytill:
Typos in Time Series queries often lead to ugly error messages because erlang terms get `~p` formatted into the error string.

``` erlang
171> riakc_ts:query(Pid, "SELECT * FROM tab4 WHERE a > 0 AND a( < 1000 AND a1 = 2").
{error,<<"Message decoding error: \"Unexpected token '<<\\\"<\\\">>'\"">>}
```

With the change the error is.

``` erlang
273> riakc_ts:query(Pid, "SELECT * FROM tab4 WHERE a > 0 AND a( < 1000 AND a1 = 2").
{error,{1020,<<"Unexpected token '<'">>}}
```

During the decode step queries, create table and describe statements are parsed using riak_ql. The decode step does not support an error return that it does not format with `~p` so the error is passed to the `riak_kv_pb_timeseries:process/2` which returns it as a result. I experimented with supporting a formatted error string that riak_api could detect with the type `{error, {formatted_error, string() | binary()}}` but preferred this way where there are no new error types. Also discoverabilty for that error return for users is poor.

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -57,6 +57,7 @@
 -define(E_ACTIVATE,          1017).
 -define(E_BAD_QUERY,         1018).
 -define(E_TABLE_INACTIVE,    1019).
+-define(E_PARSE_ERROR,       1020).
 
 -define(FETCH_RETRIES, 10).  %% TODO make it configurable in tsqueryreq
 -define(TABLE_ACTIVATE_WAIT, 30). %% ditto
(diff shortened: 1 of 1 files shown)


src/riak_kv_pb_timeseries.erl
```diff
@@ -818,20 +824,18 @@ make_tscolumndescription_list(ColumnNames, ColumnTypes) ->
...
-
-decoder_parse_error_resp({LineNo, riak_ql_parser, Msg}) when is_integer(LineNo) ->
-    flat_format("~ts", [Msg]);
-decoder_parse_error_resp({Token, riak_ql_parser, _}) ->
-    flat_format("Unexpected token '~p'", [Token]);
-decoder_parse_error_resp(Error) ->
+make_decoder_error_response({LineNo, riak_ql_parser, Msg}) when is_integer(LineNo) ->
+    make_rpberrresp(?E_PARSE_ERROR, flat_format("~ts", [Msg]));
+make_decoder_error_response({Token, riak_ql_parser, _}) when is_binary(Token) ->
+    make_rpberrresp(?E_PARSE_ERROR, flat_format("Unexpected token '~s'", [Token]));
```
hazen, line 830: Do we know that `Token` will always be a string? Just preparing for the degenerate case in which something unexpected actually blows up the error statement.

andytill, line 830: A binary can always be formatted using `~s`, had to check though.

``` erlang
1> io_lib:format("hey ~s", [<<"there">>]).
[104,101,121,32,"there"]
2> io:format("hey ~s", [<<"there">>]).    
hey thereok
3> io:format("hey ~s", [<<0>>]).      
hey ^@ok
4> io_lib:format("~s", [<<0>>]).
[[0]]
5> io_lib:format("~s", [<<4444>>]).
["\\"]
6> io_lib:format("~s", [<<44448988>>]).
["Ü"]
7> io_lib:format("~s", [<<444489889999>>]).
["Ï"]
```

andytill:
@borshop merge

PR #1338, opened 2016-01-29
Catch division by zero errors during finalisation

andytill:
Division by zero errors were not being caught by the `riak_kv_qry_worker` processes during finalisation. The result was that the request would hang for the timeout period then return a timeout error for the query below.

``` sql
SELECT AVG(pressure) / 0 FROM WeatherData WHERE time > 0 AND time < 1000 AND myfamily = 'family1' AND myseries = 'series1'
```

Now a `divide_by_zero` error is returned immediately.


macintux:
Rub some test on it? Looks good otherwise

hazen:
@borshop merge

PR #1341, opened 2016-02-04
Return an error when deleting a key that does not exist RTS-894

andytill:
Currently `ok` is returned when a single key delete is issued against a key that does not exist.

With this change `{error, {1021, <<"notfound">>}}` is returned instead. 1021 is a new error code for the case where the key does not exist.


javajolt:
:+1: 9dbc89e

hazen:
:+1: f791bc0

hazen:
@borshop merge

PR #1342, opened 2016-02-05
HTTP client API for TS

hmmr:
RTS-258

Implement these HTTP API calls:

```
GET       /ts/1/table/Table          single-key get
DELETE    /ts/1/table/Table          single-key delete
PUT       /ts/1/table/Table          batch put
GET       /ts/1/table/Table/keys     list_keys
GET       /ts/1/coverage             coverage for a query
GET/POST  /ts/1/query                execute an SQL query (SELECT or DESCRIBE with GET,
                                     CREATE TABLE with POST)
```

The keys, data and query for arguments to particular calls are to be submitted in request body, as used in these examples: https://gist.github.com/hmmr/94abdb4d648a44009e93.

Note: list_keys is broken.

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -265,10 +265,14 @@ sub_putreq_common(Mod, Table, Data, State) ->
     case catch riak_kv_ts_util:validate_rows(Mod, Data) of
         [] ->
             case riak_kv_ts_util:put_data(Data, Table, Mod) of
-                0 ->
+                ok ->
                     {reply, #tsputresp{}, State};
-                ErrorCount ->
-                    {reply, failed_put_response(ErrorCount), State}
+                {error, {some_failed, ErrorCount}} ->
+                    {reply, failed_put_response(ErrorCount), State};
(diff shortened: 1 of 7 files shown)


hazen:
@hmmr Do we have a riak_test for this?

hmmr:
@javajolt It's in the plans, too. One thing to consider is, whether to use native httpc facilities for requests or directly run curl (because that is the ultimate way this interface will be used).

hmmr:
There are a couple of technical notes to mention regarding the way API functions are invoked from webmachine callbacks.
1. It appears that webmachine does not call `accept_doc_body` for GET requests, making it problematic, for example, to get at the key data in `curl -G ts/1/Table --data {key: [1,2,3]}`. I worked around this idiosyncrasy by consolidating the entire request-api-call-response cycle in a single, all-in-one function. This way, we do the request validation and permission checks in that function, instead of relying on the webmachine callbacks.
2. There is no chunking of data supported on either upload or download path. Something to implement yet.

hmmr:
Will have to resubmit to riak_ts-develop-1.3.

PR #1345, opened 2016-02-09
Remove key field filter in `riak_kv_ts_util:make_ts_keys/3`

andytill:
We currently have the restriction that keys had to be same order that they are specified in the table definition.

When tested they also had to be the first keys in the table definition and so, also consecutive in the table definition.

Functions `riak_ql_ddl:get_local_key/3` and `riak_ql_ddl:get_partition_key/3` require all the object fields, some of which are filtered by the list comprehension. Removing the filter provides the correct input to `riak_ql_ddl`.

### DIFF
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -231,30 +231,30 @@ make_ts_keys_1_test() ->
     ).
 
 % a two element key, still using the table definition field order
-make_ts_keys_2_test() ->
-    {DDL, Mod} = helper_compile_def_to_module(
-        "CREATE TABLE table1 ("
-        "a SINT64 NOT NULL, "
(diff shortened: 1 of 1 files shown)


hazen:
Looks like the restrictions are tested in `riak_kv_ts_util:make_ts_keys_2_test/0` and `riak_kv_ts_util:make_ts_keys_3_test/0` need to be removed for the unit tests to pass on this one, @andytill

andytill:
Thanks @javajolt I was testing against my riak_ql flexible keys branch, which passes the tests.

I will be using those tests imminently so just lazily commented them instead of deleting the code.

hazen:
:+1:  89bf378

hazen:
@borshop merge

PR #1347, opened 2016-02-11
Fix for 1.2 performance degradation RTS-898

andytill:
The performance team discovered that the performance degradation was caused by a call to `riak_core_claimant:bucket_type_status/1` introduced in 1.2. This checks if the bucket is active.

This has been replaced with a call to `riak_core_bucket:get_bucket/1` which returns the bucket properties or an error if the bucket is not active. It calls into the local meta data rather than check the ring.


javajolt:
@borshop merge

andytill:
@borshop merge

PR #1349, opened 2016-02-15
Flexible keys

andytill:
Remove restrictions on partition and local keys:

```
One to N elements log.
Local key must start with partition key but can have extra fields.
Key fields do *not *have to be in the same order as defined in the table definition.
A quantum function in the partition key is not mandatory.
```

Relies on riak_ql PR https://github.com/basho/riak_ql/pull/104

### DIFF
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -977,7 +977,7 @@ bucket_type_create_with_timeseries_table_error_when_write_once_set_to_false_test
         )
     ).
 
-bucket_type_create_with_timeseries_table_error_with_short_primary_key_test() ->
+bucket_type_create_with_timeseries_table_with_two_element_key_test() ->
     TableDef =
         <<"CREATE TABLE my_type (",
           "user   varchar   not null, ",
--- a/src/riak_kv_console.erl
+++ b/src/riak_kv_console.erl
@@ -978,22 +978,22 @@ bucket_type_create_with_timeseries_table_error_when_write_once_set_to_false_test
     ).
 
 bucket_type_create_with_timeseries_table_error_with_short_primary_key_test() ->
-    Ref = make_ref(),
     TableDef =
         <<"CREATE TABLE my_type (",
-        "user   varchar   not null, ",
-        "time   timestamp not null, ",
-        "PRIMARY KEY ((user, quantum(time, 15, m)), "
-        "user, time))">>,
+          "user   varchar   not null, ",
+          "time   timestamp not null, ",
+          "PRIMARY KEY ((user, quantum(time, 15, m)), user, time))">>,
     JSON = json_props([{bucket_type, my_type},
                        {table_def, TableDef}]),
-    ?assertEqual(
-        error,
-        bucket_type_create(
-            fun(Props) -> put(Ref, Props) end,
-            <<"my_type">>,
-            mochijson2:decode(JSON)
-        )
+    Result = bucket_type_create(
+        fun(Props) -> Props end,
+        <<"my_type">>,
+        mochijson2:decode(JSON)
+    ),
+    % just assert that this returns a ddl prop
+    ?assertMatch(
+        [{ddl, _}|_],
+        Result
     ).
 
 bucket_type_create_with_timeseries_table_error_with_misplaced_quantum_test() ->
@@ -1024,10 +1024,9 @@ bucket_type_and_table_error_local_key_test() ->
         "user   varchar   not null, ",
         "time   timestamp not null, ",
         "other  varchar   not null, ",
-        "PRIMARY KEY ((series, user, quantum(time, 15, m)), "
-        "series, user, time, other))">>,
+        "PRIMARY KEY ((series, user, quantum(time, 15, m)), seriesd, user, time, other))">>,
     JSON = json_props([{bucket_type, my_type},
-        {table_def, TableDef}]),
+                       {table_def, TableDef}]),
     ?assertEqual(
         error,
         bucket_type_create(
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -454,12 +454,24 @@ fold_indexes_fun(FoldIndexFun) ->
             end
     end.
 
+build_list({_K, _V}=KV, Acc) ->
+    [KV | Acc].
+
 range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
                                               ref=Ref}) ->
     {_, Bucket, Qry} = proplists:lookup(index, Opts),
     ?SQL_SELECT{'WHERE'    = W,
                 helper_mod = Mod,
                 local_key  = LK} = Qry,
+    %% Work out what the elements of the local key are after the partitioning key.
+    %% Used below to add dummy fields to pad out the StartK2/EndK2 fields.
+    LKAST = LK#key_v1.ast,
+    ExtraLK = case length(LKAST) > 3 of
+		  true ->
+		      lists:nthtail(3, LKAST);
+		  false ->
+		      []
+	      end,
     %% this is all super-fugly
     {startkey, StartK} = proplists:lookup(startkey, W),
     {endkey,   EndK}   = proplists:lookup(endkey, W),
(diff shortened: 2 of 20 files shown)


src/riak_kv_console.erl
```diff
@@ -978,22 +978,22 @@ bucket_type_create_with_timeseries_table_error_when_write_once_set_to_false_test
...
 
 bucket_type_create_with_timeseries_table_error_with_short_primary_key_test() ->
-    Ref = make_ref(),
     TableDef =
         <<"CREATE TABLE my_type (",
-        "user   varchar   not null, ",
-        "time   timestamp not null, ",
-        "PRIMARY KEY ((user, quantum(time, 15, m)), "
-        "user, time))">>,
+          "user   varchar   not null, ",
```
jonmeredith, line 983: Can you rename the calling test - it's no longer an error - perhaps `bucket_type_create_with_timeseries_table_with_two_element_key_test`

src/riak_kv_eleveldb_backend.erl
```diff
@@ -47,13 +47,12 @@
...
          status/1,
          callback/3]).
-
 -export([data_size/1]).
 
--compile({inline, [
-                   to_object_key/2, from_object_key/1,
-                   to_index_key/4, from_index_key/1
-                  ]}).
+% -compile({inline, [
```
jonmeredith: This is a performance optimization - any reason to leave it commented or is this a development change that survived to PR?

src/riak_kv_eleveldb_backend.erl
```diff
@@ -454,47 +453,31 @@ fold_indexes_fun(FoldIndexFun) ->
...
-    StartKey = to_object_key(Bucket, StartK4),
-    EndK2 = [{Field, Val} || {Field, _Type, Val} <- EndK],
-    EndK3 = riak_ql_ddl:make_key(Mod, LK, EndK2),
-    EndK4 = riak_kv_ts_util:encode_typeval_key(EndK3),
-    EndKey = to_object_key(Bucket, EndK4),
-    FoldFun = fun({K, V}, Acc) ->
-                      [{K, V} | Acc]
-              end,
+    LocalKeyLen = length(LKAST),
+    StartKey = key_prefix(Bucket, [element(3, E) || E <- StartK], LocalKeyLen),
```
jonmeredith: Is using `element(3,E)` a performance optimization?  The code would probably read better and be easier to understand the intent of changes if it used pattern matching on the key.

`key_prefix(Bucket, [Value || {_Name,_Type,Value} <- StartK], LocalKeyLen)`

src/riak_kv_eleveldb_backend.erl
```diff
@@ -964,21 +976,16 @@ orig_to_object_key(Bucket, Key) ->
...
 %%
-to_object_key({TableName, TableName}, {Family, Series, Timestamp}) ->
-    EncodedBucketType = % sext:encode(BucketType),
-        EncodedBucketName = sext:encode(TableName),
-    EncodedFamily = sext:encode(Family),
-    EncodedSeries = sext:encode(Series),
-    EncodedTimestamp = sext:encode(Timestamp),
+to_object_key({TableName, TableName}, LocalKey) when is_tuple(LocalKey) ->
+    EncodedBucketType = EncodedBucketName = sext:encode(TableName),
+    EncodedFamily = sext:encode(LocalKey),
```
jonmeredith: EncodedLocalKey is a more accurate name after refactor.

src/riak_kv_qry_compiler.erl
```diff
@@ -436,7 +443,24 @@ col_index_and_type_of(Fields, ColumnName) ->
             {Position, Type}
     end.
 
-expand_where(Where, #key_v1{ast = PAST}, MaxSubQueries) ->
+%%
+-spec expand_where(filter(), #key_v1{}, integer()) ->
+        [where_props()] | {error, any()}.
+expand_where(Where, PartitionKey, MaxSubQueries) ->
+    case find_quanta_function_in_key(PartitionKey) of
+        [{[QField], QSize, QUnit}] ->
```
jonmeredith: This is somewhat fragile to only permitting a single quantization in the partitioning - which is fair for now.  Generally we may want to allow multiple quantization function in the partition key, e.g. if you wanted to split a high volume ingest stream by sharding an eventid to spread the load, or possible if you multiple timestamps like a created/ingested and wanted to quantize on both.

Just a note, I don't think this requires a change to land the PR.

src/riak_kv_qry_compiler.erl
```diff
@@ -436,7 +443,24 @@ col_index_and_type_of(Fields, ColumnName) ->
...
+
+%%
+find_quanta_function_in_key(#key_v1{ ast = PKAST }) ->
+    [{X, Y, Z} || #hash_fn_v1{mod = riak_ql_quanta, fn = quantum,
+                              args = [#param_v1{name = X}, Y, Z]} <- PKAST].
+
+%%
+hash_timestamp_to_quanta(QField, QSize, QUnit, MaxSubQueries, Where) ->
     GetMaxMinFun = fun({startkey, List}, {_S, E}) ->
                            {element(3, lists:last(List)), E};
```
jonmeredith: Again, a little opaque on what's being dug out - it seems like it's the value of the last element of the composite key. Maybe a helper function would be useful for digging out name / type / value from the startkeys, otherwise you'd have to wrap in a begin/end block which would probably be worse. 

Your choice - at minimum a comment explaining we're digging out the min/max timestamps to compute the quanta on from the start key.

jonmeredith: I can see this is just a refactor of the original code - so understood this would just be refactoring/cleanup. Non-critical.

src/riak_kv_qry_compiler.erl
```diff
@@ -641,6 +663,7 @@ break_out_timeseries(Filters1, LocalFields1, [QuantumFields]) ->
         {_, {{_,_,{_,Starts}}, {_,_,{_,Ends}}}} when is_integer(Starts),
                                                      is_integer(Ends),
                                                      Starts > Ends ->
+            %% FIXME reliance on three element key!?
```
jonmeredith: Fixme still here - and it does look like it relies on the three element key structure.

src/riak_kv_qry_compiler.erl
```diff
@@ -652,13 +675,12 @@ break_out_timeseries(Filters1, LocalFields1, [QuantumFields]) ->
             error({lower_and_upper_bounds_are_equal_when_no_equals_operator,
                    ?E_TSMSG_LOWER_AND_UPPER_BOUNDS_ARE_EQUAL_WHEN_NO_EQUALS_OPERATOR});
         {Filters2, {Starts, Ends}} ->
-            %% remove the quanta from the local fields, this has alreadfy been
+            %% remove the quanta from the local fields, this has already been
             %% removed from the fields
-            [F1, F2, _] = LocalFields1,
-            LocalFields2 = [F1,F2],
+            % PartitionFields2 = lists:sublist(PartitionFields1, length(PartitionFields1)), % TODO DUUUUDE
```
jonmeredith: What's left `TODO DUUUDE` :)

src/riak_kv_qry_compiler.erl
```diff
@@ -1982,10 +2001,110 @@ compile_query_with_arithmetic_type_error_2_test() ->
...
+        "a1 SINT64 NOT NULL, "
+        "a TIMESTAMP NOT NULL, "
+        "b VARCHAR NOT NULL, "
+        "c VARCHAR NOT NULL, "
+        "d SINT64 NOT NULL, "
+        "PRIMARY KEY  ((a1, quantum(a, 15, 's')), a1, a, b, c, d))"),
+    {ok, Q} = get_query(
+          "SELECT * FROM tab4 WHERE a > 0 AND a < 1000 AND a1 = 1"),
+    ?assertMatch(
+        {ok, [#riak_select_v1{}]},
```
jonmeredith, line 2063: is the compiled query correct on start/end range and that kind of thing?

src/riak_kv_qry_compiler.erl
```diff
@@ -1982,10 +2001,110 @@ compile_query_with_arithmetic_type_error_2_test() ->
...
+        "a SINT64 NOT NULL, "
+        "b TIMESTAMP NOT NULL, "
+        "PRIMARY KEY  ((a,b), a,b))"),
+    ?assertEqual(
+        no_quanta,
+        quantum_field_name(DDL)
+    ).
+
+%% short key, partition and local keys are the same
+no_quantum_in_query_1_test() ->
```
jonmeredith, line 2090: Could you add a variant with a single partition key / local key without a quanta and one with?  If this is covered in a riak test or somewhere else, just say - I don't want to overly duplicate testing effort.

src/riak_kv_w1c_worker.erl
```diff
@@ -112,7 +112,7 @@ put(RObj0, Options) ->
                 Preflist :: term()) ->
                        {ok, {reference(), atom()}}.
 
-async_put(RObj, W, PW, Bucket, NVal, {_PK, LK}, EncodeFn, Preflist) ->
+async_put(RObj, W, PW, Bucket, NVal, {_PK, LK}, EncodeFn, Preflist) when is_tuple(LK) ->
```
jonmeredith, line 115: Why is the guard needed now?

src/riak_kv_qry_compiler.erl
andytill: Removing the restriction that the quantum has to be the last field in the key is still on the to do list, I'll clean it up at that point.

src/riak_kv_w1c_worker.erl
andytill, line 115: Otherwise a two element key falls into the `{PK,LK}` guard, will comment.

src/riak_kv_qry_compiler.erl
andytill: Realised afterwards that this is matching AST which is also three elements, added a test to be sure.

src/riak_kv_qry_compiler.erl
andytill: Haha ooops, removed.

src/riak_kv_eleveldb_backend.erl
andytill: Just a development change so I can trace these functions. Does it really improve performance?

jonmeredith: For the inlining, looks like it was added around 2I time - I doubt we have
any record of benchmarks.

src/riak_kv_qry_compiler.erl
jonmeredith: So it works up until the query point.

```
riak-shell(39)>CREATE TABLE t6 (ts timestamp NOT NULL, f VARCHAR NOT NULL, eventid SINT64 NOT NULL, PRIMARY KEY ((quantum(ts,1,h), f), ts, f, eventid));
riak-shell(41)>DESCRIBE t6;
+-------+---------+-------+-----------+---------+
|Column |  Type   |Is Null|Primary Key|Local Key|
+-------+---------+-------+-----------+---------+
|  ts   |timestamp| false |     1     |    1    |
|   f   | varchar | false |     2     |    2    |
|eventid| sint64  | false |           |    3    |
+-------+---------+-------+-----------+---------+
riak-shell(46)>SELECT * FROM t6 WHERE ts > 0 AND ts < 10 AND f = 'a';
Error (1001): missing_param: Missing parameter ts in where clause.
```

PR #1350, opened 2016-02-17
HTTP client API for TS

hmmr:
RTS-258

Implement these HTTP API calls (intending eventually to comply with https://docs.google.com/document/d/19cxGPEo7rQdjsC3jUw_40eiICwAN_OzPuT_F-eLXTt8):

```
GET       /ts/v1/table/Table          single-key get
DELETE    /ts/v1/table/Table          single-key delete
PUT       /ts/v1/table/Table          batch put
GET       /ts/v1/table/Table/keys     streaming list_keys
GET       /ts/v1/coverage             coverage for a query *
GET/POST  /ts/v1/query                execute an SQL query (SELECT or DESCRIBE with GET,
                                      CREATE TABLE with POST)
```

(*) commented out for now.

The keys, data and query for arguments to particular calls are to be submitted in request body, as in the real-world use case with curl: https://gist.github.com/hmmr/94abdb4d648a44009e93 and in r_t module https://github.com/basho/riak_test/pull/996.

The required TS extensions to riak-erlang-http-client are in https://github.com/basho/riak-erlang-http-client/pull/55. A tweak in riak-erlang-client for `riakc_ts:get/3` return value is in https://github.com/basho/riak-erlang-client/pull/261. (These are not dependencies of the present PR.)

The get/put/delete calls do not yet support specification of keys directly in URL, pending discussions in the draft on whether to accept keys as family/series and/or explicit identifiers.

### DIFF
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -56,7 +56,7 @@
 -define(E_BAD_QUERY,         1018).
 -define(E_TABLE_INACTIVE,    1019).
 -define(E_PARSE_ERROR,       1020).
--define(E_DELETE_NOTFOUND,   1021).
+-define(E_NOTFOUND,          1021).
 
 -define(FETCH_RETRIES, 10).  %% TODO make it configurable in tsqueryreq
 -define(TABLE_ACTIVATE_WAIT, 30). %% ditto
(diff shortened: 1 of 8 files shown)


lehoff:
This should not be merged back to riak_ts-develop-1.3 before the feature is complete. 
It is okay to merge from riak_ts-develop-1.3 to the feature branch.

Furthermore, it would be good to try out the proposed fork-and-pull-request model that we aim for in the new Dev Cycle process.

hmmr:
Need to resubmit with a different branch name.

PR #1354, opened 2016-02-23
Fix for negating an aggregation `-COUNT(*)` RTS-928

andytill:
The negation AST node was not properly being handled in the pre-scan for extracting aggregation functions, which allowed the function AST to slip through into the select compiler which threw a `function_clause`.


gordonguthrie:
+1 e2811ca

hazen:
@borshop merge

PR #1355, opened 2016-02-27
TS HTTP client API

hmmr:
RTS-258

Implement these HTTP API calls (intending eventually to comply with https://docs.google.com/document/d/19cxGPEo7rQdjsC3jUw_40eiICwAN_OzPuT_F-eLXTt8):

```
GET       /ts/v1/tables/Table/keys        single-key get
DELETE    /ts/v1/tables/Table/keys        single-key delete
POST      /ts/v1/tables/Table/keys        batch put
GET       /ts/v1/tables/Table/list_keys   streaming list_keys
GET       /ts/v1/coverage             coverage for a query *
GET/POST  /ts/v1/query                execute an SQL query (SELECT or DESCRIBE with GET,
                                      CREATE TABLE with POST)
```

(*) disabled for now.

The keys for `get` and `delete` are to be given on the URL, either as "/F1/V1/F2/V2/F3/V3", where key-constituent field-value pairs can be in arbitrary order, or as bare values: "/V1/V2/V3", with Vn value for the corresponding field Fn.

The data and query for `put` and `query` should be submitted in request body, as in the real-world use case with curl: https://gist.github.com/hmmr/94abdb4d648a44009e93 and in r_t module https://github.com/basho/riak_test/pull/1008.

The required TS extensions to riak-erlang-http-client are in https://github.com/basho/riak-erlang-http-client/pull/55. A tweak in riak-erlang-client for `riakc_ts:get/3` return value is in https://github.com/basho/riak-erlang-client/pull/261. (These are not dependencies of the present PR.)


hmmr:
Development switched to https://github.com/basho/riak_kv/pull/1382.

PR #1361, opened 2016-03-01
Added a version of update that takes a Callback

fadushin:
This callback gets called after the hashtree has been snapshotted, but before the hash tree update is initiated.  This callback is currently being used by the batching feature of Yokozuna, to minimize differences between the YZ and KV AAE hash trees.

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -48,6 +48,7 @@
          hash_index_data/1,
          hash_object/2,
          update/2,
+         update/3,
          start_exchange_remote/4,
          delete/2,
          async_delete/2,
@@ -70,6 +71,7 @@
 -type proplist() :: proplists:proplist().
 -type riak_object_t2b() :: binary().
 -type hashtree() :: hashtree:hashtree().
+-type update_callback() :: fun(() -> term()).
 
 -record(state, {index,
                 vnode_pid,
@@ -145,7 +147,14 @@ start_exchange_remote(FsmPid, From, IndexN, Tree) ->
 %% @doc Update all hashtrees managed by the provided index_hashtree pid.
 -spec update(index_n(), pid()) -> ok | not_responsible.
 update(Id, Tree) ->
-    gen_server:call(Tree, {update_tree, Id}, infinity).
+    update(Id, Tree, undefined).
+
+%% @doc Update all hashtrees managed by the provided index_hashtree pid.
+-spec update(index_n(), pid(), undefined | update_callback()) -> ok | not_responsible.
+update(Id, Tree, undefined) ->
+    gen_server:call(Tree, {update_tree, Id, undefined}, infinity);
+update(Id, Tree, Callback) when is_function(Callback) ->
+    gen_server:call(Tree, {update_tree, Id, Callback}, infinity).
 
 %% @doc Return a hash bucket from the tree identified by the given tree id
 %%      that is managed by the provided index_hashtree.
@@ -296,21 +305,15 @@ handle_call({delete, Items}, _From, State) ->
 handle_call(get_trees, _From, #state{trees=Trees}=State) ->
     {reply, Trees, State};
 
-handle_call({update_tree, Id}, From, State) ->
+handle_call({update_tree, Id, Callback}, From, State) ->
     lager:debug("Updating tree: (vnode)=~p (preflist)=~p", [State#state.index, Id]),
     apply_tree(Id,
-               fun(Tree) ->
-                       {SnapTree, Tree2} = hashtree:update_snapshot(Tree),
-                       Self = self(),
-                       spawn_link(
-                         fun() ->
-                                 _ = hashtree:update_perform(SnapTree),
-                                 gen_server:cast(Self, {updated, Id}),
-                                 gen_server:reply(From, ok)
-                         end),
-                       {noreply, Tree2}
-               end,
-               State);
+        fun(Tree) ->
+            NewTree = snapshot_and_async_update_tree(Tree, Id, From, Callback),
+            {noreply, NewTree}
+            end,
+        State
+    );
 
 handle_call({exchange_bucket, Id, Level, Bucket}, _From, State) ->
     apply_tree(Id,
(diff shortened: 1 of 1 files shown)


src/riak_kv_index_hashtree.erl
```diff
@@ -296,11 +302,12 @@ handle_call({delete, Items}, _From, State) ->
 handle_call(get_trees, _From, #state{trees=Trees}=State) ->
     {reply, Trees, State};
 
-handle_call({update_tree, Id}, From, State) ->
+handle_call({update_tree, Id, Callback}, From, State) ->
     lager:debug("Updating tree: (vnode)=~p (preflist)=~p", [State#state.index, Id]),
     apply_tree(Id,
                fun(Tree) ->
                        {SnapTree, Tree2} = hashtree:update_snapshot(Tree),
+                       catch Callback(),
```
JeetKunDoug: Wonder if perhaps we shouldn't move this inside the existing spawn_link() call below, so that the callback can't unnecessarily block the hashtree either?

zeeshanlakhani: +1 @JeetKunDoug

src/riak_kv_index_hashtree.erl
```diff
@@ -296,11 +302,12 @@ handle_call({delete, Items}, _From, State) ->
 handle_call(get_trees, _From, #state{trees=Trees}=State) ->
     {reply, Trees, State};
 
-handle_call({update_tree, Id}, From, State) ->
+handle_call({update_tree, Id, Callback}, From, State) ->
```
zeeshanlakhani:
```erlang
handle_call({update_tree, Id, undefind}, From, State) -> ...
```

For the one w/ the Callback.. add a guard -> is_function(Callback)

src/riak_kv_index_hashtree.erl
```diff
@@ -145,7 +146,12 @@ start_exchange_remote(FsmPid, From, IndexN, Tree) ->
 %% @doc Update all hashtrees managed by the provided index_hashtree pid.
 -spec update(index_n(), pid()) -> ok | not_responsible.
 update(Id, Tree) ->
-    gen_server:call(Tree, {update_tree, Id}, infinity).
+    update(Id, Tree, fun() -> ok end).
```
zeeshanlakhani: turn fun()-> ok end into `undefined`

src/riak_kv_index_hashtree.erl
zeeshanlakhani: once you move the Callback into the spawn, you'll want to `case` on `undefined`.

src/riak_kv_index_hashtree.erl
zeeshanlakhani, line 308: For the one w/ the Callback.. add a guard -> `is_function(Callback)` or == undefined.

fadushin, line 308: I think the guard needs to go on the API call.  In this particular case, there is a default/catch-all handle_call, which just returns ok.  But in this case I think we want the (API) call to fail if the type params don't match.

(I verified that if there were not a catch-all for handle_call, and we put a guard on the handle_call({update_re, ...), then the gen_server would crash, which is also probably what we don't want.)

zeeshanlakhani, line 308: @fadushin sure, makes sense for the API call. +1. think the diff just made it more confusing. you have the right call. Just make sure we have the `_` handling of update then to throw an `lager:error` in that case.

zeeshanlakhani:
@fadushin this looks good to me minus the dialyzer error it seems via bors... what's that about?. Since @JeetKunDoug commented earlier, let's see what he says... and then we'll get this merged. Good job.

JeetKunDoug:
@borshop: retry

src/riak_kv_index_hashtree.erl
```diff
@@ -1029,3 +1039,8 @@ maybe_get_vnode_lock(SrcPartition, Pid) ->
         false ->
             ok
     end.
+
+maybe_callback(undefined) ->
+    ok;
+maybe_callback(Callback) ->
+    Callback().
```
JeetKunDoug, line 1057: This should probably be `catch Callback().` unless we're expecting the consumer to be able to cancel the update by throwing (which would seem more appropriate to do with a return value from Callback() anyway).

zeeshanlakhani, line 1057: Yep, I thought Fred switched it for the reasoning you mentioned Doug... as it was a catch in the first commit. 

Zeeshan Lakhani
programmer | 
software engineer at @basho | 
org. member/founder of @papers_we_love | paperswelove.org
twitter => @zeeshanlakhani

src/riak_kv_index_hashtree.erl
```diff
@@ -296,7 +305,7 @@ handle_call({delete, Items}, _From, State) ->
 handle_call(get_trees, _From, #state{trees=Trees}=State) ->
     {reply, Trees, State};
 
-handle_call({update_tree, Id}, From, State) ->
+handle_call({update_tree, Id, Callback}, From, State) ->
     lager:debug("Updating tree: (vnode)=~p (preflist)=~p", [State#state.index, Id]),
     apply_tree(Id,
                fun(Tree) ->
```
zeeshanlakhani: after talking w/ @JeetKunDoug, it seems best to include a try/catch around maybe_callback below, but move this :deciduous_tree: fun into a separate named_fun.

src/riak_kv_index_hashtree.erl
zeeshanlakhani, line 1057: After discussion w/ @JeetKunDoug @fadushin... look at https://github.com/basho/riak_kv/pull/1361/files#r55219339 for cleanup and where to move the try/catch.

zeeshanlakhani:
I'm good w/ this. @JeetKunDoug? If so, then maybe rebase this into 1 commit @fadushin and then we'll start the +1 process.

JeetKunDoug:
Yup - I'm good. +1 once rebased.

JeetKunDoug:
:+1: b40628f

JeetKunDoug:
@borshop merge

PR #1363, opened 2016-03-01
Allow readrepair to ignore object size and sibling limits

bsparrow435:
Added a new readrepair element to the putargs record which will be set to true IIF an incoming put has the `rr` property in the put Options proplist. The `rr` option is only added in `riak_kv_vnode:readrepair` which is only called by the `riak_kv_get_fsm` if readrepair is in the list of final actions for the FSM. 

Additionally, a new function `actual_put/7` was added to allow for setting the MaxCheckFlag(`no_max_check`|`do_max_check`) based on elements from the putargs record. Handoff currently uses the MaxCheckFlag to bypass object limits but this could be refactored to be a bit more friendly such as just having a boolean MaxCheckFlag instead of the max_check atoms.


bsparrow435:
https://github.com/basho/riak_test/pull/1007

jadeallenx:
See associated test at basho/riak_test#1009

jadeallenx:
:+1: cdc9983

nickelization:
Since @mrallen1 +1'd and the build error is now fixed, I'm assuming it's fine if I go ahead and merge this.

nickelization:
@borshop merge

PR #1364, opened 2016-03-02
RTS-889: Add SQL INSERT Statement

hazen:
- Requires https://github.com/basho/riak_ql/pull/110
- Based largely on @jonmeredith's POC

### DIFF
--- a/include/riak_kv_ts.hrl
+++ b/include/riak_kv_ts.hrl
@@ -66,8 +66,8 @@
 -record(riak_sql_insert_v1,
         {
           'INSERT'      = <<>>  :: binary(),
-          fields                :: [identifier()],
-          values                :: [[value_type()]],
+          fields                :: [field_identifier()],
+          values                :: [[riak_ql_ddl:data_value()]],
           helper_mod            :: atom()
         }).
 
--- a/include/riak_kv_ts.hrl
+++ b/include/riak_kv_ts.hrl
@@ -26,7 +26,7 @@
 %% For dialyzer types
 -include_lib("riak_ql/include/riak_ql_ddl.hrl").
 
-%% the result type of a query, rows means to return all mataching rows, aggregate
+%% the result type of a query, rows means to return all matching rows, aggregate
 %% returns one row calculated from the result set for the query.
 -type select_result_type() :: rows | aggregate.
 
@@ -59,7 +59,14 @@
 
 -record(riak_sql_describe_v1,
         {
-          'DESCRIBE' = <<>>  :: binary()
+          'DESCRIBE'    = <<>>  :: binary()
+        }).
+
+-record(riak_sql_insert_v1,
+        {
+          'INSERT'      = <<>>  :: binary(),
+          fields                :: [{identifier, binary()}],
+          values                :: list(list(term()))
         }).
 
 -define(SQL_SELECT, #riak_select_v1).
--- a/include/riak_kv_ts.hrl
+++ b/include/riak_kv_ts.hrl
@@ -29,6 +29,7 @@
 %% the result type of a query, rows means to return all matching rows, aggregate
 %% returns one row calculated from the result set for the query.
 -type select_result_type() :: rows | aggregate.
+-type value_type() :: binary() | float() | integer() | boolean().
 
 -record(riak_sel_clause_v1,
         {
@@ -65,8 +66,9 @@
 -record(riak_sql_insert_v1,
         {
           'INSERT'      = <<>>  :: binary(),
-          fields                :: [{identifier, binary()}],
-          values                :: list(list(term()))
+          fields                :: [identifier()],
+          values                :: [[value_type()]],
+          helper_mod            :: atom()
         }).
 
 -define(SQL_SELECT, #riak_select_v1).
--- a/include/riak_kv_ts.hrl
+++ b/include/riak_kv_ts.hrl
@@ -29,13 +29,12 @@
 %% the result type of a query, rows means to return all matching rows, aggregate
 %% returns one row calculated from the result set for the query.
 -type select_result_type() :: rows | aggregate.
--type value_type() :: binary() | float() | integer() | boolean().
 
 -record(riak_sel_clause_v1,
         {
           calc_type        = rows :: select_result_type(),
           initial_state    = []   :: [any()],
-          col_return_types = []   :: [field_type()],
+          col_return_types = []   :: [riak_ql_ddl:field_type()],
           col_names        = []   :: [binary()],
           clause           = []   :: [riak_kv_qry_compiler:compiled_select()],
           finalisers       = []   :: [skip | function()]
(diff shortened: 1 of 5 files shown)


include/riak_kv_ts.hrl
```diff
@@ -59,7 +59,14 @@
...
         {
-          'DESCRIBE' = <<>>  :: binary()
+          'DESCRIBE'    = <<>>  :: binary()
+        }).
+
+-record(riak_sql_insert_v1,
+        {
+          'INSERT'      = <<>>  :: binary(),
+          fields                :: [{identifier, binary()}],
+          values                :: list(list(term()))
```
hmmr: There's a type for that: `riak_pb_ts_codec:ldbvalue()`, which would be appropriate for the inner `term()`.

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,108 @@ wait_until_active(Table, State, Seconds) ->
...
+%% INSERT statements, called from check_table_and_call.
+%%
+-spec make_insert_response(module(), #riak_sql_insert_v1{}) ->
+                           #tsqueryresp{} | #rpberrorresp{}.
+make_insert_response(Mod, #riak_sql_insert_v1{'INSERT' = Table, fields = Fields, values = Values}) ->
+    case lookup_field_positions(Mod, Fields) of
+	{error, FieldReason} ->
+	    make_rpberrresp(?E_BAD_QUERY, FieldReason);
+	{ok, Positions} ->
+        Empty = make_empty_row(Mod),
```
hmmr, line 273: Indentation.

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,108 @@ wait_until_active(Table, State, Seconds) ->
...
+        case xlate_insert_to_putdata(Values, Positions, Empty) of
+            {error, ValueReason} ->
+                make_rpberrresp(?E_BAD_QUERY, ValueReason);
+            {ok, Data} ->
+                Response = sub_putreq_common(Mod, Table, Data, #state{}),
```
hmmr: You are producing a default State for `sub_putreq_common`, which happens to be OK because there's no useful fields in `#state{}` for the put request, only to be discarded in `xlate_put_to_qry_resp`. I would suggest simply copying the outer case statement from `sub_putreq_common` right here, or perhaps as a separate local helper function, which would produce proper `#tsqueryresp{}` instead of `#tsputresp{}`.

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,108 @@ wait_until_active(Table, State, Seconds) ->
...
+    Err.
+
+
+%%
+%% Return an all-null empty row ready to be populated by the values
+%%
+-spec make_empty_row(module()) -> tuple(undefined).
+make_empty_row(Mod) ->
+    Positions = Mod:get_field_positions(),
+    list_to_tuple(lists:duplicate(length(Positions), undefined)).
```
hmmr, line 303: Just to make sure, is `undefined` the proper value for NULL?

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,108 @@ wait_until_active(Table, State, Seconds) ->
...
+                   {Good, [flat_format("undefined field ~s", [FieldName]) | Bad]};
+               true ->
+                   {[Mod:get_field_position(FieldName) | Good], Bad}
+		   end
+	   end, {[], []}, FieldIdentifiers)
+    of
+	{Positions, []} ->
+	    {ok, lists:reverse(Positions)};
+	{_, Errors} ->
+	    %% Only returns the first error, could investigate returning multiple.
```
hmmr: I suggest collecting just the bad fields in the foldl, and then formatting those for error message in a single `flat_format("undefined fields: ~s", [string:join(Fields, ", ")])`.

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,108 @@ wait_until_active(Table, State, Seconds) ->
...
+                                               [Reason, RowNum]),
+                         {Good, [Reason1 | Bad], RowNum + 1}
+                 end
+             end,
+    Converted = lists:foldl(ConvFn, {[], [], 1}, Values),
+    case Converted of
+        {PutData, [], _} ->
+            {ok, lists:reverse(PutData)};
+        {_, Errors, _} ->
+            %% Only returns the first error, could investigate returning multiple.
```
hmmr: Same suggestion as for reporting undefined fields, above.

include/riak_kv_ts.hrl
hazen: Really? Don't we break out `integer()` and `float()`, for example?  We also support lists of lists of `list()`?

``` erlang
-type ldbvalue() :: binary() | number() | boolean() | list().
```

src/riak_kv_pb_timeseries.erl
hazen: I thought about that.  Yes, that does sound cleaner.

include/riak_kv_ts.hrl
hmmr: Interesting! Last time I looked, there was no `list()` in `ldbvalue()`. That type was specifically meant to represent all types valid for passing to/from eleveldb, and has been used as such everywhere since 1.0.

As for `number()` standing for `integer()` and `float()`, it's perfectly fine for the purpose of type checking by dialyzer, and that's all that matters here.

src/riak_kv_pb_timeseries.erl
hmmr: .. or modify `sub_putreq_common` to serve the needs of all three callers, as you see fit.

hazen:
@hmmr If you have some time to investigate:

```
riak_kv_pb_timeseries.erl:280: The pattern {'error', FieldReason} can never match the type {'ok',[]}
riak_kv_pb_timeseries.erl:366: The pattern <[{_Type, Val} | Values], [Pos | Positions], Row> can never match the type <_,[],{'undefined'}>
```

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,112 @@ wait_until_active(Table, State, Seconds) ->
...
+	   fun({identifier, FieldName}, {Good, Bad}) ->
+		   case Mod:is_field_valid(FieldName) of
+               false ->
+                   {Good, [FieldName | Bad]};
+               true ->
+                   {[Mod:get_field_position(FieldName) | Good], Bad}
+		   end
+	   end, {[], []}, FieldIdentifiers)
+    of
+        {Positions, []} ->
```
hmmr, line 323: Adding a (seemingly redundant) guard here did help: `when length(Positions) > 0`.

src/riak_kv_pb_timeseries.erl
```diff
@@ -255,6 +261,112 @@ wait_until_active(Table, State, Seconds) ->
...
+        {PutData, [], _} ->
+            {ok, lists:reverse(PutData)};
+        {_, Errors, _} ->
+            {error, flat_format("too many values in row index(es) ~s",
+                                [string:join(lists:reverse(Errors), ", ")])}
+    end.
+
+-spec make_insert_row([] | [riak_ql_ddl:data_value()], [] | [pos_integer()], tuple()) ->
+                      {ok, tuple()} | {error, string()}.
+make_insert_row([], _Positions, Row) when is_tuple(Row) ->
```
hmmr: Still working out the problem here...

src/riak_kv_pb_timeseries.erl
andytill, line 323: This requires an iteration of `Positions`, this is probably a small list but `Positions /= []` doesn't require the iteration.

src/riak_kv_pb_timeseries.erl
hmmr, line 358: Seems to be the effect of a butterfly flapping wings in riak_ql_ddl.hrl.

gordonguthrie:
+1 dbc872c

PR #1366, opened 2016-03-04
Finish TS batching

macintux:
This PR completes the incremental extension of batch uploads of data all the way through to the target vnodes. Previously, data was batched until it was divvied up into single records to be pushed via the `w1c_worker` processes.

Corresponding `eleveldb` and `riak_test` PRs to follow. Still waiting on final benchmarking numbers from @BrianMMcClain but an earlier version of this set of changes with a max batch size of 200 records showed a dramatic performance boost.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -155,6 +155,10 @@ start(_Type, _StartArgs) ->
                                           [true, false],
                                           false),
 
+            riak_core_capability:register({riak_kv, w1c_batch_vnode},
+                                          [true, false],
+                                          false),
+
             %% mapred_system should remain until no nodes still exist
             %% that would propose 'legacy' as the default choice
             riak_core_capability:register({riak_kv, mapred_system},
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -393,6 +393,8 @@ estimated_row_count(SampleRowSize, MaxBatchSize) ->
     RowSizeFudged = (SampleRowSize * 10) div 9,
     MaxBatchSize div RowSizeFudged.
 
+create_batches(Rows, MaxSize) when length(Rows) =< MaxSize ->
+    [Rows];
 %% May be a more efficient way to do this. Take a list of arbitrary
 %% data (expected to be a list of lists for this use case) and create
 %% a list of MaxSize lists.
@@ -989,4 +991,15 @@ batch_3_test() ->
     ?assertEqual([[1, 2, 3], [4, 5, 6], [7, 8, 9]],
                  create_batches([1, 2, 3, 4, 5, 6, 7, 8, 9], 3)).
 
+batch_undersized1_test() ->
+    ?assertEqual([[1, 2, 3, 4, 5, 6]],
+                 create_batches([1, 2, 3, 4, 5, 6], 6)).
+
+batch_undersized2_test() ->
+    ?assertEqual([[1, 2, 3, 4, 5, 6]],
+                 create_batches([1, 2, 3, 4, 5, 6], 7)).
+
+batch_almost_undersized_test() ->
+    ?assertEqual([[1, 2, 3, 4, 5], [6]],
+                 create_batches([1, 2, 3, 4, 5, 6], 5)).
 -endif.
--- a/src/riak_kv_pb_timeseries.erl
+++ b/src/riak_kv_pb_timeseries.erl
@@ -63,9 +63,6 @@
 -define(FETCH_RETRIES, 10).  %% TODO make it configurable in tsqueryreq
 -define(TABLE_ACTIVATE_WAIT, 30). %% ditto
 
--define(MIN_PUT_BATCH_SIZE, 10).  %% TODO make it configurable somewhere
--define(MAX_PUT_BATCH_SIZE, 200). %% ditto
-
 -record(state, {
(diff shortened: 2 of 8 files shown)


src/riak_kv_pb_timeseries.erl
```diff
@@ -362,6 +374,41 @@ row_to_key(Row, DDL, Mod) ->
...
+create_batches(Rows, MaxSize) when length(Rows) =< MaxSize ->
+    [Rows];
+%% May be a more efficient way to do this. Take a list of arbitrary
+%% data (expected to be a list of lists for this use case) and create
+%% a list of MaxSize lists.
+create_batches(Rows, MaxSize) ->
+    create_batches(Rows, MaxSize, MaxSize, [], []).
+
+create_batches([], _Counter, _Max, ThisBatch, AllBatches) ->
+    AllBatches ++ [ThisBatch];
```
hmmr: Does the order of batch groups matter here? I wonder if you could reassemble list elements with a little more efficient `[ThisElement|Rest]`.

macintux: My expectation is that this will be invoked rarely, and I thought the VM had optimized this idiom pretty well, but no, I don't think order matters. Even if all batches are retained in the correct order, there's no guarantee they'll arrive at the remote vnode (and hence the backend) in that same order.

Willing to change it, unconvinced it's worth doing.

macintux:
More changes resulting from @hmmr's question are pending; branch has been benchmarked, need to merge it into this branch

macintux:
This is ready for final review. Requires the eleveldb branch from basho/eleveldb#183 (RIAK-1406) (RIAK-1520) (RIAK-1919) for compilation/dialyzation.

src/riak_kv_app.erl
```diff
@@ -155,6 +155,10 @@ start(_Type, _StartArgs) ->
                                           [true, false],
                                           false),
 
+            riak_core_capability:register({riak_kv, w1c_batch_vnode},
```
hmmr, line 158: If the batching of operations is transparent to the callers of riak_kv_ts_api:put_data (as I understand it is), then it is inconsequential whether a node has this capability or not. What are the circumstances where you expect this capability to be queried?

macintux, line 158: Rolling upgrades. If a batch is sent to a 1.3 node, it must not deliver batches to 1.2 nodes in the cluster.

hmmr, line 158: Ok, got it. And, I should have scrolled down to see the capability check.

src/riak_kv_pb_timeseries.erl
```diff
@@ -316,21 +321,28 @@ put_data(Data, Table, Mod) when is_binary(Table) ->
...
-
-                                        {ok, ReqId} =
-                                            riak_kv_w1c_worker:async_put(
-                                              RObj, W, PW, Bucket, NVal, LK,
-                                              EncodeFn, Preflist),
-                                        {[ReqId | PartReqIds], PartErrors}
-                                end,
-                                {[], 0}, Records),
-                          {GlobalReqIds ++ Ids, GlobalErrorsCnt + Errs};
+                          DataForVnode = pick_batch_option(SendFullBatches,
```
hmmr, line 453: That's toodamnhigh an indentation. Can you do something to reduce the vast white space on the left?

src/riak_kv_vnode.erl
```diff
@@ -840,6 +842,24 @@ handle_command({get_index_entries, Opts},
...
+%% For now, ignore async_put
+handle_command(?KV_W1C_BATCH_PUT_REQ{objs=Objs, type=Type},
+                From, State=#state{mod=Mod, idx=Idx, modstate=ModState}) ->
+    StartTS = os:timestamp(),
+    Context = {w1c_batch_put, From, Type, Objs, StartTS},
+    case Mod:batch_put(Context, Objs, [], ModState) of
+        {ok, UpModState} ->
+            lists:foreach(
+              fun({{Bucket, Key}, EncodedVal}) ->
+                      update_hashtree(Bucket, Key, EncodedVal, State),
```
hmmr: It's not safe to skip on intermediate updates and only call `update_hashtree` once, I assume?

macintux: I'll double check thanks

macintux: It does look like we can optimize away at least half of the function calls that originate here. Good catch!

src/riak_kv_pb_timeseries.erl
macintux, line 453: Regrettably, I can't find a good way to fix this. There is a massive number of bindings in this function which have to be curried around, so neither the outer anonymous function for the `foldl` nor the inner anonymous functions for the `invoke_async_put` invocation can be extracted into named functions without creating a different kind of monster.

Ideas greatly encouraged, because I definitely don't like the way this turned out, speaking as a fan of one-line functions.

macintux:
It appears `riak_ql` has changed underneath me, this no longer builds properly because `riak_ql_ddl.hrl` no longer defines the types that `pb_timeseries` needs.

macintux:
Fixed. Fingers crossed.

hmmr:
@borshop: retry

hazen:
:+1: f5820fa

hazen:
@borshop merge

PR #1367, opened 2016-03-07
Output some query debug information

andytill:
Utility functions for printing basic debug information about queries to the shell.

To use, attach a riak shell and call `riak_kv_ts_util:explain_query_print/1` with a select query. The target table must exist and be activated.

Example output:

``` erlang
(dev1@127.0.0.1)36> riak_kv_ts_util:explain_query_print("SELECT * FROM table2 WHERE a = 1 AND b = 1 AND c >= 1999 AND c <= 5000 AND d = 'lol' AND e > 10 AND d = true").
SUB QUERY 1
>  1/1/1999/ FILTER d = 'lol' AND e > 10 AND d = true
<  1/1/2000/
SUB QUERY 2
>  1/1/2000/ FILTER d = 'lol' AND e > 10 AND d = true
<  1/1/3000/
SUB QUERY 3
>  1/1/3000/ FILTER d = 'lol' AND e > 10 AND d = true
<  1/1/4000/
SUB QUERY 4
>  1/1/4000/ FILTER d = 'lol' AND e > 10 AND d = true
<  1/1/5000/
SUB QUERY 5
>  1/1/5000/ FILTER d = 'lol' AND e > 10 AND d = true
<  1/1/5000/
```

### DIFF
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -238,6 +238,7 @@ make_ts_keys(CompoundKey, DDL = #ddl_v1{local_key = #key_v1{ast = LKParams},
 encode_typeval_key(TypeVals) ->
     list_to_tuple([Val || {_Type, Val} <- TypeVals]).
 
+%% Print the query explanation to the shell.
 explain_query_print(QueryString) ->
     explain_query_print2(1, explain_query(QueryString)).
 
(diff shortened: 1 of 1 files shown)


hmmr:
I'd rather see it somewhere in riak_shell proper. Or if this needs to be in riak_kv_ts_util, it could decompose the WHERE parts into some abstract records and return that (for its callers in riak_shell to do the eventual io:format).

hazen:
@hmmr I could see this be useful in other contexts as well and perhaps even introduce a QUERY PLAN-like directive

hmmr:
.. which is the other alternative I was suggesting: make it break down the WHERE parts and let the callers print it out directly (as riak_shell would) or process it further for other purposes.

andytill:
I was thinking along the lines of an SQL `EXPLAIN` command. This could be extended to be an API with some riak_pb support.

For now I just need it for the dev work I'm doing though and can be deleted when proper support is available.

hazen:
:+1: 57848bd

hazen:
@borshop merge

PR #1369, opened 2016-03-11
bucket type properties in CREATE WITH clause

hmmr:
RTS-865

This PR is a refurbished #1328, obsoletes #1332. Requires https://github.com/basho/riak_ql/pull/113.

KV support for setting custom table properties in an optional WITH clause of the CREATE TABLE query. This allows clients to create TS tables with custom properties via query API call.

There is also a more strict and accurate validation of property names accepted in WITH clause (as well as in json supplied to `riak-admin bucket-type create` command. Now an attempt to create a bucket type with an unknown property will result in appropriate error message.

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_kv_bucket: bucket validation functions
 %%
-%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2016 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -24,6 +24,7 @@
 -module(riak_kv_bucket).
 
 -export([validate/4]).
+-export([is_valid_property/1]).
 
 -include("riak_kv_types.hrl").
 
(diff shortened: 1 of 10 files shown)


src/riak_kv_bucket.erl
```diff
@@ -43,6 +44,43 @@
 
 -export_type([props/0]).
 
+-define(VALID_PROPERTIES,
```
andytill: Some of these properties cannot be used by Time Series, or are not supported by write once.

Only `dw`, `n_val`, `pr`, `pw`, `r`, `rw` and `w` should be allowed.

src/riak_kv_bucket.erl
```diff
@@ -43,6 +44,43 @@
...
+         "small_vclock",
+         "w",
+         "write_once",
+         "young_vclock",
+         %% TS-specific ones:
+         "ddl",
+         "table_def"
+        ]).
+
+-spec is_valid_property(string() | binary()) -> boolean().
```
macintux: Tough call. We've not constrained bucket properties in the past, and it's unclear this is a good reason to start doing so.

On the other hand, we've certainly wanted to document all possible bucket properties to get out of the JSON-on-the-CLI business for riak-admin bucket-type create, so this work may prove valuable even if we choose not to use it.

macintux: Upon further reflection, will have to reject this, at least as a global change to `erlify_bucket_props`. Buckets are a riak_core concept that applications can use even if not written by Basho, hence we don't and can't have a whitelist of properties.

hmmr: Quite simply, there's a list_to_existing_atom down the line, and here's why
I attempt to collect all possible values for the bucket properties here.

src/riak_kv_ts_util.erl
```diff
@@ -186,26 +191,36 @@ maybe_parse_table_def(BucketType, Props) ->
...
     ok;
-assert_type_and_table_name_same(BucketType1, #ddl_v1{table = BucketType2}) ->
+assert_type_and_table_name_same(BucketType1, ?DDL{table = BucketType2}) ->
     throw({error,
            {table_name,
-            "The bucket type and table name must be the same\n"
-            "    bucket type was: " ++ binary_to_list(BucketType1) ++ "\n"
-            "     table name was: " ++ binary_to_list(BucketType2) ++ "\n"}}).
+            flat_format(
+              "Time series bucket type and table name mismatch (~s != ~s)",
```
andytill: I prefer the original message, it is very easy to compare vertical columns.

If this one is here to stay it should read, Time series table and bucket type do not match.

src/riak_kv_bucket.erl
macintux: Understood, but we still need to remove all of this. Catch the badarg if you need to.

hmmr:
@borshop: retry

macintux:
One last bit I'd like to see before this is merged: I don't see any tests to provoke the new error tuples from `erlify_bucket_prop`: `bad_linkfun_modfun`, `bad_linkfun_key`, and `bad_chash_keyfun`. A unit test for these would be lovely.

macintux:
:+1: df1862e

hmmr:
@borshop: merge

macintux:
@borshop merge

PR #1370, opened 2016-03-11
Wrap the call to update_index_info in global:trans [JIRA: RIAK-2441]

nickelization:
This fixes a race condition that was causing stats corruption if two
processes tried to update the data for the same index at the same time.

As detailed in the code comments, this may not have the best possible
performance characteristics, but it seems to be fine in practice and is a
very easy workaround for the problems we found.

For future reference, we originally found and reproduced this bug using
yz_aae_test. It didn't always trigger the race, but adding a short
timer:sleep call in between the lookup and insert calls made it very
easy to catch.

### DIFF
--- a/src/riak_kv_entropy_info.erl
+++ b/src/riak_kv_entropy_info.erl
@@ -174,12 +174,10 @@ update_index_info_impl(Key, Cmd) ->
     %% contention if lots of different pieces of code are all relying on
     %% it at once.
     %%
-    %% If we do ever want to remove the call to global:trans and replace
-    %% it with something better, I think we could come up with some code
-    %% to allow concurrent updates by writing multiple versions of the
-    %% record to ETS, and then we could detect that on read and resolve
-    %% the conflict based on how we want the stats to be calculated. This
-    %% would be trickier to implement, but should yield great performance.
+    %% If we find that the call to global:trans is a bottleneck, we can
+    %% explore other options (probably either synchronizing updates through
+    %% a pool of workers, or maybe some sort of CRDT-style conflict
+    %% resolution mechanism).
     Info = case ets:lookup(?ETS, {index, Key}) of
                [] ->
                    #index_info{};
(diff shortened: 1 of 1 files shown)


src/riak_kv_entropy_info.erl
```diff
@@ -158,6 +158,28 @@ exchanges(Index, IndexN) ->
...
+    %% We are fixing it by wrapping this code in a call to global:trans,
+    %% which feels like a quick and dirty fix, but works great and seems
+    %% to have fine enough performance characteristics from our tests.
+    %% We'll need to look out for any additional future use of the global
+    %% module though, since all calls to set_lock/trans on a given node
+    %% use the same ETS table behind the scenes, which could lead to
+    %% contention if lots of different pieces of code are all relying on
+    %% it at once.
+    %%
+    %% If we do ever want to remove the call to global:trans and replace
```
JeetKunDoug: I'd rewrite the "If we ever do" part as simply:

```
If we find that global:trans/3 is a bottleneck, we should explore other options.
```

And leave it at that. I'm still not sure a CRDT-style conflict resolution system is necessary for this - if nothing, the "put a process in front of the updates" approach, or if it's too slow, a pool of processes, is a simpler approach than attempting to do conflict resolution here.

Otherwise, +1

nickelization: @JeetKunDoug I see your point, but I did want to leave some reference to a conflict resolution approach, rather than just burying the idea. With all the focus on optimization lately, it seems like it's worth preserving an idea that could potentially be faster than using a worker pool, particularly in an area that could become a bottleneck. I'll definitely tone down the comment a bit, but I'd still like to leave a passing mention to the idea in case it turns out to be a good approach in the future. Deal?

JeetKunDoug:
Have discussed this with both Torben and Jon - unless our performance testing finds an issue, will move forward with this.

JeetKunDoug:
:+1: 7058ad7

JeetKunDoug:
@borshop merge

JeetKunDoug:
create jira issue

Basho-JIRA:
Merged to 2.0 - needs forward merge 

**_[posted via JIRA by Douglas Rohrer]**_

Basho-JIRA:
Forward merge complete. Marking as done. 

**_[posted via JIRA by Nick Marino]**_

PR #1371, opened 2016-03-14
Feature th http client for RTS-258

lehoff:
Re-worked to use Webmachine better.

Open issues:
- The query API only accepts POST and uses a ?query="..." query string in the URL to pass in the query. This is to make it consistent across HTTP methods if we want to match the side effects of the query with the HTTP method.
- The output from list_keys is some hrefs inside a `<html>` block. Perhaps we should change the design and provide something more useful. It is not of much use when doing a curl or http (httpie.org) request on the command line.
- Dialzyer is not fully happy about the listkeys module, but it smells like a Dialyzer bug - see comment in that file.
- The commits have not been squashed - open to doing that before doing the final PR.


src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,375 @@
...
+resource_exists(Path, 'DELETE', RD, #ctx{mod=Mod}=Ctx) ->
+    %% Since reading the object is expensive we will assume for now that the
+    %% object exists for a delete, but if it turns out that it does not then the
+    %% processing of the delete will return 404 at that point.
+    try
+        Key = validate_key(Path, Mod),
+        {true, RD, Ctx#ctx{key=Key}}
+    catch
+        _:Reason ->
+            Resp = riak_kv_wm_ts_util:set_error_message("lookup on ~p failed due to ~p",
```
hmmr: Using `handle_error` here (and elsewhere in this module) would be logical, instead of custom setting of response body.

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,375 @@
...
+     try riak_kv_ts_api:delete_data(Key, Table, Mod, Options) of
+        ok ->
+             Json = result_to_json(ok),
+              Resp = riak_kv_wm_ts_util:set_json_response(Json, RD),
+              {true, Resp, Ctx};
+         {error, notfound} ->
+              {{halt, 404}, RD, Ctx}
+     catch
+         _:Reason ->
+             lager:log(info, self(), "delete_resource failed: ~p", [Reason]),
```
hmmr: A `lager:debug` would do (or perhaps no logging at all, as I guess this is a remnant of your debugging sessions).

There are many other occurrences of lager:log in this PR that could be cleaned up.

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,375 @@
...
+        Ctx#ctx{timeout = Timeout,
+                options = [{timeout, Timeout}]}
+    catch
+        _:_ ->
+            throw(riak_kv_wm_ts_util:flat_format("timeout not an integer value: ~s", [TimeoutStr]))
     end;
-validate_ts_records(_) ->
-    undefined.
+extract_params(Params, _Ctx) ->
+    throw(riak_kv_wm_ts_util:flat_format("incorrect paramters: ~p", [Params])).
```
hmmr: `flat_format` is too generic and too simple to be placed in an external module and exported.

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,375 @@
...
-    case riak_core_security:check_permission(
-           {riak_kv_ts_util:api_call_to_perm(Call), Table}, Security) of
-        {false, Error, _} ->
-            handle_error(
-              {not_permitted, utf8_to_binary(Error)}, RD, Ctx);
+        GoodValue when GoodValue > 0 ->
+            GoodValue;
         _ ->
-            {true, RD, Ctx}
+            throw(url_key_bad_value)
```
hmmr: Up the stack under the catch in `resource_exists`, you are throwing both a string (in `match_path`), and, here, an atom.

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,375 @@
...
-            prepare_data_in_body(RD, Ctx#ctx{result = ok});
-        {error, {bad_key_length, Got, Need}} ->
-            handle_error({key_element_count_mismatch, Got, Need}, RD, Ctx);
-        {error, notfound} ->
-            handle_error(notfound, RD, Ctx);
-        {error, Reason} ->
-            handle_error({riak_error, Reason}, RD, Ctx)
-    end.
+%% @todo: might be better if the DDL helper module had a
+%% valid_field_value(Field, Value) -> boolean() function.
```
hmmr, line 389: There is one, called `Mod:validate_obj/1`. It operates on an entire record, though, but you can still consider using it.

src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -77,164 +74,122 @@ service_available(RD, Ctx = #ctx{riak = RiakProps}) ->
...
 
 -spec content_types_provided(#wm_reqdata{}, #ctx{}) ->
-                                    {[{ContentType::string(), Producer::atom()}],
-                                     #wm_reqdata{}, #ctx{}}.
+                                    cb_rv_spec([{ContentType::string(), Producer::atom()}]).
 %% @doc List the content types available for representing this resource.
 content_types_provided(RD, Ctx) ->
-    {[{"application/json", produce_doc_body}], RD, Ctx}.
-
+      {[{"text/html", produce_doc_body}], RD, Ctx}.
```
hmmr: Is "text/html" suitable here? Streamed data are supposed to be consumed by some machinery and not rendered in a browser.

src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -77,164 +74,122 @@ service_available(RD, Ctx = #ctx{riak = RiakProps}) ->
...
-                {ok, ReqId} = riak_client:stream_list_keys(
-                                {Table, Table}, undefined, Client),
-                stream_keys(ReqId)
-        end,
-    {{stream, {<<>>, F}}, RD, Ctx}.
-
-stream_keys(ReqId) ->
+    {ok, ReqId} = riak_client:stream_list_keys(
+                    {Table, Table}, undefined, Client),
+    lager:log(info, self(), "in produce_doc_body ~p", [Table]),
```
hmmr: Don't forget to remove these debugging printouts.

src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -77,164 +74,122 @@ service_available(RD, Ctx = #ctx{riak = RiakProps}) ->
...
+    {{halt, 200}, wrq:set_resp_body({stream, prepare_stream(ReqId, Table, Mod)}, RD), Ctx}.
+
+prepare_stream(ReqId, Table, Mod) ->
+    {<<"<html>">>, fun() -> stream_keys(ReqId, Table, Mod) end}.
```
hmmr: Embedding json in html is, aside of questions regarding its purposefulness, requires proper escaping of random characters that can be interpreted as html tags.

src/riak_kv_wm_timeseries_query.erl
```diff
@@ -54,23 +54,25 @@
...
+          compiled_query :: undefined | #ddl_v1{} | #riak_sql_describe_v1{} | #riak_select_v1{},
+          result         :: undefined | ok | {Headers::[binary()], Rows::[ts_rec()]} |
+                            [{entry, proplists:proplist()}]
+         }).
 
 -define(DEFAULT_TIMEOUT, 60000).
 -define(TABLE_ACTIVATE_WAIT, 30).   %% wait until table's bucket type is activated
 
--define(CB_RV_SPEC, {boolean()|atom()|tuple(), #wm_reqdata{}, #ctx{}}).
+-type cb_rv_spec(T) :: {T, #wm_reqdata{}, #ctx{}}.
```
hmmr, line 74: These types could also be moved to riak_kv_wm_ts_utils.

src/riak_kv_wm_timeseries_query.erl
```diff
@@ -126,371 +144,171 @@ forbidden(RD, Ctx) ->
...
+        {ddl, _ } = Res ->
+            Res;
+        {Type, Compiled} when Type==select; Type==describe ->
+            {ok, SQL} =  riak_kv_ts_util:build_sql_record(
+                           Type, Compiled, undefined),
+            {Type, SQL}
     end.
 
+%% @todo: should really be in riak_ql somewhere
+table_from_sql(#ddl_v1{table=Table})                    -> Table;
```
hmmr, line 291: Consider using (and extending) https://github.com/basho/riak_kv/blob/feature-az-http_client/src/riak_kv_ts_util.erl#L129.

src/riak_kv_wm_ts_util.erl
```diff
@@ -0,0 +1,121 @@
...
+table_from_request(RD) ->
+    utf8_to_binary(
+      mochiweb_util:unquote(
+        wrq:path_info(table, RD))).
+
+%% move to util module.
+utf8_to_binary(S) ->
+    unicode:characters_to_binary(S, utf8, utf8).
+
+flat_format(Format, Args) ->
```
hmmr, line 55: This one-liner is too simple and too generic for an exported function. And, by placing it in riak_kv_wm_ts_util` you encourage callers in many other places in riak to call it from this module (otherwise, there is little sense in exporting it).

hmmr:
General notes:
1. The coverage request is missing.
2. An r_t module is needed to see it in operation.

jonmeredith:
Raw notes from our discussion...

```
riak_kv_wm_timeseries_query
     - why bother with prefix, it is unused.
     - why bother with riak, it is unused.
   - may as well just do C = riak_client:new(node(), undefined).

service_available
     - KV endpoints always use the the local riak_client.  That can coordinate with the cluster even if the local KV/TS service is down until shutdown.
     - if you wanted to be pedantic you could use init:get_status() and check in started state.

malformed_request
    - if a new statement type is added, it will throw a non-match in the case statement.

content_types_accepted
   - what happens if you supply a content-type in the query - is it ignored?

resource_exists
  - on create, no differentiation between exists/does not exist

process_post
  - would have expected to look in the request body.

checkpoint
     - what's the value of this function versus a -define(checkpoint, lager:info), with the current system the module and location information is not picked up by the lager parse transform.  You would get more data by directly calling lager:info() or using the define above if you wanted to segregate difference message types.
   - is the intent to make this at debug level eventually? At the moment this seems way to chatty for an informational message in the main logs - service_available and malformed_request will be called for each HTTP request.

Finally got this to work.
curl --trace-ascii - -X POST 'http://localhost:8098/ts/1/query?query=CREATE%20TABLE%20t%20%28f%20VARCHAR%20NOT%20NULL%2C%20s%20VARCHAR%20NOT%20NULL%2C%20ts%20TIMESTAMP%20NOT%20NULL%2C%20PRIMARY%20KEY%20%28%28f%2Cs%2Cquantum%28ts%2C1%2Cd%29%29%2Cf%2Cs%2Cts%29%29'

which you can do as
 curl -G -X POST http://localhost:8098/ts/1/query --trace-ascii - --data-urlencode "query=SELECT * FROM t WHERE f='f' AND s='s' AND ts > 0 AND ts < 10"
```

src/riak_kv_wm_timeseries.erl
lehoff: I think it is better to keep the error close to where it is happening. @jonmeredith - what do you think?

src/riak_kv_wm_timeseries.erl
lehoff: The lager things should have been removed before doing the PR - I will remove them all before asking for a re-review.

src/riak_kv_wm_timeseries.erl
lehoff: A utility function used several places should be in its own module. @jonmeredith - what are your thoughts?

I will try to avoid using it here as it might be good enough to use an io_lib:format/2.

src/riak_kv_wm_timeseries.erl
lehoff: Good point - I will make the throws consistent.

src/riak_kv_wm_timeseries.erl
lehoff, line 389: Good hint, still not sure what works best, so I will re-visit this when it is time to work on the DDL helper modules with merl.

src/riak_kv_wm_timeseries_listkeys.erl
lehoff: Good question. Not sure what the exact use case is.
@jonmeredith can you clarify what we want out of this?

Perhaps we only want to use streaming on the inside and collect all keys before returning the response?  This is what the code does now.

src/riak_kv_wm_timeseries_listkeys.erl
lehoff: There will be a total purge of all of them! Thx!

src/riak_kv_wm_timeseries_listkeys.erl
lehoff: Discussed return format with @jonmeredith - we will encode everything as JSON. Which makes this inconsistency go away.

src/riak_kv_wm_timeseries_query.erl
lehoff, line 74: Yes and no.

If we move it, it has to be

``` Erlang
-type cb_rv_spec(T, C) :: {T, #wm_reqdat{}, C)
```

since the `#ctx{}` record is different in the modules.

Given than I am leaning towards the local definition.

src/riak_kv_wm_timeseries_query.erl
lehoff, line 291: I think we need to move this into riak_ql somewhere - it seems like something that should be contained inside riak_ql.

src/riak_kv_wm_ts_util.erl
lehoff, line 55: I don't think "too simple and too generic" is a reason to exclude a function from being exported.

It is ment for `riak_kv_wm_timeseries` modules. If others don't have the discipline to raise this to a higher level it is their problem.

@jonmeredith - what is your take on the above?

But I will investigate if I can avoid exporting it as I don't think I use it that much.

lehoff:
@jonmeredith 
- prefix and riak removed from #ctx{}
- init_get:get_status/0 used in service_available
- 503 when a new SQL type is encountered
- resource_exists now issues a 409 when trying to create an existing table
- process_post does not have to look in the body as we have to look at the body in malformed_quest where we store things to be used later
- checkpoint removed
- will turn all responses into JSON later today

lehoff:
@hmmr 
- coverage is not part of the HTTP API, so we are safe there.
- a RT test has been created and will be updated when I change the return formats to be JSON.

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,382 @@
...
-                               mochiweb_util:unquote(
-                                 wrq:path_info(table, RD)))
-                    }};
+        {ok, _C} ->
+            Table = riak_kv_wm_ts_util:table_from_request(RD),
+            Mod = riak_ql_ddl:make_module_name(Table),
+            {true, RD, Ctx#ctx{table=Table, mod=Mod}};
         {error, Reason} ->
-            handle_error({riak_client_error, Reason}, RD, Ctx)
+            Resp = riak_kv_wm_ts_util:set_error_message("Unable to connect to Riak: ~p", [Reason], RD),
```
jonmeredith: I don't think the error message reflects the problem - it's that the node is starting or stopping, not that you are unable to connect.

src/riak_kv_wm_timeseries_query.erl
```diff
@@ -54,69 +55,82 @@
...
+                            mod=Mod}}
+    catch
+        throw:{query, Reason} ->
+            Response = riak_kv_wm_ts_util:set_error_message("bad query: ~s", [Reason], RD),
+            {true, Response, Ctx};
+        throw:{unsupported_sql_type, Type} ->
+            Response = riak_kv_wm_ts_util:set_error_message(
+                         "The ~p query type is not supported over the HTTP API yet",
+                         [Type], RD),
+            {{halt, 503}, Response, Ctx}
```
jonmeredith, line 119: Exceptions thrown by parser still a bit sad. Perhaps a general catch all error for the non-thrown errors.

```
 curl -X POST http://localhost:8098/ts/1/query -v --data "OHNOES A DANGLING QUOTE ' "
*   Trying ::1...
* connect to ::1 port 8098 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8098 (#0)
> POST /ts/1/query HTTP/1.1
> Host: localhost:8098
> User-Agent: curl/7.43.0
> Accept: */*
> Content-Length: 26
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 26 out of 26 bytes
< HTTP/1.1 500 Internal Server Error
< Server: MochiWeb/1.1 WebMachine/1.10.8 (that head fake, tho)
< Date: Thu, 24 Mar 2016 18:56:16 GMT
< Content-Type: text/html
< Content-Length: 1112
<
<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1>The server encountered an error while processing this request:<br><pre>{error,<<"Unexpected token '''.">>,
    [{riak_ql_lexer,yyaction,4,[{file,"src/riak_ql_lexer.xrl"},{line,131}]},
     {riak_ql_lexer,string,4,
         [{file,
              "/opt/erlang/OTP_R16B02_basho10-rel/lib/parsetools-2.0.10/include/leexinc.hrl"},
          {line,40}]},
     {riak_ql_lexer,lex,1,[{file,"src/riak_ql_lexer.xrl"},{line,157}]},
     {riak_ql_lexer,get_tokens,1,[{file,"src/riak_ql_lexer.xrl"},{line,138}]},
     {riak_kv_wm_timeseries_query,compile_query,1,
         [{file,"src/riak_kv_wm_timeseries_query.erl"},{line,268}]},
     {riak_kv_wm_timeseries_query,malformed_request,2,
         [{file,"src/riak_kv_wm_timeseries_query.erl"},{line,102}]},
     {webmachine_resource,resource_call,3,
         [{file,"src/webmachine_resource.erl"},{line,186}]},
     {webmachine_resource,do,3,
* Connection #0 to host localhost left intact
         [{file,"src/webmachine_resource.erl"},{line,142}]}]}</pre><P><HR><ADDRESS>mochiweb+webmachine web server</ADDRESS></body></html>
```

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,382 @@
...
-        {true, RD, Ctx#ctx{mod = Mod, ddl = DDL}}
+        JsonStr = binary_to_list(wrq:req_body(RD)),
+        Json = mochijson2:decode(JsonStr),
+        DDLFieldTypes = ddl_fields_and_types(Mod),
+        extract_records(Json, DDLFieldTypes)
     catch
-        error:undef ->
-            handle_error({no_such_table, Table}, RD, Ctx)
+        Error:Reason ->
+            lager:log(info, self(), "extract_data: ~p:~p", [Error, Reason]),
```
jonmeredith: I don't think we should log here or a problem on a production system would fill the logs.  Error should be returned to the client, and could be included in an audit log.

src/riak_kv_wm_timeseries.erl
```diff
@@ -29,572 +34,382 @@
...
+              {true, Resp, Ctx};
+         {error, notfound} ->
+             Resp = riak_kv_wm_ts_util:set_error_message(
+                      "resource ~p does not exist - impossible to delete",
+                      [wrq:path(RD)],
+                      RD),
+              {{halt, 404}, Resp, Ctx}
+     catch
+         _:Reason ->
+             lager:log(info, self(), "delete_resource failed: ~p", [Reason]),
```
jonmeredith: Left over log statement - please remove.

src/riak_kv_wm_timeseries_query.erl
```diff
@@ -126,371 +140,196 @@ forbidden(RD, Ctx) ->
...
-            error_out({halt, 400},
-                      "Inappropriate method ~s for SQL query type", [Method], RD, Ctx)
-    end.
-
-flat_format(Format, Args) ->
-    lists:flatten(io_lib:format(Format, Args)).
-
-utf8_to_binary(S) ->
-    unicode:characters_to_binary(S, utf8, utf8).
+%% log(Format, Args) ->
```
jonmeredith: Looks like dead code - please remove.

src/riak_kv_wm_timeseries.erl
```diff
@@ -22,6 +22,11 @@
 
 %% @doc Resource for Riak TS operations over HTTP.
 %%
+%% This resource is responsible for everything under
+%% ```
+%% ts/v1/table/Table/keys
```
jonmeredith: URL is /tables/, not /table/ - please fix here and examples below.

jonmeredith:
I think the list_keys endpoint may need something of a rethink.  Here is what you get when you query it.

```
 curl -v -v -v  http://localhost:8098/ts/v1/tables/t/list_keys
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8098 (#0)
> GET /ts/v1/tables/t/list_keys HTTP/1.1
> Host: localhost:8098
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Vary: Accept-Encoding
< Transfer-Encoding: chunked
< Server: MochiWeb/1.1 WebMachine/1.10.8 (that head fake, tho)
< Date: Thu, 24 Mar 2016 22:39:12 GMT
< 
* Connection #0 to host localhost left intact
<html>{"url":"http://127.0.0.1:8098/ts/v1/tables/t/keys/f/f/s/s/ts/789/"}{"url":"http://127.0.0.1:8098/ts/v1/tables/t/keys/f/f/s/s/ts/456/"}{"url":"http://127.0.0.1:8098/ts/v1/tables/t/keys/f/f/s/s/ts/123/"}</html>
```

Content type is given as JSON, but the output is not JSON.  It's malformed HTML without a head/body element containing some fragments of JSON.  I think most languages would be sad if they had to parse this.

The problem is that the output could be large so we would want to chunk the data.  Ideally we would have a paginated interface for queries so we could just return a single json object.

Options:
- Generate a real HTML document with links to the keys.
- Just write out a list of CR/LF delimited URLs to the keys
- Send chunks made up of JSON arrays, delimited by \r\n or something like that.

hmmr:
@lehoff have a look how keys are sent in standard riak: https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_wm_keylist.erl#L230, and how they are received: https://github.com/basho/riak-erlang-http-client/blob/riak_ts-develop/src/rhc_listkeys.erl#L86.

There is also my branch in riak-erlang-http-client, https://github.com/basho/riak-erlang-http-client/tree/feature/az/http-client, where I have implemented all the TS-specific API calls. I then used riakhttpc as the alternative client in ts_cluster_comprehensive (https://github.com/basho/riak_test/pull/1016) to run the same scenario as it normally does with the protobuffer client (riakc). The wrappers in riak-erlang-http-client will probably need to be modified per your changes in riak_kv.

src/riak_kv_wm_timeseries.erl
lehoff: Changed the message to:
`"Node not ready: ~p"`

src/riak_kv_wm_timeseries_query.erl
lehoff, line 119: Changed to a nicer error without internal server error.

src/riak_kv_wm_timeseries.erl
lehoff: Removed.

src/riak_kv_wm_timeseries.erl
lehoff: Removed.

src/riak_kv_wm_timeseries_query.erl
lehoff: Removed.

lehoff:
It should be stream - fix for that coming up.
For the list_keys the idea was to return a stream of `{"url": "http://..."}` pairs. 
The html tag is a leftover that should have been removed.
All the other timeseries HTTP endpoints return JSON, so it might be best to continue that.
So chunks in JSON arrays seems best. But should it just be an array of `["http://...", "http://..."] or should we tag them with "url" in pairs I originally envisioned?

hmmr:
The query callback still needs some tweaks to accept the recently implemented WITH-clause for the CREATE query. The changes are in https://github.com/lehoff/riak_kv/pull/2, for you to consider and pull.

With those changes, I can confirm the test in https://github.com/basho/riak_test/pull/1016 passes.

Edit: The riak_test PR is 1015, not 1016.

hmmr:
+1 383e049

The tests (https://github.com/basho/riak_test/pull/1015) all pass.

jonmeredith:
... and the dialyzer says no, with a strange unrelated error to eleveldb.

```
eleveldb.erl:324: Call to missing or unexported function lager:error/2
```

hmmr:
Okaay, it appears @lehoff forgot to accept https://github.com/lehoff/riak_kv/pull/3, something I had presumed he did..

I have opened a yet another PR, with identical contents as this one plus the last commit to fix the dialyzer issue: https://github.com/basho/riak_kv/pull/1382.

lehoff:
Should be covered by #1398

PR #1372, opened 2016-03-15
start key has timestamp of first value RTS-977

andytill:
Riak TS uses the start key to hash to a vnode and does not take into account the start_inclusive filter.

``` sql
SELECT * FROM table1 WHERE a = 1 AND b = 1 AND c > 1999 AND c <= 5000;
```

This query produces a first sub query with a key from 1999..3000, this hashes to the 1999 quantum which will not contain the results from 2000..3000. The result set will have missing objects.

**This bug only manifests when using the `>` operator with a right hand value one less than the quanta boundary.**

This change increments the start key by one and removing the `{start_inclusive,false}` filter when it exists to produce a start key of 2000..3000 in the first sub query, the start key hashes to the vnode where the data is actually stored. Fitler `>=` is not affected because the right hand value is

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -509,10 +509,7 @@ hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, MaxSubQueries, Where1) ->
     {NoSubQueries, Boundaries} =
         riak_ql_quanta:quanta(Min2, Max2, QSize, QUnit),
     if
-        NoSubQueries == 1 ->
-            [Where2];
-        NoSubQueries > 1 andalso (MaxSubQueries == undefined orelse
-                                  NoSubQueries =< MaxSubQueries) ->
+        MaxSubQueries == undefined orelse NoSubQueries =< MaxSubQueries ->
             %% use the maximum value that has not been incremented, we still use
             %% the end_inclusive flag because the end key is not used to hash
             make_wheres(Where2, QField, Min2, Max1, Boundaries);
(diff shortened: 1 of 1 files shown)


src/riak_kv_qry_compiler.erl
```diff
@@ -1063,23 +1069,18 @@ simple_with_filter_1_test() ->
...
     [[StartKey, EndKey |_]] =
-        test_data_where_clause(<<"Scotland">>, <<"user_1">>, [{3000, 5000}]),
-    Where = [
+        test_data_where_clause(<<"Scotland">>, <<"user_1">>, [{3001, 5000}]),
+    ExpectedWhere = [
              StartKey,
              EndKey,
-             {filter, {'=', {field, <<"weather">>, varchar}, {const, <<"yankee">>}}},
-             {start_inclusive, false}
+             {filter, {'=', {field, <<"weather">>, varchar}, {const, <<"yankee">>}}}
```
hmmr, line 1076: It would probably be nice to have a non-optional `start_inclusive` flag.

src/riak_kv_qry_compiler.erl
```diff
@@ -485,23 +485,34 @@ hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, MaxSubQueries, Where) ->
...
+    %% from it.
+    case lists:keytake(start_inclusive, 1, Where1) of
+        {value, {start_inclusive, false}, WhereX}  ->
+            Where2 = WhereX,
+            Min2 = Min1 + 1;
+        _ ->
+            Where2 = Where1,
+            Min2 = Min1
+    end,
+    Max2 =
```
hmmr, line 504: Moving `Max` should not be generally a concern, except in an extreme case where the range includes a single time tick (which would make to the range collapse to 0, by incrementing the `Min`). It's just a note to make sure I read it right.

hmmr:
+1 daaffa5874

Posted two comments, but no change in code is expected.

andytill:
@borshop merge

PR #1373, opened 2016-03-16
relax DDL check in is_type_compiled

hmmr:
The check for type availability uses a match on DDL bodies of argument
DDL and that previously compiled and stored in metadata. That check is
not strictly necessary, but more importantly, it is bound to fail when
DDLs are produced from the same schema but in different #ddl_vX{} versions.

Relaxing this check will allow post-1.2 nodes to see DDLs compiled by
different ddl record versions, and thus do not be a blocker in
upgrade/downgrade scenarios.

### DIFF
--- a/src/riak_kv_metadata_store_listener.erl
+++ b/src/riak_kv_metadata_store_listener.erl
@@ -56,8 +56,7 @@ handle_event(_Event, State) ->
     {ok, State}.
 
 handle_call({is_type_compiled, [BucketType, DDL]}, State) ->
-    IsCompiled = (riak_kv_compile_tab:get_state(BucketType) == compiled andalso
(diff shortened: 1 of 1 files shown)


macintux:
It's necessary because riak core metadata updates are not guaranteed to arrive in the original order. It's possible that two different nodes could have two different versions of the DDL if the bucket type is redefined.

hmmr:
Then, we would need to compare the two DDLs after converting each to the version this node is running, I suppose. And, write the upgrade/downgrade functions, of course. Work to do.

hmmr:
I gather the concern of this PR is no longer valid, now that DDLs do happily upgrade and downgrade. One last thing to do here is, then, to GC this PR.

PR #1374, opened 2016-03-16
WIP: Refactor riak_kv_app:start/2 and Add Capabilities for the Query Pipeline

hazen:


src/riak_kv_app.erl
```diff
@@ -49,8 +49,37 @@
 %% @doc The application:start callback for riak.
 %%      Arguments are ignored as all configuration is done via the erlenv file.
 start(_Type, _StartArgs) ->
-    riak_core_util:start_app_deps(riak_kv),
```
hazen, line 52: Breaking this up seems like a great idea

src/riak_kv_app.erl
```diff
@@ -49,8 +49,37 @@
...
+    %% Register our cluster_info app callback modules, with catch if
+    %% the app is missing or packaging is broken.
+    catch cluster_info:register_app(riak_kv_cinfo),
+
+    %% print out critical env limits for support/debugging purposes
+    catch riak_kv_env:doc_env(),
+
+    %% Spin up supervisor
+    %% TODO isn't this a problem? the capabilities should be registered BEFORE
+    %% the top sup starts, surely?
```
hazen, line 68: That seems to make sense

src/riak_kv_app.erl
```diff
@@ -74,40 +103,169 @@ start(_Type, _StartArgs) ->
...
+    riak_core_capability:register({riak_kv, crdt_epoch_versions},
+                                  [?E2_DATATYPE_VERSIONS, ?E1_DATATYPE_VERSIONS],
+                                  ?E1_DATATYPE_VERSIONS),
+    
+    riak_core_capability:register({riak_kv, put_fsm_ack_execute},
+                                  [enabled, disabled],
+                                  disabled),
+    
+%    %% register the pipeline capabilities
+    riak_core_capability:register({query_pipeline, create_table},
```
hazen: You'll need to refresh my memory why components have different versions and how we would deal with mixed upgrades, i.e., some but not all components have a greater version. Just trying to wrap my head around this granularity.

gordonguthrie: Each stage in the pipeline will need to declare what inputs it can receive by means of a capability and what outputs it can emit by checking the downstream capability. I think it will also have to tag what it emits somehow (a record? a tagged tuple) so the downstream can know how to process it...

yorkshire-steve:

PR #1377, opened 2016-03-29
RTS-980,RTS-981,RTS-982: Recompile DDL

hazen:
Store the DDL compiler version in DETS and at startup if the version does not match, recompile it.  Also update all pre-1.2 DETS entries to include the legacy version so they update cleanly as well.  Any DETS entries which were stuck in the `compiling` state when the node went down are now marked as `retrying` and recompiled.  The versions are registered as a `riak_core_capability` but are not used as of yet.

**NOTE:** This PR requires https://github.com/basho/riak_ql/pull/115

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -209,10 +209,10 @@ start(_Type, _StartArgs) ->
                                           disabled),
 
             riak_core_capability:register({riak_kv, riak_ql_ddl_version},
-                                           riak_ql_ddl:get_compiler_capabilities(),
-                                           lists:last(riak_ql_ddl:get_compiler_capabilities())),
+                                           riak_ql_ddl_compiler:get_compiler_capabilities(),
+                                           lists:last(riak_ql_ddl_compiler:get_compiler_capabilities())),
 
-            riak_kv_ts_newtype:recompile_ddl(riak_ql_ddl:get_compiler_version()),
+            riak_kv_ts_newtype:recompile_ddl(riak_ql_ddl_compiler:get_compiler_version()),
 
             HealthCheckOn = app_helper:get_env(riak_kv, enable_health_checks, false),
             %% Go ahead and mark the riak_kv service as up in the node watcher.
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -127,6 +127,13 @@ start(_Type, _StartArgs) ->
     %% print out critical env limits for support/debugging purposes
     catch riak_kv_env:doc_env(),
 
+    %% TODO: Fix when DETS table is opened. Currently it's in riak_kv_ts_sup:init/0
+    %% Convert all legacy DDL compilation records to current format
+    %% riak_kv_compile_tab:upgrade_legacy_records(riak_ql_ddl:get_legacy_compiler_version()),
+
+    %% Clean up any lingering tables stuck in the compiling state
+    %% riak_kv_compile_tab:mark_compiling_for_retry(),
+
     %% Spin up supervisor
     case riak_kv_sup:start_link() of
         {ok, Pid} ->
@@ -207,6 +214,14 @@ start(_Type, _StartArgs) ->
             riak_core_capability:register({riak_kv, put_fsm_ack_execute},
                                           [enabled, disabled],
                                           disabled),
+
+            riak_core_capability:register({riak_kv, riak_ql_ddl_version},
+                                           riak_ql_ddl:get_compiler_capabilities(),
+                                           lists:last(riak_ql_ddl:get_compiler_capabilities())),
+
+            %% Q: Does KV need to be registered before getting a core_capability?
+            NegotiatedCompilerVersion = riak_core_capability:get({riak_kv, riak_ql_ddl_version}),
+            riak_kv_ts_newtype:recompile_ddl(NegotiatedCompilerVersion),
 
             HealthCheckOn = app_helper:get_env(riak_kv, enable_health_checks, false),
             %% Go ahead and mark the riak_kv service as up in the node watcher.
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -127,13 +127,6 @@ start(_Type, _StartArgs) ->
     %% print out critical env limits for support/debugging purposes
     catch riak_kv_env:doc_env(),
 
-    %% TODO: Fix when DETS table is opened. Currently it's in riak_kv_ts_sup:init/0
-    %% Convert all legacy DDL compilation records to current format
-    %% riak_kv_compile_tab:upgrade_legacy_records(riak_ql_ddl:get_legacy_compiler_version()),
-
-    %% Clean up any lingering tables stuck in the compiling state
-    %% riak_kv_compile_tab:mark_compiling_for_retry(),
-
     %% Spin up supervisor
     case riak_kv_sup:start_link() of
         {ok, Pid} ->
@@ -219,9 +212,7 @@ start(_Type, _StartArgs) ->
                                            riak_ql_ddl:get_compiler_capabilities(),
                                            lists:last(riak_ql_ddl:get_compiler_capabilities())),
 
-            %% Q: Does KV need to be registered before getting a core_capability?
-            NegotiatedCompilerVersion = riak_core_capability:get({riak_kv, riak_ql_ddl_version}),
-            riak_kv_ts_newtype:recompile_ddl(NegotiatedCompilerVersion),
+            riak_kv_ts_newtype:recompile_ddl(riak_ql_ddl:get_compiler_version()),
 
             HealthCheckOn = app_helper:get_env(riak_kv, enable_health_checks, false),
             %% Go ahead and mark the riak_kv service as up in the node watcher.
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -23,18 +23,15 @@
 -module(riak_kv_compile_tab).
 
 -export([
-         cleanup_old_tables/1,
          delete_dets/1,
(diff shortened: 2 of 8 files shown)


macintux:
Skeptical of adding an unused capability since the system is designed to easily add one retroactively (any node which doesn't register it is centrally assigned the fallback value). Is there a good reason to add this one before we need it?

src/riak_kv_compile_tab.erl
```diff
@@ -22,37 +22,66 @@
...
-    dets:open_file(?TABLE, [{type, set}, {repair, force}, {file, FilePath}]).
+    lager:debug("Opening DDL DETS table ~s", [FilePath]),
+    Result = dets:open_file(?TABLE, [{type, set}, {repair, force}, {file, FilePath}]),
+    %% Only update the DETS database if it opens successfully
+    case Result of
+        {ok, _} ->
+            %% Convert all legacy DDL compilation records to current format
+            upgrade_legacy_records(),
+
+            %% Clean up any lingering tables stuck in the compiling state
```
macintux: Two nitpicks: suggest using "records" instead of "tables" since the latter is a bit ambiguous in a DETS context.

Also would like to see an explanation of why these compiling records are lingering (node reboot during compilation). Same with the function comment.

src/riak_kv_compile_tab.erl
```diff
@@ -71,35 +100,75 @@ is_compiling(BucketType) ->
         compiling_state() | notfound.
 get_state(BucketType) when is_binary(BucketType) ->
     case dets:lookup(?TABLE, BucketType) of
-        [{_,_,_,State}] ->
+        [{_,_,_,_,State}] ->
```
macintux, line 140: Does it make sense to add a pattern match against the 4-tuple version, do a `lager:error` and/or try to recover?

Former seems like an easy thing to do, at least, and I assume would avoid a process crash.

macintux, line 140: Guess that means we'd have to do it everywhere it could crash, which is probably overkill.

src/riak_kv_compile_tab.erl
```diff
@@ -71,35 +100,75 @@ is_compiling(BucketType) ->
...
     end.
 
+%% Mark any lingering compilations as being retried
+-spec mark_compiling_for_retry() -> ok.
+mark_compiling_for_retry() ->
+    CompilingPids = dets:match(?TABLE, {'_','_','_','$1',compiling}),
+    lists:foreach(fun([Pid]) -> update_state(Pid, retrying) end, CompilingPids).
+
+%% Get the list of tables which need to be recompiled
+-spec get_tables_needing_recompiling(DDLVersion :: riak_ql_ddl:compiler_version_type()) ->
```
macintux: Skeptical about using "tables" in the function/binding names here, same as earlier comment.

src/riak_kv_compile_tab.erl
hazen: How about "bucket types"? Or "table records"?  It is confusing since it is records about the state of database tables stored in, well, a database.

src/riak_kv_compile_tab.erl
hazen, line 140: The whole idea is to clean up these before we get into the code.  This is the "upgrade once" idea which then in the rest of the code we assume to be already upgraded.

macintux, line 140: I understand, I was just contemplating the consequences if that cleanup doesn’t occur.

-John

src/riak_kv_compile_tab.erl
hazen: How about `get_ddl_needing_recompilation`? Or `get_bucket_types_needing_recompilation`?

src/riak_kv_app.erl
```diff
@@ -207,6 +207,12 @@ start(_Type, _StartArgs) ->
                                           [enabled, disabled],
                                           disabled),
 
+            riak_core_capability:register({riak_kv, riak_ql_ddl_version},
+                                           riak_ql_ddl:get_compiler_capabilities(),
+                                           lists:last(riak_ql_ddl:get_compiler_capabilities())),
+
```
javajolt: Agree that this registration is probably premature, so we really should just reschedule https://bashoeng.atlassian.net/browse/RTS-982 for 1.4+

src/riak_kv_compile_tab.erl
macintux: `records` seems both self-evident and less awkward, but perhaps I'm missing the ambiguity there.

macintux:
Having problems with the `ts_simple_select` test, the table isn't activating.

See this error:

```
2016-04-07 08:37:40.922 [error] <0.445.0>@riak_kv_ts_newtype:maybe_compile_ddl:140 Unknown DDL version type {ddl_v1,<<"GeoCheckin">>,[{riak_field_v1,<<"myfamily">>,1,varchar,false},{riak_field_v1,<<"myseries">>,2,varchar,false},{riak_field_v1,<<"time">>,3,timestamp,false},{riak_field_v1,<<"weather">>,4,varchar,false},{riak_field_v1,<<"temperature">>,5,double,true}],{key_v1,[{param_v1,[<<"myfamily">>]},{param_v1,[<<"myseries">>]},{hash_fn_v1,riak_ql_quanta,quantum,[{param_v1,[<<"time">>]},15,m],timestamp}]},{key_v1,[{param_v1,[<<"myfamily">>]},{param_v1,[<<"myseries">>]},{param_v1,[<<"time">>]}]}} for bucket type GeoCheckin (expecting ddl_v1 with old version notfound, new version 2)
```

macintux:
Both compiler tables appear to be empty, so `riak_kv_compile_tab:get_compiled_ddl_version` returns `notfound`.

src/riak_kv_ts_newtype.erl
```diff
@@ -111,20 +116,40 @@ code_change(_OldVsn, State, _Extra) ->
...
-                      riak_kv_compile_tab:get_ddl(BucketType)).
-
-maybe_compile_ddl(_BucketType, NewDDL, NewDDL) ->
+    maybe_compile_ddl(BucketType,
+                      retrieve_ddl_from_metadata(BucketType),
+                      riak_kv_compile_tab:get_ddl(BucketType),
+                      riak_ql_ddl:get_compiler_version(),
+                      riak_kv_compile_tab:get_compiled_ddl_version(BucketType)).
+
+maybe_compile_ddl(BucketType, NewDDL, NewDDL, NewVsn, NewVsn) ->
```
macintux, line 125: Need a new header where the last argument is `notfound`. New type.

src/riak_kv_compile_tab.erl
```diff
@@ -22,45 +22,112 @@
...
+%% Open both old and new copies of the DETS table.  Assume old table is
+%% canonical and force entries in the new table to be recompiled.
+-spec new_table_result({ok, dets:tab_name()} | {error, any()},
+                       {ok, dets:tab_name()} | {error, any()}) ->
+                       {ok, dets:tab_name(), dets:tab_name()} | {error, any(), any()}.
+new_table_result({ok, LegacyDets}, {ok, Dets}) ->
+    %% Convert all legacy DDL compilation records to current format
+    upgrade_legacy_records(),
+
+    %% TODO: Use this once 1.2 support has been removed
```
macintux: This TODO looks out of place and is confusing. I assume you meant something above the `upgrade_legacy_records` invocation, and a verb other than "Use"

src/riak_kv_compile_tab.erl
```diff
@@ -71,35 +138,77 @@ is_compiling(BucketType) ->
...
+    MismatchedTables = dets:select(?TABLE, [{{'$1','$2','_','_',compiled},[{'/=','$2', DDLVersion}],['$$']}]),
+    RetryingTables = dets:match(?TABLE, {'$1','$2','_','_',retrying}),
+    Tables = [hd(X) || X <- MismatchedTables ++ RetryingTables],
+    lager:info("Recompile the DDL of these bucket types ~p", [Tables]),
+    Tables.
+
+%% TODO: Remove this once 1.2 support has been removed
+%% Convert all pre-1.3 records to include the DDL compiler version in the
+%% DETS table.  Simply recompile everything all the time since we don't
+%% know if this is a downgrade or not.
```
macintux: Comment is incorrect. Rather, we don't know if this is an upgrade following a downgrade.

We could determine whether this was a downgrade from a later compiler version by comparing version fields in the 1.3 table.

src/riak_kv_compile_tab.erl
```diff
@@ -71,35 +138,77 @@ is_compiling(BucketType) ->
...
+%% Mark any lingering compilations as being retried
+-spec mark_compiling_for_retry() -> ok.
+mark_compiling_for_retry() ->
+    CompilingPids = dets:match(?TABLE, {'_','_','_','$1',compiling}),
+    lists:foreach(fun([Pid]) -> update_state(Pid, retrying) end, CompilingPids).
+
+%% Get the list of records which need to be recompiled
+-spec get_ddl_records_needing_recompiling(DDLVersion :: riak_ql_ddl_compiler:compiler_version()) ->
+    [binary()].
+get_ddl_records_needing_recompiling(DDLVersion) ->
```
macintux, line 190: Should we try `failed` DDLs as well?

src/riak_kv_compile_tab.erl
hazen: In a previous iteration this was commented out, but decided to use it even though it's less useful when recompiling everything.

src/riak_kv_compile_tab.erl
hazen, line 140: Since we now have both old and new copies, just query the new copies with the 5-tuple.

macintux:
:+1: da7d02f

PR #1381, opened 2016-04-04
Postcommit support for timeseries

macintux:
Add support for postcommit hooks. This is a streamlined version of the KV postcommit hooks.

Also provides a new batch put interface designed for batches of data that are already msgpack-encoded.
Required for basho/riak_repl#738 (RIAK-1546) (RIAK-1823)

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -46,6 +46,7 @@
          waiting_local_vnode/2,
          waiting_remote_vnode/2,
          postcommit/2, finish/2]).
+-export([get_hooks/2, get_hooks/3]).
 
 -type detail_info() :: timing.
 -type detail() :: true |
@@ -394,6 +395,7 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
                                       robj = RObj0,
                                       n=N, bucket_props = BucketProps,
                                       trace = Trace,
+                                      bkey = {Bucket, _Key},
                                       preflist2 = Preflist2}) ->
     Timeout = get_option(timeout, Options0, ?DEFAULT_TIMEOUT),
     PW0 = get_option(pw, Options0, default),
@@ -446,7 +448,7 @@ validate(timeout, StateData0 = #state{from = {raw, ReqId, _Pid},
                 end,
             Postcommit =
                 if Disable -> [];
-                   true -> get_hooks(postcommit, BucketProps, StateData0)
+                   true -> get_hooks(postcommit, BucketProps, Bucket)
                 end,
             {VNodeOpts0, ReturnBody} =
                 case get_option(returnbody, Options) of
@@ -980,9 +982,9 @@ get_hooks(HookType, BucketProps) ->
             Hooks
     end.
 
-get_hooks(postcommit, BucketProps, #state{bkey=BKey}) ->
+get_hooks(postcommit, BucketProps, Bucket) ->
     BaseHooks = get_hooks(postcommit, BucketProps),
-    CondHooks = riak_kv_hooks:get_conditional_postcommit(BKey, BucketProps),
+    CondHooks = riak_kv_hooks:get_conditional_postcommit(Bucket, BucketProps),
     BaseHooks ++ (CondHooks -- BaseHooks).
 
 get_option(Name, Options) ->
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -46,7 +46,6 @@
          waiting_local_vnode/2,
          waiting_remote_vnode/2,
          postcommit/2, finish/2]).
--export([get_hooks/2, get_hooks/3]).
 
 -type detail_info() :: timing.
 -type detail() :: true |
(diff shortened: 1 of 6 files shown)


macintux:
Single-value puts are currently broken in these PRs, am working on a fix.

macintux:
Bugs corrected in most recent commit. Ready for review.

src/riak_kv_w1c_worker.erl
```diff
@@ -149,12 +155,37 @@ ts_batch_put(RObjs, W, PW, Bucket, NVal, EncodeFn, Preflist) ->
...
+      {batch_put, Bucket, BucketProps, EncodedVals, ReqId, DocIdx, Preflist,
        #rec{w=W, pw=PW, n_val=NVal, from=self(),
             start_ts=StartTS,
             size=Size}}),
     {ok, {ReqId, Worker}}.
 
+%% Primarily for repl support. Put a list of objects represented as
+%% tuple({Bucket, LocalKey}, msgpack-encoded riak object) into the
+%% correct partition
+ts_batch_put_encoded([{{Bucket, _LK}, _RObj0}|_Rest]=RObjs, DocIdx) ->
```
gordonguthrie, line 182: The funcs ts_batch_put_encoded and ts_batch_put contain a lot of common code which could be refactored out - both funs being a 'prep' stage for a do_put function

src/riak_kv_w1c_worker.erl
```diff
@@ -108,13 +108,13 @@ put(RObj0, Options) ->
...
                 NVal :: pos_integer(),
-                Key :: binary()|{binary(), binary()},
+                Key :: tuple()|{binary(), tuple()},
                 EncodeFn :: fun((riak_object:riak_object()) -> binary()),
                 Preflist :: term()) ->
                        {ok, {reference(), atom()}}.
 
-async_put(RObj, W, PW, Bucket, NVal, {PK, LK}, EncodeFn, Preflist) when is_tuple(LK) ->
-    async_put(RObj, W, PW, Bucket, NVal, PK, LK, EncodeFn, Preflist);
+async_put(RObj, W, PW, Bucket, NVal, {PartIdx, LK}, EncodeFn, Preflist) when is_tuple(LK) ->
```
nickelization, line 131: Something feels off about this. I can see in `riak_kv_ts_api:put_data_to_partitions` where you put in the change to pass in `{PartIdx, LK}`, but then the other place that `async_put` is called (in `riak_kv_w1c_worker:put`) it looks like it's still passing in something else. I'm not quite sure if/what the mismatch here is stemming from, or if there's a bug being introduced somewhere, so I will keep investigating and try to figure out what's going on here.

src/riak_kv_put_fsm.erl
```diff
@@ -46,6 +46,7 @@
          waiting_local_vnode/2,
          waiting_remote_vnode/2,
          postcommit/2, finish/2]).
+-export([get_hooks/2, get_hooks/3]).
```
bsparrow435: Dont think these functions need to be exported. I like the changes to seperate the Bucket out of the Bkey for the logic to select the hook but get_hook is only called within this module. Any reason i'm missing to make it public?

src/riak_object.erl
```diff
@@ -129,6 +130,19 @@ new(B, K, V, MD) when is_binary(B), is_binary(K) ->
 newts(B, K, V, MD) ->
     new_int2(B, K, V, MD).
 
+-spec is_ts(riak_object()) -> {'true', pos_integer()} | 'false'.
+is_ts(RObj) ->
+    check_for_ddl(get_contents(RObj)).
+
+check_for_ddl([{Metadata, _V}]) ->
```
bsparrow435, line 137: Do we need to worry about siblings with TS data? `get_contents` can return more than a 1 element list which would make `check_for_ddl` always return true.

macintux, line 137: Correction: would always return false, neh?

Struggling to think through w1c and siblings. Good question.

src/riak_kv_put_fsm.erl
macintux: I suspect the exports are left over from a very early draft. I'll remove them.

src/riak_object.erl
macintux, line 137: Will add a pattern match for the tail of a list (but ignore it) because it's not reasonable to assume the riak object will never have siblings.

andytill, line 137: Write Once objects never have siblings. Maybe better to crash than to silently ignore that, and lose data presumably if it ever happens.

macintux, line 137: Held a brief discussion with Fred, we’re going to do more digging but it does not appear safe to assume they never have siblings.

-John

macintux, line 137: Rethinking this now. The reason I originally wrote it that way was to avoid invoking `dict:find` on every object with siblings, so now we have a potentially significant performance hit on fullsync.

JeetKunDoug:
:+1: 8f6449e

JeetKunDoug:
:+1: 0751235

macintux:
@borshop merge

JeetKunDoug:
@borshop merge

PR #1383, opened 2016-04-07
TS filter deleted rows [JIRA: RIAK-2493]

andytill:
When a row is deleted (currently single key deletes) it's key will not be removed before leveldb compaction.  Previously the coordinating `riak_kv_qry_worker` would convert these deleted records into empty lists.  This would cause problems when encoding the lists into protobuffer format.

Now the `riak_kv_qry_worker` removes the deleted records, making them appear to not exist.  Filtering them here avoids having checks all the way through the selection funs and in riak_pb.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -157,18 +157,17 @@ run_sub_qs_fn([{{qry, Q}, {qid, QId}} | T]) ->
     {ok, _PID} = riak_kv_index_fsm_sup:start_index_fsm(node(), [{raw, QId, Me}, Opts]),
     run_sub_qs_fn(T).
 
-decode_results(KVList) ->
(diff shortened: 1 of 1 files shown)


macintux:
:+1: 59345da

macintux:
Let's sit tight on merging this until after HTTP lands.

javajolt:
OK. I was going to retest it today and possibly merge today.

hazen:
Yes, it looks like the test failures I saw yesterday were not related to the PR.  The TTB I suspect blew things up.

macintux:
@borshop merge

DSomogyi:
create jira issue

Basho-JIRA:
PR:
https://github.com/basho/riak_kv/pull/1383 

**_[posted via JIRA by Derek Somogyi]**_

PR #1384, opened 2016-04-08
Added back row encoding

erikleitch:
Single-line change to add back row encoding.  This got dropped in the recent TTB changes.  

The only change in this PR is that query rows are returned to the client as a list of tuples instead of a list of lists.  This is the format in which rows were previously returned by the PB encoding layer, and by the earlier version of TTB encoding.


hazen:
This helped the testing situation, but we still have many failures:

```
19 Tests Failed
41 Tests Passed
That's 68.33333333333333% for those keeping score
```

I'll look at this continued failures

hazen:
Clearly this does not completely solve the problem, but it helps, so I'll merge it for now.

hazen:
👍 f2c5290

PR #1389, opened 2016-04-12
make outgoing PB messages TTB-friendly

hmmr:
Ensure that
- all error messages in `#rpberrresp` are binaries;
- present NULLs as `[]` not `undefined` in DESCRIBE query returned table;
- also, present rows as tuples, not lists.

### DIFF
--- a/src/riak_kv_ts_svc.erl
+++ b/src/riak_kv_ts_svc.erl
@@ -801,7 +801,7 @@ check_table_and_call(Table, Fun, TsMessage, State) ->
 -spec make_rpberrresp(integer(), string()) -> #rpberrorresp{}.
 make_rpberrresp(Code, Message) ->
     #rpberrorresp{errcode = Code,
-                  errmsg = lists:flatten(Message)}.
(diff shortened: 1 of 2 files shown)


src/riak_kv_ts_svc.erl
```diff
@@ -801,7 +801,7 @@ check_table_and_call(Table, Fun, TsMessage, State) ->
 -spec make_rpberrresp(integer(), string()) -> #rpberrorresp{}.
 make_rpberrresp(Code, Message) ->
     #rpberrorresp{errcode = Code,
-                  errmsg = lists:flatten(Message)}.
+                  errmsg = iolist_to_binary(Message)}.
```
hazen, line 804: Do we need to flatten first?

hmmr, line 804: Nope. `iolist_to_binary` has digestive system strong enough to crunch that: http://erldocs.com/R16B02/erts/erlang.html?i=1&search=iolist#iolist_to_binary/1.

hazen:
👍 0c67ac9

PR #1392, opened 2016-04-13
Mark next rebuild to full in case the node is shutdown cleanly while the tree is being updated.

JeetKunDoug:


jvoegele:
:+1: 7eb2fcfa27c322e78d4efd8c2aaceb170b32724a

fadushin:
@borshop merge

PR #1396, opened 2016-04-17
Prepare internal API for shared use between PB and WM callbacks [1/3]

hmmr:
This is Part 1/3 of #1382 broken up and reassembled into new commits.

After the merge of TTB branch, and in anticipation of the merge of HTTP API branch, this PR proposes a code reorg consisting of the following steps:
1. Move functions shared between PB and WM callbacks from riak_kv_ts_svc to riak_kv_ts_util;
2. Further move functions `get_data`, `put_data`, `delete_data`, `query` and `compile_to_per_quantum_queries` (used to produce coverage results, with `riak_kv_ts_util:sql_to_cover`) into the newly created module `riak_kv_ts_api`. These will constitute the **TS internal API**.
3. Move code serving the INSERT query from `riak_kv_ts_svc` into `riak_kv_qry`. Now the call to `riak_kv_qry:submit("INSERT ...")` will actually take effect (previously, the 'submit' part was a no-op while all the validations, conversions and the eventual call of `riak_kv_ts_util:put_data` happened as a side effect).

PR #1382, opened 2016-04-06
A copy of HTTP client API for TS (#1371)

hmmr:
This is the same as https://github.com/basho/riak_kv/pull/1371 plus one additional commit to fix one dialyzer issue. That fix would be merged from https://github.com/lehoff/riak_kv/pull/3 into Torben's branch.

PR #1398, opened 2016-04-17
HTTP client API all-in-one [3/3]

hmmr:
This is Part 3/3 of #1382 broken up and reassembled into new commits.

Requires (includes) #1396, #1397 (with basho/riak_pb#190 (RIAK-2492)).

A two month worth of commits by @hmmr and @lehoff, squashed.

The odd change in riak_kv_qry_worker (preserving lists for records in the return value of query) goes with the list_to_tuple conversion now happening in riak_kv_ts_api. Native lists from riak_kv_qry:submit are immediately consumable by mochijson2:encode, and this is the form we would like to have the query results in wm_timeseries_query.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -279,7 +279,7 @@ prepare_final_results(#state{
 prepare_final_results2(#riak_sel_clause_v1{col_return_types = ColTypes,
                                            col_names = ColNames}, Rows) ->
     %% filter out empty records
-    FinalRows = [list_to_tuple(R) || R <- Rows, R /= [[]]],
+    FinalRows = [R || R <- Rows, R /= [[]]],
     {ColNames, ColTypes, FinalRows}.
 
 %%%===================================================================
@@ -292,10 +292,8 @@ prepare_final_results2(#riak_sel_clause_v1{col_return_types = ColTypes,
 
 prepare_final_results_test() ->
     Rows = [[12, <<"windy">>], [13, <<"windy">>]],
-    RowsAsTuples = [{12, <<"windy">>}, {13, <<"windy">>}],
-    % IndexedChunks = [{1, Rows}],
     ?assertEqual(
-        {[<<"a">>, <<"b">>], [sint64, varchar], RowsAsTuples},
+        {[<<"a">>, <<"b">>], [sint64, varchar], Rows},
         prepare_final_results(
             #state{
                 qry =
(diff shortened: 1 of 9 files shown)


src/riak_kv_qry_worker.erl
```diff
@@ -292,10 +292,8 @@ prepare_final_results2(#riak_sel_clause_v1{col_return_types = ColTypes,
 
 prepare_final_results_test() ->
     Rows = [[12, <<"windy">>], [13, <<"windy">>]],
```
hazen, line 294: Shouldn't this be tuples then?  I'll test and find out...

hmmr, line 294: `list_to_tuple` now happens in `riak_kv_ts_svc`, right before the records are delivered to the PB/TTB code and further to the client. This is so in order to avoid converting tuples back to lists in `riak_kv_wm_timeseries_query`, where lists are required for converting the data to json.

Which is why we still have native records-as-lists here.

src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -0,0 +1,202 @@
...
+            _ = riak_kv_keys_fsm:ack_keys(From),
+            {ts_keys_to_body(Keys, Table, Mod), fun() -> stream_keys(ReqId, Table, Mod) end};
+        {ReqId, {keys, Keys}} ->
+            {ts_keys_to_body(Keys, Table, Mod), fun() -> stream_keys(ReqId, Table, Mod) end};
+        {ReqId, done} ->
+            {<<>>, done};
+        {ReqId, {error, timeout}} ->
+            {mochijson2:encode({struct, [{error, timeout}]}), done};
+        _Weird ->
+            %% @todo: should we log this?
```
hazen: Probably a good idea. Might be useful to see if bad requests are being sent.

src/riak_kv_wm_utils.erl
```diff
@@ -355,6 +355,8 @@ jsonify_bucket_prop({search_extractor, {M, F, Arg}}) ->
                               {?JSON_ARG, Arg}]}};
 jsonify_bucket_prop({name, {_T, B}}) ->
     {<<"name">>, B};
+jsonify_bucket_prop({ddl, _DDL}) ->
+    {<<"ddl">>, <<"riak_ql_to_string:sql_to_string might be useful here.">>};
```
hazen, line 359: Were we actually going to support `<<"ddl">>` or not yet?

hmmr, line 359: It's a stub, effectively. And, it's not really important because the DDL structure can is available via a DESCRIBE query.

hazen:
👍 94db162

hazen:
@borshop merge

PR #1399, opened 2016-04-18
Account for n_vals associated with bucket types when determining responsible preflists

fadushin:
The KV and YZ hash tree subsystems build hash trees based on the size of preflists for all buckets stored in the ring.  However, the calculation of preflists does not currently account for bucket types, resulting in an incorrect set of hash trees if a bucket type specifies an n_val different from the default.

This patch adds the n_vals from all bucket types to the responsible preflists calculation.

In addition, this PR add convenience `clear_trees` and `expire_trees` operations to the `riak_kv_entropy_manager` module.

This PR requires https://github.com/basho/riak_core/pull/832

### DIFF
--- a/src/riak_kv_ensembles.erl
+++ b/src/riak_kv_ensembles.erl
@@ -167,34 +167,10 @@ bootstrap_preflists(Ring, CHBin) ->
     ok.
 
 required_ensembles(Ring) ->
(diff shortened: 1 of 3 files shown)


nickelization:
:+1: 47a553e

fadushin:
@borshop merge

PR #1400, opened 2016-04-19
restore batching put-capable version of riak_kv_ts_api:put_data

hmmr:
This PR restores the version of `put_data` introduced in https://github.com/basho/riak_kv/pull/1366, which was accidentally in code migration from `riak_kv_pb_timeseries` to `riak_kv_ts_svc` and eventually to `riak_kv_ts_api`.

ts_cluster_comrehensive passes.

### DIFF
--- a/src/riak_kv_ts_api.erl
+++ b/src/riak_kv_ts_api.erl
@@ -106,10 +106,18 @@ put_data(Data, Table, Mod) ->
     end.
 
 put_data_to_partitions(Data, Bucket, BucketProps, DDL, Mod) ->
+    DDL = Mod:get_ddl(),
+    BucketProps = riak_core_bucket:get_bucket(Bucket),
     NVal = proplists:get_value(n_val, BucketProps),
+
     PartitionedData = partition_data(Data, Bucket, BucketProps, DDL, Mod),
     PreflistData = add_preflists(PartitionedData, NVal,
                                  riak_core_node_watcher:nodes(riak_kv)),
+
+    SendFullBatches = riak_core_capability:get({riak_kv, w1c_batch_vnode}, false),
+    %% Default to 1MB for a max batch size to not overwhelm disterl
+    CappedBatchSize = app_helper:get_env(riak_kv, timeseries_max_batch_size,
+                                         1024 * 1024),
     EncodeFn =
         fun(O) -> riak_object:to_binary(v1, O, msgpack) end,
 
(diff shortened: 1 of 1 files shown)


macintux:
👍 5505611

hazen:
@macintux Do we have unit and/or riak_test tests for batch puts?

macintux:
No. @hmmr pointed out in the original PR that we should have something in riak_test that leverages intercepts to verify that batches are actually used; today we only check that all data arrives as intended, so if it reverts to single-item puts, we'd have no way of noticing that.

The time required didn't seem worthwhile, as I expected things to break if somehow batching was lost; in a sense I was correct, in that MDC went haywire, but otherwise tests seem to pass.

src/riak_kv_ts_api.erl
```diff
@@ -106,10 +106,18 @@ put_data(Data, Table, Mod) ->
...
 put_data_to_partitions(Data, Bucket, BucketProps, DDL, Mod) ->
+    DDL = Mod:get_ddl(),
+    BucketProps = riak_core_bucket:get_bucket(Bucket),
     NVal = proplists:get_value(n_val, BucketProps),
+
     PartitionedData = partition_data(Data, Bucket, BucketProps, DDL, Mod),
     PreflistData = add_preflists(PartitionedData, NVal,
                                  riak_core_node_watcher:nodes(riak_kv)),
+
+    SendFullBatches = riak_core_capability:get({riak_kv, w1c_batch_vnode}, false),
```
hazen, line 117: Where does this ever get registered?

src/riak_kv_ts_api.erl
```diff
@@ -184,8 +196,54 @@ build_object(Bucket, Mod, DDL, Row, PK) ->
     RObj = riak_object:newts(
              Bucket, PK, Obj,
              dict:from_list([{?MD_DDL_VERSION, ?DDL_VERSION}])),
-    {RObj, LK}.
-
+    {LK, RObj}.
```
hazen, line 199: This reordering of terms is intentional, right?

src/riak_kv_ts_api.erl
hmmr, line 117: @javajolt See discussion in the original https://github.com/basho/riak_kv/pull/1366/files

macintux, line 117: sigh. In a commit that also apparently didn't make it in. Good catch, need to dig deeper.

macintux, line 117: wait, I lied, wrong branch. riak_kv_app.erl has it.

hmmr, line 117: It is:https://github.com/basho/riak_kv/pull/1366/files#diff-6aee27f2dd08e05d1365029f3c722b2aL508

src/riak_kv_ts_api.erl
macintux, line 199: Yes, corresponds to line 235 in this file and `riak_kv_w1c_worker:ts_batch_put/7` (note the spec). Pretty sure I had it reversed in an earlier version of the code.

macintux:
👍 5505611

macintux:
Screw it, merged manually

PR #1401, opened 2016-04-22
Descending Keys

andytill:
Relies on https://github.com/basho/riak_ql/pull/117 to build.

Tests in https://github.com/basho/riak_test/pull/1055

Descending keys adds a new keyword `DESC` that can be applied to elements in the local key.

``` sql
CREATE TABLE table1 (
a SINT64 NOT NULL,
b SINT64 NOT NULL,
c TIMESTAMP NOT NULL,
PRIMARY KEY ((a,b,quantum(c, 1, 's')), a,b,c DESC));
```

This signals that field `c` should have it's value inverted when it is written as a key. This means that for `sint64` and `timestamp` values the value is negated so 10 becomes -10. For varchars `bnot` is run on each byte. This means that the keys will be sorted in the opposite, descending order by leveldb. Without the `DESC` keyword, keys with the following values will be sorted in the order:

``` erlang
{1,1,1}
{1,1,2}
{1,1,3}
```

When the descending keys transform is applied, the key values will be modified to sort like the following:

``` erlang
{1,1,-3}
{1,1,-2}
{1,1,-1}
```

When the values are queried they will come out in this order, as if the query had `ORDER BY c DESC`.  Currently Riak TS does not support ORDER BY but even when it does, it will allow query result sets to be ordered and streamed back (streaming still to be implemented also). An ORDER BY requires that all of the results be retrieved from disk and sorted before anything is returned. If the user knows what queries are going to be issued then the key can be designed including use of `DESC` to return the results using streaming.

Note that the transform is only applied to the **local key** and not the value stored in the object. This affects single key operations and list keys as well as the more obvious writes and queries.  The **partition key** does **not** get transformed so the vnode hashing is done on the original key values.


erikleitch:
**Summary as I understand it**

**include/riak_kv_ts_svc.hrl, src/riak_kv_ts_svc.erl**
- the state record has been modified to include a key transformation function to map from keys as-written to leveldb to keys as-would-be-written by a user.  This is used in process_stream to convert results from listkeys that may have had fields swapped for descending keys back to user-space keys.    All seems fine.

**src/riak_kv_qry.erl**
- describe function modified to describe field ordering.  Seems fine.

**src/riak_kv_eleveldb_backend.erl**
- key_to_storage_format_key function added to convert start/end keys for range_scan to possibly-swapped-for-descending-order keys.  Note added that in the case of descending keys, the sense of the start/end keys will already have been swapped on entry to this function.  All fine.

**src/riak_kv_qry_compiler.erl**
- This code addresses my chief concern, that in addition to swapping the start and end keys in range_scans, the subqeries should also be executed in reverse order for descending keys.  This appears to be done correctly, and in my manual testing, the subqueries are reversed as I would expect them to be.

**src/riak_kv_qry_coverage_plan.erl**
- Now takes the min of the start/end key, since for descending keys the start/end keys will be swapped on input to this function.

**General Notes**

1) There's a lot of key manipulation going on, and a proliferation of functions to convert from:
- local key to partition key
- local key to storage-format key (normal to descending order)
- storage-format key to local key (descending to normal order)
- streamed listkeys to user-order (descending to normal order)

some of these are passed around as key conversion fns in records, others are built into modules, others are part of riak_kv_ts_util.

I feel like this is becoming a potential maintenance liability -- it's increasingly difficult when reading through the code to keep track of what version of the key is being used where and where the relevant key conversion functions can be found.  I don't know if these can be consolidated in one place (riak_kv_ts_util? Mod?), or just better named (maybe rigorously identify which key is being converted to what, both in the function name and the argument lists, rather than using 'Key' to mean different things in different contexts), or what, but I feel that it may be worth some effort to think about streamlining it.

2) All of this manipulation of keys to support descending key order begs the question of a) what problem we were trying to solve by adding support for descending keys in the first place (was this written down anywhere in a design doc?), and b) if the point is to optimize performance for queries to return the last N vals (LIMIT BY behavior), have we achieved that optimization in practice or are we killing any benefit derived from ordering with the extra overhead needed to map and remap these keys?

andytill:
> I feel like this is becoming a potential maintenance liability -- it's increasingly difficult when reading through the code to keep track of what version of the key is being used where and where the relevant key conversion functions can be found. I don't know if these can be consolidated in one place (riak_kv_ts_util? Mod?), or just better named (maybe rigorously identify which key is being converted to what, both in the function name and the argument lists, rather than using 'Key' to mean different things in different contexts), or what, but I feel that it may be worth some effort to think about streamlining it.

I agree that keys are becoming increasingly complicated, because of the number of things they can mean. Local key, partition key or descending local key.  I used a fun to partially apply `Mod` and `ColumnInfo`, instead of passing them around in the state record, looks like I forgot to remove the `column_info` field from the record. I'll fix that.

> 2) All of this manipulation of keys to support descending key order begs the question of a) what problem we were trying to solve by adding support for descending keys in the first place (was this written down anywhere in a design doc?)

The section in the doc is here https://docs.google.com/document/d/1kuw5XkZbmn0slpBvi8rqyqoyvsjelpdZMlRUmpU-b4o/edit#heading=h.xwmi4hmotixy (restricted access, sorry).

The aim was to allow rows to be returned in a user defined order, in a way that could use streamed queries coming in 1.4. This returns sub query results as soon as they come into the coordinator. Using ORDER BY means retrieving all results from all sub queries to calculate the total order.

>  b) if the point is to optimize performance for queries to return the last N vals (LIMIT BY behavior), have we achieved that optimization in practice or are we killing any benefit derived from ordering with the extra overhead needed to map and remap these keys?

I didn't have LIMIT in mind when I did this but it would work for getting the bottom N rows. Not sure of the performance impact, but the transform takes place per sub query rather than per row, so it isn't that much extra work. Except for list keys which is per-key.

andytill:
Closing, this is change is now part of PR https://github.com/basho/riak_kv/pull/1500

PR #1402, opened 2016-04-23
allow non-3-tuple keys when assisted sext decoding in eleveldb_backend fold_object fun

hmmr:
RTS-1083

The hand-written sext-decoder in `riak_kv_eleveldb_backend:from_object_key` does some binary matching to skip one call to `sext:decode` when converting the object to a `{Bucket, Key}` tuple before it is handed over to key folding function. That matching expects the key to be a three-tuple.

Because flexible keys work by @andytill has lifted the 3-field restriction on TS keys, we should, correspondingly, allow the 5th byte of the binary (the number of elements in the encoded tuple) to be any number.

With the tuple size hard-coded at 3, keys with the number of fields other than 3 were treated as non-TS data, and were returned already sext-decoded, eventually breaking the client code which attempted decoding on its own. In particular, there was a workaround for this, which is now being removed (the change in `riak_kv_wm_timeseries_listkeys:ts_keys_to_body`).

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -1059,7 +1059,6 @@ orig_from_object_key(LKey) ->
 %% so that the key in the riak object structure will be a binary on decode.
 %% All support tooling/usual methods for visiting objects should still work.
 %%
-%% 2016-04-23 @hmmr:
 %% Because flexible keys work by @andytill has lifted the 3-field
 %% restriction on TS keys, we should, correspondingly, allow the 5th
 %% byte of the binary (which is the number of elements in the encoded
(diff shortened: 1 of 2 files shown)


src/riak_kv_eleveldb_backend.erl
```diff
@@ -1059,12 +1059,17 @@ orig_from_object_key(LKey) ->
 %% so that the key in the riak object structure will be a binary on decode.
 %% All support tooling/usual methods for visiting objects should still work.
 %%
+%% 2016-04-23 @hmmr:
+%% Because flexible keys work by @andytill has lifted the 3-field
```
andytill, line 1062: Names dates can be removed from comments. Git blame for that.

andytill:
This could do with some unit tests since it is quite testable within the backend module. Maybe convert erlang terms to the key binary and then decode them back to confirm the round trip is successful. Does this resolve a failing riak test also or was it found through manual testing?

andytill:
Tested before and after in the shell and it has fixed the issue.

hmmr:
@andytill As things currently stand, there are no special checks for non-3-element keys in any r_t module. The bug was found in the course of investigations of failures on that hand-crafted miniature DDL which @jonmeredith has reported while testing the HTTP client API.

hmmr:
@borshop: retry f0a49e1

src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -159,20 +159,15 @@ stream_keys(ReqId, Table, Mod) ->
...
-
     URLs =
         lists:map(
-          fun(Key) when is_tuple(Key) ->  %% simple single-field keys like {1}, seen in the wild
-                  format_url(BaseUrl, KeyTypes, tuple_to_list(Key));
-             (Key) when is_binary(Key) -> %% sext-encoded ones
-                  format_url(BaseUrl, KeyTypes, tuple_to_list(sext:decode(Key)))
+          fun(Key) when is_binary(Key) ->
+                  format_url(BaseUrl, KeyTypes, tuple_to_list(sext:decode(Key)));
+             (Key) ->
```
hazen, line 166: Did you want a `is_tuple` guard here?

hmmr, line 166: `Key` is a binary all round -- just what `sext:decode/1` expects.

src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -159,20 +159,15 @@ stream_keys(ReqId, Table, Mod) ->
...
+                  %% this clause is to please dialyzer (dialyzer is
+                  %% within his rights as we have no way to 'declare'
+                  %% the type of Key)
+                  format_url(BaseUrl, KeyTypes, tuple_to_list(Key))
```
hazen, line 170: But down here `Key` is just a tuple, right?

hmmr, line 170: This one is dead code, for dialyzer to feel good. Because `Keys` are received from other processes, dialyzer has to consider `Key` as being of type `term()`, and needs to know how the non-binary values of `Key` going to be disposed of. Even if we fancy putting a guard on the receive branch, we can at most make sure `Keys`, in plural, is a list.

If we remove the `is_binary` guard, which I have also tried, then dialyzer complains about something else.

And, if I put an `is_tuple` guard on the second branch, then dialyzer will complain about the case of `Key` being not a binary and not a tuple.

andytill:
+1 f0a49e1

hmmr:
@borshop merge

PR #1404, opened 2016-04-26
Bugfix/listkeys filters

erikleitch:
**Context**

stream_list_keys has always returned inconsistent results when used with TS data.  I've tracked this down to the coverage filter functions that are applied during the fold over keys on a vnode.  These filters check whether a key should be included, by checking if its hash falls within a specified hash range for a vnode.  These filters are applied to the storage keys in leveldb.  For TS, however, these filters need to be applied not to the storage key (which is the local key), but instead to the partition key, which can be derived from the local key by an appropriate transformation.

This PR addresses this inconsistency by changing the filter functions to use a key conversion function on the storage key prior to hashing. It is one of three PRs: riak_kv (https://github.com/basho/riak_kv/pull/1404), riak_core (https://github.com/basho/riak_core/pull/834) and riak_pipe (https://github.com/basho/riak_pipe/pull/105).  For reference, I include my diagram of the listkeys code path here:

 RiakTS Listkeys Path

**Changes in this repo**
- **riak_kv_ts_svc.erl**<br>
  - sub_tslistkeysreq now constructs a key conversion
    function (KeyConvFn) when invoked, that takes the decoded object key, sext
    decodes it, and converts to a partition key, using riak_kv_ts_api:row_to_key with the
    Mod and DDL specific to the table for which listkeys has been invoked.  This is passed as an argument to riak_client:stream_list_keys
- **riak_kv_ts_api**<br>
  - row_to_key/3 exported because it is needed in sub_tslistkeysreq when building the key conversion function
- **riak_client**<br>
  - stream_list_keys/5 has been added, with KeyConvFn as 3rd arg, as 
    the function that all variants of stream_list_keys call down to.
    This calls riak_kv_keys_fsm_sup:start_keys_fsm with KeyConvFn as an
    extra argument in the fsm startup arg list
  - stream_list_keys/4, which in the riak_ee codebase was only called by sub_tslistkeysreq, has been modified to take KeyConvFn as its 3rd argument
  - the interfaces to stream_list_keys/2 and stream_list_keys/3, which are used  elsewhere in
    the riak_ee codebase, remain unchanged.  On invocation,
    they construct a no-op key conversion function and pass it into stream_list_keys/4
  - list_keys/4 has been modified to call start_keys_fsm with a no-op key conversion function, to ensure that list_keys also continues to work with this change
- **riak_kv_keys_fsm**<br>
  - init uses the KeyConvFn argument to construct a
    coverage plan function that returns a proplist including the key
    conversion function, in addition to the usual vnode filter list.  This will
    be needed later in riak_kv_vnode:handle_coverage_fold to construct the
    vnode filter function applied during the listkeys fold.
- **riak_kv_index_fsm.erl, riak_kv_buckets_fsm.erl**<br>
  - Because riak_core_coverage_fsm now expects a plan fn rather than a plan mod, init functions for these modules have also been modified to return a plan fun, so that we don't break range scans and bucket listing
- **riak_kv_vnode.erl**<br>
  - handle_coverage_fold has been modified to extract the key
    conversion function from the FilterVNodes proplist, and pass it into
    riak_kv_coverage_filter:build_filter, where it is used to construct an appropriate
    filter
- **riak_kv_coverage_filter.erl**<br>
  - build_filter/4 has been added to use an explicit key conversion function
  - build_filter/3 has been modified to pass a no-op key conversion function, to  preserve the functionality of other code that uses build_filter/3 directly
  - build_preflist_fun and build_subpartition_fun now take KeyConvFn as an extra argument, and use it to convert the key before hashing

**Notes**
- I have verified in manual testing that streaming list keys now works consistently for TS buckets (ie, the changes here do what they are intended to do)
- I have verified in manual testing that streaming list keys still works (and returns consistent results) for non-TS buckets (ie, the code changes here haven't broken listkeys for non-TS buckets, and incidentally that there isn't a separate problem with non-TS listkeys)
- I have verified in riak_test that verify_listkeys now passes (note that there is also a bugfix/listkeys_filters branch of riak_ee that can be used in riak_test to verify the changes)
- I still want to verify that stream_list_buckets works, but I haven't gotten to that yet

### DIFF
--- a/src/riak_kv_coverage_filter.erl
+++ b/src/riak_kv_coverage_filter.erl
@@ -152,6 +152,33 @@ build_item_filter(FilterInput) ->
     %% FilterInput is a list of {Module, Fun, Args} tuples
     compose(FilterInput).
 
+%% The functions build_preflist_fun and build_subpartition_fun are
+%% used to compute the hash and/or vnode index corresponding to a key
+%% read from the backend.
+%%
+%% For example, the function returned in build_preflist_fun is the
+%% PrefListFun called in the filter returned by build_vnode_filter
+%% above.  That filter is the Filter arg used in
+%% riak_kv_vnode:fold_fun variants to determine if a key should be
+%% added to a results buffer during a fold.
+%%
+%% These functions have been modified to accept a key conversion
+%% function, KeyConvFn, which allows for an arbitrary transformation
+%% to be performed on the storage key prior to computing a hash.  
+%%
+%% This is required for timseries keys for example, because the local
+%% key read from the backend must be transformed to a partition key to
+%% determine what hash partition it belongs to.
+%%
+%% The key conversion function is constructed once when the fold is
+%% initialized and passed into these functions to avoid incurring
+%% per-key overheads (e.g., extracting the DDL from a TS bucket every
+%% time a key is read from the backend, when instead the DDL can be
+%% retrieved once and built into the key conversion function)
+%%
+%% For ordinary keys, KeyConvFn will usually be a no-op, i.e.,
+%% KeyConvFn = fun(Key) -> Key end.  We leave it here as a generic
+%% hook to allow transformations on any type of key.
 
 %% @private
 build_preflist_fun(Bucket, CHBin, KeyConvFn) ->
--- a/src/riak_kv_coverage_filter.erl
+++ b/src/riak_kv_coverage_filter.erl
@@ -73,8 +73,8 @@ build_filter(Bucket, ItemFilterInput, {_Hash, _Mask}=SubP, KeyConvFn) ->
             SubpartitionFun = build_subpartition_fun(Bucket, KeyConvFn),
             compose_sub_filter(SubP, SubpartitionFun);
         true -> % key and vnode filtering
-            SubpartitionFun = build_subpartition_fun(Bucket),
-            compose_sub_filter(SubP, SubpartitionFun, ItemFilter, KeyConvFn)
+            SubpartitionFun = build_subpartition_fun(Bucket, KeyConvFn),
+            compose_sub_filter(SubP, SubpartitionFun, ItemFilter)
     end;
 build_filter(Bucket, ItemFilterInput, FilterVNode, KeyConvFn) ->
     ItemFilter = build_item_filter(ItemFilterInput),
--- a/src/riak_kv_coverage_filter.erl
+++ b/src/riak_kv_coverage_filter.erl
@@ -30,7 +30,7 @@
 -module(riak_kv_coverage_filter).
 
 %% API
--export([build_filter/1, build_filter/3]).
+-export([build_filter/1, build_filter/3, build_filter/4]).
 
 -export_type([filter/0]).
 
@@ -60,19 +60,23 @@
 build_filter(Filter) ->
     build_item_filter(Filter).
 
-
 -spec build_filter(bucket(), filter(), [index()]|subpartition()) -> filter().
-build_filter(Bucket, ItemFilterInput, {_Hash, _Mask}=SubP) ->
+build_filter(Bucket, ItemFilterInput, FilterVNode) ->
+    KeyConvFn = fun(StorageKey) -> StorageKey end,
+    build_filter(Bucket, ItemFilterInput, FilterVNode, KeyConvFn).
+
+-spec build_filter(bucket(), filter(), [index()]|subpartition(), function()) -> filter().
+build_filter(Bucket, ItemFilterInput, {_Hash, _Mask}=SubP, KeyConvFn) ->
     ItemFilter = build_item_filter(ItemFilterInput),
     if
         (ItemFilter == none) -> % only subpartition filtering required
-            SubpartitionFun = build_subpartition_fun(Bucket),
+            SubpartitionFun = build_subpartition_fun(Bucket, KeyConvFn),
             compose_sub_filter(SubP, SubpartitionFun);
         true -> % key and vnode filtering
             SubpartitionFun = build_subpartition_fun(Bucket),
-            compose_sub_filter(SubP, SubpartitionFun, ItemFilter)
+            compose_sub_filter(SubP, SubpartitionFun, ItemFilter, KeyConvFn)
     end;
-build_filter(Bucket, ItemFilterInput, FilterVNode) ->
+build_filter(Bucket, ItemFilterInput, FilterVNode, KeyConvFn) ->
     ItemFilter = build_item_filter(ItemFilterInput),
 
     if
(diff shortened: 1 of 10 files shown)


src/riak_kv_ts_api.erl
```diff
@@ -32,6 +32,7 @@
          get_data/2, get_data/3, get_data/4,
          delete_data/2, delete_data/3, delete_data/4, delete_data/5,
          query/2,
+         row_to_key/3,
```
hmmr: I would rather this function be moved over to `riak_kv_ts_util`. It is not really part of internal API (in the sense `get_data` and the rest are).

andytill: Isn't this a function in the riak_ql helper module?

hmmr: No, it's different. And it's a simple convenience thing: https://github.com/basho/riak_kv/pull/1404/files#diff-4f95cc846bf31c44a5533591faee0ef6L181.

src/riak_kv_coverage_filter.erl
```diff
@@ -150,26 +154,26 @@ build_item_filter(FilterInput) ->
 
 
 %% @private
-build_preflist_fun(Bucket, CHBin) ->
+build_preflist_fun(Bucket, CHBin, KeyConvFn) ->
     fun({o, Key, _Value}) -> %% $ index return_body
-            ChashKey = riak_core_util:chash_key({Bucket, Key}),
+            ChashKey = riak_core_util:chash_key({Bucket, KeyConvFn(Key)}),
```
hmmr, line 186: Theoretically, you could obtain a DDL from `Bucket` and call `riak_ql_ddl:get_partition_key` on the spot here, without the need to create that function in riak_kv_ts_svc:sub_tslistkeysreq` and tread it all the way here. This may or may not cause any performance overhead, but it would make the diff a lot smaller. I'm not urging you to do it this way, though; just, what were your arguments for the design decision you have taken?

src/riak_kv_coverage_filter.erl
```diff
@@ -150,26 +154,26 @@ build_item_filter(FilterInput) ->
...
             chashbin:responsible_index(ChashKey, CHBin);
        ({_Value, Key}) ->
-            ChashKey = riak_core_util:chash_key({Bucket, Key}),
+            ChashKey = riak_core_util:chash_key({Bucket, KeyConvFn(Key)}),
```
hmmr, line 189: Is it necessary to apply the identity function on non-TS keys? The `KeyConvFn` here can only be the dumb `fun(X) -> X end`, can it?

src/riak_kv_ts_svc.erl
```diff
@@ -332,9 +332,28 @@ sub_tsdelreq(Mod, _DDL, #tsdelreq{table = Table,
...
+
+    %% Construct a function to convert from TS local key to TS
+    %% partition key.
+    %%
+    %% This is needed because coverage filter functions must check the
+    %% hash of the partition key, not the local key, when folding over
+    %% keys in the backend.
+
+    KeyConvFn = 
+	fun(Key) when is_binary(Key) ->
```
gordonguthrie: This code is indented with tabs - should be spaces as per Basho standard...

andytill: This function should be moved out into a module level function and then made into a fun using the function reference syntax e.g. `fun ts_row_to_key/1`. Function references show what the function is when the fun term is displayed.

```erlang
1> fun(Fmt, Args) -> io:format(Fmt, Args) end.
#Fun<erl_eval.12.80484245>
2> fun io:format/2.
#Fun<io.format.2>
```

It is much easier to work out what the fun is using the function reference style when debugging.

src/riak_kv_coverage_filter.erl
erikleitch, line 186: Design was predicated on two things:
1. I didn't want the performance overhead of constructing the DDL _every time a key is evaluated_.  This is unnecessary, and doesn't scale.  Rather I wanted to construct a simple conversion function once, where the DDL was already available.
2. build_preflist_fun and build_subpartition_fun  shouldn't have to know anything about details of the bucket, in particular whether or not it is a TS bucket.  Although I guess we have this in the fn as a pattern match on {o, Key, Value}, like the rest of the riak_core_coverage_fsm/riak_kv_coverage_filter code, I wanted to preserve as much as possible the model of doing TS-specific things in TS-specific code, and having the rest of the codepath be TS/KV-agnostic.  My feeling is that the more in-place checks for TS vs non-TS buckets and keys and encodings we pepper throughout the codebase, the more unmaintainable it becomes.

src/riak_kv_coverage_filter.erl
erikleitch, line 189: Yes, this is true.  I modified this code before I remembered that {o, Key, _Value} was a TS-specific pattern match, but decided to leave it in as a generic hook to allow transformations on any type of key.  

However, one thing we do need to think about is whether we are allowing mixed TS and non-TS keys to be written to a TS bucket, and what happens when we do.  In that case, we may actually need to differentiate between types of keys encountered in the same bucket, and only do transforms on one of them.

I'll investigate what happens if I try to write both types of keys to a TS bucket... that will inform what we do here.

src/riak_kv_ts_svc.erl
```diff
@@ -332,9 +332,28 @@ sub_tsdelreq(Mod, _DDL, #tsdelreq{table = Table,
...
+    %%
+    %% This is needed because coverage filter functions must check the
+    %% hash of the partition key, not the local key, when folding over
+    %% keys in the backend.
+
+    KeyConvFn = 
+        fun(Key) when is_binary(Key) ->
+                riak_kv_ts_api:row_to_key(sext:decode(Key), DDL, Mod);
+           (Key) ->
+                %% Key read from leveldb will always be binary.  This
```
gordonguthrie: This path should be logged

erikleitch:
Have made the following changes in response to review comments:
- added extensive notes to riak_kv_coverage_filter:build_preflist_fun/build_supartition_fun to explain why key conversion was necessary and why we decided to do things this way, for future reference
- added error logging in riak_kv_ts_svc:sub_tslistkeysreq KeyConvFn if unexpected branch is ever exercised
- Moved row_to_key from riak_kv_ts_api to riak_kv_ts_util module, and changed references in riak_kv_ts_svc and riak_kv_ts_api to point to that function

Tests:
- I have tested the updated versions of riak_kv and riak_core in riak_test, and confirmed that they behave as expected.  
- I have also tested mixing KV and TS keys into the TS bucket, to see if the key conversion code breaks things.  The upshot is that mixing KV and TS keys _does_ break things, but it is not the key conversion code that is the source of the error, it is the sext:decode() and subsequent riak_pb_ts_codec:encode_rows calls in riak_kv_ts_svc:process_stream that break things.  This function processes the keys returned from the fold before sending to the client, and it expects all keys to be sext-encoded TS keys.

hmmr:
@borshop retry 7754045

hmmr:
@borshop: retry 7754045

hmmr:
@borshop merge

hmmr:
@borshop merge

hmmr:
+1 7754045

hmmr:
+1 49c0a34

hmmr:
@borshop merge

PR #1406, opened 2016-04-28
Remove unit tests where the quantum is not the last element

andytill:
Remove unit tests where the quantum is not the last element in the partition key. Also where the quantum field is not type timestamp.

This is no longer allowed.


hmmr:
+1  0dc13ea

PR #1409, opened 2016-05-01
Handle case where there is no key_conv_fn defined

macintux:
`riak_kv_vnode:handle_coverage_fold/8` may not have a `key_conv_fn` option supplied, which results in a serious 2i failure without this fix.


hmmr:
It is papering over a deeper issue with index_fsm, for which a fix is proposed in #1410.

It can be merged as a last resort, but it will only help to avoid the crash in secondary index listing, with these symptoms:
`{error,\n    {worker_crash,\n        {{badfun,undefined},\n         [{riak_kv_coverage_filter,'-build_preflist_fun/3-fun-0-',4,\n ...`

hazen:
+1 eae9c78

PR #1410, opened 2016-05-02
supply lk2pk key conv fun for index_fsm init, to fix 2i listing

hmmr:
As detailed by @erikleitch in a3076a22b8, the coverage plan for list_keys operation in TS was computed incorrectly because it must use PK and not LK. The proposed fix used a function to derive PK from the LK (immediately available in TS object being iterated), passed to the call site of `riak_kv_coverage_filter:build_filter` (and further to `:build_preflist_fun`) from the caller of `riak_client:list_keys`.

Because index_fsm facilities (as well as keys_fsm and bucket_fsm) are shared between TS and standard riak, and because 2i functionality is not enabled in TS, the fix by Erik was OK for the TS uses but intrusive enough to subtly break things for standard riak operations.

Following up and complementing this work, this commit:
1. unbreaks 2i list_keys by supplying an identity fun for the non-TS uses of index_fsm;
2. creates the key conversion function to build the correct coverage plan in queries with pre-ordered coverage contexts (possibly fixing the sporadic failures of ts_cluster_coverage).


macintux:
This PR is insufficient because 2i requests with pre-calculated coverage plans do not invoke the custom `create_plan` functionality implemented in `riak_kv_index_fsm`.

macintux:
Will give this a +1 once my PR is merged.

macintux:
:+1: f65b4c2

macintux:
yolo-merged

PR #1414, opened 2016-05-06
Refactor riak_kv_qry_worker

andytill:
Refactor of `riak_kv_qry_worker`.
- Remove the unused `#state.status` field that was written to but never checked.
- Remove the unused `#state.name` field that was written to but never checked. This is the registered name for the process and available from it's process info.
- Functions to extract data from the `?SQL_SELECT` records to clean up function headers.
- Extract code into it new functions e.g. in case statement bodies.
#### Motivation

Remove cruft and prepare for all the changes that will arrive for temporary tables and streaming queries.

There are more changes in this style in the streaming queries WIP https://github.com/basho/riak_kv/pull/1415.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -168,7 +168,6 @@ make_key_conversion_fun(Table) ->
             Key
     end.
 
-
 decode_results([]) ->
     [];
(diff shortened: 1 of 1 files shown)


andytill:
@borshop merge

PR #1417, opened 2016-05-10
better is_crdt checks

zeeshanlakhani:
- make is_crdt check usable for deps or times when additional values/sibs are involved
- add more exports for diff. use-cases in external applications


JeetKunDoug:
+1 9adc0bd

zeeshanlakhani:
@borshop merge

Issue #1178, opened 2015-09-02
get/put FSM supervisor PID leak when overload protection is enabled [JIRA: RIAK-2157]


ian-mi:
Calls to `riak_kv_get_fsm_sup:start_get_fsm` leave abandoned PIDs in the `riak_kv_get_fsm_sup` state because FSMs are linked to a sidejob supervisor rather than the get FSM supervisor. A similar issue applies to the put FSM supervisor. This can cause extended shutdown times as the supervisor attempts to iterate through millions of dead PIDs.

jonmeredith:
Do you have a reproducer or customer ticket this came from?

ian-mi:
I've reproduced the issue by doing the following from riak attach on a new node:

supervisor:count_children(riak_kv_get_fsm_sup).
[{specs,1},{active,0},{supervisors,0},{workers,0}]

riak_kv_get_fsm_sup:start_get_fsm(node(), [<<"">>, <<"bucket">>, <<"key">>, []]).
{ok,<0.5520.0>}

process_info(list_to_pid("<0.5520.0>")).
undefined

supervisor:count_children(riak_kv_get_fsm_sup).  
[{specs,1},{active,0},{supervisors,0},{workers,1}]

jonmeredith:
The riak_kv_get_fsm_sup should be deprecated. The only consumer I could find is `riak_repl_util:reap/3` called in `riak_repl_util:do_repl_put/3` which could be modified to just call start_link instead.

If using sidejob it would obey overload protection which is a good thing and not create a link to the calling process.  The case where overload protection is disabled is more tricky - if the FSM crashes it would be linked to the caller.  `reap/3` could call unlink, but there is a race.

We should check with cliserv if there has been a case where we have disabled sidejob.  It was a 'get-out-of-jail' option while the code proved itself in production, so we could clean all of that stuff up to just require sidejob.

/cc: @JeetKunDoug

ian-mi:
It seems incorrect that riak_kv_get_fsm:start_link does not link to the caller when sidejob is enabled as that breaks the expected semantics of start_link. One fix I've been thinking of is to manually call link when sidejob is used. However, the reason start_link has been used seems to be just an artifact of avoiding the get_fsm_sup for performance reasons. Based on my discussions with others in cliserv, we have not seen anyone disabling the FSM limit, so I agree that the cleanest solution would be to always use sidejob and perhaps rename the start function to reflect that it does not link to its caller.

[...]

PR #1420, opened 2016-05-19
Clean up get/put FSM start code and remove non-sidejob code paths

nickelization:
This PR removes a bunch of code that allowed us to run riak_kv with sidejob disabled. None of our customers run with overload protection disabled, and now that sidejob has been pretty thoroughly vetted in production, we no longer need an escape hatch. Removing the non-sidejob code (as well as associated test code) will cut down on risk and maintenance costs. Performance should also be improved a little bit since we avoid having to do the extra check for sidejob vs non-sidejob when starting a get/put FSM.

These changes also remove the obsolete riak_kv_get_fsm and riak_kv_put_fsm supervisors, and renames the get/put FSM start_link functions to "start", since they don't actually link to the calling process. (Although the start_link function name is kept as an alias to avoid any potential problems during rolling upgrades.)


nickelization:
Also see basho/riak_kv#1178 (RIAK-2157) (RIAK-2567) for more background. We'll also have to update repl to not use the get FSM supervisor, and instead just start a get FSM directly, but that'll be a nice easy one-line fix.

jadeallenx:
Nice work

nickelization:
Thanks!

ian-mi:
Looks good to me. I'm glad you cleaned up the naming; start_link is currently a lie. I also approve of tearing out the get / put monitor code.

This might be a good opportunity to examine the riak_client behaviour as well. It seems unnecessary to use proc_lib:spawn_link to call the fsm start function rather than making a synchronous call. Perhaps the reason for this is outdated?

nickelization:
@ian-mi Thanks! Yeah, I'm not sure about the riak_client code, that's a good point. I'd like to look into that but I think due to time constraints I probably won't get to it until after 2.2 is out. I'll put it on my TODO list though for sure.

If all the changes look good, could I get a +1 so I can go ahead and merge this? Once this is merged I'll also go into riak_repl and add a PR to switch it over to use the new API instead of starting things under the now-defunct FSM supervisor.

ian-mi:
Sounds good.

+1 3cea5ae

bsparrow435:
+1 3cea5ae202e9bdda72bab4cb0649c2c7265869ad

nickelization:
@borshop merge

PR #1423, opened 2016-05-25
give reasons why 'process/2' takes SQL records and not #tsqueryreq{}

hmmr:
`riak_kv_ts_svc:process/2` has heads matching on `#tssomethingreq{}`, as it does elsewhere, but also four heads matching on `#riak_sql_querytype{}`. This cheeky wee PR adds a comment detailing why this is so.

### DIFF
--- a/src/riak_kv_ts_svc.erl
+++ b/src/riak_kv_ts_svc.erl
@@ -136,7 +136,11 @@ process(M = #tslistkeysreq{table = Table}, State) ->
 process(M = #tscoveragereq{table = Table}, State) ->
     check_table_and_call(Table, fun sub_tscoveragereq/4, M, State);
 
-%% this is tsqueryreq, subdivided per query type in its SQL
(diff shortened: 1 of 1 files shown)


src/riak_kv_ts_svc.erl
```diff
@@ -136,7 +136,11 @@ process(M = #tslistkeysreq{table = Table}, State) ->
 process(M = #tscoveragereq{table = Table}, State) ->
     check_table_and_call(Table, fun sub_tscoveragereq/4, M, State);
 
-%% this is tsqueryreq, subdivided per query type in its SQL
+%% The following heads of `process` are all, in terms of protobuffer
```
macintux: Let's do the Erlang edoc thing, even though this won't be an edoc, and use the proper backtick/apostrophe combination around process, as well as #tsqueryreq{} and other literals below.

macintux:
:+1:    dfb8453

hmmr:
@borshop merge

PR #1424, opened 2016-05-26
in wm_timeseries_listkeys: urlencode ' ' as %20, not as a '+'

hmmr:
RTS-1106

For the purposes of percent-encoding in the output of "http://server/ts/table/T/list_keys", we use `mochiweb_util:quote_plus/2`. Normally, though, that function will encode a space as a plus sign, which happens to be undesirable. In this PR, we borrow that function and tweak it to don't take exceptions for `' '` and encode the space as %20.

Let @christophermancini and @lehoff be the reviewers.

### DIFF
--- a/src/riak_kv_wm_timeseries_listkeys.erl
+++ b/src/riak_kv_wm_timeseries_listkeys.erl
@@ -178,7 +178,7 @@ format_url(BaseUrl, KeyTypes, Key) ->
 
 key_to_string(KFTypes) ->
     string:join(
-      [[Field, $/, mochiweb_util:quote_plus(value_to_url_string(Key, Type))]
+      [[Field, $/, http_uri:encode(value_to_url_string(Key, Type))]
        || {Key, {Field, Type}} <- KFTypes],
       "/").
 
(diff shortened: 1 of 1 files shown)


src/riak_kv_wm_timeseries_listkeys.erl
```diff
@@ -189,6 +189,34 @@ value_to_url_string(V, sint64) ->
 value_to_url_string(V, timestamp) ->
     integer_to_list(V).
 
+%% `urlencode2' is copied from mochiweb_util:quote_plus/2 and modified
+%% to convert a space to %20 (not to a '+')
+-define(PERCENT, 37).  % $\%
+-define(FULLSTOP, 46). % $\.
+-define(IS_HEX(C), ((C >= $0 andalso C =< $9) orelse
```
andytill: Is `?IS_HEX` used?

andytill:
Requires a common test.

andytill:
`http_uri:encode/1` does this and is in the standard library. From my shell.

```
2> http_uri:encode(" ").
"%20"
```

andytill:
eunit tested instead of ct.

andytill:
+1 55a0fa1

hmmr:
@borshop merge

PR #1425, opened 2016-05-26
TS coverage replacement

macintux:
Corresponding test: basho/riak_test#1078
- New coverage query document
- Support TS coverage chunk replacement
- Support future normalization of coverage chunk data formats between KV and TS

See `docs/coverage-queries.md` for much more detail.

### DIFF
--- a/docs/coverage-queries.md
+++ b/docs/coverage-queries.md
@@ -1,0 +1,254 @@
+## Context
+
+Historically Riak's coverage plans were strictly internal constructs
+to support 2i, mapreduce, and similar mechanisms that needed to know
+the minimal set of servers required to extract a single copy of every
+key in the system.
+
+With Spark integration for Basho's data platform and timeseries, this
+became an externally-accessible API so that data could be queried in
+parallel by clients, bypassing Riak's usual central coordination of
+bulk queries.
+
+The high level flow:
+
+1. Client asks for a segmented coverage plan.
+    * For timeseries, a SQL query must be included with the request;
+      coverage plans for timeseries are **not** independent of the
+      data, unlike Riak KV.
+2. Each chunk that is returned contains an opaque binary to be sent
+   back to Riak and a description of the node for which that chunk is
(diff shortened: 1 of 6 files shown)


hmmr:
I could not find anything demonstrably wrong in this PR set. The test is well documented, and is passing.

I was specifically looking to see how the cover replacement deals with the situation where the vnode the user claims to be unreachable (by mentioning it in the `Unavail` argument to `riak_kv_ts_util:sql_to_cover`, is actually _not_ unavailable (that is, the user's claim is false). The answer is, that vnode is simply considered to be down (https://github.com/basho/riak_kv/pull/1425/files#diff-97ee51daf178d2355af3f4adc560c04fR54), and no attempt is made to verify the user's claim or prove the user wrong. I am not sure this behavior is not correct, but it may be worth a five-minute effort to add a statement in `docs/cover-queries.md` to clarify that the user's `DownNodes` may actually be all well up.

And secondly, concerning the documentation, I think the `Riak client (..)` sections should better be moved to the riak-erlang-client repo. Should any changes happen in future in either repo, such changes will then be self-contained and come in a single PR, instead of fixing code in the client and fixing the documentation in the riak_kv repo, resulting in two PRs.

macintux:
The reason we don't verify the unavailable server(s): just because Riak can see them, doesn't mean the client can talk to them. Network partitions are weird beasts.

macintux:
And in this case, "Riak client" is `riak_client.erl` in KV, not the remote Erlang client, so that doc section stays here.

srgg:
The following can't be treated as a blocker for the 'liar' detection mentioned by @hmmr, it is  just another 2 cents: we are speculating on this trusted behavior in the Java Client and Spark Connector as well,  it allows us to simplify our failover tests.

hmmr:
+1 197625d

hmmr:
+1 062b4d7

macintux:
@borshop merge

PR #1426, opened 2016-05-27
Fix broken handle_cast

JeetKunDoug:
Was returning a reply message, as it was a handle_call before.


fadushin:
+1 3ffe420

JeetKunDoug:
@borshop merge

PR #1427, opened 2016-05-31
correctly convert LK to PK for coverage plan calculation in folding fun for list_keys

hmmr:
RTS-1126

This is #1404 done right (#1410 was a fixup for some other regression in 1404).

A r_t module for this issue is in https://github.com/basho/riak_test/pull/1081.

We now use a newly written function `lk_to_pk`, which does what its name suggests, and which is a pared down version of `riak_kv_ts_util:make_ts_keys/3`. It will be used as key conversion function for proper coverage plan calculation instead of `riak_kv_ts_util:row_to_key/3`, in list_keys code path.

As its argument, the old key conversion function (`row_to_key`) used to take the LK values while it logically treated the argument as a list of _all_ values of a TS record. For the (common) case where key fields all appear at the beginning of DDL, and they do so in the order they are defined, `row_to_key` happens to work. It will break, however, when that condition does not hold.

@erikleitch and/or @andytill (optionally also @macintux) will make great reviewers of this PR.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -161,7 +161,7 @@ make_key_conversion_fun(Table) ->
     Mod = riak_ql_ddl:make_module_name(Table),
     DDL = Mod:get_ddl(),
     fun(Key) when is_binary(Key) ->
-            riak_kv_ts_util:row_to_key(sext:decode(Key), DDL, Mod);
+            riak_kv_ts_util:lk_to_pk(sext:decode(Key), DDL, Mod);
        (Key) ->
             lager:error("Key conversion function "
                         "encountered a non-binary object key: ~p", [Key]),
--- a/src/riak_kv_ts_svc.erl
+++ b/src/riak_kv_ts_svc.erl
@@ -350,7 +350,7 @@ sub_tslistkeysreq(Mod, DDL, #tslistkeysreq{table = Table,
 
     KeyConvFn =
         fun(Key) when is_binary(Key) ->
-                riak_kv_ts_util:row_to_key(sext:decode(Key), DDL, Mod);
+                riak_kv_ts_util:lk_to_pk(sext:decode(Key), DDL, Mod);
            (Key) ->
                 %% Key read from leveldb should always be binary.
                 %% This clause is just to keep dialyzer quiet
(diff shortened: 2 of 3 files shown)


andytill:
If the partition key is always a prefix of the local key, can't we just take the first part of the local key? The length can be found in the number of key ast elements in the DDL.

This also needs some ct tests.

hmmr:
@andytill We still need the full record as this is what `riak_ql_ddl:get_partition_key/3` accepts as argument. It is in that function that the real timestamp value gets reset to the beginning of the quantum in which it belongs. The patch is not so much about truncating LK fields to match PK length but rather resetting the timestamp.

As for the tests, I used this branch for testing: https://github.com/basho/riak_test/commit/508174e71c, but I'm going to prepare a dedicated and more thorough test.

andytill:
Yeah, it seems like a lot of work has to be done to convert the key to a format which `riak_ql_ddl:get_partition_key/3` accepts.

Any comments on how this fits in with descending keys?

There is a streaming list keys test in the descending keys branch for riak_test here https://github.com/basho/riak_test/pull/1055

erikleitch:
+1 

This PR creates a new function riak_kv_ts_util:lk_to_pk to derive the
partition key from the local key fields, required for converting local
keys to partition keys for comparison with subpartition filters during
a streaming listkeys fold.

lk_to_pk uses the DDL to construct a dummy object containing all
fields in order, but with LK fields filled in from the actual local
key and void fields for all others.  This creates an object that is
appropriate for passing to riak_ql_ddl:get_partition_key().

I've verified that this PR does indeed fix the problem I originally
described in https://bashoeng.atlassian.net/browse/RTS-1126.  (Note:
you need a cluster with nval > 1 and populated with data spanning
multiple partitions to see the problem in the first place, since no
filter at all is constructed otherwise).  

I've tested both scenarios where the PK fields are a subset of the LK, 
and where the PK fields use fields in a different order than defined in 
the schema.

**NB** lk_to_pk will still have to reverse any negated
fields to work with descending-key schema, since the values are
negated when the local key is written to disk, but should _not_ be
negated when creating a partition key from them.  I think this will
require some extra work to ensure that the dummy object fields are
correctly ordered before passing to riak_ql_ddl:get_partition_key.

hmmr:
@andytill There's a comment on possible implications the descending keys may have on the correct lk2pk calculation: https://github.com/basho/riak_ql/pull/117#issuecomment-228019822.

hazen:
+1 09d2f67

hmmr:
@borshop merge

PR #1430, opened 2016-06-30
Recognize TS data arriving via handoff

macintux:
Currently, TS data that arrives via handoff is not stored under a recognizable key.

The key is sext-encoded twice: once as pulled from disk on the sending side, then again when sent to the backend on the receiving side.

This patch addresses that problem, although it deserves more thought for the next major release.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -2383,6 +2383,13 @@ encode_and_put_no_sib_check(Obj, Mod, Bucket, Key, IndexSpecs, ModState,
                         false ->
                             ok
                     end,
+
+                    %% Timeseries data arrives during handoff with
+                    %% sext-encoded local keys. Unless care is taken,
+                    %% those keys are doubly-encoded, so here we will
+                    %% identify them and unencode the key.
+                    %%
+                    %% Not the ideal solution but sufficient.
                     PutRet = case timeseries_data(Key, Obj) of
                                  true ->
                                      RawKey = sext:decode(Key),
@@ -2396,10 +2403,13 @@ encode_and_put_no_sib_check(Obj, Mod, Bucket, Key, IndexSpecs, ModState,
             end
     end.
 
-%% Look for sext-encoded tuple (key) because handoff lifts it straight out of
-%% the backend and gives it to us as is. If this isn't sext-encoded
-%% then this ain't timeseries so we don't need to dig through the
-%% object metadata for a DDL.
+%% This function takes a key and the riak object from the handoff call
+%% chain and verifies whether or not the object is from timeseries.
+%%
+%% The binary pattern match is quick and easy: does this look like a
+%% sext-encoded tuple?
+%%
+%% Only if that is true do we examine the object for DDL metadata.
 timeseries_data(<<16,0,0,0, _Rest/binary>>, Object) ->
     {IsTs, _Ver} = riak_object:is_ts(Object),
     IsTs;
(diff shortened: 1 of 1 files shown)


src/riak_kv_vnode.erl
```diff
@@ -2383,12 +2383,29 @@ encode_and_put_no_sib_check(Obj, Mod, Bucket, Key, IndexSpecs, ModState,
...
                     {PutRet, EncodedVal}
             end
     end.
 
+%% Look for sext-encoded tuple (key) because handoff lifts it straight out of
+%% the backend and gives it to us as is. If this isn't sext-encoded
+%% then this ain't timeseries so we don't need to dig through the
+%% object metadata for a DDL.
+timeseries_data(<<16,0,0,0, _Rest/binary>>, Object) ->
+    {IsTs, _Ver} = riak_object:is_ts(Object),
```
paegun: The comment above says "we don't need to dig ... for a DDL", but riak_object:is_ts/1 does just that. The cost of "digging" pales in comparison to losing data.

If we are 100% certain that only TS data is handed off as is, then the comment should remain and the verification that the object contains TS data should be removed. Otherwise, the comment should be removed and the verification should remain.

Besides the comparison between plain-old KV objects, we should consider CRDT objects.

I recommend we err on the side of preserving data, but jump on testing the various object types. We can more readily test the performance impact of digging into the riak_object for DDL on handoff.

macintux: I'll clarify the comment. The point is that we _do_ check for a DDL via `is_ts/1` but we only do that if the key looks like it's a sext-encoded tuple.

Does that alleviate your concern? I don't think any code changes are warranted, simply improving the commentary.

paegun:
+1 abc5050

src/riak_kv_vnode.erl
```diff
@@ -2383,12 +2383,39 @@ encode_and_put_no_sib_check(Obj, Mod, Bucket, Key, IndexSpecs, ModState,
...
+                    %% Timeseries data arrives during handoff with
+                    %% sext-encoded local keys. Unless care is taken,
+                    %% those keys are doubly-encoded, so here we will
+                    %% identify them and unencode the key.
+                    %%
+                    %% Not the ideal solution but sufficient.
+                    PutRet = case timeseries_data(Key, Obj) of
+                                 true ->
+                                     RawKey = sext:decode(Key),
+                                     Mod:put(Bucket, RawKey, IndexSpecs, EncodedVal,
```
JeetKunDoug: Discussing w/ Torben this morning - we believe that using sync_put in this case is most likely a premature optimization and may end up causing us grief (scheduler collapse anyone?). Would much prefer to see this case statement removed and create a function that just gets the final key for put using the timeseries_data function and sext decoding:

``` erlang
get_final_key(<<16,0,0,0, _Rest/binary>> = Key, Object) ->
    case riak_object:is_ts(Object)  of
        {true,_ } ->
           sext:decode(Key);
        _ ->
           Key
   end;
get_final_key(Key, _) ->
    Key.
```

@erikleitch have you made any final decisions about sync_put vs. put here?

hmmr:
Please also create a separate PR of this branch against `riak_ts-develop-1.3`.

src/riak_kv_vnode.erl
macintux: Changes under testing

macintux:
Ran some simple tests with KV keys mirroring the `sext` keys we're looking for, and everything seems correct after the latest commit.

hmmr:
+1 8180b04

macintux:
@borshop merge

PR #1437, opened 2016-07-12
Feature/gsets [JIRA: RIAK-2691]

binarytemple-external:
@russelldb - first of the pull requests to enable gset support.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -28,7 +28,7 @@
         {eper, ".*", {git, "git://github.com/basho/eper.git", {tag, "0.78"}}},
         {sext, ".*", {git, "git://github.com/basho/sext.git", {tag, "1.1p3"}}},
         {riak_pipe, ".*", {git, "git://github.com/basho/riak_pipe.git", {branch, "2.0"}}},
-        {riak_api, ".*", {git, "git://github.com/basho/riak_api.git", {branch, "2.0"}}},
-        {riak_dt, ".*", {git, "git://github.com/basho/riak_dt.git", {branch, "2.0"}}},
+        {riak_api, ".*", {git, "git://github.com/basho/riak_api.git", {branch, "feature/gsets"}}},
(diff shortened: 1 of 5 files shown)


zeeshanlakhani:
@binarytemple-bet365 maybe it wasn't mentioned in the HC, but I don't think this will hit the 2.0 series _officially_, but I'll let @JeetKunDoug, @fadushin refer to which branch it should hit off before we review.

binarytemple-external:
Thanks @zeeshanlakhani

rebar.config
```diff
@@ -28,7 +28,7 @@
         {eper, ".*", {git, "git://github.com/basho/eper.git", {tag, "0.78"}}},
         {sext, ".*", {git, "git://github.com/basho/sext.git", {tag, "1.1p3"}}},
         {riak_pipe, ".*", {git, "git://github.com/basho/riak_pipe.git", {branch, "2.0"}}},
-        {riak_api, ".*", {git, "git://github.com/basho/riak_api.git", {branch, "2.0"}}},
-        {riak_dt, ".*", {git, "git://github.com/basho/riak_dt.git", {branch, "2.0"}}},
+        {riak_api, ".*", {git, "git://github.com/basho/riak_api.git", {branch, "feature/gsets"}}},
```
fadushin, line 31: These two dependencies need to be on a non-feature branch.

zeeshanlakhani, line 31: @fadushin maybe we can just focus and give @binarytemple-bet365 info how which branch to make this PR against too?

fadushin:
Should PR off develop-2.2.  Also, looking at the diff set here, the same needs to be done for riak_api and riak_dt (looks like there are already PRs into 2.0).

bashopatricia:
create jira issue

rebar.config
binarytemple-external, line 31: @fadushin ping

binarytemple-external:
re-targeting against 2.0-develop. closing for now.

PR #1438, opened 2016-07-14
RTS-1005: Add Quantum Data to DESCRIBE command

hazen:
Rather than adding additional PBC records, simply add two additional columns with metadata about the quantum column.

Two new columns are added: `Interval` and `Unit`.


paegun:
+1 2a1cd64

Looks great w/i riak-shell. The quantum display fields Interval and Unit clearly describe the quantum hash function as long as the units for time do not overlap w/ future support for geo, which should be the case. If not, adding a display field for hash function type is as easy as adding the Interval and Unit fields here.

hazen:
@borshop merge

PR #1439, opened 2016-07-17
RTS-1160: Build the necessary SQL record when a string is supplied

hazen:
This fixes `riak_kv_query:submit/2` for the string case.

### DIFF
--- a/src/riak_kv_qry.erl
+++ b/src/riak_kv_qry.erl
@@ -62,7 +62,10 @@ submit(SQLString, DDL) when is_list(SQLString) ->
             Error;
         {'EXIT', Reason} ->  %% lexer problem
             {error, Reason};
-        {ok, SQL} ->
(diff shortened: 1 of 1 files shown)


paegun:
+1 ce18b37

Manually tested by loading the modified beam and following the steps to repro:
0. precondition: a table named tmetric exists. (The table name does not matter, just using a name for consistency w/i the steps.) 
1. w/i attached riak shell, execute the following:

``` erlang
M1 = riak_ql_ddl:make_module_name(<<"tmetric">>).
D1 = M1:get_ddl().
Res = riak_kv_qry:submit("describe tmetric", D1).
```

Prior to the fix, the lex/parse step is repeated. With the fix, the result (Res) matches the expected value, { ok, {[ ... describe table columns/rows ... ]}} .

hazen:
@borshop merge

PR #1440, opened 2016-07-18
RTS-1149: Add missing query_insert to riak_kv_ts_api:api_call_to_perm/1

hazen:


andytill:
Should we differentiate between put and insert?

hazen:
I think so. We differentiate between get and query, right?

andytill:
We do but I don't think it is correct, can't think of any use cases where someone would need put privileges but not insert privileges.

hazen:
My understanding based on @ph07's comments in today's mumble that we want to keep this level of granularity

hazen:
Naturally we also need #1443

hmmr:
+1 171249c

macintux:
Rejecting this approach in favor of #1441. Will close when testing is done

hazen:
@macintux Awesome time for me to grab lunch.  We need to discuss now.

PR #1441, opened 2016-07-18
converge permissions for INSERT query onto "riak_ts.put"

hmmr:
Because an INSERT query is essentially a batch put, the two API calls should be identified by a single permission atom (`"riak_ts.put"`).

@javajolt can deal with in a snap.

### DIFF
--- a/src/riak_kv_ts_api.erl
+++ b/src/riak_kv_ts_api.erl
@@ -64,7 +64,7 @@ api_call_to_perm(put) ->
 api_call_to_perm(delete) ->
     "riak_ts.delete";
 api_call_to_perm(listkeys) ->
-    "riak_ts.listkeys";
+    "riak_ts.list_keys";
 api_call_to_perm(coverage) ->
     "riak_ts.coverage";
 api_call_to_perm(query_create_table) ->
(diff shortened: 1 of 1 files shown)


macintux:
I like this approach but there are more changes needed to fix the `list_keys` problem. E.g. https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_wm_timeseries_listkeys.erl#L91

hmmr:
`riak_kv_ts_api:api_call_to_perm(listkeys)` now correctly returns `"riak_ts.list_keys"`. What else can be the problem here?

macintux:
You're right, I wasn't thinking clearly.

It would be nice to see the string and atom be consistent, but I can live with it.

hazen:
@hmmr We should be sure that `listkeys` is changed to `list_keys` everywhere else

hazen:
In fact, I would prefer to see this as two separate PRs: one for insert permission and the other for the listkey name change

hmmr:
@javajolt The spelling for this atom is corrected in https://github.com/basho/riak_kv/pull/1443.

macintux:
:+1: 741257a

macintux:
@borshop merge

Issue #1188, opened 2015-09-17
Put FSM Redirect Retries and Siblings [JIRA: RIAK-2191]


shino:
### Symptom and Cause

Single put from client may create many, for example O(10), siblings
when a node in cluster is too slow or almost frozen. No specific
options in put is required.

Another flavor is non-zero PW option is used and some nodes are
stopped. In this case, siblings are not created, but many Put FSMs are
spawned.

These are caused by Put FSM coordinator redirection and ack logic.
Redirections may occur for redirected requests and/or ack is not
sent in error responses to clients.
### Workaround or only Mitigation
- Avoid redirection retries by `riak_kv.retry_put_coordinator_failure=false`.
- [Any good idea? Please add them to here :)]
### Diagram for slow node case

Assuming:
- Put is attmpted without specific options.
- Primary vnodes are on node P, Q and R, first fallback is on node S.
- node-R is too slow or almost frozen.
- Choosing redirect target node is randomized in actual code, the last
  node is always chosen in the diagram below as one of possible sequences.
  (maybe it's worst case)

```
client        node-A              node-R         node-S
   ---(Put)-->
             Compute PL
               = P, Q and R
             Redirect to R --->  [frozen]
             |
             | 3 sec timeout
             V
             Compute new PL excluding R
               = P, Q and S
             Redirect to S --------------------> Compute PL without
             |                                     any knowlege about R (at this point)
             |                                     = P, Q and R
             |                                   Redirect to R  ---+
             |                                   |                 |
             |                 [what happnes?] <-|-----------------+
             |                                   | 3 sec timeout
             |                                   V
             |                                   Compute new PL excluding R
             |                                     = P, Q and S
             |                                   I'm coordinator this time
             |                                   Execute put
             V 3 sec timeout
             Compute new PL again
               [continues]
```

Steps to reproduce: https://gist.github.com/shino/b65ff551aa5e5a6750a2
### Diagram for Non-zero PW with some stopped nodes

Assuming:
- Put is attmpted with option `PW=all`
- Primary vnodes are on node P, Q and R, first fallback is on node S.
- node-R is stopped, every ohter node know about it is stopped
- Choosing redirect target node is randomized in actual code, the last
  node is always chosen in the diagram below as one of possible sequences.

```
client        node-A              node-R                  node-S
   ---(Put)-->                    [stooped]
             Compute PL
               = P, Q and S
             Redirect to S -----------------------------> Compute PL
             |                                              = P, Q and S
[...]

PR #1442, opened 2016-07-18
Remove redudant ack_execute option.

ian-mi:
The ack_execute option is already added by make_ack_options when the
corresponding capability exists. Adding the option here is redundant and
causes unnecessary acks to be sent.


JeetKunDoug:
Nice catch - we also found it working on #1188 (RIAK-2191) and will fix it there, so I'm going to close this PR.

PR #1444, opened 2016-07-22
Allow ISO 8601 strings as timestamps for select and insert

macintux:
Ready for review. Requires https://github.com/basho/riak-erlang-client/pull/303 and https://github.com/basho/riak_shell/pull/43 for testing

### DIFF
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -153,7 +153,7 @@ timestamp_to_epoch(Str) ->
                    jam:compile(
                      jam_iso8601:parse(Str), ProcessOptions)
                   ), second),
-    jam:to_epoch(Datetime, 1000).
+    jam:to_epoch(Datetime, 3).
 
 %% Useful key extractors for functions (e.g., in get or delete code
 %% paths) which are agnostic to whether they are dealing with TS or
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -146,11 +146,14 @@ timestamp_to_epoch(Str) ->
         [
          {default_timezone,
           riak_core_metadata:get({riak_core, timezone},
-                                 string, [{default, "+00"}])},
-         {target_accuracy, second}
+                                 string, [{default, "+00"}])}
         ],
-    {_, DT} = jam:normalize(jam:process(jam_iso8601:parse(Str), ProcessOptions)),
-    jam:to_epoch(DT, 1000).
+    Datetime = jam:expand(
+                 jam:normalize(
+                   jam:compile(
+                     jam_iso8601:parse(Str), ProcessOptions)
+                  ), second),
+    jam:to_epoch(Datetime, 1000).
 
 %% Useful key extractors for functions (e.g., in get or delete code
 %% paths) which are agnostic to whether they are dealing with TS or
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -116,7 +116,7 @@ build_sql_record_int(insert, SQL, _Cover) ->
         false ->
             {error, <<"Must provide exactly one table name">>}
     end;
-build_sql_record(show_tables, _SQL, _Cover) ->
+build_sql_record_int(show_tables, _SQL, _Cover) ->
     {ok, #riak_sql_show_tables_v1{}}.
 
 convert_where_timestamps(_Mod, []) ->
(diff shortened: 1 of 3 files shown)


macintux:
eunit reveals much unhappiness with the AST walking, which doesn't surprise me: I only tested on simple cases. Again, WIP.

macintux:
Now ready for review, assuming the tests pass.

macintux:
Needs basho/jam#10 to address one dialyzer error. Other seems unrelated to my work, investigating.

src/riak_kv_ts_util.erl
```diff
@@ -97,15 +110,111 @@ build_sql_record(insert, SQL, _Cover) ->
...
+        true ->
+            %% jam returns atoms if processing gives an error
+            throw(Epoch);
+        false ->
+            Epoch
+    end.
+
+maybe_convert_to_epoch({'EXIT', _}, Value, _Op) ->
+    {binary, Value};
+maybe_convert_to_epoch(timestamp, Value, '>') ->
```
hmmr, line 158: What about '<'?

macintux, line 158: Strictly greater-than is the only scenario where we need to adjust the user input.

"< 2016" is easy: expand 2016 to 2016-01-01 00:00:00, convert to epoch, and we're done.

"> 2016" is harder: increment 2016 to 2017 and then expand.

src/riak_kv_ts_util.erl
```diff
@@ -97,15 +110,111 @@ build_sql_record(insert, SQL, _Cover) ->
...
+    CompleteFun = fun(DT) -> jam:expand(jam:increment(DT, 1), second) end,
+    {integer, do_epoch(Value, CompleteFun, 3)};
+maybe_convert_to_epoch(timestamp, Value, _Op) ->
+    CompleteFun = fun(DT) -> jam:expand(DT, second) end,
+    {integer, do_epoch(Value, CompleteFun, 3)};
+maybe_convert_to_epoch(_Type, Value, _Op) ->
+    {binary, Value}.
+
+%% I haven't investigated why the values are a nested list
+convert_insert_timestamps(Mod, Fields, [Values]) ->
```
hmmr: The third argument is a list of rows, each a list of values. As is, you are only taking a single row -- this will crash on statements inserting multiple rows (like "INSERT INTO t8 VALUES (1,7),(2,8),(3,9)").

macintux: Excellent, that explains why the values are a nested list and I was unwise to not have dug deeper. Good catch, thanks.

Do we have a test for multi-insert somewhere?

src/riak_kv_ts_util.erl
```diff
@@ -648,4 +756,77 @@ validate_rows_bad_2_test() ->
...
+        "a SINT64 NOT NULL, "
+        "b TIMESTAMP NOT NULL, "
+        "c SINT64 NOT NULL, "
+        "PRIMARY KEY((a, quantum(b, 15, 's')), a, b))"),
+    ?assertEqual(timestamp, Mod:get_field_type([<<"b">>])),
+    Lexed = riak_ql_lexer:get_tokens(BadQuery),
+    {ok, Q} = riak_ql_parser:parse(Lexed),
+    ?assertMatch({error, <<"Invalid date/time string">>},
+                 build_sql_record(select, Q, undefined)).
+
```
hmmr, line 874: A test for "INSERT" would be desirable, and it should be inserting multiple rows.

src/riak_kv_ts_util.erl
hmmr: There is tests/ts_simple_insert. Alternatively, you can use the quick and dirty HTTP client (https://gist.github.com/hmmr/94abdb4d648a44009e93).

src/riak_kv_ts_util.erl
```diff
@@ -97,15 +110,111 @@ build_sql_record(insert, SQL, _Cover) ->
...
+    {integer, do_epoch(Value, CompleteFun, 3)};
+maybe_convert_to_epoch(_Type, Value, _Op) ->
+    {binary, Value}.
+
+%% I haven't investigated why the values are a nested list
+convert_insert_timestamps(Mod, Fields, [Values]) ->
+    %% Inserted timestamps cannot be partial, so our function to
+    %% complete incomplete datetime values will throw an error
+    Types = lists:map(fun({identifier, [Column]}) ->
+                              catch Mod:get_field_type([Column])
```
hmmr: I believe you can omit the `catch` here because the parser will at this point have verified the fields are valid.

src/riak_kv_ts_util.erl
macintux, line 874: Added a good and a bad insert test after fixing multi-row inserts.

hmmr:
+1 57e05efc

macintux:
@borshop merge

PR #1445, opened 2016-07-22
SQL `GROUP BY` RTS-732

andytill:
Dependent PRs, riak_ql and riak_core are required as deps.
- riak_ql https://github.com/basho/riak_ql/pull/132
- riak_test https://github.com/basho/riak_test/pull/1109
- riak_core https://github.com/basho/riak_core/pull/848
- rfc https://github.com/basho/rfc/pull/6

Allow `GROUP BY` to be specified in an SQL statement,.

``` sql
SELECT c, COUNT(*) FROM tab1
WHERE a = 1 AND b = 2 GROUP BY b, a
```

See RFC for details.

This PR also adds upgrade and downgrade support to the select record because a field is added and then then the record is sent to other nodes, which may be on a different TS version. Before the record is sent to the other nodes by the `riak_kv_qry_worker`.  In the receiving vnode, the record is converted to the version that the node actually uses.  If the nodes are on the same version there is no conversion.


include/riak_kv_ts.hrl
```diff
@@ -29,12 +29,12 @@
 %% For dialyzer types
 -include_lib("riak_ql/include/riak_ql_ddl.hrl").
 
--define(SQL_SELECT, #riak_select_v1).
--define(SQL_SELECT_RECORD_NAME, riak_select_v1).
+-define(SQL_SELECT, #riak_select_v2).
```
hazen, line 32: I'm assuming all of this versioning will be addressed in the upgrade/downgrade PR? 😉

src/riak_kv_eleveldb_backend.erl
```diff
@@ -483,11 +483,12 @@ build_list({_K, _V}=KV, Acc) ->
...
 range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
                                               ref=Ref}) ->
-    {_, Bucket, Qry} = proplists:lookup(index, Opts),
+    {_, Bucket, Qry1} = proplists:lookup(index, Opts),
+    Qry2 = riak_kv_select:convert(riak_kv_select:current_version(), Qry1),
     ?SQL_SELECT{'WHERE'    = W,
                 helper_mod = _Mod,
                 local_key  =  #key_v1{ast = LKAST},
-                partition_key = #key_v1{ast = _PKAST}} = Qry,
+                partition_key = #key_v1{ast = _PKAST}} = Qry2,
```
hazen: I take the `partition_key` is just included for completeness since its value is unused, right?

src/riak_kv_qry_compiler.erl
```diff
@@ -58,17 +58,32 @@
     {ok, [?SQL_SELECT{}]} | {error, any()}.
 compile(?DDL{}, ?SQL_SELECT{is_executable = true}, _MaxSubQueries) ->
     {error, 'query is already compiled'};
-compile(?DDL{} = DDL,
-        ?SQL_SELECT{is_executable = false, 'SELECT' = Sel} = Q, MaxSubQueries) ->
-    if Sel#riak_sel_clause_v1.clause == [] ->
-            {error, 'full table scan not implemented'};
```
hazen, line 64: Is full table scan check handled somewhere else now?

src/riak_kv_qry_compiler.erl
```diff
@@ -58,17 +58,32 @@
...
+    Mod = riak_ql_ddl:make_module_name(T),
+    case maybe_compile_group_by(Mod, compile_select_clause(DDL, Q1), Q1) of
+        {ok, Q2} ->
+            compile_where_clause(DDL, Q2, MaxSubQueries);
+        {error, _} = Error ->
+            Error
+    end.
+
+%%
+maybe_compile_group_by(_, {error,_} = E, _) ->
```
hazen, line 72: I know these are internal functions, but could we get a `spec` for some of these? Always helpful in documenting expected inputs and outputs.

src/riak_kv_qry_compiler.erl
```diff
@@ -58,17 +58,32 @@
...
+    E;
+maybe_compile_group_by(Mod, {ok, Sel3}, ?SQL_SELECT{ group_by = GroupBy } = Q1) ->
+    compile_group_by(Mod, GroupBy, [], Q1?SQL_SELECT{ 'SELECT' = Sel3 }).
+
+%%
+compile_group_by(_, [], Acc, Q) ->
```
hazen, line 78: Looks like the output is a list of tuples of positions and field names in the group by, if it has a GROUP BY clause. Or an error tuple if the field name is not in the table.

src/riak_index.erl
```diff
@@ -360,8 +360,6 @@ apply_continuation(Q, C) ->
 %% @doc upgrade a query to the current latest version
```
hazen, line 360: Why is upgrade/downgrade stuff in `riak_index.erl`? I know you are only updating what is already here.  Thought this is for 2i.

src/riak_kv_qry_compiler.erl
```diff
@@ -1911,7 +1933,7 @@ sum_sum_finalise_test() ->
       ).
 
 extract_stateful_function_1_test() ->
-    {ok, #riak_select_v1{ 'SELECT' = #riak_sel_clause_v1{ clause = [Select] } }} =
+    {ok, ?SQL_SELECT{ 'SELECT' = #riak_sel_clause_v1{ clause = [Select] } }} =
```
javajolt: Is this an upgrade test? If not, perhaps the record name should also be `?SQL_SELECT`

src/riak_kv_qry_coverage_plan.erl
```diff
@@ -73,8 +73,13 @@ do_plan({Query, Bucket}, NVal, _PVC, ReqId, NodeCheckService, _Request, DownNode
     UnavailableVNodes =
         riak_core_coverage_plan:identify_unavailable_vnodes(CHBin, RingSize,
                                                             NodeCheckService),
-
-    Key = make_key(Query),
+    %% access the select record using element so that we don't have to use the
+    %% record syntax, the record may have been upgraded or downgraded at this
+    %% point
+    HelperMod = element(?SQL_SELECT.helper_mod, Query),
```
hazen, line 79: I wonder if these should be accessor functions? Then they could be reused.  Just a suggestion.

src/riak_kv_qry_worker.erl
```diff
@@ -54,7 +54,7 @@
           qid           = undefined           :: undefined | {node(), non_neg_integer()},
           sub_qrys      = []                  :: [integer()],
           receiver_pid                        :: pid(),
-          result        = []                  :: [{non_neg_integer(), list()}] | [{binary(), term()}],
+          result        = []                  :: [{non_neg_integer(), list()}] | [{binary(), term()}] | any(),
```
hazen: Where does the `any()` come from?

src/riak_kv_qry_worker.erl
```diff
@@ -168,6 +170,13 @@ make_key_conversion_fun(Table) ->
             Key
     end.
 
+%% Convert the sql select record to a version that is safe to pass around the
+%% cluster. Do not treat the result as a record in the local node, just as a
+%% tuple.
+convert_query_to_cluster_version(Query) ->
+    Version = riak_core_capability:get({riak_kv, sql_select_version}),
```
hazen, line 182: If this is a mixed cluster, could this be `undefined`?

src/riak_kv_select.erl
```diff
@@ -0,0 +1,134 @@
...
+    upgrade_select([To|Tail], Select3).
+
+%% Iterate over the versions backwards to downgrade, from 3 to 2 then 2 to 1 etc.
+downgrade_select([_], Select) ->
+    Select;
+downgrade_select([2,1=To|Tail], Select1) ->
+    Select2 = setelement(1, Select1, riak_select_v1),
+    %% The group_by field might not be the default value but it is safe to
+    %% downgrade because it is not used by remote vnodes.
+    Select3 = erlang:delete_element(?SQL_SELECT.group_by, Select2),
```
hmmr: Because in this function clause you are dealing with a specific version of the record, don't you have to refer to it by the explicit record name and not by what `?SQL_SELECT` expands to? When we introduce version 3 (even though the `group_by` field is not likely to be dropped in it), its index may change, and we will be deleting the wrong field.

src/riak_kv_ts_util.erl
```diff
@@ -68,18 +68,21 @@ sql_record_to_tuple(?SQL_SELECT{'FROM'   = From,
 %% Convert the proplist obtained from the QL parser
 build_sql_record(select, SQL, Cover) ->
     T = proplists:get_value(tables, SQL),
-    F = proplists:get_value(fields, SQL),
+    Fields = proplists:get_value(fields, SQL),
```
hmmr: Oi! Don't you break ma alignments!

src/riak_kv_select.erl
```diff
@@ -0,0 +1,134 @@
...
+            upgrade_select(VersionSteps, Select);
+        Version < CurrentVersion ->
+            VersionSteps = lists:seq(CurrentVersion, Version, -1),
+            downgrade_select(VersionSteps, Select)
+    end.
+
+%% Iterate over the versions and upgrade the record, from 1 to 2, 2 to 3 etc.
+upgrade_select([_], Select) ->
+    Select;
+upgrade_select([1,2 = To|Tail], Select1) ->
```
hazen: Would it be safer to use records instead of munging the tuple? Not sure what best Erlang practices are.

src/riak_kv_qry_worker.erl
hmmr, line 182: You can use `riak_kv_select:current_version/0` for that, can you?

src/riak_kv_qry_worker.erl
```diff
@@ -131,29 +131,31 @@ new_state() ->
...
-    Table = Q?SQL_SELECT.'FROM',
+run_sub_qs_fn([{{qry, ?SQL_SELECT{cover_context = undefined} = Q1}, {qid, QId}} | T]) ->
+    Table = Q1?SQL_SELECT.'FROM',
     Bucket = riak_kv_ts_util:table_to_bucket(Table),
     %% fix these up too
     Timeout = {timeout, 10000},
     Me = self(),
     KeyConvFn = make_key_conversion_fun(Table),
-    Opts = [Bucket, none, Q, Timeout, all, undefined, {Q, Bucket}, riak_kv_qry_coverage_plan, KeyConvFn],
+    Q2 = convert_query_to_cluster_version(Q1),
```
hmmr, line 146: These subqueries were all compiled in `riak_kv_qry:do_select/2` (https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_qry.erl#L267). Are they not, because of this, guaranteed to be the current version?

src/riak_kv_qry_compiler.erl
```diff
@@ -58,17 +58,32 @@
...
+    compile_group_by(Mod, GroupBy, [], Q1?SQL_SELECT{ 'SELECT' = Sel3 }).
+
+%%
+compile_group_by(_, [], Acc, Q) ->
+    {ok, Q?SQL_SELECT{ group_by = lists:reverse(Acc) }};
+compile_group_by(Mod, [{identifier,FieldName}|Tail], Acc, Q)
+        when is_binary(FieldName) ->
+    case Mod:get_field_position([FieldName]) of
+        undefined ->
+            {error, {group_by_field_invalid, FieldName}};
```
hmmr: Eventually, for this error to be properly reported (https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_ts_svc.erl#L460), the second element should become a description string (binary).

src/riak_kv_qry_compiler.erl
hmmr, line 64: It's actually redundant because the parser would have errored out on queries without any field to select.

src/riak_kv_qry_compiler.erl
hmmr: Eventually, for this error to be properly reported (as in https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_ts_svc.erl#L460), the second element must become a human-readable message (and I guess a new entry needs to be added to riak_kv_ts_error_msgs.hrl).

src/riak_kv_eleveldb_backend.erl
hmmr: True programmers embed comments in variable names!

include/riak_kv_ts.hrl
andytill, line 32: All upgrade and downgrade is done in this PR.

hazen, line 32: Sorry, I saw that later on.  Should have deleted that question.

src/riak_kv_qry_compiler.erl
hazen: Brewing up a human-readable error would be pretty easy

src/riak_kv_qry_worker.erl
hazen, line 182: @hmmr That would just get the latest version of the select record, not the agreed upon current version in the cluster.

src/riak_kv_select.erl
hazen: Interesting point, @hmmr. That also prompted my question about using records instead of mucking with the details of a tuple.  It's a bit more verbose, but might be safer and more explicit.

src/riak_kv_eleveldb_backend.erl
andytill: They are left over from a previous change, I have removed them in 6bca9b51ff27364808333192e08865d233fabdd6.

src/riak_kv_qry_coverage_plan.erl
andytill, line 79: We should be using the record syntax for most cases, where we use select record. This is a special case because the records may already downgraded.

src/riak_kv_qry_compiler.erl
andytill, line 72: The second function head is very specific though.

src/riak_kv_qry_worker.erl
andytill: Added named types for this spec in af482fa88d9f5845eb36195f90172c47eacfefcf

src/riak_kv_qry_worker.erl
andytill, line 146: They are converted to the highest version supported by the whole cluster here.  This node might be dealing in a version that the other nodes cannot understand yet, if it is running in a mixed version cluster.

src/riak_kv_qry_worker.erl
andytill, line 182: If all nodes in the cluster do not define the capability, then the default value of 1 will be used.  Nodes that don't define it won't be calling it of course, and will just be using version 1 because that is all they know.

src/riak_kv_select.erl
andytill: @hmmr the transformations are incremental so it will be 1 - 2, then 2 - 3, same with downgrades. We will know what has changed in each version. Also changes to field indexes should not be taken lightly. Can you elaborate on a scenario that you are concerned about?

src/riak_kv_qry_worker.erl
hmmr, line 146: In other words, the effective 'current' version may be different from that defined in the code. I understand now.

src/riak_kv_select.erl
hmmr: This function clause deals with specific version of the record (2). When, in order to get a field position, you do `#riak_select_v2.group_by`, your only assumption is that the declaration of `#riak_select_v2` never changes. When you refer to it by `?SQL_SELECT.field`, you have additionally to assume that no subsequent versions adds or removes a field before `group_by`. I contend that the first case wins because there are fewer ways to break things here by doing changes somewhere else. (Disclaimer: I am a registered antiglobalist.)

andytill: OK I get you, ?SQL_SELECT.field may return a different index in a future definition.  Changed this so that it specifies the index in 17adc61db46256bf31db6a3469846ce3e3453a99.

src/riak_kv_qry_worker.erl
```diff
@@ -49,12 +49,17 @@
 -define(NO_MAX_RESULTS, no_max_results).
 -define(NO_PG_SORT, undefined).
 
+%% accumulators for different types of query results.
+-type rows_acc()      :: [{non_neg_integer(), list()}].
```
hazen, line 53: Yeah! Types! Thank you!

src/riak_kv_select.erl
```diff
@@ -0,0 +1,135 @@
...
+    upgrade_select([To|Tail], Select3).
+
+%% Iterate over the versions backwards to downgrade, from 3 to 2 then 2 to 1 etc.
+downgrade_select([_], Select) ->
+    Select;
+downgrade_select([2,1=To|Tail], Select1) ->
+    Select2 = setelement(1, Select1, riak_select_v1),
+    %% The group_by field might not be the default value but it is safe to
+    %% downgrade because it is not used by remote vnodes.
+    V2GroupByIndex = 13,
```
hazen: 😟 Hard-coded magic numbers make me sad.  Still think records would be more maintainable in the long run.  More verbose and less speedy, I would agree, but probably more readable.

hmmr: It relates to another issue: whether to keep older `#riak_select_v*` versions around? If we do (as is done, for example here: https://github.com/basho/riak_kv/blob/riak_ts-develop/include/riak_kv_vnode.hrl) then `#riak_select_v2{}` will be available for use in this function, and can be used to get the field index. Else, hardcoding is the way.

andytill: Please leave emotions out of reviews.  Would it be more maintainable in the long run, do you have evidence for that?

I don't think it is fair to say it is a 'magic' number because it is purposely uses a named binding to describe it.

I have removed the v1 record to keep the size of the header file down, there will be one line difference, this is a big reduction in code size which would otherwise just be used by one line in this module.  Upgrade and downgrade is intentionally isolated so that multiple versions only get handled in this module.  Sharing them is not required for the code to function.

src/riak_kv_qry_compiler.erl
```diff
@@ -170,27 +182,33 @@ compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause =
...
     {ColTypes, Errors} = my_mapfoldl(
         fun(ColASTX, Errors) ->
             infer_col_type(DDL, ColASTX, Errors)
         end, [], Sel),
 
-    case sets:is_element(aggregate, ResultTypeSet) of
-        true  ->
-            Q3 = Q2#riak_sel_clause_v1{
+    IsGroupBy = (Q?SQL_SELECT.group_by /= []),
+    IsAggregate = sets:is_element(aggregate, ResultTypeSet),
```
gordonguthrie, line 193: You could generate a value CalcType with the case clause and not repeat the record 3 times and make this 1/3 the length

andytill, line 193: The initial state is different for `rows`.

src/riak_kv_select.erl
hazen: Sorry to have offended you, Andy.  My point is simply that without the original record it's less clear where the differences are.

andytill: No worries.

I have added a comment to the select record stating which version the `group_by` field was added in, in commit 69327da3b042a4c2ccc2aa17fe4d0d1f65be983f.

src/riak_kv_eleveldb_backend.erl
```diff
@@ -483,11 +483,10 @@ build_list({_K, _V}=KV, Acc) ->
...
 range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
                                               ref=Ref}) ->
-    {_, Bucket, Qry} = proplists:lookup(index, Opts),
+    {_, Bucket, Qry1} = proplists:lookup(index, Opts),
+    Qry2 = riak_kv_select:convert(riak_kv_select:current_version(), Qry1),
     ?SQL_SELECT{'WHERE'    = W,
-                helper_mod = _Mod,
-                local_key  =  #key_v1{ast = LKAST},
-                partition_key = #key_v1{ast = _PKAST}} = Qry,
+                local_key  =  #key_v1{ast = LKAST}} = Qry2,
```
gordonguthrie, line 489: Good catch - this is a case of code of mine containing _variables that were probably in there at dev time but which just clutter things up.

Wonder if it is worth scanning the code base for things like this?

src/riak_kv_qry_compiler.erl
gordonguthrie, line 193: could still be shrunk with a case that returns `{CalcType, InitialState}`

andytill, line 193: That creates a new type though and allocates a tuple.

include/riak_kv_ts.hrl
gordonguthrie, line 32: I vote for explicity retaining the old version records - but lets see what the consensus is

include/riak_kv_ts.hrl
```diff
@@ -29,12 +29,12 @@
...
+-define(SQL_SELECT_RECORD_NAME, riak_select_v2).
 
 %% the result type of a query, rows means to return all matching rows, aggregate
 %% returns one row calculated from the result set for the query.
--type select_result_type() :: rows | aggregate.
+-type select_result_type() :: rows | aggregate | group_by.
```
gordonguthrie, line 37: There is an interesting question about changing the type of a record - that presumably implies that there are changes the data passed in that record that might make the record incompatible over the wire.

This would imply bumping #riak_sel_clause_v1 to #riak_sel_clause_v2 instead of extending the type. 

I suspect that this is the case here - but it is being covered by the fact that we have bumped the version of the record that contains this record...

Should we have a general policy on 'upgrades in place' in records by type extension? Not sure this question can/should be fixed/addressed in this PR - but raising it anyhoo...

andytill, line 37: This change was made knowing that we know that the result type is not checked by the other vnodes, in the future we may either need this information to execute on the vnode, or have to use capabilities to make sure we are sending truly compatible records.

src/riak_index.erl
```diff
@@ -381,8 +379,13 @@ upgrade_query(#riak_kv_index_v2{
         end_inclusive=EndInclusive,
         return_body=ReturnBody};
 upgrade_query(Q) when is_tuple(Q) ->
-    {ok, Q2} = make_query(Q, ?KV_INDEX_Q{}),
-    Q2.
+    case riak_kv_select:is_sql_select_record(Q) of
```
gordonguthrie, line 382: Covered by general upgrade/downgrade discussion

src/riak_kv_app.erl
```diff
@@ -212,6 +212,11 @@ start(_Type, _StartArgs) ->
                                            riak_ql_ddl_compiler:get_compiler_capabilities(),
                                            lists:last(riak_ql_ddl_compiler:get_compiler_capabilities())),
 
+            riak_core_capability:register({riak_kv, sql_select_version},
```
gordonguthrie, line 215: Other versioned things like 2i indices and the object format use v1, v2 and v3 for versions not 1, 2, 3 - we should stick with that convention

v1, v2, and v3 behave with >, < and = like 1, 2, 3 - the list generator is just a big clunkier...

src/riak_kv_index_fsm.erl
```diff
@@ -78,10 +78,15 @@ use_ack_backpressure() ->
 
 %% @doc Construct the correct index command record.
 -spec req(binary()|tuple(binary()), term(), term()) -> term().
-req(Bucket, _ItemFilter, ?SQL_SELECT{} = Q) ->
-    #riak_kv_sql_select_req_v1{bucket=Bucket,
-                               qry=Q};
 req(Bucket, ItemFilter, Query) ->
```
gordonguthrie, line 81: with explicit records this case construction becomes a guard...

src/riak_kv_qry_compiler.erl
```diff
@@ -58,19 +58,30 @@
...
-            case compile_select_clause(DDL, Q) of
-                {ok, S} ->
-                    compile_where_clause(DDL, Q?SQL_SELECT{'SELECT' = S}, MaxSubQueries);
-                {error, _} = Error ->
-                    Error
-            end
+compile(?DDL{ table = T} = DDL,
+        ?SQL_SELECT{is_executable = false} = Q1, MaxSubQueries) ->
+    Mod = riak_ql_ddl:make_module_name(T),
+    case maybe_compile_group_by(Mod, compile_select_clause(DDL, Q1), Q1) of
```
gordonguthrie, line 64: assing through the error from maybe_compile_group_by is clunky - either a double case or a single case like

```
    case {compile_select, compile_group_by, compile_where} of
        {ok, S}, {ok, G}, {ok, W} -> record
        {E1, E2, E2}              -> make_error(E1, E2, E3)
    end
```

src/riak_kv_qry_compiler.erl
```diff
@@ -58,19 +58,30 @@
...
+    E;
+maybe_compile_group_by(Mod, {ok, Sel3}, ?SQL_SELECT{ group_by = GroupBy } = Q1) ->
+    compile_group_by(Mod, GroupBy, [], Q1?SQL_SELECT{ 'SELECT' = Sel3 }).
+
+%%
+compile_group_by(_, [], Acc, Q) ->
+    {ok, Q?SQL_SELECT{ group_by = lists:reverse(Acc) }};
+compile_group_by(Mod, [{identifier,FieldName}|Tail], Acc, Q)
+        when is_binary(FieldName) ->
+    Pos = Mod:get_field_position([FieldName]),
```
gordonguthrie, line 82: This is a bit clunky as `Mod:get_field_position(<<"*">>)` will return `undefined` for a wild card like `count(*)` here

Kinda using `get_field_position` as it is not meant to be used, hmmm

(a minor point)

gordonguthrie, line 82: Maybe it just needs a comment that this will handle `<<*>>` field identifiers as it took me a bit of noodling to prove that to my satisfaction...

src/riak_kv_qry_compiler.erl
```diff
@@ -170,27 +182,33 @@ compile_select_clause(DDL, ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{ clause =
...
-    case sets:is_element(aggregate, ResultTypeSet) of
-        true  ->
-            Q3 = Q2#riak_sel_clause_v1{
+    IsGroupBy = (Q?SQL_SELECT.group_by /= []),
+    IsAggregate = sets:is_element(aggregate, ResultTypeSet),
+    if
+        IsGroupBy ->
+            Sel2 = Sel1#riak_sel_clause_v1{
+                   calc_type = group_by,
+                   col_names = get_col_names(DDL, Q) };
```
gordonguthrie, line 198: This should also set the initial state for `group by` which is currently being done in `riak_kv_qry_worker:execute_query/2` at line 210

src/riak_kv_qry_compiler.erl
```diff
@@ -2229,6 +2247,50 @@ two_element_key_range_cannot_match_test() ->
         compile(DDL, Q, 100)
     ).
 
+group_by_one_field_test() ->
```
gordonguthrie, line 2250: Update unit tests to cover all error cases as per the docs:
https://docs.google.com/document/d/1wkXdv2L9dMFp_YWf-C_MMMNhJZpScM9ud0AAUs4Oog0/edit

(apologies - these error and success cases were added to the spec after the code was written)

src/riak_kv_qry_worker.erl
```diff
@@ -186,20 +200,27 @@ pop_next_query() ->
...
+execute_query({query, ReceiverPid, QId, [Qry1|_] = SubQueries, _},
               #state{ run_sub_qs_fn = RunSubQs } = State) ->
     Indices = lists:seq(1, length(SubQueries)),
     ZQueries = lists:zip(Indices, SubQueries),
     %% all subqueries have the same select clause
-    ?SQL_SELECT{'SELECT' = Sel} = Qry,
-    #riak_sel_clause_v1{initial_state = InitialState} = Sel,
+    ?SQL_SELECT{'SELECT' = Sel} = Qry1,
+    #riak_sel_clause_v1{initial_state = InitialState1} = Sel,
+    InitialState2 =
```
gordonguthrie: `riak_sel_clause_v1.initial_state` should be set in `riak_kv_qry_compiler` line 198 surely?

src/riak_kv_ts_util.erl
```diff
@@ -80,19 +80,21 @@ build_sql_record(Command, SQL, Cover) ->
 %% Convert the proplist obtained from the QL parser
 build_sql_record_int(select, SQL, Cover) ->
     T = proplists:get_value(tables, SQL),
-    F = proplists:get_value(fields, SQL),
+    Fields = proplists:get_value(fields, SQL),
```
gordonguthrie: If you are going to expand a variable from a letter to a word, do 'em all or don't do 'em

...the pedantic king of vertical alignment writes...

src/riak_kv_select.erl
```diff
@@ -0,0 +1,135 @@
...
+
+%% Convert a select record to a different version.
+-spec convert(integer(), Select1::tuple()) -> Select2::tuple().
+convert(Version, Select) when is_integer(Version) ->
+    CurrentVersion = select_record_version(element(1, Select)),
+    if
+        Version == CurrentVersion ->
+            Select;
+        Version > CurrentVersion ->
+            VersionSteps = lists:seq(CurrentVersion, Version),
```
gordonguthrie: If we are to use v1, v2, v3 to style then these list generators will need to be expanded...

src/riak_kv_app.erl
andytill, line 215: What benefit do atoms give?

src/riak_kv_index_fsm.erl
andytill, line 81: This would then have to be modified for every version increment, it was a real surprise to me that this record made it into riak_kv_index_fsm. If we keep the checks in riak_kv_select the changes will be isolated.

src/riak_kv_qry_compiler.erl
andytill, line 82: You can't use \* in select, or group by when group by is specified.

src/riak_kv_qry_compiler.erl
andytill, line 2250: I can only see one, that is covered here https://github.com/basho/riak_ql/pull/132/files#diff-53ba1ef30ed23e05f486e07b0a995b46R223

src/riak_kv_ts_util.erl
andytill: Changed back to `F` in 648f89e5de8cf39411d17f44f92dfa66b18d79de.

include/riak_kv_ts.hrl
andytill, line 32: Concensed in e172b82.

src/riak_kv_app.erl
andytill, line 215: Implemented in ab6ddb837e263f217d79eef06f5a3299470c425a.

src/riak_kv_qry_compiler.erl
andytill, line 64: Discussed and going to leave as is.

src/riak_kv_qry_compiler.erl
andytill, line 198: As discussed, added comment in de9fdfd1b667f09d60c340023060289b43e29425.

src/riak_kv_qry_compiler.erl
andytill: Discussed, leaving as is.

src/riak_kv_qry_compiler.erl
andytill, line 82: UPDATE: I thought that `*` could not be used in a group by was validated, but it was not. Fixed in https://github.com/basho/riak_ql/pull/132/commits/57ef47d3bd2d4140ebbdebc585f4b97645ccbfc4

gordonguthrie:
+1 de9fdfd

PR #1446, opened 2016-07-22
Adapt AAE throttling to use new riak_core_throttle module

jvoegele:
Related to https://bashoeng.atlassian.net/browse/RIAK-2626

Refactor AAE throttling logic to use the new `riak_core_throttle` module introduced in https://github.com/basho/riak_core/pull/849


jvoegele:
The build currently fails because it depends on changes in https://github.com/basho/riak_core/pull/849 that are not visible here.

bsparrow435:
The changes look good and I much prefer this method of using a module in core to wrap the env's than a bunch of env variables just floating around. Now checking out Riak core PR.

priv/riak_kv.schema
```diff
@@ -185,25 +185,7 @@
...
-         [{Mbox - 1, Delay}|Settings]
-     end, [], TierNames)),
-   case Throttles of
-       %% -1 is a magic "minimum" bound and must be included, so if it
-       %% isn't present we call it invalid
-       [{-1,_}|_] -> Throttles;
-       _ -> cuttlefish:invalid("anti_entropy.throttle tiers must include a tier with mailbox_size 0")
-   end
-end
+ riak_core_throttle:create_limits_translator_fun("anti_entropy", "mailbox_size")
```
JeetKunDoug: Make sure to switch out `riak_kv.aae_throttle_kill_switch` to `riak_kv.aae_throttle_enabled` here and in the riak_core_throttle module so that we can stop the cognitive dissonance.

src/riak_kv_entropy_manager.erl
```diff
@@ -819,19 +802,7 @@ get_aae_throttle_limits() ->
...
-            application:set_env(riak_kv, aae_throttle_limits, Limits),
-            ok;
-        {Else1, Else2} ->
-            {error, [{negative_one_length_is_present, Else1},
-                     {all_sleep_times_are_non_negative_integers, Else2}]}
-    end.
+    riak_core_throttle:set_limits(?AAE_THROTTLE_ENV_KEY, Limits).
 
 query_and_set_aae_throttle(#state{last_throttle=LastThrottle} = State) ->
     case get_aae_throttle_kill() of
```
JeetKunDoug: When we invert the kill/enabled thing, need to change this function name and just use the `enabled` value - switch the case statement around, etc. Need to review the module across the board whenever we're using the get_aae_throttle_kill() function.

src/riak_kv_cinfo.erl
```diff
@@ -54,7 +54,7 @@ aae_throttle(CPid) ->
                         [ThrottleLimits]).
 
 get_aae_throttle() ->
-    [{throttle_kill, riak_kv_entropy_manager:get_aae_throttle_kill()},
+    [{throttle_enabled, riak_kv_entropy_manager:is_aae_throttle_enabled()},
```
JeetKunDoug: This value seems to be thrown away in all cases I've found (riak_test and riak proper, the function, while exported, is only called from this module). Can we check to see if we can just drop the throttle_enabled key/value in this proplist altogether please?

JeetKunDoug:
`riak_kv_exchange_fsm` should really not be using `riak_kv_entropy_manager:get_aae_throttle/0` - it should just call a wrapped version of `riak_core_throttle:throttle(App, Key)` (probably still implemented in `riak_kv_entropy_manager`). The way it's implemented right now it will ignore the throttle enabled/disabled flag from what I can see.

See https://github.com/basho/yokozuna/pull/662/files#diff-1752d2a588dd638b5d126f54901866e6R525 for correct implementation in YZ.

JeetKunDoug:
+1 592057a

jvoegele:
@borshop merge

PR #1448, opened 2016-07-26
RTS-1051: Add the SHOW TABLES command

hazen:
Return a single-columned table with a list of all compiled tables when the query string "SHOW TABLES" is given

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -24,7 +24,7 @@
 
 -export([
          delete_dets/1,
-         get_all_compiled_ddls/0,
+         get_all_table_names/0,
          get_compiled_ddl_version/1,
          get_ddl/1,
          get_state/1,
@@ -167,13 +167,14 @@ get_ddl(BucketType) when is_binary(BucketType) ->
     end.
 
 %%
--spec get_all_compiled_ddls() -> term() | notfound.
-get_all_compiled_ddls() ->
-    case dets:match(?TABLE, {'_','_','$1','_','compiled'}) of
+-spec get_all_table_names() -> term() | notfound.
+get_all_table_names() ->
+    case dets:match(?TABLE, {'$1','_','_','_','compiled'}) of
         [] ->
             [];
         Matches ->
-            [DDL || [DDL] <- Matches]
+            Tables = [DDL || [DDL] <- Matches],
+            lists:usort(Tables)
     end.
 
 %% Update the compilation state using the compiler pid as a key.
@@ -361,8 +362,8 @@ get_all_compiled_ddls_test() ->
             ok = insert(<<"my_type4">>, 8, {ddl_v1}, Pid4, compiled),
 
             ?assertEqual(
-                [{ddl_v1},{ddl_v1}],
-                get_all_compiled_ddls()
+                [<<"my_type2">>,<<"my_type4">>],
+                get_all_table_names()
             )
         end).
 -endif.
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -171,7 +171,7 @@ get_ddl(BucketType) when is_binary(BucketType) ->
 get_all_compiled_ddls() ->
     case dets:match(?TABLE, {'_','_','$1','_','compiled'}) of
         [] ->
-            notfound;
+            [];
         Matches ->
             [DDL || [DDL] <- Matches]
     end.
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -24,6 +24,7 @@
 
 -export([
          delete_dets/1,
+         get_all_compiled_ddls/0,
          get_compiled_ddl_version/1,
          get_ddl/1,
          get_state/1,
@@ -163,6 +164,16 @@ get_ddl(BucketType) when is_binary(BucketType) ->
             DDL;
         [] ->
             notfound
+    end.
+
+%%
+-spec get_all_compiled_ddls() -> term() | notfound.
+get_all_compiled_ddls() ->
+    case dets:match(?TABLE, {'_','_','$1','_','compiled'}) of
+        [] ->
+            notfound;
+        Matches ->
+            [DDL || [DDL] <- Matches]
     end.
 
 %% Update the compilation state using the compiler pid as a key.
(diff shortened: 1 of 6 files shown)


src/riak_kv_compile_tab.erl
```diff
@@ -165,6 +166,16 @@ get_ddl(BucketType) when is_binary(BucketType) ->
             notfound
     end.
 
+%%
+-spec get_all_compiled_ddls() -> term() | notfound.
+get_all_compiled_ddls() ->
+    case dets:match(?TABLE, {'_','_','$1','_','compiled'}) of
```
andytill: Does this give one result per DDL per DDL version?

src/riak_kv_ts_api.erl
```diff
@@ -46,15 +46,16 @@
...
+-type query_api_call() :: query_create_table | query_select | query_describe | query_insert | query_show_tables.
 -type api_call() :: get | put | delete | list_keys | coverage | query_api_call().
 -export_type([query_api_call/0, api_call/0]).
 
 -spec api_call_from_sql_type(riak_kv_qry:query_type()) -> query_api_call().
-api_call_from_sql_type(ddl)      -> query_create_table;
-api_call_from_sql_type(select)   -> query_select;
-api_call_from_sql_type(describe) -> query_describe;
-api_call_from_sql_type(insert)   -> query_insert.
+api_call_from_sql_type(ddl)         -> query_create_table;
```
andytill, line 54: This function should be like the snippet below to avoid big diffs like this.

``` erlang
api_call_from_sql_type(ddl) ->
    query_create_table.
```

src/riak_kv_compile_tab.erl
```diff
@@ -165,6 +166,16 @@ get_ddl(BucketType) when is_binary(BucketType) ->
             notfound
     end.
 
+%%
+-spec get_all_compiled_ddls() -> term() | notfound.
```
paegun: Instead of `term()` in the spec, isn't the type `list(ddl_v1())`?

andytill: It isn't guaranteed to be a `ddl_v1` record in the future.

src/riak_kv_compile_tab.erl
hazen: IIRC you can only have one version of a table at any given time so each table will therefore return only a single DDL record

andytill: There will be multiple versions in there when we support upgrade/downgrade of DDLs.

andytill: The first element of the dets records' is the bucket type name as a binary, why not use that, and push it to a set?

paegun:
+1 84bb61d

src/riak_kv_compile_tab.erl
hazen: I wasn't sure if we needed additional metadata about a table

hazen: Can more than one version of a table be active at the same time on the same node?

andytill: No, but all versions of the DDL will be stored in this dets.  So that on downgrade the correct version can be picked out, because it may not be possible to have a downgrade transform for ddl_v10 when we only know about version ddl_v8 for example.

src/riak_kv_compile_tab.erl
```diff
@@ -167,13 +167,14 @@ get_ddl(BucketType) when is_binary(BucketType) ->
     end.
 
 %%
--spec get_all_compiled_ddls() -> term() | notfound.
-get_all_compiled_ddls() ->
-    case dets:match(?TABLE, {'_','_','$1','_','compiled'}) of
+-spec get_all_table_names() -> term() | notfound.
+get_all_table_names() ->
+    case dets:match(?TABLE, {'$1','_','_','_','compiled'}) of
```
andytill, line 172: Does this have to be a case expression, it is safe for a list comprehension to iterate an empty list.

paegun:
I wasn't able to assemble the case when multiple table versions exists (non-distinct table name), but based on the `dets:match/2` underlying, matching on $1 to distinctify the match, as long as the $1 version of the table is not ~"garbage collected", this solution will hold. For these reasons, I recommend the following:
1. Add a test that demonstrates as well as tests the case of multiple table versions.
2. Add a comment about this match on $1 w/ a warning about this failing if versions are ~"garbage collected".

andytill:
@paegun how do you mean garbage collected?

btw multiple versions will come in, in the DDL upgrade work I am doing for descending keys. There is only one DDL version at the moment so this can't happen but no point in writing future problems.

paegun:
@andytill I'm unfamiliar with the various ways that table versions are created, nor how long old versions will be valid until, but given future time is infinite, typically old versions that are no longer useful are garbage, so collected, it seems likely that counting on version one to always be present is not a 100% certain solution.

andytill:
No plan to delete old versions right now, but I guess there must be along an infinitely long time span :)

paegun:
+1 101869fa00f6aea6a89ab3c695e94369b1fcdf23
- With @andytill explanation re: how tables version, a test should be added when upgrade/downgrade test-ability is present.
- Considering ~"garbage collection" a feature for later.

src/riak_kv_compile_tab.erl
hazen, line 172: You are correct, Andy. I'll make a new PR for this.

hazen:
@borshop merge

src/riak_kv_ts_svc.erl
```diff
@@ -152,7 +152,13 @@ process(M = #riak_sql_describe_v1{'DESCRIBE' = Table}, State) ->
     check_table_and_call(Table, fun sub_tsqueryreq/4, M, State);
 
 process(M = #riak_sql_insert_v1{'INSERT' = Table}, State) ->
-    check_table_and_call(Table, fun sub_tsqueryreq/4, M, State).
+    check_table_and_call(Table, fun sub_tsqueryreq/4, M, State);
+
+process(#riak_sql_show_tables_v1{}, State) ->
+    {ok, {ColNames, ColTypes, LdbNativeRows}} =
+        riak_kv_ts_api:query("SHOW TABLES", ?DDL{}),
```
hmmr, line 159: Instead of calling `riak_kv_ts_api:query/2` with a hard-coded SQL statement, you could call `riak_kv_qry:submit(M, ?DDL{})` (where `M` would be the `#riak_sql_show_tables_v1{}` argument matched in function head, as in other `process` heads). By the time execution reaches `riak_kv_ts_svc:process/2`, the original SQL statement will have already been parsed (resulting in `#riak_sql_show_tables_v1{}`, which is why we are here), and there's no need to create a new, identical statement and have it parsed again.

PR #1450, opened 2016-07-27
RTS-1051: Clean up a few outstanding issues

hazen:
Both @hmmr and @andytill found a couple of optimizations, which have been addressed here.

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -167,15 +167,11 @@ get_ddl(BucketType) when is_binary(BucketType) ->
     end.
 
 %%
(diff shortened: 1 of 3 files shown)


hmmr:
+1 e31b887

src/riak_kv_compile_tab.erl
```diff
@@ -169,13 +169,9 @@ get_ddl(BucketType) when is_binary(BucketType) ->
 %%
 -spec get_all_table_names() -> term() | notfound.
```
andytill: The spec here is incorrect.

hmmr:
+1 1ad180f

hazen:
@borshop merge

PR #1453, opened 2016-08-01
Small tidy-up: removing dead record definitions

gordonguthrie:
An incidental tidy of commented out and dead records from the TS PoC arising from comparative code reviews.


gordonguthrie:
Badly structured PR - killed

PR #1456, opened 2016-08-01
Support incomplete datetime values for insert

macintux:
Per team discussions, we have decided to allow incomplete strings to be used for insert statements to reduce the friction for users attempting to explore Riak TS.

For tests to pass, requires https://github.com/basho/jam/pull/11 also.


hazen:
:+1: ee9adc8

hazen:
@borshop merge

PR #1457, opened 2016-08-02
Address key logic bugs regarding comparisons of ISO 8601 strings

macintux:
See also https://github.com/basho/riak_test/pull/1115 to help validate this (and earlier) code.

I realized today that there were unhandled edge cases when expanding incomplete dates. Two key problems with the existing code:
- `<=` also requires incrementing an incomplete date along with `>`. `time <= 2016` means we need to increment the structure to 2017 before expansion and conversion to epoch.
- After incrementing for either `<=` or `>` logic against incomplete dates, we need to decrement the resulting epoch value by 1 ms (so, in the above example, we're comparing against `<= 2016-12-31 23:59:59.999`). The other approach would be to convert the operator `<=` to `<` and `>` to `>=`, but this was easier.

### DIFF
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -131,7 +131,7 @@ replace_ast_timestamps(Mod, {Op, FieldName, {binary, Value}}) ->
 replace_ast_timestamps(_Mod, {Op, Item1, Item2}) ->
     {Op, Item1, Item2}.
 
-do_epoch(String, CompleteFun, Exponent) ->
+do_epoch(String, CompleteFun, PostCompleteFun, Exponent) ->
     Normal = timestamp_to_normalized(String),
     case jam:is_valid(Normal) of
         false ->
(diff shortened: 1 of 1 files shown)


src/riak_kv_ts_util.erl
```diff
@@ -155,16 +155,22 @@ do_epoch(String, CompleteFun, Exponent) ->
...
-    %% "2017". "2016-07-05T11" becomes "2016-07-05T12".
+maybe_convert_to_epoch(timestamp, Value, Op) when Op == '>'; Op == '<=' ->
+    %% For strictly greater-than or less-than-equal-to, we increment
+    %% any incomplete date/time strings by one "unit", where the unit
+    %% is the least significant value provided by the user. "2016"
+    %% becomes "2017". "2016-07-05T11" becomes "2016-07-05T12".
+    %%
+    %% The result is then decremented by 1ms to maintain the proper
+    %% semantics. The alternative would be to convert the operator in
+    %% the syntax tree (from > to >=, or from <= to <).
```
hmmr, line 166: The alternative you are offering to consider would, hands down, be more efficient (I'm not willing to suggest this needs to be implemented now, though).

hmmr:
+1 6dc721e

hazen:
@borshop merge

PR #1459, opened 2016-08-09
Add async job switches

tburghart:
Adds switches for the following operations arriving through HTTP or protobuf interfaces:
- list_buckets
- list_keys
- map_reduce_js
- map_reduce
- secondary_index (search)
- stream_list_buckets
- stream_list_keys
- stream_secondary_index (search)

_Caveats:_
- `map_reduce_js`, meant to control map-reduce jobs using embedded javascript, is not handled distinctly from `map_reduce`, as we have not yet identified an efficient way to determine whether or not a submitted job actually uses JS at the outer boundary where filtering is implemented.


tburghart:
@borshop retry

tburghart:
@borshop: retry

src/riak_kv_pb_bucket.erl
```diff
@@ -94,42 +94,54 @@ encode(Message) ->
...
+    Accept = riak_core_util:job_class_enabled(Class),
+    _ = riak_core_util:report_job_request_disposition(
+            Accept, Class, ?MODULE, process, ?LINE, protobuf),
+    case Accept of
         true ->
             case check_bucket_type(Type) of
                 {ok, GoodType} ->
                     do_list_buckets(GoodType, T, S, Req, State);
                 error ->
-                    {error, {format, "No bucket-type named '~s'", [Type]}, State}
```
nickelization: Small request for the future: if/when you break up long lines, could you do it in its own separate commit? It adds a little bit of friction to the review process when I have to look at the diff and mentally filter out what's an actual change vs what's just being reformatted.

Basho-JIRA:
I would not have linked this to a repo, for multiple reasons.

First, it affects 4, 6, or 7 repos, depending how you count.

Second, linking to a repo copies internal conversations to public display on github.

Last, the issue references private repos and google docs.

I think we should have (or maybe already do, as we've had prior discussions on the subject) a policy on linking to repos; my personal opinion is that we shouldn't, as we often comment on issues in a manner that we may not want published. If we DO have a policy to link to repos, we need to make it abundantly clear that internal discussions are not to be entered into Jira.

– Ted 

 

***[posted via JIRA by Ted Burghart]***

src/riak_kv_pb_bucket.erl
```diff
@@ -96,28 +94,60 @@ encode(Message) ->
...
-        error ->
-            {error, {format, "No bucket-type named '~s'", [Type]}, State}
+process(#rpblistbucketsreq{type = Type, timeout = T, stream = S} = Req, State) ->
+    Class = case S of
+        true ->
+            {riak_kv, stream_list_buckets};
+        _ ->
+            {riak_kv, list_buckets}
+    end,
+    Accept = riak_core_util:job_class_enabled(Class),
```
JeetKunDoug: It seems like there's an opportunity for some significant reduction in code here if riak_core_util:job_class_enabled just took an "Accepted fun" and a "Rejected fun" and wrapped up all the calls to `request_disposition` and `job_class_disabled_message`.  May require an "error format" parameter as well (although it looks here like all the error messages are in binary format?). I assume a similar pattern appears in other PRs like this?

tburghart: Yes and no - there's certainly _some_ commonality, but the `process` (for PB) and `forbidden` (for WM) functions differ enough for various operations that it could get fairly convoluted.  It certainly could be changed to a different pattern, but I wanted to keep complexity down for the first pass.
I also opted to try to keep more of the logic out at the edge for clarity, assuming there'll be a good deal of refactoring for the 2.3 jobs API.

src/riak_kv_pb_index.erl
```diff
@@ -109,7 +107,7 @@ process(#rpbindexreq{} = Req, State) ->
                     maybe_perform_query(QueryVal, Req, State)
             end;
         false ->
-            {error, "Secondary index queries have been disabled in the configuration", State}
```
nickelization: Does it really get us much of a benefit to make all the error messages look similar? It seems to me that the most important thing in an error message is that it clearly describes what went wrong. IMO the former sentence does a better job of explaining exactly what happened than the one that replaced it here.

I'll admit I'm being kind of nit-picky, so if you disagree then you don't have to change it, but I wanted to at least revisit this decision briefly.

rebar.config
```diff
@@ -1,37 +1,69 @@
...
+    {"(XC - UC) || (XU - X - B - \"(cluster_info|dtrace|yz_kv|yz_stat)\" : Mod)", []}
+]}.
 
 {erl_first_files, [
-                   "src/riak_kv_backend.erl"
-                  ]}.
+    "src/riak_kv_backend.erl"
+]}.
 
 {deps, [
```
nickelization, line 27: I definitely agree that we should be moving away from so many tags in our dev branches, but I don't know if I like that these changes are being snuck into an unrelated pull request.

Some of these are relatively innocuous; it looks like sidejob and bitcask are just going to end up pointing to the same commit as they are now (for the time being anyway).

The sext change makes me a little more nervous - any changes to the way stuff gets encoded could require special care and attention, so this is one case where I might be somewhat inclined to leave it as a tag. I guess this is probably okay in this case anyway since we're pointing at our own forked copy of sext, so at least we control when it's updated...

Oh, actually though, I just determined that we shouldn't be updating sext here because it tries to pull in a new version of edown. This will cause a version conflict with the "setup" and "parse_trans" repos, both of which still specifically request the 0.5 tag. We could try updating things on all three repos at once, but I think we should do that as a separate series of PRs instead of trying to tack it onto this effort.

It looks like we also get some new code pulled in for eper and eunit_formatters, but at least those are only debug/test tools, so I'm a little less worried about that. Did you test it with the new versions of those deps by any chance though, just to make sure nothing big breaks?

tburghart, line 27: Been using that config since the start, per commit 06b4218 (July 17th) comment, so yes, all testing was done with it.  As alluded to in that commit's comment, I _did_ review what the change pulled in ... which was mostly nothing.

The entirety of the sext change is the bump in edown version, and as far as I can tell the only place edown is used is in exometer_core.  In that case if its not found we pull the latest commit from some other fork, so I don't think we care about version specifics.  edown is, after all, a self-contained script for doc manipulation, not a runtime component.

test/riak_kv_schema_tests.erl
```diff
@@ -374,6 +405,21 @@ all_backend_multi_test() ->
...
+test_job_class_enabled({true, RCSchema}) when erlang:is_list(RCSchema) ->
+    Config = cuttlefish_unit:generate_templated_config(
+        [RCSchema, "../priv/riak_kv.schema"], [],
+        riak_core_schema_tests:context() ++ context()),
+
+    cuttlefish_unit:assert_config(
+        Config, "riak_core.job_accept_class",
+        lists:sort(?DEFAULT_ENABLED_JOB_CLASSES)),
+    ok;
+test_job_class_enabled(_) ->
```
nickelization: I'd like to see a log message here, so that it's clear why the test was skipped.

Even in the "normal" case where it just can't find the riak_core schema, it'd be good to make it explicit why the test is being skipped so that the user has a chance to notice and set things up to correctly run this test.

On the flip side, if some unexpected error occurs, the existing code would also silently swallow it, so it might also be good to have two different log messages: one for the schema-not-found case, and a separate warning for any unexpected errors.

src/riak_kv_pb_bucket.erl
JeetKunDoug: If nothing else, we should have a common function _in PB_ that does this - would remove tons of repetitive code in these modules, and would mean, if the API _does_ change in 2.3, we would only need to change it in one place.

rebar.config
JeetKunDoug, line 27: First, a general comment - I'd agree that it would be nice to have this set of changes in a separate PR. Making unrelated rebar.config changes in a PR like this makes it harder for this PR to move forward _and_ to get the rebar changes done - separate reviews would make this process smoother.

Second, to the actual changes:
Any changes that would allow a library that has on-disk format impacts to float on a branch is bad - thus, I agree that we shouldn't be floating sext on a branch, maybe ever. I understand that _this particular_ change won't, today, break anything if it's really only an edown change, but if the master branch changes (for example, if TS decides to make and merge a change we're not aware of), we could break Riak for users with existing data on-disk. To keep this PR moving, I'd be fine with any of the changes to the riak_\* modules (that are pointing at develop-2.2 branches, as obviously the develop-2.2 branch of riak_kv wants those), sidejob (because I'm pretty sure we need to bump it anyway), and eper/eunit_formatters as debugging/unit testing tools, but I really don't want to see sext, or any other library that may cause on-disk incompatibility issues, to be floated on a branch at this point. That also means Bitcask should probably remain pinned to a tag so we can make conscious decisions about moving it, rather than getting "whatever's latest" and breaking something.

nickelization:
+1 8e0233b

nickelization:
@borshop merge

Basho-JIRA:
already merged 

**_[posted via JIRA by Patricia Brewer]**_

PR #1460, opened 2016-08-11
Fix unhappy path for too_many_values wrt bad timestamp in INSERT

hmmr:
RTS-1299

In order to convert ISO timestamps given as strings, rather than doing this in `riak_kv_ts_util:build_sql_record_int`, we now pass values for INSERT as is and relegate the timestamp conversion to `riak_kv_qry:do_insert`. The conversion will thus take place after checking for too many values in INSERT query wrt the number of fields in the DDL. Otherwise, this check would need to be done twice as the timestamp conversion code needs to have the field types available (and check that `length(Types) == length(Values)` and do something if it is not).

cc @macintux

### DIFF
--- a/src/riak_kv_qry.erl
+++ b/src/riak_kv_qry.erl
@@ -79,7 +79,7 @@ submit(SQL = ?SQL_SELECT{}, DDL) ->
 submit(#riak_sql_show_tables_v1{} = _SQL, _DDL) ->
     do_show_tables();
 submit(#riak_sql_explain_query_v1{'EXPLAIN' = Select}, DDL) ->
-    do_explain_query(DDL, Select).
+    do_explain(DDL, Select).
 
 %% ---------------------
 %% local functions
@@ -322,10 +322,10 @@ build_show_tables_result(Tables) ->
 %%
 %% EXPLAIN statement
 %%
--spec do_explain_query(DDL :: ?DDL{},
-                       Select :: ?SQL_SELECT{}) ->
-                       {ok, query_tabular_result()} | {error, term()}.
-do_explain_query(DDL, Select) ->
+-spec do_explain(DDL :: ?DDL{},
+                 Select :: ?SQL_SELECT{}) ->
+                        {ok, query_tabular_result()} | {error, term()}.
+do_explain(DDL, Select) ->
     ColumnNames = [
         <<"Subquery">>,
         <<"Range Scan Start Key">>,
@@ -419,7 +419,7 @@ explain_query_test() ->
     SQL = "SELECT a,b,c FROM tab WHERE b > 0 AND b < 2000 AND a=319 AND c='hola' AND (d=15 OR (e=true AND f='adios'))",
     {ok, Q} = riak_ql_parser:parse(riak_ql_lexer:get_tokens(SQL)),
     {ok, Select} = riak_kv_ts_util:build_sql_record(select, Q, undefined),
-    Res = do_explain_query(DDL, Select),
+    Res = do_explain(DDL, Select),
     ExpectedRows =
         [[1,"c = 'hola', b = 1",false,"c = 'hola', b = 1000",false,
             "(((d = 15) OR ((e = true) AND (f = 'adios'))) AND (a = 319))"],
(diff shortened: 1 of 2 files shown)


hmmr:
@borshop retry

src/riak_kv_qry.erl
```diff
@@ -95,11 +95,15 @@ do_insert(#riak_sql_insert_v1{'INSERT' = Table,
...
-                    {error, Reason};
+            Types = [catch Mod:get_field_type([Column]) || {identifier, [Column]} <- Fields],
+            try xlate_insert_to_putdata(Values, Positions, Empty, Types) of
                 {ok, Data} ->
-                    insert_putreqs(Mod, Table, Data)
+                    insert_putreqs(Mod, Table, Data);
+                {error, Reason} ->
+                    {error, Reason}
+            catch
+                throw:Reason ->
```
andytill, line 105: What is using `throw/1`, rather than `error/1`?

src/riak_kv_qry.erl
```diff
@@ -161,11 +165,17 @@ lookup_field_positions(Mod, FieldIdentifiers) ->
...
+%% Additionally, we will convert Value in {varchar, Value} into an
+%% integer when that value is a timestamp. We would do it as we
+%% repackage list to tuple here, rather than separately as a
+%% post-processing step on Data, to avoid another round of
+%% list_to_tuple(tuple_to_list())
+-spec xlate_insert_to_putdata([[riak_ql_ddl:data_value()]], [pos_integer()], tuple(undefined),
+                              [riak_ql_ddl:simple_field_type()]) ->
                               {ok, [tuple()]} | {error, string()}.
-xlate_insert_to_putdata(Values, Positions, Empty) ->
+xlate_insert_to_putdata(Values, Positions, Empty, FieldTypes) ->
```
andytill, line 176: What does xlate mean? I know this is a pre-existing function.

src/riak_kv_qry.erl
```diff
@@ -180,18 +190,29 @@ xlate_insert_to_putdata(Values, Positions, Empty) ->
...
     %% Out of entries in the value - row is populated with default values
     %% so if we run out of data for implicit/explicit fieldnames can just return
     {ok, Row};
-make_insert_row(_, [], Row) when is_tuple(Row) ->
-    %% Too many values for the field
-    {error, too_many_values};
-%% Make sure the types match
-make_insert_row([{_Type, Val} | Values], [Pos | Positions], Row) when is_tuple(Row) ->
-    make_insert_row(Values, Positions, setelement(Pos, Row, Val)).
+make_insert_row([TypedVal | Values], [Pos | Positions], Row, FieldTypes) ->
```
andytill, line 205: This would be better as a function in the table helper module.

``` erlang
make_insert_row([A,B,C]) ->
    {A, maybe_convert_timestamp(B), C}.
```

That would avoid a lot of conversion, iteration, tuple setting etc.

src/riak_kv_qry.erl
hmmr, line 105: https://github.com/basho/riak_kv/blob/fix-az-unhappy_path_for_toofewvalues_wrt_badtimestamp_in_insert/src/riak_kv_ts_util.erl#L220, and possibly some code in jam, too.

src/riak_kv_qry.erl
hmmr, line 176: "translate"?

src/riak_kv_qry.erl
andytill, line 105: ok for control flow, thanks.

src/riak_kv_qry.erl
hmmr, line 205: It's interesting but tricky. For one thing, as used here, the row may be incomplete and the argument values may come in arbitrary order (which happens when INSERT comes with field names). Or perhaps your suggestion is to have the `Mod:maybe_convert_timestamp` take the full row?

I'm still reluctant to take it on, at least now, because doing so would raise the demons of DDL version change and new capability handling. Also, calling an external, non-OTP function from a helper module is something I can't quite see the full consequences of.

src/riak_kv_qry.erl
macintux, line 176: I was a bit worried that was something I started using without precedent, but a quick search shows it's relatively common. IBM has used it, there's a company called xlate that does translation services, etc.

macintux:
:+1: 9fb6065

hmmr:
@borshop merge

PR #1461, opened 2016-08-12
Remove the legacy mapping for the select capability.

andytill:
Remove the legacy mapping for the select capability, their is no legacy values to map.


hmmr:
+1 5741e86

The legacy parameter is only used in situations where some negotiating nodes will have no such capability registered at all. When this is the case, there will be a `rpc:call` somewhere to query the 'legacy' value. And, because the ﻿⁠select﻿⁠ capability has always been there and is guaranteed to be present in all upgrade scenarios, we can safely remove the legacy value. It's like removing dead code.

andytill:
@borshop merge

PR #1465, opened 2016-08-19
index_fsm:init: avoid passing fun objects as args to remote call sites

hmmr:
A partial fix for future verify_2i_mixed_cluster failures.

Because instances of index fsm may be started on different nodes (particularly, on different nodes with differing versions of riak_kv_index_fsm.beam), passing a fun object to where the call is made is likely to cause a `badfun` error the moment the function object is called.  This happens to work when the module beam on the remote node contains the compiled code the received function object refers to; but all chances are off when 'index' and/or 'uniq' properties (as returned by `erlang:fun_info/1`) are different between the caller and target beams.

To avoid badfuns, we export the function we are passing as an argument, for the caller to refer to it safely.

This won't fix badfuns in a mixed cluster with 1.3 nodes, where clients call API functions involving list_keys, but will prevent this from happening going forward.

See www.javalimit.com/2010/05/passing-funs-to-other-erlang-nodes.html for details and discussion.

c/o @andytill.

### DIFF
--- a/src/riak_kv_index_fsm.erl
+++ b/src/riak_kv_index_fsm.erl
@@ -48,7 +48,8 @@
          plan/2,
          process_results/3,
          process_results/2,
-         finish/2]).
+         finish/2,
+         identity/1]).
 -export([use_ack_backpressure/0,
          req/3]).
 
@@ -98,6 +99,19 @@ index_req(Bucket, ItemFilter, Query) ->
                                   qry=Query}
     end.
 
+%% Because instances of index fsm may be started on different nodes
+%% (particularly, on different nodes with differing versions of the
+%% beams), passing a fun object to where the call is made is likely to
+%% cause a badfun error the moment the function object is called.
+%% This happens to work when the module beam on the remote node
+%% contains the compiled code the received function object refers to;
+%% but all chances are off when 'index' and/or 'uniq' properties are
+%% different between the caller and target beams.
+%%
+%% To avoid badfuns, we export the function we are passing as an
+%% argument, for the caller to refer to it safely.
+identity(X) -> X.
+
 %% @doc Return a tuple containing the ModFun to call per vnode,
 %% the number of primary preflist vnodes the operation
 %% should cover, the service to use to check for available nodes,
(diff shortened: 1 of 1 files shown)


andytill:
Nice find.  I am still getting test failures in `verify_2i_mixed_cluster`.

``` erlang
08:38:37.002 [error] <0.2.0> 
================ verify_2i_mixed_cluster failure stack trace =====================
{{assertEqual_failed,[{module,secondary_index_tests},
                      {line,137},
                      {expression,"ActualKeys"},
                      {expected,[<<"obj05">>,<<"obj06">>,<<"obj07">>,
                                 <<"obj08">>,<<"obj09">>]},
                      {value,[]}]},
 [{secondary_index_tests,'-assertExactQuery/5-fun-4-',2,
                         [{file,"tests/secondary_index_tests.erl"},
                          {line,137}]},
  {secondary_index_tests,'-assertExactQuery/5-lc$^0/1-0-',5,
                         [{file,"tests/secondary_index_tests.erl"},
                          {line,117}]},
  {secondary_index_tests,'-assertExactQuery/5-lc$^0/1-0-',5,
                         [{file,"tests/secondary_index_tests.erl"},
                          {line,117}]},
  {verify_2i_mixed_cluster,confirm,0,
                           [{file,"tests/verify_2i_mixed_cluster.erl"},
                            {line,51}]},
  {riak_test_runner,return_to_exit,3,
                    [{file,"src/riak_test_runner.erl"},{line,196}]}]}
==================================================================================
```

I am also concerned about the fun in riak_kv_ts_svc, https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_ts_svc.erl#L359.

If riak_kv_ts_svc has a different module hash then this will fail too right? Because it is not a function reference.

hmmr:
@andytill sadly, the key conversion function (the other function that gets passed as an argument) cannot be exported as it depends on two vars from the environment, and needs to be constructed locally.

andytill:
+1 b1ab3a7

andytill:
@borshop merge

Issue #1464, opened 2016-08-17
Riak TS: Performing single row fetches sometimes results in "timeout" [JIRA: RTS-1328]


alexmoore:
Sometimes when performing a single row fetch operation over the TTB interface, I get back the error "timeout" with code 1010. 

The issue seems to resolve itself fairly quickly, as subsequent fetches and queries will pull back the row successfully.  

Probably EC in action, but a more descriptive error message would be better. 

Will update issue with a reproducer test link.

alexmoore:
See https://github.com/basho/riak-java-client/blob/kv-1464-reproducer/src/test/java/com/basho/riak/client/api/commands/itest/ITestTimeSeriesFetchBug.java for a reproducer test.  

Test can be run with:
`mvn test -Dtest=ITestTimeSeriesFetchBug -Dcom.basho.riak.pbcport=10017 -Dcom.basho.riak.timeseries=true`

May take many runs to reproduce, have had more success with larger devrel clusters (4-5 devrel nodes).

alexmoore:
Error is thrown here: https://github.com/basho/riak_kv/blob/a1ba1f4bad805a9e043b43495a74dac1d05267a8/src/riak_kv_ts_svc.erl#L309-L310

Basho-JIRA:
Fix pushed to https://github.com/basho/riak_kv/pull/1466 

**_[posted via JIRA by Andrei Zavada]**_

Basho-JIRA:
Following a discussion, it was decided to disallow 0 for a timeout value (and thus, no 'infinity') in get, delete and list_keys API calls. No change is required in riak_kv; and [~amoore] will take care of diagnosing and handling this situation in clients. 

**_[posted via JIRA by Andrei Zavada]**_

PR #1466, opened 2016-08-22
get_fsm: treat a timeout value of 0 as infinity

hmmr:
RTS-1328 (#1464)

When a client calls a `get` with a value of 0 for its `Timeout` parameter, get_fsm treats that value literally and immediately triggers a timeout (https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_get_fsm.erl#L544). One can argue that this is indeed expected behaviour, but the FSM still manages to serve some `get`s (which becomes possible under certain racy conditions between state changes and message passing to self), and that can be really confusing for the client.

This PR explicitly treats the case of `Timeout = 0` the same as `Timeout = infinity`, for the `get` API call.

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -1077,7 +1077,12 @@ recv_timeout(Options) ->
         undefined ->
             %% If no reply timeout given, use the FSM timeout + 100ms to give it a chance
             %% to respond.
-            proplists:get_value(timeout, Options, ?DEFAULT_TIMEOUT) + 100;
+            case proplists:get_value(timeout, Options, ?DEFAULT_TIMEOUT) of
+                infinity ->
+                    infinity;
+                Finity ->
+                    Finity + 100
+            end;
         Timeout ->
             %% Otherwise use the directly supplied timeout.
             Timeout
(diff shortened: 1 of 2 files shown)


andytill:
Suggest validating that the input is a positive integer, as per the current dialyzer spec. This enforces the current behaviour rather than changing it.

hmmr:
@andytill A zero timeout is allowed for other calls (delete, list_keys). Disallowing 0 for the get timeout would make it an odd one out.

Due to apparently different ways zero timeout is treated in those other calls, more extensive work is needed to make it uniformly mean infinity, which this PR does not claim to do.

andytill:
This change is making it the odd one out though, by rewriting the timeout.

It is just strange to rewrite the timeout because zero is the exact opposite of infinity.

hmmr:
Here's another take. Zeros are now converted to `'infinity'` in riak_kv_ts_svc (i.e., underlying functions don't see 0 at all), for all relevant API calls (get, delete, list_keys).

jvoegele:
Just a gut reaction here, but treating `0` as `infinity` feels a bit too much like magic to me. I’d prefer more explicit behavior, such as disallowing `0`.

hmmr:
@jvoegele That's fair, but how do you specify `infinity` then?

jvoegele:
With the atom `infinity`? Maybe I'm missing something here...

andytill:
Currently `infinity` doesn't get passed to the fsm by riak_client, it only passes the given or default timeout.

hmmr:
The protobuff messages for the affected calls have an uint32 for it.

hmmr:
A decision was reached to disallow 0 _and_ `;infinity'` as valid timeout values.

PR #1467, opened 2016-08-23
Remove the legacy compile dets table support

andytill:
The legacy dets tab is what Riak TS used to store compile state in 1.2 and previous.

We do not support upgrade/downgrade to 1.2 now because of breaking client APIs, so it can be removed.  The table itself is not removed.

This is in support of the DDL upgrade work. Removing it means we do not have to support the legacy table in ongoing changes.

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -47,7 +47,7 @@
 
 %% 
 -spec new(file:name()) ->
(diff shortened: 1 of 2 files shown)


hazen:
Yeah! Great to see the clean up.

hmmr:
+1 a21b6ce

andytill:
@borshop merge

PR #1468, opened 2016-08-23
Refactored `riak_kv_pb_index` and `riak_kv_pb_bucket`

lehoff:
Refactored to reduce code duplication and increase clarity.

### DIFF
--- a/src/riak_kv_pb_bucket.erl
+++ b/src/riak_kv_pb_bucket.erl
@@ -66,89 +66,86 @@ init() ->
 %% @doc decode/2 callback. Decodes an incoming message.
 decode(Code, Bin) ->
     Msg = riak_pb_codec:decode(Code, Bin),
-    case Msg of
-        rpblistbucketsreq ->
-            %% backwards compat
-            {ok, Msg, {"riak_kv.list_buckets", <<"default">>}};
-        #rpblistbucketsreq{} ->
-            Type = case Msg#rpblistbucketsreq.type of
-                       undefined ->
-                           <<"default">>;
-                       T ->
-                           T
-                   end,
-            {ok, Msg, {"riak_kv.list_buckets", Type}};
-        #rpblistkeysreq{} ->
-            Type = case Msg#rpblistkeysreq.type of
-                       undefined ->
-                           <<"default">>;
-                       T ->
-                           T
-                   end,
-            {ok, Msg, {"riak_kv.list_keys", {Type,
-                                                Msg#rpblistkeysreq.bucket}}}
-    end.
+    handle_decoded(Msg).
+
+handle_decoded(rpblistbucketsreq = Msg) ->
+    %% backwards compat
+    {ok, Msg, {"riak_kv.list_buckets", <<"default">>}};
+handle_decoded(#rpblistbucketsreq{} = Msg) ->
+    Type = convert_type(Msg#rpblistbucketsreq.type),
+    {ok, Msg, {"riak_kv.list_buckets", Type}};
+handle_decoded(#rpblistkeysreq{} = Msg) ->
+    Type = convert_type(Msg#rpblistkeysreq.type),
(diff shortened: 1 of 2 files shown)


src/riak_kv_pb_bucket.erl
```diff
@@ -66,90 +66,87 @@ init() ->
...
+determine_class_and_listing(#rpblistkeysreq{}) ->
+    {{riak_kv, stream_list_keys}, keys}.
+
+maybe_stream_list_keys(#rpblistkeysreq{type = Type, bucket = B, timeout = T} = Req,
+                       #state{client = Client} = State) ->
+    case check_bucket_type(Type) of
+        {ok, GoodType} ->
+            Bucket = maybe_create_bucket_type(GoodType, B),
+            {ok, ReqId} = Client:stream_list_keys(Bucket, T),
+            {reply, {stream, ReqId},
```
nickelization, line 126: Small formatting nitpick, but this line no longer really needs to be split up across multiple lines now that it's been cut and pasted out of the nested case statement where it used to be.

Same applies for the bodies of `error_no_bucket_type` and `error_accept` below, as well as the final clause of `validate_request` farther down.

src/riak_kv_pb_bucket.erl
```diff
@@ -66,90 +66,87 @@ init() ->
...
+%% @doc process/2 callback. Handles an incoming request message.
+process(Req, State) ->
+    {Class, Listing} = determine_class_and_listing(Req),
+    Accept = determine_accept_and_report_job_disposition(Class),
+    case {Accept, Listing} of
+        {true, buckets} ->
+            maybe_do_list_buckets(Req, State);
+        {true, keys} ->
+            %% at present list-keys always streams
+            %% Start streaming in list keys
```
nickelization, line 100: I would remove this "Start streaming in list keys" comment. Doesn't seem to really add anything, given the next function is called "maybe_stream_list_keys".

src/riak_kv_pb_bucket.erl
```diff
@@ -66,90 +66,87 @@ init() ->
...
-    Class = {riak_kv, stream_list_keys},
+%% @doc process/2 callback. Handles an incoming request message.
+process(Req, State) ->
+    {Class, Listing} = determine_class_and_listing(Req),
+    Accept = determine_accept_and_report_job_disposition(Class),
+    case {Accept, Listing} of
+        {true, buckets} ->
+            maybe_do_list_buckets(Req, State);
+        {true, keys} ->
+            %% at present list-keys always streams
```
nickelization, line 99: I think this "at present list-keys always streams" comment is useful, but might be better located down in the `determine_class_and_listing" function. It doesn't seem super relevant to the code here, whereas when I read the body of`determine_class_and_listing` my immediate response was "where's the clause for non-streaming keylists?"

nickelization:
Other than those tiny issues with formatting and comments, it all looks great to me. Much cleaner and more readable than the old code.

nickelization:
Talked with @JeetKunDoug and he suggested I go ahead and make my review changes myself, at which point he will +1 them. Will have the changes posted shortly.

src/riak_kv_pb_index.erl
```diff
@@ -56,81 +56,86 @@ init() ->
...
+        re:compile(TermRe)
+  end.
+
 %% @doc process/2 callback. Handles an incoming request message.
 process(#rpbindexreq{stream = S} = Req, State) ->
     Class = case S of
         true ->
             {riak_kv, stream_secondary_index};
-        _ ->
+        false ->
```
nickelization, line 106: Missed this during my first sweep earlier. This change makes me a bit nervous. Are we positive that this value will only ever be `true` or `false`? Could it be, say, `undefined`?

nickelization, line 106: Alright, talked with Doug and he said the protobuf specs claim that for optional fields, it should still default to false even if it's not specifically set either way.

PR #1470, opened 2016-08-31
Bump bitcask up to the develop branch

nickelization:
This is needed to pull in the switch over to the cuttlefish develop branch. Other than that it's the same version of bitcask that we were previously using from the tag.


hazen:
:+1: 7cc2a58

nickelization:
@borshop merge

PR #1473, opened 2016-09-02
Add v0 of versioning to AAE hashtree locks, upgrades, and object hashing

bsparrow435:
This PR adds version 0 of AAE hashing to riak_kv_entropy_manager, riak_kv_index_hashtree, riak_object, and the riak_kv_vnode. Specifically, we've added logic which requires getting a hashtree lock to provide an appropriate version and then our hashtree version determines what version of riak_object:hash will be used for the incoming data.

Upgrading is completed automatically for the user. The upgrade uses a combination of riak_core_capability with a new riak_kv variable object_hash_version and logic within the riak_kv_entropy manager to intelligently upgrade hashtrees once all conditions have been met. THe upgrade flow is as follows:
1. User upgrade all nodes to 2.2.
2. Each node locally starts checking to begin upgrade in riak_kv_entropy_manager within the management tick.
3. IIF all partitions are built and have exchanged with all sibling partitions, we'll mark the riak_kv_entropy_manager state with a new pending_version.(need to exchange all siblings to catch any pre 2.1 data potentially suffering from kv679).
4. Each poke on a riak_kv_index_hashtree check the pending version of the manager and if it's set, calls the vnode to upgrade it's local tree. 
5. Vnode destroys and makes a new hashtree with the `upgrade` atom included in the start opts.
6. Once all upgraded hashtrees have registered with entropy_manager, flip our entropy_manager version to 0 to indicate this node has completed the upgrade for all local nodes.

Open areas for discussion:
1. How best to handle default case of riak_object:hash without provided version(yz hashing). Current logic flips the hash when the entropy_manager notes all local trees have upgraded.
2. hashtree:destroy only removes data within the partition dir(i.e. ./data/antie_entropy/PARTID/*). The partition ID's still need to be cleaned up out of the anti-entropy directory.

### DIFF
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -583,7 +583,7 @@ tick(State) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     State1 = query_and_set_aae_throttle(State),
     State2 = maybe_reload_hashtrees(Ring, State1),
-    State3 = maybe_start_upgrade(State2),
+    State3 = maybe_start_upgrade(Ring, State2),
     State4 = lists:foldl(fun(_,S) ->
                                  maybe_poke_tree(S)
                          end, State3, lists:seq(1,10)),
@@ -598,15 +598,16 @@ maybe_poke_tree(State) ->
     riak_kv_index_hashtree:poke(Tree),
     State2.
 
--spec maybe_start_upgrade(state()) -> state().
-maybe_start_upgrade(State=#state{version=undefined, pending_version=undefined}) ->
+-spec maybe_start_upgrade(riak_core_ring(),state()) -> state().
+maybe_start_upgrade(Ring, State=#state{trees=Trees, version=undefined, pending_version=undefined}) ->
+    Indices = riak_core_ring:my_indices(Ring),
     case riak_core_capability:get({riak_kv, object_hash_version}, undefined) of
-        0 ->
+        0 when length(Trees) == length(Indices) ->
             maybe_upgrade(State);
-        undefined ->
+        _ ->
             State
     end;
-maybe_start_upgrade(State) ->
+maybe_start_upgrade(_Ring, State) ->
     State.
 
 -spec maybe_upgrade(state()) -> state().
@@ -616,7 +617,6 @@ maybe_upgrade(State=#state{trees_version = VTrees}) ->
             State;
         %% No trees have been upgraded, need to wait for exchanges
         Trees when length(Trees) == length(VTrees) ->
-            lager:info("Checking exchanges to attempt hashtree upgrade"),
             check_exchanges_and_upgrade(State);
         %% Upgrade already started, set pending_version
         _ ->
@@ -627,10 +627,9 @@ maybe_upgrade(State=#state{trees_version = VTrees}) ->
 check_exchanges_and_upgrade(State) ->
     case riak_kv_entropy_info:all_sibling_exchanges_complete() of
         true ->
-            lager:info("Starting upgrade"),
+            lager:info("Starting AAE hashtree upgrade"),
             State#state{pending_version=0};
         false ->
-            lager:info("Exchanges incomplete, delaying upgrade"),
             State
     end.
 
@@ -650,6 +649,7 @@ check_upgrade(State=#state{pending_version=Version,trees_version=VTrees}) ->
                 [] ->
                     State;
                 Trees when length(Trees) == length(VTrees) ->
+                    lager:info("Local AAE hashtrees have completed upgrade to version: ~p",[Version]),
                     State#state{version=Version, pending_version=undefined}
             end;
         _ ->
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -613,8 +613,9 @@ maybe_start_upgrade(_Ring, State) ->
 -spec maybe_upgrade(state()) -> state().
 maybe_upgrade(State=#state{trees_version = VTrees}) ->
     case [Idx || {Idx, undefined} <- VTrees] of
+        %% Upgrade is done already, set version in state
         [] ->
-            State;
+            State#state{version=0};
         %% No trees have been upgraded, need to wait for exchanges
         Trees when length(Trees) == length(VTrees) ->
             check_exchanges_and_upgrade(State);
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -35,7 +35,7 @@
          get_lock/2,
          release_lock/1,
          requeue_poke/1,
-         start_exchange_remote/3,
+         start_exchange_remote/4,
          exchange_status/4,
          expire_trees/0,
          clear_trees/0]).
@@ -117,15 +117,16 @@ release_lock(Pid) ->
 %%      then forwarded to the relevant index_hashtree to acquire a tree lock.
 %%      If both locks are acquired, the pid of the remote index_hashtree is
 %%      returned.
--spec start_exchange_remote({index(), node()}, index_n(), pid())
+-spec start_exchange_remote({index(), node()}, index_n(), pid(), non_neg_integer())
                            -> {remote_exchange, pid()} |
                               {remote_exchange, anti_entropy_disabled} |
                               {remote_exchange, max_concurrency} |
                               {remote_exchange, not_built} |
-                              {remote_exchange, already_locked}.
-start_exchange_remote(_VNode={Index, Node}, IndexN, FsmPid) ->
+                              {remote_exchange, already_locked} |
+                              {remote_exchange, bad_version}.
+start_exchange_remote(_VNode={Index, Node}, IndexN, FsmPid, Version) ->
     gen_server:call({?MODULE, Node},
-                    {start_exchange_remote, FsmPid, Index, IndexN},
+                    {start_exchange_remote, FsmPid, Index, IndexN, Version},
                     infinity).
 
 %% @doc Used by {@link riak_kv_index_hashtree} to requeue a poke on
@@ -267,7 +268,7 @@ handle_call(disable, _From, State) ->
 handle_call({get_lock, Type, Pid}, _From, State) ->
     {Reply, State2} = do_get_lock(Type, Pid, State),
     {reply, Reply, State2};
-handle_call({start_exchange_remote, FsmPid, Index, IndexN}, From, State) ->
+handle_call({start_exchange_remote, FsmPid, Index, IndexN, Version}, From, State) ->
     case {enabled(),
           orddict:find(Index, State#state.trees)} of
         {false, _} ->
@@ -279,7 +280,7 @@ handle_call({start_exchange_remote, FsmPid, Index, IndexN}, From, State) ->
                 {ok, State2} ->
                     %% Concurrency lock acquired, now forward to index_hashtree
                     %% to acquire tree lock.
-                    riak_kv_index_hashtree:start_exchange_remote(FsmPid, From, IndexN, Tree),
+                    riak_kv_index_hashtree:start_exchange_remote(FsmPid, Version, From, IndexN, Tree),
                     {noreply, State2};
                 {Reply, State2} ->
                     {reply, {remote_exchange, Reply}, State2}
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -471,7 +471,8 @@ add_hashtree_pid(true, Index, Pid, State=#state{trees=Trees, trees_version=VTree
             VTrees2 = orrdict:store(Index, riak_kv_index_hashtree:get_version(Pid), VTrees),
             State2 = State#state{trees=Trees2, trees_version=VTrees2},
             State3 = add_index_exchanges(Index, State2),
-            State3
+            State4 = check_upgrade(State3),
+            State4
     end.
 
 -spec do_get_lock(any(),pid(),state())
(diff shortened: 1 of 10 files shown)


src/riak_kv_entropy_manager.erl
```diff
@@ -116,16 +124,27 @@ release_lock(Pid) ->
 %%      will try to acquire a concurrency lock. If successsful, the request is
 %%      then forwarded to the relevant index_hashtree to acquire a tree lock.
 %%      If both locks are acquired, the pid of the remote index_hashtree is
-%%      returned.
+%%      returned. This function assumes an undefined version of the hashtree
```
nickelization, line 128: Instead of stating what the function does, I feel like this comment would be more helpful if it explained why it does it. Something like "this is left over for compatibility in mixed clusters with pre-2.2 nodes, from before we added versioning to the hashtree."

src/riak_kv_entropy_manager.erl
```diff
@@ -189,6 +208,15 @@ expire_trees() ->
 clear_trees() ->
     gen_server:cast(?MODULE, clear_trees).
 
+get_version() ->
```
nickelization, line 219: Would be nice to have type specs on these as long as we're adding new API functions. I wouldn't let it stop this from passing review since none of the other API calls in this module seem to have type specs, but just thought I'd suggest it in case you're waiting for more review comments and you have a chance to add them :)

src/riak_kv_entropy_manager.erl
```diff
@@ -550,6 +598,66 @@ maybe_poke_tree(State) ->
...
+-spec check_upgrade(state()) -> state().
+check_upgrade(State=#state{pending_version=undefined}) ->
+    State;
+check_upgrade(State=#state{pending_version=Version,trees_version=VTrees}) ->
+    %% Verify we have all partitions registered with a hashtree, version
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    Indices = riak_core_ring:my_indices(Ring),
+    Existing = dict:from_list(VTrees),
+    MissingIdx = [Idx || Idx <- Indices,
+        not dict:is_key(Idx, Existing)],
```
nickelization: If we wanted to simplify the code a little bit, we could skip the extra step of converting to a `dict` and just use `lists:keymember`.

src/riak_kv_entropy_manager.erl
```diff
@@ -418,7 +453,12 @@ reload_hashtrees(Ring, State=#state{trees=Trees}) ->
     Existing = dict:from_list(Trees),
     MissingIdx = [Idx || Idx <- Indices,
                          not dict:is_key(Idx, Existing)],
-    _ = [riak_kv_vnode:request_hashtree_pid(Idx) || Idx <- MissingIdx],
```
nickelization, line 421: What's the purpose of this change? Seems like the old version was simpler and would have the same effect, unless I'm missing something.

src/riak_kv_entropy_manager.erl
nickelization: Oh, or even just `orddict:is_key`, since we seem to be treating `trees_version` as an `orddict`.

src/riak_kv_entropy_manager.erl
```diff
@@ -66,16 +70,20 @@
 -type vnode() :: {index(), node()}.
 -type exchange() :: {index(), index(), index_n()}.
 -type riak_core_ring() :: riak_core_ring:riak_core_ring().
+-type version() :: undefined | non_neg_integer().
 
 -record(state, {mode           = automatic :: automatic | manual,
                 trees          = []        :: [{index(), pid()}],
                 tree_queue     = []        :: [{index(), pid()}],
+                trees_version  = []        :: [{index(), version()}],
```
nickelization: Since we're using this as an `orddict`, we should have the type spec indicate that for clarity. i.e. `orddict:orddict(index(), version())` instead of `[{index(), version()}]`

src/riak_kv_entropy_manager.erl
bsparrow435: This is left over from a WIP commit. I was originally doing some logic in the special empty list case before returning ok but I removed it and forgot to revert this as well.

src/riak_kv_entropy_manager.erl
bsparrow435: Yep absolutely, I was reusing this logic from some ring logic i added and it should be using the orddict functions.

src/riak_kv_entropy_manager.erl
```diff
@@ -550,6 +598,66 @@ maybe_poke_tree(State) ->
...
+        %% Upgrade already started, set pending_version
+        _ ->
+            State#state{pending_version=0}
+    end.
+
+-spec check_exchanges_and_upgrade(state()) -> state().
+check_exchanges_and_upgrade(State) ->
+    case riak_kv_entropy_info:all_sibling_exchanges_complete() of
+        true ->
+            lager:info("Starting AAE hashtree upgrade"),
```
nickelization: Should we make the level for this `notice` instead of `info`? Seems like a rare/significant enough event to warrant the extra visibility.

src/riak_kv_entropy_manager.erl
bsparrow435: Yep, agreed, will change and do the same with trees since the code always treats trees as an orddict as well.

src/riak_kv_entropy_manager.erl
bsparrow435: Will change.

src/riak_kv_entropy_manager.erl
bsparrow435: Yeah, I'll add it to all these api calls. You're right, public facing API calls should absolutely have specs.

src/riak_kv_entropy_manager.erl
```diff
@@ -550,6 +598,66 @@ maybe_poke_tree(State) ->
...
+            State#state{pending_version=0};
+        false ->
+            State
+    end.
+
+-spec check_upgrade(state()) -> state().
+check_upgrade(State=#state{pending_version=undefined}) ->
+    State;
+check_upgrade(State=#state{pending_version=Version,trees_version=VTrees}) ->
```
nickelization: Could we name this variable something like `PendingVersion` instead? It seems a little confusing to have two state fields named `version` and `pending_version` and then bind `Version` to `pending_version` and not `version`.

src/riak_kv_entropy_manager.erl
```diff
@@ -550,6 +598,66 @@ maybe_poke_tree(State) ->
...
+    %% Verify we have all partitions registered with a hashtree, version
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    Indices = riak_core_ring:my_indices(Ring),
+    Existing = dict:from_list(VTrees),
+    MissingIdx = [Idx || Idx <- Indices,
+        not dict:is_key(Idx, Existing)],
+    case MissingIdx of
+        [] ->
+            case [Idx || {Idx, V} <- VTrees, V == Version] of
+                [] ->
```
nickelization, line 665: I see the `[]` case for when none of the local hashtrees have been upgraded, and then the following case where all local trees have been upgraded, but what if only some have been upgraded? Is that a case we might run into and then end up crashing here?

src/riak_kv_get_core.erl
```diff
@@ -284,6 +288,36 @@ num_pr(GetCore = #getcore{num_pok=NumPOK, idx_type=IdxType}, Idx) ->
             GetCore
     end.
 
+%% @private Print a warning if objects are not equal. Only called on case of no read-repair
+%% This situation could happen with pre 2.1 vclocks in very rare cases. Fixing the object
+%% requires the user to rewrite the object in 2.1+ of Riak. Logic is disabled when using
+%% a defined version of aae hashing as it's no longer needed.
+maybe_log_old_vclock(Results) ->
+    case riak_core_capability:get({riak_kv, object_hash_version}, undefined) of
```
nickelization, line 297: So right now, as I understand it, we don't start the hashtree upgrades as soon as the capability flips, because we want to make sure all partitions have exchanged first so that we get a chance to log any inconsistent data here in this function. However, it looks like we don't do any logging here once the capability flips, right? I might be missing something but it seems like we may want to be checking something other than the capability to decide whether to check for mismatched data.

nickelization, line 297: Oh wait, so actually I misread the code there. The comment up above this function says "logic is disabled when using a defined version of aae hashing as it's no longer needed," but it looks like we actually disable it when we're _not_ using a defined version of aae hashing, right? Is this actually what we want to do though? Feels like something is backwards here but maybe it's just my brain hitting late Friday afternoon and starting to stall out on me ;)

src/riak_kv_get_core.erl
```diff
@@ -284,6 +288,36 @@ num_pr(GetCore = #getcore{num_pok=NumPOK, idx_type=IdxType}, Idx) ->
...
+        undefined ->
+            ok;
+        0 ->
+            Version = riak_kv_entropy_manager:get_version(),
+            case [RObj || {_Idx, {ok, RObj}} <- Results] of
+                [] ->
+                    ok;
+                [_|[]] ->
```
nickelization: Very small nitpick, but the pattern I normally see to match a singleton list is just `[_]`. I'm sure this works just as well, but feels a little bit unidiomatic.

src/riak_kv_index_hashtree.erl
```diff
@@ -81,7 +84,8 @@
                 path,
                 build_time,
                 trees,
-                use_2i = false :: boolean()}).
+                use_2i = false :: boolean(),
+                version :: undefined | non_neg_integer()}).
```
nickelization: Technically the `undefined |` is implicit in the type spec for any Erlang record field that doesn't include a default value. Maybe you already knew this and just added it for clarity, in which case I have no problem with that, but thought I'd point that out in case you'd rather cut down on the verbosity a little.

src/riak_kv_index_hashtree.erl
```diff
@@ -198,13 +202,31 @@ get_trees({test, Pid}) ->
...
+-spec get_lock(pid(), any(), non_neg_integer()) -> ok | not_built | already_locked | bad_version.
+get_lock(Tree, Type, Version) ->
+    get_lock(Tree, Type, Version, self()).
 
 %% @doc Acquire the lock for the specified index_hashtree if not already
 %%      locked, and associate the lock with the provided pid.
--spec get_lock(pid(), any(), pid()) -> ok | not_built | already_locked.
-get_lock(Tree, Type, Pid) ->
-    gen_server:call(Tree, {get_lock, Type, Pid}, infinity).
+-spec get_lock(pid(), any(), non_neg_integer(), pid()) -> ok | not_built | already_locked.
```
nickelization: Should update this type spec with `bad_version` as a possible return value.

src/riak_kv_entropy_manager.erl
bsparrow435: Good feedback, will change.

src/riak_kv_entropy_manager.erl
bsparrow435: Yep, that makes more sense, will change.

src/riak_object.erl
```diff
@@ -623,12 +623,37 @@ get_value(Object=#r_object{}) ->
...
+legacy_hash(Obj=#r_object{}) ->
+    UpdObj = riak_object:set_vclock(Obj, lists:sort(vclock(Obj))),
+    Hash = erlang:phash2(to_binary(v0, UpdObj)),
+    term_to_binary(Hash).
+
+%% @private return the hash of the vclock
+-spec vclock_hash(riak_object()) -> binary().
+vclock_hash(Obj=#r_object{}) ->
+    Hash = erlang:phash2(to_binary(v0, lists:sort(vclock(Obj)))),
+    term_to_binary(Hash).
```
nickelization, line 656: As long as we're changing the hashing format, could we take this opportunity to optimize our hashing code any further?

For examples, it looks like `erlang:phash2` accepts arbitrary terms, so perhaps we don't need to wrap the result of `lists:sort(vclock(Obj))` in a call to `to_binary` - not sure if this would perform better or worse, but it seems like there's a good chance it would perform better since we'd be taking an extra encoding step out of the process.

Also, do we need to use `term_to_binary` at the end, or could we just do something more lightweight to convert it to a binary like `<<Hash:32>>`? That would almost certainly be faster, and yields smaller binaries too.

src/riak_kv_entropy_manager.erl
bsparrow435: Good catch. This was causing some crashes in a slower test I ran last night. My normal tests allow for all trees to upgrade in the same tick for speed and was missing this error.

src/riak_kv_get_core.erl
bsparrow435: So I think the comment is a bit misleading and I have changed it for clarity. We only want to trigger this heavy riak_object comparison and logging when we are in the process of the hashtree upgrade. So, if the user has manually overridden the capability to set `undefined` version, we want this logic off.  If the capability has been set to 0(all nodes are upgraded to 2.2) and the riak_kv_entropy_manager:get_version call returns undefined(anything but 0) then we know the upgrade has not yet finished and we should log in specific situations where we have multiple replica objects in the Result set and they are also not equal.

nickelization:
@bsparrow435 So it occurred to me this morning that we should probably have an option to disable automatically upgrading to the new tree format. It doesn't necessarily have to be documented or supported, but I think it would be good to provide an escape hatch. If nothing else, it may make testing much easier if we want to, say, write a riak_test module that starts with 2.2 and then performs a downgrade (e.g. something like `verify_riak_object_reformat`).

nickelization:
+1 0e6bb3e

nickelization:
@borshop merge

Issue #1472, opened 2016-09-01
Crashing errors when using date/time strings in riak-shell [JIRA: RIAK-2782]


macintux:
This appears to be a consequence of a nonsensical query (> noon and < 1pm, when using reduced accuracy times, is effectively the same as `>= X andalso < X`.

```
riak-shell(15)>select * from training where time > '2014-05-03 12' and time < '2014-05-03 13';
SQL Lexer error {{{case_clause,
                      {error,
                          <<"Error processing incoming message: error:function_clause:[{riak_ql_quanta,\n                                                           gen_quanta,\n                                                           [0,1399122000000,\n                                                            3600000,[]],\n                                                           [{file,\n                                                             \"src/riak_ql_quanta.erl\"},\n                                                            {line,89}]},\n                                                          {riak_ql_quanta,\n                                                           quanta,4,\n                                                           [{file,\n                                                             \"src/riak_ql_quanta.erl\"},\n                                                            {line,81}]},\n                                                          {riak_kv_qry_compiler,\n                                                           hash_timestamp_to_quanta,\n                                                           6,\n                                                           [{file,\n                                                             \"src/riak_kv_qry_compiler.erl\"},\n                                                            {line,547}]},\n                                                          {riak_kv_qry_compiler,\n                                                           expand_query,4,\n                                                           [{file,\n                                                             \"src/riak_kv_qry_compiler.erl\"},\n                                                            {line,108}]},\n                                                          {riak_kv_qry,\n                                                           do_select,2,\n                                                           [{file,\n                                                             \"src/riak_kv_qry.erl\"},\n                                                            {line,289}]},\n                                                          {riak_kv_ts_svc,\n                                                           sub_tsqueryreq,4,\n                                                           [{file,\n                                                             \"src/riak_kv_ts_svc.erl\"},\n                                                            {line,434}]},\n                                                          {riak_api_pb_server,\n                                                           process_message,4,\n                                                           [{file,\n                                                             \"src/riak_api_pb_server.erl\"},\n                                                            {line,388}]},\n                                                          {riak_api_pb_server,\n                                                           connected,2,\n                                                           [{file,\n                                                             \"src/riak_api_pb_server.erl\"},\n                                                            {line,226}]}]">>}},
                  [{connection_srv,handle_call,3,
                       [{file,"src/connection_srv.erl"},{line,96}]},
                   {gen_server,handle_msg,5,
                       [{file,"gen_server.erl"},{line,585}]},
[...]

PR #1474, opened 2016-09-06
Address select exception at quantum boundaries

macintux:
Addresses #1472 (RIAK-2782). `riak_test` to follow.

1.4.0 shipped with a exception on selecting at quantum boundaries that was exacerbated by the new time parsing and not-entirely-intuitive handling of reduced accuracy times.

`select * from foo where time > '2016-01-01 05' and time < '2016-01-01 06'` e.g. seems like a reasonable query but in effect both time values are the same.

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -26,6 +26,10 @@
 -export([finalise_aggregate/2]).
 -export([run_select/2, run_select/3]).
 
+-ifdef(TEST).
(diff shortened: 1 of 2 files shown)


hazen:
Could we get a unit test for this one?

macintux:
Unit test added @javajolt

hazen:
:+1: 83673bc

hazen:
@borshop merge

PR #1475, opened 2016-09-07
Handle the case with no fsm_limit being specified

nickelization:
Prior to #1420, an fsm_limit setting of `undefined` meant that overload
protection would be disabled. Since those code paths have been removed,
we had originally figured that everyone would have to specify an FSM
limit now.

However, this wouldn't be terribly friendly to anyone with an existing
cluster set up that had disabled overload protection. To smooth over
this transition and still allow existing config to work, we once again
allow a setting of `undefined` to mean that there's no limit.

There is no option in sidejob to allow us to specify "no limit", but by
setting the limit to the maximum process limit on the node, we can
pretty much guarantee that the limit will never get hit, making it all
effectively unlimited.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -49,20 +49,7 @@
 start(_Type, _StartArgs) ->
     riak_core_util:start_app_deps(riak_kv),
 
-    FSM_Limit = case app_helper:get_env(riak_kv, fsm_limit, ?DEFAULT_FSM_LIMIT) of
-                    undefined ->
-                        %% If it's explicitly set to 'undefined', that means "no limit".
-                        %% sidejob doesn't have an option to specify "infinity" but if
-                        %% we make the limit equal to the node's global process limit,
-                        %% then we're pretty much guaranteed to never exceed that:
-                        erlang:system_info(process_limit);
-                    Limit when is_integer(Limit) ->
-                        Limit;
-                    BadValue ->
-                        lager:critical("Bad value provided for riak_kv.fsm_limit - ~p. "
-                                       "Must be an integer or 'undefined'", [BadValue]),
-                        throw({error, bad_fsm_limit})
-                end,
+    FSM_Limit = find_fsm_limit(),
 
     sidejob:new_resource(riak_kv_put_fsm_sj, sidejob_supervisor, FSM_Limit),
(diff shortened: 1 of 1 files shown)


src/riak_kv_app.erl
```diff
@@ -49,7 +49,21 @@
 start(_Type, _StartArgs) ->
     riak_core_util:start_app_deps(riak_kv),
 
-    FSM_Limit = app_helper:get_env(riak_kv, fsm_limit, ?DEFAULT_FSM_LIMIT),
+    FSM_Limit = case app_helper:get_env(riak_kv, fsm_limit, ?DEFAULT_FSM_LIMIT) of
```
JeetKunDoug: Please refactor this code into its own function - it's now too much to just embed in another function. Maybe still get the env here, but pass it to a "determine_fsm_limit/1" that has different function heads for each case?

src/riak_kv_app.erl
```diff
@@ -49,7 +49,21 @@
...
+                    undefined ->
+                        %% If it's explicitly set to 'undefined', that means "no limit".
+                        %% sidejob doesn't have an option to specify "infinity" but if
+                        %% we make the limit equal to the node's global process limit,
+                        %% then we're pretty much guaranteed to never exceed that:
+                        erlang:system_info(process_limit);
```
JeetKunDoug: @angrycub would it be wise to do something like 90% of the process limit, vs. all of it? Or subtract some minimal number (1k? 5k?) to reserve so if things go really south we're still providing some protection to the node?

angrycub: To be honest, it's no more _dangerous_ then infinite has always been.  The user would be specifically turning off the governor.  I'm not sure if there is a good way to know or even estimate the number of processes necessary to allow the node not to get sad when you are running out this far.   I'm for safer where we can get it, but this option has always been a bit "Hey Y'all, Watch this..."

nickelization: I can definitely see your point here, but I think I would object to that change on principle. If we're going to allow people to say "never trigger overload no matter what", it seems like we ought to honor that, even if 99.9% of the time it's a stupid thing to do.

bsparrow435:
+1 351112b245ece9224a6871a5d6b886eef1c320aa

nickelization:
@borshop merge

PR #1476, opened 2016-09-08
add correct cap for second level cap when HLLs are not around

zeeshanlakhani:


nickelization:
+1 5693064

zeeshanlakhani:
@borshop merge

PR #1477, opened 2016-09-09
Changed job control switches to be commented with default values

fadushin:
This PR changes the job control settings added in 2.2 so that they are commented with their associated default values.  Doing so reduces friction in the (unlikely) case that a new install is downgraded to a previous version.


jvoegele:
+1 ff12939

fadushin:
@borshop merge

PR #1478, opened 2016-09-13
RTS-1391: qry:do_explain: correctly spell column type as 'sint64' not 'integer'

hmmr:
RTS-1391

In `riak_kv_qry:explain`, the integer type for column "Subquery" was misspelled as `integer`, which causes a function_clause error in `riak_pb_ts_codec:encode_field_type/1` when it fails to match here: https://github.com/basho/riak_pb/blob/develop/src/riak_pb_ts_codec.erl#L70.

**Update:** Note that the bug is only triggered when the client uses the old PB encoding method (instead of TTB), as is the case with the Python client. This explains why it has gone undetected (the relevant r_t module, ts_simple_explain_query passes both with and without the fix).

### DIFF
--- a/src/riak_kv_qry.erl
+++ b/src/riak_kv_qry.erl
@@ -467,12 +467,12 @@ explain_query_test() ->
     ?assert(meck:validate(riak_core_claimant)),
     Res = do_explain(DDL, Select),
     ExpectedRows =
-        [[1,"dev1@127.0.0.1/0, dev1@127.0.0.1/1, dev1@127.0.0.1/1",
-            "c = 'hola', b = 1",false,"c = 'hola', b = 1000",false,
-            "(((d = 15) OR ((e = true) AND (f = 'adios'))) AND (a = 319))"],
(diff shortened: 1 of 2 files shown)


andytill:
Discussed with Andrei and agreed that this should have a rt test where the client specifies the `use_ttb=false` option to verify the issue and fix.

hazen:
@andytill Great idea.  I'll look at beefing up the `EXPLAIN` tests.

hmmr:
There used to be a switch in ts_cluster_comprehensive, to the effect of
repeating the whole battery with `{encoding, pb}` (or something), which
should do the trick.

andytill:
It would be better if it was not mixed up with cluster comprehensive, that test is very difficult to work with, and debug.

hazen:
@hmmr I found that the strings need to binaries, too, to be able to encode for protocol buffers.

hazen:
Here is the updated corresponding `riak_test`: https://github.com/basho/riak_test/pull/1150

andytill:
+1 709b4e5

hmmr:
@borshop merge

PR #1479, opened 2016-09-14
Enable query buffers (temp tables) to support LIMIT, OFFSET and ORDER BY

hmmr:
RTS-733, RTS-870, RTS-1006

Depends on https://github.com/basho/riak_ql/pull/139, https://github.com/basho/riak_pb/pull/208; r_t module in https://github.com/basho/riak_test/pull/1152.

This PR implements https://github.com/basho/rfc/pull/5:
1. Introduces query buffers, a node-local temporary storage for SELECT queries, which can now be of arbitrary size and span arbitrary range (thus lifting the restriction on max subqueries).
2. Enables sorting of SELECT query results with the new keyword ORDER BY.
3. Enables the LIMIT and OFFSET keywords.
4. Allows reuse of query buffers for subsequent matching queries, controlled by the newly introduced flag `#tsqueryreq.allow_qbuf_reuse` (off by default).


hmmr:
Reverting to WIP status as I and @andytill have discovered issues around lifting the MaxSubqueries limitation.

hmmr:
In 6eca18d, I am replacing the limit on max subqueries with a total query result size limit. It is calculated as `TotalQuerySize = AverageChunkSize * NSubqueries` for regular queries (those not backed by the query buffers). For query buffer-based queries, `TotalQuerySize` is the LIMIT value times average chunk size.

If the total query result size is projected to exceed a certain, configurable value (hardcoded to be 5 MiB), the query is aborted.

The sizes are computed after incorporation of each chunk, starting from chunk 2, in and around https://github.com/basho/riak_kv/pull/1479/files#diff-85d69bbf5fbaa852650866d125af628aR287.

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+maybe_inform_waiting_process(_IsReady, _Pid, _QBufRef) ->
+    nop.
+
+
+do_set_ready_waiting_process(QBufRef, Pid, #state{qbufs = QBufs0} = State0) ->
+    case get_qbuf_record(QBufRef, QBufs0) of
+        false ->
+            {reply, {error, bad_qbuf_ref}, State0};
+        #qbuf{is_ready = true} ->
+            %% qbuf is already ready: don't wait for the bext
```
gordonguthrie: 'bext' should be 'next'

include/riak_kv_ts.hrl
```diff
@@ -52,8 +52,8 @@
           'SELECT'              :: #riak_sel_clause_v1{},
           'FROM'        = <<>>  :: binary() | {list, [binary()]} | {regex, list()},
           'WHERE'       = []    :: [riak_ql_ddl:filter()],
-          'ORDER BY'    = []    :: [riak_kv_qry_compiler:sorter()],
-          'LIMIT'       = []    :: [riak_kv_qry_compiler:limit()],
+          'ORDER BY'            :: undefined | [riak_kv_qry_compiler:sorter()],
```
gordonguthrie: These elements are empty lists because the values are mini-LISP programmes that are to be evaluated elsewhere

include/riak_kv_ts.hrl
```diff
@@ -83,6 +83,27 @@
...
+          partition_key = none  :: none | #key_v1{},
+          %% indicates whether this query has already been compiled to a sub query
+          is_executable = false :: boolean(),
+          type          = sql   :: sql | timeseries,
+          cover_context = undefined :: term(), %% for parallel queries
+          allow_qbuf_reuse = false  :: boolean(),  %% control reuse of query buffers
+          %% prolly a mistake to put this here - should be in DDL
+          local_key,
+          %% since v2
+          group_by = ?GROUP_BY_DEFAULT :: [{identifier, binary()}] | [{FieldPos::integer(), FieldName::binary()}]
```
gordonguthrie: This should be 'GROUP BY' by convention we preserve SQL clauses upper case in the SQL records - the rational is that these fields are semantically the same as the original SQL clauses (or can be so reconciled) whilst containing syntactic differences pertaining to execution. It should also be insert up into the record with the rest of the SQL clauses.

src/riak_kv_app.erl
```diff
@@ -213,7 +213,7 @@ start(_Type, _StartArgs) ->
                                            lists:last(riak_ql_ddl_compiler:get_compiler_capabilities())),
 
             riak_core_capability:register({riak_kv, sql_select_version},
-                                          [v2,v1],
+                                          [v3,v2,v1],
```
gordonguthrie, line 220: **Note** implies upgrade/downgrade

src/riak_kv_qry.erl
```diff
@@ -281,22 +281,62 @@ count_to_position(Col, [_ | Rest], Pos) ->
...
                 {ok, SubQueries} ->
-                    maybe_await_query_results(
-                      riak_kv_qry_queue:put_on_queue(self(), SubQueries, DDL))
+                    FullCycle =
+                        fun(QBufRef) ->
+                            maybe_await_query_results(
+                              riak_kv_qry_queue:put_on_queue(self(), SubQueries, DDL, QBufRef))
+                        end,
+                    case maybe_create_query_buffer(SQL, length(SubQueries), DDL, []) of
+                        {error, _Reason} ->
```
gordonguthrie: If it fails, it fails and should be logged and reported as an error instead of being left to go on and fail somewhere else

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+-behaviour(gen_server).
+
+-export([start_link/0,
+         init/1,
+         handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3
+        ]).
+
+-export([
+         batch_put/2,           %% emulate INSERT (new Chunk collected by worker for a query Ref)
```
gordonguthrie, line 53: There are two seperate API's here:

Used API:
• `data_row`
• `fetch_limit`
• `qbuf_ref`
• `batch_put`
• `delete_qbuf`
• `set_ready_waiting_process`
• `get_or_create_qbuf`

Developer API:
• `get_qbuf_orig_query`
• `get_qbuf_size`
• `get_qbufs`
• `is_qbuf_ready`
• `set_qbuf_expiry`
• `make_qref`

If you have a developer/testing API you have a couple of options - certainly have a different export statement - and consider decorating the function names with things like `make_qref_DEV`

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+-spec start_link() -> {ok, pid()} | ignore | {error, term()}.
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+
+-spec init(proplists:proplist()) -> {ok, #state{}}.
+%% @private
+init(Options) ->
+    RootPath =
+        filename:join(
```
gordonguthrie: @The 'design' of the data directory is a UX thing (we expose stuff to customers/CSEs) so lets get @angrycub 's view on this.

Also the fileroot should be passed in as an option - this means that a test suite can be written which writes to /tmp or a memory disk or whatever...

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+               root_path =
+                   proplists:get_value(root_path, Options, RootPath)
+              },
+    Self = self(),
+    _TickerPid = spawn_link(fun() -> schedule_tick(Self) end),
+    {ok, State}.
+
+schedule_tick(Pid) ->
+    Pid ! tick,
+    timer:sleep(1000),
```
gordonguthrie: Macroise 1000 and setup at to the top the rest of your timeouts, watermarks etc...

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+    schedule_tick(Pid).
+
+-spec handle_call(term(), pid() | {pid(), term()}, #state{}) -> {reply, term(), #state{}}.
+%% @private
+handle_call(get_qbufs, _From, State) ->
```
gordonguthrie: this handle_call/do_call pattern is a waste of space...

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+
+
+-spec handle_cast(term(), #state{}) -> {noreply, #state{}}.
+%% @private
+handle_cast(_Msg, State) ->
+    lager:warning("Not handling cast message ~p", [_Msg]),
+    {noreply, State}.
+
+
+-spec handle_info(term(), #state{}) -> {noreply, #state{}}.
```
gordonguthrie, line 335: I would do the tick in a `call` or `cast` not `info`

`info` is for messages that your process gets in some only slightly expected way from any other process (or some random problem messages)

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+%% @doc Make a query ref (or hash), to identify this (and similar) queries as
+%%      having certain SELECT, FROM, WHERE and ORDER BY clauses.
+make_qref(?SQL_SELECT{'SELECT'   = #riak_sel_clause_v1{col_names = ColNames},
+                      'FROM'     = From,
+                      'WHERE'    = Where,
+                      'ORDER BY' = OrderBy})
+  when OrderBy /= undefined ->
+    Part0 = erlang:phash2(From),
+    Part1 = erlang:phash2(ColNames),
+    Part2 = erlang:phash2(lists:sort(Where)),
```
gordonguthrie: The WHERE clause is canonicalised in the lexer/parser - all the clauses should be canonical coming out of that...

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+                    end
+            end;
+        {error, Reason} ->
+            {reply, {error, Reason}, State0}
+    end.
+
+maybe_ensure_qref(?SQL_SELECT{allow_qbuf_reuse = true} = SQL, QBufs) ->
+    ensure_qref(SQL, QBufs);
+maybe_ensure_qref(?SQL_SELECT{allow_qbuf_reuse = false}, _QBufs) ->
+    AlwaysUniqueRef = crypto:rand_bytes(8),
```
gordonguthrie, line 414: Don't use hash for unique values - use References - first class Erlang terms

http://erlang.org/doc/man/erlang.html#make_ref-0

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+            end
+    end.
+
+get_ordby_field_qualifiers({_, Dir, NullsGroup}) ->
+    {Dir, NullsGroup}.
+
+
+maybe_inform_waiting_process(true, Pid, QBufRef) when is_pid(Pid) ->
+    lager:debug("notifying waiting pid ~p", [Pid]),
+    Pid ! {qbuf_ready, QBufRef};
```
gordonguthrie: @lehoff recommends not passing in PIDs but using fns like this as a reference

```
%% Caputure Pid in a closuse
RefToMeFn = fun(Msg) -> 
        Pid ! {qbuf_read, Msg),
        ok
end,
setup_query_buffer(Blah, Bleh, Bloh, RefToMeFn)
...
```

And then in your module you do

```
ok = ReplyFn(MyMsg)
```

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
+%%-------------------------------------------------------------------
```
gordonguthrie, line 1: **Note** this module needs a test - prolly a common test or mebbies EQC

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+    Now = os:timestamp(),
+    QBufs9 =
+        lists:filter(
+          fun({_QBufRef, #qbuf{is_ready = true,
+                               ldb_ref = LdbRef,
+                               ddl = ?DDL{table = Table},
+                               expire_msec = ExpireMsec,  %% qbuf-specific, possibly overriden
+                               last_accessed = LastAccessed}}) ->
+                  ExpiresOn = advance_timestamp(LastAccessed, ExpireMsec),
+                  if ExpiresOn < Now ->
```
gordonguthrie: use a case statement

src/riak_kv_qry_buffers.erl
gordonguthrie: What is the GUI design for specifying the total system resource limits as per @matthewvon 's spec for this? We need an associated `clique` change

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+%% @private Create a DDL to accommodate the SELECT. Note that the
+%% original query comes here not compiled and therefore, has fields
+%% appearing as `{identifier, Field}` rather than `Field` (and has no
+%% types).
+sql_to_ddl(OrigDDL, OrigSQL) ->
+    %% We compile the original SQL again (only the parts we need,
+    %% i.e., without WHERE or GROUP BY). We can match on {ok, SQL}
+    %% because any errors will have been detected when the compilation
+    %% chain was run on the query in riak_kv_qry_compiler.
+    {ok, CompiledSelect} = riak_kv_qry_compiler:compile_select_clause(OrigDDL, OrigSQL),
```
gordonguthrie: This looks like a code smell - we have already compiled the SQL and we are now recompiling it to re-extract bits...

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+        riak_kv_qry_compiler:compile_order_by(
+          {ok, OrigSQL?SQL_SELECT{'SELECT' = CompiledSelect}}),
+    ?DDL{table         = make_qbuf_id(From, Select, OrderBy),
+         fields        = make_fields_from_select(Select),
+         partition_key = none,
+         %% this ensures the right natural order in the newly created
+         %% eleveldb
+         local_key     = make_lk_from_orderby(OrderBy)}.
+
+make_qbuf_id(From, Select, OrderBy) ->
```
gordonguthrie, line 644: This query buffer ID is not unique because the timestamp isn't - `os:now()` is guaranteed to be unique but has a cost and will be deprecated in more modern versions of Erlang

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,172 @@
...
+-module(riak_kv_qry_buffers_ldb).
+
+-export([new_table/2,
+         delete_table/3,
+         add_rows/5,
+         fetch_rows/3]).
+
+
+%% leveldb instance parameters
+-define(LDB_WRITE_BUFFER_SIZE, 10*1024*1024).  %% 10 M should be enough for everybody
```
gordonguthrie, line 35: What is this? How does it relate to @matthewvon 's params?

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,172 @@
...
+
+
+-type errors() :: ldb_put_failed.
+-export_type([errors/0]).
+
+-spec new_table(binary(), string()) -> {ok, eleveldb:db_ref()} | {error, term()}.
+new_table(Table, Root) ->
+    Path = filename:join(Root, binary_to_list(Table)),
+    _ = os:cmd(fmt("mkdir -p '~s'", [Path])),
```
gordonguthrie: Don't use `os:cmd`

http://erlang.org/doc/man/file.html#make_dir-1

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,815 @@
...
+
+-spec init(proplists:proplist()) -> {ok, #state{}}.
+%% @private
+init(Options) ->
+    RootPath =
+        filename:join(
+          app_helper:get_prop_or_env(data_root, Options, eleveldb),  %% reuse eleveldb own root_path
+          "query_buffers"),
+    %% don't bother recovering any leftover tables
+    os:cmd(
```
gordonguthrie: Don't use `os:cmd`

http://erlang.org/doc/man/file.html#del_dir-1

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,172 @@
...
+               {compression, false},
+               %% this prevents leveldb from autoexpiring records:
+               {is_internal_db, true}
+              ],
+    case eleveldb:open(Path, Options) of
+        {ok, LdbRef} ->
+            lager:info("new LdbRef ~p in ~p", [LdbRef, Path]),
+            {ok, LdbRef};
+        {error, _} = ErrorReason ->
+            _ = os:cmd(fmt("rm -rf '~s'", [Path])),
```
gordonguthrie: Don't use `os:cmd` ;-)

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,172 @@
...
+            ErrorReason
+    end.
+
+-spec delete_table(binary(), eleveldb:db_ref(), string()) -> ok.
+delete_table(Table, LdbRef, Root) ->
+    ok = eleveldb:close(LdbRef),
+    _ = os:cmd(
```
gordonguthrie: Don't use `os:cmd` ;-)

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,172 @@
...
+                  RowEnc = sext:encode(Row),
+                  ok = eleveldb:put(LdbRef, KeyEnc, RowEnc, [{sync, true}])
+          end,
+          RowsIndexed)
+    catch
+        error:badmatch ->
+            {error, ldb_put_failed}
+    end.
+
+maybe_negate([], desc, nulls_first) ->
```
gordonguthrie: `mabye_negate` is what you are doing - why you are doing it is `make_sorted_keys`

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,172 @@
...
+        catch
+            {break, Acc} ->
+                {ok, Acc}
+        end,
+    case FoldRes of
+        {ok, Fetched} ->
+            Decoded =
+                [sext:decode(Row) || Row <- lists:reverse(Fetched)],
+            {ok, Decoded}
+        %% {error, Reason} ->
```
gordonguthrie: If it ain't happening, delete it...

src/riak_kv_qry_compiler.erl
```diff
@@ -43,35 +45,61 @@
...
+-type offset()           :: non_neg_integer().
 -type operator()         :: [binary()].
--type sorter()           :: term().
+-type sorter()           :: {binary(), asc|desc, nulls_first|nulls_last}.
 
 
 -export_type([
     combinator/0,
-    limit/0,
+    limit/0, offset/0,
```
gordonguthrie: go vertical or go horizontal

src/riak_kv_qry_compiler.erl
```diff
@@ -43,35 +45,61 @@
...
-compile(?DDL{}, ?SQL_SELECT{is_executable = true}, _MaxSubQueries) ->
+compile(?DDL{}, ?SQL_SELECT{is_executable = true}) ->
     {error, 'query is already compiled'};
-compile(?DDL{ table = T} = DDL,
-        ?SQL_SELECT{is_executable = false} = Q1, MaxSubQueries) ->
+compile(?DDL{table = T} = DDL,
+        ?SQL_SELECT{is_executable = false} = Q1) ->
     Mod = riak_ql_ddl:make_module_name(T),
-    case maybe_compile_group_by(Mod, compile_select_clause(DDL, Q1), Q1) of
+    case compile_order_by(
```
gordonguthrie, line 74: time to break up this set of nested mebbies into a single compile function that branches on matching heads

src/riak_kv_qry_compiler.erl
```diff
@@ -43,35 +45,61 @@
...
+    case non_unique_identifiers(OrderBy) of
+        [] ->
+            {ok, Q?SQL_SELECT{'ORDER BY' = OrderBy}};
+        WhichNonUnique ->
+            {error, {non_unique_orderby_fields, ?E_NON_UNIQUE_ORDERBY_FIELDS(hd(WhichNonUnique))}}
+    end.
+
+non_unique_identifiers(FF) ->
+    Occurses = [{F, occurs(F, FF)} || {F, _, _} <- FF],
+    [F || {F, N} <- Occurses, N > 1].
```
gordonguthrie: Hooray @hmmr tripped up on his English [does small jig in seat] - not sure 'occurses' is a word ;-)

Revenge for https://github.com/basho/riak_shell/commit/83d7ca28d132af1046964f738b17c3ae6c4ee9de

src/riak_kv_qry_compiler.erl
```diff
@@ -518,7 +543,7 @@ find_quantum_field_index_in_key2([_|Tail], Index) ->
     find_quantum_field_index_in_key2(Tail, Index+1).
 
 %%
-hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, MaxSubQueries, Where1) ->
+hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, Where1) ->
```
gordonguthrie, line 654: Need to test this and see if there is a max-max subqueries limit we should put in...

src/riak_kv_qry_compiler.erl
```diff
@@ -547,16 +572,11 @@ hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, MaxSubQueries, Where1) ->
...
-        MaxSubQueries == undefined orelse NoSubQueries =< MaxSubQueries ->
-            %% use the maximum value that has not been incremented, we still use
-            %% the end_inclusive flag because the end key is not used to hash
-            make_wheres(Where2, QField, Min2, Max1, Boundaries);
-        NoSubQueries > MaxSubQueries ->
-            {error, {too_many_subqueries, ?E_TOO_MANY_SUBQUERIES(NoSubQueries)}}
-    end.
+    %% use the maximum value that has not been incremented, we still use
+    %% the end_inclusive flag because the end key is not used to hash
+    make_wheres(Where2, QField, Min2, Max1, Boundaries).
```
gordonguthrie, line 697: There should be unit tests for queries with `ORDER BY` and `LIMIT` in here - we are not testing the `ORDER BY` compiler...

src/riak_kv_qry_queue.erl
```diff
@@ -67,18 +67,18 @@
 %%% API
 %%%===================================================================
 
--spec put_on_queue(pid(), [qry()], ?DDL{}) ->
+-spec put_on_queue(pid(), [qry()], ?DDL{}, riak_kv_qry_buffers:qbuf_ref() | undefined) ->
```
gordonguthrie, line 70: not that happy about using `undefined` to mean `no local query buffer`

src/riak_kv_qry_worker.erl
```diff
@@ -134,9 +152,9 @@ code_change(_OldVsn, State, _Extra) ->
 new_state() ->
     #state{ }.
 
-run_sub_qs_fn([]) ->
-    ok;
-run_sub_qs_fn([{{qry, ?SQL_SELECT{cover_context = undefined} = Q1}, {qid, QId}} | T]) ->
+prepare_fsm_options([], Acc) ->
```
gordonguthrie, line 154: Sub-queries are always of one type so this endless partial recalculation is a bit redundant, no?

src/riak_kv_select.erl
```diff
@@ -117,19 +134,36 @@ downgrade_select([v2,v1=To|Tail], Select1) ->
         cover_context = Select1#riak_select_v2.cover_context,
         local_key     = Select1#riak_select_v2.local_key
     },
+    downgrade_select([To|Tail], Select2);
+downgrade_select([v3,v2=To|Tail], Select1) ->
+    Select2 = #riak_select_v2{
+        'SELECT'      = Select1#riak_select_v3.'SELECT',
```
gordonguthrie: **Note** check if the record formats are moved around (they are aren't they) and is the upgrade/downgrade OK

src/riak_kv_qry_buffers.erl
```diff
@@ -0,0 +1,865 @@
...
+    [#riak_field_v1{name     = ColName,
+                    position = Pos,
+                    type     = ColReturnType,
+                    optional = false} || {Pos, ColName, ColReturnType} <- WithPositions].
+
+%% when Descending keys by @atill gets merged, we will be able to
+%% store OrderingDirection in ?SQL_PARAM where it belongs. Until then,
+%% we keep the ordering qualifiers in a makeshift #qbuf{} field outside DDL.
+make_lk_from_orderby(OrderBy) ->
+    #key_v1{ast = [?SQL_PARAM{name = [Field]} || {Field, _Dir, _NullsGroup} <- OrderBy]}.
```
gordonguthrie: `Dir` is a confusing abbreviation for `Direction` as I read it as `Directory`

src/riak_kv_qry.erl
```diff
@@ -281,22 +281,60 @@ count_to_position(Col, [_ | Rest], Pos) ->
...
+                              QBufRef,
+                              SQL?SQL_SELECT.'LIMIT', SQL?SQL_SELECT.'OFFSET')
+                    end
             end;
         {false, Errors} ->
             {error, {invalid_query, format_query_syntax_errors(Errors)}}
     end.
 
+
+-spec maybe_create_query_buffer(?SQL_SELECT{}, pos_integer(), ?DDL{}, proplists:proplist()) ->
```
gordonguthrie: the word `maybe` in a function name usually indicates a side-effect which may or may not happen - is it right here? is `undefined` really an appropriate return value from this function?

src/riak_kv_qry_buffers_ldb.erl
```diff
@@ -0,0 +1,173 @@
...
+    fetch_rows(LdbRef, 0, Limit);
+fetch_rows(LdbRef, Offset, LimitOrUndef) ->
+    %% Because the database is read-only, the plan is to keep a cache
+    %% of iterators for faster seeking (to the nearest stored),
+    %% ideally also enable eleveldb to do the folding from stored
+    %% iterators rather than from 'first'.
+    FetchLimitFn =
+        fun(_KV, {Off, Lim, Pos, Acc}) when Pos < Off ->
+                {Off, Lim, Pos + 1, Acc};
+           %% Fetching K (let alone V) while "seeking" to our Offset
```
gordonguthrie, line 133: I am not sure this is a correct description of the seek function in leveldb - we need to raise this with @matthewvon and seek his opinion before embarking on this

include/riak_kv_ts.hrl
```diff
@@ -54,8 +54,8 @@
           'SELECT'              :: #riak_sel_clause_v1{},
           'FROM'        = <<>>  :: binary() | {list, [binary()]} | {regex, list()},
           'WHERE'       = []    :: [riak_ql_ddl:filter()],
-          'ORDER BY'    = []    :: [riak_kv_qry_compiler:sorter()],
-          'LIMIT'       = []    :: [riak_kv_qry_compiler:limit()],
+          'ORDER BY'            :: undefined | [riak_kv_qry_compiler:sorter()],
+          'LIMIT'               :: undefined | riak_kv_qry_compiler:limit(),
```
andytill: We shouldn't change old records, because they still don't change on the old versioned nodes that create them.

include/riak_kv_ts.hrl
```diff
@@ -71,8 +71,8 @@
...
```
andytill: We shouldn't change old records, because they still don't change on the old versioned nodes that create them.

gordonguthrie (commented):
Lots of not very major comments about style and readabillity - couple of major things not covered.

First up (as raised seperately) the code can be made to blow up by querying a comedy amount of quanta.

Secondly we don't have a complete set of error conditions - which we will need for documentation, training and review of the riak_test suites.

src/riak_kv_qry_buffers.erl
angrycub: I do feel like there are usability benefits for a configurable path. A user might want to store this data on AWS instance storage. Configurable paths would prevent symlink hell for users that want to take advantage.  A riak.conf default of `${platform_data_dir}/query_buffers` would align nicely with AAE data path configurations (https://github.com/basho/riak_kv/blob/develop/priv/riak_kv.schema#L123-L128).   Might be good to get an additional CSE eyeball on this.

src/riak_kv_qry_compiler.erl
hmmr: **Abstract**

English influences in modern Erlang are subtle and not well understood.

**Introduction**

`Occurses` may not be correct English but it _is_ correct programmese.

**Discussion**

There is logic and consistency in calling `Poses` a list of positions, each denoted by variable `Pos` elsewhere in your code.

**Conclusions**

I beg to disagree.

include/riak_kv_ts.hrl
hmmr: The composition of the record did not change, only the default values did. Because 1.4 nodes don't inspect `'ORDER BY'` and `'LIMIT'`, I don't expect things will break because of this change. An upgrade/downgrade test is in the plans.

matthewvon:
What is scenario that requires "rm -rf"?  Was there a crash?  Did you try API DestroyDB first to validate Db is not in use?

hmmr:
@matthewvon After a crash, we certainly don't need stale temporary tables, because the queries for which they were created are gone. And, the application which does the clean-up at startup, is the only user of these tables, so we are certain no-one else will need them. Plain and brutal rm -rf is just appropriate here.

src/riak_kv_qry_buffers.erl
hmmr: Did it via `gen_server:cast` in 2ee4fb1.

src/riak_kv_qry_buffers.erl
hmmr: I have removed all unused, but exported, internal API functions, in d8b39a1.

include/riak_kv_ts.hrl
hmmr: I agree in general but I would rather leave the rename to @andytill. It will obviously require a sweeping change all over the place. The current PR's diff is already big enough.

include/riak_kv_ts.hrl
hmmr: Can you clarify on that? I see `ORDER BY` as a list of fields by which sorting is to be done, annotated with asc/desc and/or nulls first/last modifiers (`LIMIT` is a simple integer).

src/riak_kv_qry.erl
hmmr: Yes, `maybe_create_query_buffer` has the intended effect for eligible queries (and is a noop for others).

Re `undefined` for the `QBufRef` it returns, it's a way to avoid two extra indent levels. An alternative would be to have a case statement differentiating on `?SQL_SELECT{'ORDER BY' = DefinedOrNot}`, where `create_query_buffer` would do its work unconditionally. And, the `FullCycle` and downstream functions have been taught to handle `undefined` for `QBufRef`.

src/riak_kv_qry_buffers.erl
hmmr: Done in 783870a.

src/riak_kv_qry_buffers.erl
hmmr: I need it to be a binary, the same size as that generated by `make_qref` (if only for it to be of the same type from dialyzer's POV).

src/riak_kv_qry_buffers.erl
hmmr: Replaced with case in 2aec11b.

src/riak_kv_qry_buffers_ldb.erl
hmmr: We now have our very own `riak_kv_ts_util:rm_rf/1` (916c59e). For the mkdir case, there is `filelib:ensure_dir/1`.

src/riak_kv_qry_buffers.erl
hmmr: Macroised (20000d0). As for parameters, they are now configurable in riak_kv.schema (0c15cd0).

src/riak_kv_qry.erl
hmmr: Rewrote to make it a query-cancelling error (26648b15).

src/riak_kv_qry_buffers.erl
hmmr: It is now configurable in riak_kv.schema (26648b15).

src/riak_kv_qry_buffers.erl
hmmr: Now using `riak_kv_ts_util:rm_rf/1` and `filelib:ensure_dir/1` (916c59e).

src/riak_kv_qry_buffers.erl
hmmr: It's a choice between putting all code in `handle_call`s, or none. Given the size of `do_get_or_create_qbuf` and others, the first is clearly not a good option.

src/riak_kv_qry_buffers.erl
hmmr: Okay (7c81ca9).

src/riak_kv_qry_buffers.erl
hmmr: Done in ec4b352.

src/riak_kv_qry_compiler.erl
hmmr: The max subqueries check was reinstated in bfb275b5 (initially in 3fc2471a, it was removed); it is now done earlier in the call chain. In 0c15cd06, it is made configurable in riak_kv.schema.

src/riak_kv_qry_queue.erl
hmmr: But.. why?

src/riak_kv_qry_buffers.erl
hmmr: Renamed (b8ed326e).

src/riak_kv_qry_buffers_ldb.erl
hmmr: This setting sets the max amount of data to keep in memory before writing to disk happens.

src/riak_kv_qry_buffers_ldb.erl
hmmr: Renamed in 78747953.

src/riak_kv_qry_buffers_ldb.erl
hmmr: Ideally, we would like to keep `{Pos, Iter}` pairs in `#qbuf` records, to seek to OFFSET positions from the nearest cached pair. As I point out in the comment to this piece of code, there is no iterator support in leveldb (at any rate, not with the expected semantics). I have confirmed this sad fact with @erikleitch.

For @matthewvon (or rather, for @erikleitch) to consider implementing iterators that can be exposed via external leveldb API, I would like to have (1) `fold` to accept an iterator as an argument, to have it start folding from that position, and (2) a `seek(Iter, Distance)` function that would return an iterator pointing to a position some `Distance` (positive or negative) relative to `Iter`, and maybe (3) the `fold` function to also return the iterator where it has finished.

Of course it is painful to watch `eleveldb:fold` fetch objects, only to be dropped as it iterates from the beginning, every damn time it is called. I really don't see how it can be instructed not to do so -- but neither do I want query buffers to be delayed by our wishful thinking.

hmmr: Poring over https://github.com/basho/eleveldb/blob/riak_ts-develop/src/eleveldb.erl#L376L377, I start to wonder if I could cache `{Pos, Key}` pairs, where each `Key` would be extracted with `lists:last(RecordsForClient)`.. Still, it quacks like a kludge.

src/riak_kv_qry_buffers_ldb.erl
hmmr: It was my attempt, inexpressive as it is, to deal with the fact that `eleveldb:fold` does not return any `{error, Reason}`. Refactored a little in 9215fd2801.

src/riak_kv_qry_compiler.erl
hmmr: Added more tests in 2abb433.

src/riak_kv_qry_buffers_ldb.erl
matthewvon: @hmmr  API for manual manipulation of iterators within Erlang exists.  See riak_core/hashtree.erl for examples.  That code is already doing much of what you are requesting.

hmmr:
After a crash, we certainly don't need stale temporary tables, because
the queries for which they were created are gone. And, the application
which does the clean-up at startup, is the only user of these tables,
so we are certain no-one else will need them. Plain and brutal rm -rf
is just appropriate here.

src/riak_kv_qry_worker.erl
hmmr: Refactored in 178cbdf.

src/riak_kv_qry_buffers.erl
hmmr: Added a sleep of 1 msec to emulate `erlang:now/1` in 2ea39988.

src/riak_kv_select.erl
hmmr: Aaand it did cause trouble, was bound to. When a 1.4 node receives an assignment to spin up an index_fsm, it eventually trips on bad `?SQL_SELECT` field index (around here https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_qry_coverage_plan.erl#L80), leading to a cryptic crash with these.. words:

```
{function_clause,
 [{riak_ql_ddl, make_key,
   [{key_v1, [{param_v1, [<<"a">>]}, {hash_fn_v1, riak_ql_quanta, quantum, [{param_v1, [<<"b">>]}, 10, s], timestamp}]},
    true,
    [{<<"a">>, 1}, {<<"b">>, 1000}]],
   [{file, "src/riak_ql_ddl.erl"}, {line, 168}]},
  {riak_kv_qry_coverage_plan, make_key, 3, [{file, "src/riak_kv_qry_coverage_plan.erl"}, {line, 129}]},
  {riak_kv_qry_coverage_plan, do_plan, 7, [{file, "src/riak_kv_qry_coverage_plan.erl"}, {line, 83}]},
  {riak_kv_index_fsm, create_plan, 8, [{file, "src/riak_kv_index_fsm.erl"}, {line, 161}]},
  {riak_core_coverage_fsm, initialize, 2, [{file, "src/riak_core_coverage_fsm.erl"}, {line, 272}]},
  {gen_fsm, handle_msg, 7, [{file, "gen_fsm.erl"}, {line, 505}]}, {proc_lib, init_p_do_apply, 3, [{file, "proc_lib.erl"}, {line, 239}]}]}
```

A fix is to append, rather than insert, new fields to `#riak_select_v3` (a7eb41a),

src/riak_kv_qry_compiler.erl
gordonguthrie: I would be Occurances ;-)

include/riak_kv_ts.hrl
gordonguthrie: They will all be implemented as operations going forward:

``` [
     {order_by, [<<"col1">>, <<"col4">>]},
     {as,           [{<<"col1">>, <<"new fancy name">>}]},
     {limit,        [1000]}
     ],
```

In this world a `NoOp` is `[ ]`

src/riak_kv_qry_queue.erl
gordonguthrie: When you see `undefined` during debugging you don't know if it is an intended or unintended value...

src/riak_kv_select.erl
gordonguthrie: Sorry, this is not acceptable - the system cannot work like this where the position of a key in a record at compile time changes the behaviour...

gordonguthrie:
+1 4ee8edd

hmmr:
@borshop retry

hmmr:
@borshop: retry

hmmr:
@borshop merge

I have manually initiated a rebuild, and it's green: http://buildbot.bos1/builders/test-riak_kv/builds/4250.

hmmr:
@borshop: merge

hmmr:
@borshop: retry 4ee8edd

andytill:
+1 0658863

hmmr:
@borshop merge

hmmr:
@borshop merge

hmmr:
@borshop merge

hmmr:
@borshop: merge

PR #1481, opened 2016-09-16
Add force update flag and rename `undefined` version to `legacy`

bsparrow435:
Rename AAE hashtree version from `undefined` to `legacy`.
Add public ETS table to handle partition version requests

### DIFF
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -41,6 +41,7 @@
          expire_trees/0,
          clear_trees/0,
          get_version/0,
+         get_partition_version/1,
          get_pending_version/0,
          get_trees_version/0]).
 -export([all_pairwise_exchanges/2]).
@@ -219,6 +220,20 @@ clear_trees() ->
 -spec get_version() -> version().
 get_version() ->
     gen_server:call(?MODULE, get_version, infinity).
+
+-spec get_partition_version(index()) -> version().
+get_partition_version(Index) ->
+    try
+        case ets:lookup(?ETS, Index) of
+            [] ->
+                legacy;
+            [{Index, Version}] ->
+                Version
+        end
+    catch
+        _:_ ->
+            legacy
+    end.
 
 -spec get_pending_version() -> version().
 get_pending_version() ->
@@ -482,8 +497,10 @@ add_hashtree_pid(true, Index, Pid, State=#state{trees=Trees, trees_version=VTree
             State;
         _ ->
             monitor(process, Pid),
+            Version = riak_kv_index_hashtree:get_version(Pid),
             Trees2 = orddict:store(Index, Pid, Trees),
-            VTrees2 = orddict:store(Index, riak_kv_index_hashtree:get_version(Pid), VTrees),
+            VTrees2 = orddict:store(Index, Version, VTrees),
+            ets:insert(?ETS, {Index, Version}),
             State2 = State#state{trees=Trees2, trees_version=VTrees2},
             State3 = add_index_exchanges(Index, State2),
             State4 = check_upgrade(State3),
@@ -550,6 +567,7 @@ maybe_clear_registered_tree(Pid, State) when is_pid(Pid) ->
         false ->
             State;
         {value, {Index, Pid}, Trees} ->
+            ets:delete(?ETS, Index),
             VTrees = orddict:erase(Index, State#state.trees_version),
             State#state{trees=Trees, trees_version=VTrees}
     end;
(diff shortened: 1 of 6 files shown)


src/riak_kv_entropy_manager.erl
```diff
@@ -219,6 +221,20 @@ clear_trees() ->
...
+-spec get_partition_version(index()) -> version().
+get_partition_version(Index) ->
+    try
+        case ets:lookup(?ETS, Index) of
+            [] ->
+                legacy;
+            [{Index, Version}] ->
+                Version
+        end
+    catch
```
JeetKunDoug: Why the try/catch here? In case you somehow end up with multiple rows in the ETS table? ets:lookup docs seem to indicate that it should never throw, and given the table is a set (can only have one entry per key) it'd never actually return more than one row. Was there something else you were defending against here?

src/riak_kv_entropy_manager.erl
```diff
@@ -644,14 +664,52 @@ maybe_upgrade(State=#state{trees_version = VTrees}) ->
...
+                    State#state{pending_version=0};
+                _ ->
+                    State
+            end;
         _ ->
             State
     end.
 
+-spec check_remote_exchanges(list(), list()) -> complete | incomplete.
+check_remote_exchanges([], Acc) ->
```
JeetKunDoug: This looks like a lists:all call in disguise - if you take the check_remote_exchanges logic and split it into a few functions you end up with something like:

``` erlang

check_all_remote_exchanges_complete(Nodes) ->
    lists:all(fun maybe_check_and_record_remote_exchange/1, Nodes).

maybe_check_and_record_remote_exchange(Node) ->
    case ets:lookup(?ETS, Node) of
         [{Node, true}] ->
             true;
         _ ->
             Result = check_remote_exchange(Node),
             ets:insert(?ETS, {Node, Result}),
             Result
     end.
```

This has a few advantages:
- Readability - it's clear what you're trying to accomplish.
- Bails on the first "false" answer
- Doesn't have to build/sort a list of results.
- We're trying to encourage the use of functions like lists:all, lists:map, etc. when they make sense, versus hand-rolling recursive functions on our own (or even using a foldl when you really mean map). This generally helps code maintainability.

src/riak_kv_index_hashtree.erl
```diff
@@ -507,24 +507,29 @@ determine_data_root() ->
...
-                    case filelib:is_dir(filename:join(filename:join(Root, "v" ++ integer_to_list(V)),integer_to_list(Index))) of
-                        true ->
-                            V;
-                        false ->
-                            undefined
-                    end;
-                _ ->
-                    undefined
-            end
+    case application:get_env(riak_kv, force_hashtree_upgrade) of
```
JeetKunDoug: Please try to break up this 4-level deep case statement into something more manageable.

Take a look at http://www.gar1t.com/blog/solving-embarrassingly-obvious-problems-in-erlang.html for some suggestions as to how you might go about it. I realize there's tons of this in our codebase, just trying to make sure we don't add more of them.

JeetKunDoug (changes requested):
Generally looks good. Please see comments about some cleanup/refactoring I think would be useful to future-us as we need to maintain this code.

src/riak_kv_entropy_manager.erl
bsparrow435: In looking around at how this was implemented in other places, I came across the riak_core_capability:get function which uses this try catch. I also looked at the docs and didn't understand why it was there and also looked in the commit history for capabilties and didn't find anything. Unssure of if I was missing something, I just added the try catch because there was no downside from a logic perspective. With that said, I agree and will remove the try/catch as I havent found a single reason why it's there.

src/riak_kv_entropy_manager.erl
bsparrow435: This is a much better way thank you

JeetKunDoug: Yea - us functional programmers always seem to start with "let's write a recursive list processing function" rather than looking at other options first :)

src/riak_kv_index_hashtree.erl
bsparrow435: Changed to:

```
%% @doc Determine the version this tree should use at startup. If update atom is in Opts
%%      then we immediately use version in capabilities. Otherwise check if capabilities
%%      has flipped to a version yet and if it has, check if we've already upgraded by
%%      looking for versioned AAE directory.
-spec determine_version(list(), index(), list()) -> version().
determine_version(Root, Index, Opts) ->
    case force_upgrade(Opts) of
        true ->
            get_cap_hash_version();
        _ ->
            find_version(Root, Index)
    end.

-spec force_upgrade(list()) -> boolean().
force_upgrade(Opts) ->
    check_upgrade_opts(Opts, check_upgrade_env()).

-spec check_upgrade_opts(lists(), boolean()) -> boolean().
check_upgrade_opts(_Opts, true) ->
    true;
check_upgrade_opts(Opts, _) ->
    lists:member(upgrade, Opts).

-spec check_upgrade_env() -> boolean().
check_upgrade_env() ->
    case application:get_env(riak_kv, force_hashtree_upgrade) of
        {ok, true} ->
            true;
        _ ->
            false
    end.

-spec get_cap_hash_version() -> version().
get_cap_hash_version() ->
    riak_core_capability:get({riak_kv, object_hash_version}, legacy).

-spec find_version(list(), index()) -> version().
find_version(Root, Index) ->
    check_root_version(Root, Index, get_cap_hash_version()).

-spec check_root_version(list(), index(), version()) -> version().
check_root_version(Root, Index, Version) when is_integer(Version) ->
    case filelib:is_dir(filename:join(filename:join(Root, "v" ++ integer_to_list(V)),integer_to_list(Index))) of
        true ->
            Version;
        false ->
            legacy
    end;
check_root_version(_Root, _Index, Version) ->
    Version.
```

src/riak_kv_index_hashtree.erl
```diff
@@ -507,31 +507,53 @@ determine_data_root() ->
...
+
+-spec check_upgrade_opts(list(), boolean()) -> boolean().
+check_upgrade_opts(_Opts, true) ->
+    true;
+check_upgrade_opts(Opts, _) ->
+    lists:member(upgrade, Opts).
+
+-spec check_upgrade_env() -> boolean().
+check_upgrade_env() ->
     case application:get_env(riak_kv, force_hashtree_upgrade) of
```
JeetKunDoug: Sorry - one more. How about using the 3-arity version of get_env and providing false as the default, rather than having to deal with the {ok, true} - can just match on true and _ at that point (in case someone put something stupid other than true in there).

JeetKunDoug (approved):

PR #1484, opened 2016-09-21
Add object lager sink

bsparrow435:
Also add a version entry into riak_kv_entropy_manager ets table so we
don't need to serialize get_version requests on the managers msgq.

### DIFF
--- a/src/riak_kv_get_core.erl
+++ b/src/riak_kv_get_core.erl
@@ -310,7 +310,7 @@ maybe_log_old_vclock(Results) ->
                         [] ->
                             ok;
                         _ ->
-                            lager:warning("Bucket: ~p Key: ~p should be rewritten to guarantee
+                            object:warning("Bucket: ~p Key: ~p should be rewritten to guarantee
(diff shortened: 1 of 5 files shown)


src/riak_kv_get_core.erl
```diff
@@ -310,7 +310,7 @@ maybe_log_old_vclock(Results) ->
                         [] ->
                             ok;
                         _ ->
-                            lager:warning("Bucket: ~p Key: ~p should be rewritten to guarantee
+                            object:warning("Bucket: ~p Key: ~p should be rewritten to guarantee
```
hazen, line 313: Dialyzer doesn't seem too happy with this.  Is this change intentional?

bsparrow435: Yes, lager3 will do a parse transform on object: to route to the correct sink. Purpose of the change in rebar.config: {lager_extra_sinks, [object]}

JeetKunDoug:
+1 05866ee

hazen:
@borshop merge

PR #1485, opened 2016-09-26
Fix issue with mixed version

bsparrow435:
Specifically add support for aae exchanges if the upgraded nodes still
have legacy trees.

### DIFF
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -146,6 +146,10 @@ start_exchange_remote(VNode, IndexN, FsmPid) ->
                               {remote_exchange, not_built} |
                               {remote_exchange, already_locked} |
                               {remote_exchange, bad_version}.
+start_exchange_remote(_VNode={Index, Node}, IndexN, FsmPid, legacy) ->
+    gen_server:call({?MODULE, Node},
+                    {start_exchange_remote, FsmPid, Index, IndexN},
+                    infinity);                        
 start_exchange_remote(_VNode={Index, Node}, IndexN, FsmPid, Version) ->
     gen_server:call({?MODULE, Node},
                     {start_exchange_remote, FsmPid, Index, IndexN, Version},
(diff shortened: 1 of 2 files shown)


src/riak_kv_entropy_manager.erl
```diff
@@ -330,24 +334,13 @@ handle_call(get_trees_version, _From, State=#state{trees=Trees}) ->
 handle_call({get_lock, Type, Pid}, _From, State) ->
     {Reply, State2} = do_get_lock(Type, Pid, State),
     {reply, Reply, State2};
+%% To support compatibility with pre 2.2 nodes.
+handle_call({start_exchange_remote, FsmPid, Index, IndexN}, From, State) ->
+    {Type, Reply, State2} = do_start_exchange_remote(FsmPid, Index, IndexN, legacy),
```
ian-mi: Why unpack this tuple? Is there a safety related motivation for this?

src/riak_kv_index_hashtree.erl
```diff
@@ -141,10 +142,12 @@ async_delete(Items, Tree) when Tree =:= undefined; Items =:= [] ->
 async_delete(Items=[{_Id, _Key}|_], Tree) ->
     catch gen_server:cast(Tree, {delete, Items}).
 
-%
 %% @doc Called by the entropy manager to finish the process used to acquire
 %%      remote vnode locks when starting an exchange. For more details,
 %%      see {@link riak_kv_entropy_manager:start_exchange_remote/3}
+-spec start_exchange_remote(pid(), term(), index_n(), pid()) -> ok.
```
ian-mi: Is this needed? Won't the entropy manager always call the new version?

ian-mi (changes requested):
Please squash these commits after addressing comments.

src/riak_kv_index_hashtree.erl
bsparrow435: Yes you're correct. THis should always be called by the local entropy_manager and never by a remote one so we can just have the version with a Version spec.

src/riak_kv_entropy_manager.erl
bsparrow435: No safety concern, was just for clarity and is not necessary.

bsparrow435:
Squashing and merging.

PR #1487, opened 2016-09-27
Don't pattern match results for `do_start_exchange_remote`

JeetKunDoug:
It sometimes returns `{noreply, State}` 2-tuple.
- Also fix spec for do_start_exchange to match reality.


fadushin:
+1 6b1ae04

JeetKunDoug:
@borshop merge

PR #1488, opened 2016-09-29
Add logic to skip upgrading a tree if already upgraded trees are not yet built

bsparrow435:
This fixes an issue where, if a user gets an unlucky queue, all trees will upgrade before doing any builds. This leads to unavailable AAE for a period of time.


src/riak_kv_entropy_manager.erl
```diff
@@ -568,6 +568,32 @@ do_get_build_token(State=#state{build_tokens=Tokens}) ->
             build_limit_reached
     end.
 
+-spec maybe_get_build_token(state()) -> build_limit_reached | {ok, state()}.
+maybe_get_build_token(State=#state{trees_version=VTrees}) ->
+    case [Idx || {Idx, Version} <- VTrees, Version /= legacy] of
+        %% All legacy trees, allow build token
+        [] ->
```
ian-mi: This case doesn't appear to be necessary.

bsparrow435:
Once tests finish should be ready for merge

PR #1495, opened 2016-10-06
Always drop dots from CRDTs, not just on merge [JIRA: RIAK-2987]

nickelization:
Prior to this change, the first write of a new CRDT would end up with a dot in its metadata, since dots were only being dropped on merge. This could potentially lead to sibling CRDTs on disk, since our merge code assumes CRDTs will never have dots.

Although we should never have sibling values for CRDTs present on disk, this bug did not normally cause problems, because using the `fetch_type` client functions would still cause the values to be merged prior to returning them to the client. However, doing a basic `get` could cause the siblings to be exposed to the client directly.

This was the issue that has been causing the `test_hll` test to intermittently fail. The issue can be consistently reproduced using the new `verify_no_datatype_siblings` test, and applying this fix causes that test to pass.

Adding the new call to `drop_the_dot` is needed in order to fix this issue, but removing the old makes no difference for new data, and would be more efficient. However, if we remove the old call then we will run into problems with existing CRDTs on disk that already have dots present, so I'm leaving the old call in place to deal with any such scenarios.

### DIFF
--- a/src/riak_kv_crdt.erl
+++ b/src/riak_kv_crdt.erl
@@ -398,7 +398,7 @@ meta(undefined, ?CRDT{ctype=CType}) ->
     M3 = dict:store(?MD_VTAG, riak_kv_util:make_vtag(Now), M2),
     dict:store(?MD_CTYPE, CType, M3);
 meta(Meta, _CRDT) ->
-    Meta.
+    drop_the_dot(Meta).
 
 %% Just a simple take the largest for meta values based on last mod
 merge_meta(CType, Meta1, Meta2) ->


russelldb:
@nickelization I don't understand your last paragraph "removing the old makes no difference" etc. Removing the old what? And what is "more efficient"?

bsparrow435:
Test fails without this change, passes with it.

nickelization:
@russelldb "removing the old" was supposed to mean "removing the existing call to `drop_the_dot`". There used to just be one place that function was called, but now in this PR we're adding a new call without removing the old call. In normal operation, the old invocation of `drop_the_dot` is made redundant by the new invocation, so there's a minor efficiency hit that we're taking by calling that function twice even though don't need to do that in most cases.

russelldb:
@nickelization are you saying `drop_the_dot` gets called twice now?

nickelization:
@russelldb Correct, this PR adds a new invocation in `meta`, but does not remove the old invocation from `merge_meta`.

russelldb:
@nickelization but merge_meta is only called if there is an existing object, I thought?

russelldb:
Ah ok. I get it now. `meta` and `merge_meta` are called. OK.

russelldb:
+1 from me

bashopatricia:
create jira issue

Basho-JIRA:
[~fdushin] We have this on the list for 2.0.8 and Brian put your name next to it for estimating.  Can you estimate this one?  Thank you! 

***[posted via JIRA by Patricia Brewer]***

Basho-JIRA:
PR for 2.0.8: https://github.com/basho/riak_kv/pull/1575 

PR #1498, opened 2016-10-10
Several `riak_kv_index_hashtree` fixes (performance and functional) [JIRA: RIAK-2813]

JeetKunDoug:
## Don't update trees before destroying them (when calling `clear_trees`).

Allow the case where we're about to destroy the hashtree to simply close,
without going through a (sometimes very lengthy) update process first. Only
used from `clear_trees`. This condition was caught by a 4-day performance
test, and is not generally reproducible without having tons of data, on
the order of 10s of millions of key/value pairs per partition.
## Handle errors in vnode fold request.

`fold_fun`'s call to `riak_core_master:sync_command` ignored the result
of the call. It is possible for `sync_command` to return {error, overload}
if the vnode is overloaded. Rather than ignore the result, make sure to
handle non-success results by failing the tree build.
## Allow build throttle to be changed while a build is ongoing.

Before, the build limits (size and wait time) could only be changed
before a build started. Once it did, you were stuck with those settings.
If the vnode was large, you may want to decrease the wait or increase
the size, but we wouldn't pick it up until the rebuild was over.

### DIFF
--- a/src/riak_kv_index_hashtree.erl
+++ b/src/riak_kv_index_hashtree.erl
@@ -612,8 +612,9 @@ hash_index_data(IndexData) when is_list(IndexData) ->
 -spec fold_keys(index(), pid(), partition(), boolean()) -> ok.
 fold_keys(Partition, HashtreePid, Index, HasIndexTree) ->
     FoldFun = fold_fun(HashtreePid, HasIndexTree),
+    {Limit, Wait} = get_build_throttle(),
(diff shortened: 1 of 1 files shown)


lehoff:
+1 dfe369c

bashopatricia:
create jira issue

JeetKunDoug:
@borshop merge

Basho-JIRA:
2.0.8 PR:

https://github.com/basho/riak_kv/pull/1574 

PR #1500, opened 2016-10-13
Descending Keys and DDL record upgrades in mixed version clusters

andytill:
Descending Keys originally reviewed in https://github.com/basho/riak_kv/pull/1401 but has been blocked because it requires DDL record upgrade/downgrade support in mixed version clusters.

RFC:

https://github.com/basho/rfc/pull/35

Dependent PRs:
- https://github.com/basho/riak_ql/pull/117
- https://github.com/basho/riak_test/pull/1055

The "descending keys" part of this PR was reviewed in https://github.com/basho/riak_kv/pull/1401


src/riak_kv_ts_svc.erl:160: The call riak_kv_qry:submi
```diff
@@ -0,0 +1 @@
...| #riak_sql_explain_query_v1{'EXPLAIN'::#riak_select_v2{'SELECT'::'undefined' | #riak_sel_clause_v1{calc_type::'aggregate' | 'group_by' | 'rows',initial_state::[any()],col_return_types::[any()],col_names::[any()],clause::[any()],finalisers::[any()]},'FROM'::binary() | {'list',[any()]} | {'regex',[any()]},'WHERE'::[any()],'ORDER BY'::[any()],'LIMIT'::[any()],helper_mod::atom(),partition_key::'none' | #key_v1{ast::[any()]},is_executable::boolean(),type::'sql' | 'timeseries',group_by::[{_,_}]}} | #riak_sql_insert_v1{'INSERT'::binary(),fields::[{'identifier',[any()]}],values::'undefined' | [[{_,_}]],helper_mod::atom()} | #riak_select_v2{'SELECT'::#riak_sel_clause_v1{calc_type::'aggregate' | 'group_by' | 'rows',initial_state::[any()],col_return_types::['boolean' | 'double' | 'sint64' | 'timestamp' | 'varchar'],col_names::[binary()],clause::[fun((_,_) -> any())],finalisers::['skip' | fun()]},'FROM'::binary() | {'list',[binary()]} | {'regex',[any()]},'WHERE'::[any()],'ORDER BY'::[any()],'LIMIT'::[any()],helper_mod::atom(),partition_key::'none' | #key_v1{ast::[{_,_,_} | {_,_,_,_,_}]},is_executable::boolean(),type::'sql' | 'timeseries',group_by::[{'identifier' | integer(),binary()}]},any())
```
gordonguthrie: This is a commit fart...

src/riak_kv_compile_tab.erl
```diff
@@ -25,155 +25,170 @@
...
-            State;
-        [] ->
-            notfound
+-spec insert(BucketType :: binary(), DDL :: term()) -> ok.
+insert(BucketType, DDL) when is_binary(BucketType), is_tuple(DDL) ->
+    lager:info("DDL DETS Update: ~p, ~p", [BucketType, DDL]),
+    DDLVersion = riak_ql_ddl:ddl_record_version(element(1, DDL)),
+    Row = #row_v3{
+         table = BucketType
+        ,table_version = {BucketType, DDLVersion}
```
gordonguthrie, line 101: [must not burn down the bikeshed]
[must not burn down the bikeshed]
[must not burn down the bikeshed]
[blinks furiously]

src/riak_kv_compile_tab.erl
```diff
@@ -25,155 +25,170 @@
 -export([
          delete_dets/1,
```
gordonguthrie, line 26: Break testing functions out into their own `-export[]` set and mark with a comment - rename them to have a name like `delete_dets_TEST` to make their utility clear

This will stop review ping-pong where the next team to work on this identify them as 'unused' fns and refactor them away - it also give the next developer a clue to the appropriate workflow for making changes to the file

If some of them are 'internal' fns that you export for development workflow they get their own `-export[]` with comment but retain their plain names - so the next developer knows they are not be used directly.

src/riak_kv_compile_tab.erl
```diff
@@ -25,155 +25,170 @@
...
+insert(BucketType, DDL) when is_binary(BucketType), is_tuple(DDL) ->
+    lager:info("DDL DETS Update: ~p, ~p", [BucketType, DDL]),
+    DDLVersion = riak_ql_ddl:ddl_record_version(element(1, DDL)),
+    Row = #row_v3{
+         table = BucketType
+        ,table_version = {BucketType, DDLVersion}
+        ,ddl = DDL},
+    ok = dets:insert(?TABLE3, Row),
+    ok = dets:sync(?TABLE3),
+    case lists:keyfind(ddl_v1, 1, riak_ql_ddl:convert(v1, DDL)) of
```
gordonguthrie, line 105: So there are table definitions that cannot be safely downgraded and in this case they are not being written to the downgraded table and are not available after a node downgrade?

src/riak_kv_compile_tab.erl
```diff
@@ -25,155 +25,170 @@
...
--spec get_state(BucketType :: binary()) ->
-        compiling_state() | notfound.
-get_state(BucketType) when is_binary(BucketType) ->
-    case dets:lookup(?TABLE, BucketType) of
-        [{_,_,_,_,State}] ->
-            State;
-        [] ->
-            notfound
+-spec insert(BucketType :: binary(), DDL :: term()) -> ok.
+insert(BucketType, DDL) when is_binary(BucketType), is_tuple(DDL) ->
```
gordonguthrie, line 96: Let us consider this scenario:
- a user upgrades to a new version of riak_ts - and inserts a table that is not downgradable
  - `mytable` is in the v3 `dets` table
  - `mytable` is _NOT_ in the v2 `dets` table

The user writes some data to `mytable`

The node is downgrade - `mytable` is not available - therefore the user scratches their head and recreates `mytable` (with a different set of fields). Does that work or do we reject that table creation?

The user writes some more data to `mytable`  - when they read `mytable` they only see data written since the table was re-created?

The node is re-updated - what do they see now? The V2 definition is read, converted and overwrites the V3 one? But there is data in the table under the old V3 form...

src/riak_kv_qry_compiler.erl
```diff
@@ -558,6 +650,8 @@ hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, MaxSubQueries, Where1) ->
             {error, {too_many_subqueries, ?E_TOO_MANY_SUBQUERIES(NoSubQueries)}}
     end.
 
+make_wheres(Where, QField, Min, Max, Boundaries) when Min > Max ->
```
gordonguthrie: `Min` and `Max` are not the right variable names here, `Bound1` and `Bound2` mebbies...

src/riak_kv_qry_coverage_plan.erl
```diff
@@ -125,7 +125,8 @@ drop_nodes(Vnodes, DownNodes) ->
 %% planner which should only know about a more logical format
 make_key(Mod, PartitionKey, Where) ->
     {startkey, StartKey} = proplists:lookup(startkey, Where),
-    StartKey2 = [{Field, Val} || {Field, _Type, Val} <- StartKey],
+    {endkey, EndKey} = proplists:lookup(endkey, Where),
```
gordonguthrie, line 128: This is because the start and end keys might be swapped because of`ASC` and `DESC` - probably needs a comment - makes sense in the context of an `ASC/DESC` PR but not otherwise

src/riak_kv_ts_newtype.erl
```diff
@@ -111,101 +113,133 @@ code_change(_OldVsn, State, _Extra) ->
...
+                        throw:unknown_ddl_version ->
+                            log_unknown_ddl_version(Table, DDL),
+                            ok
+                    end;
+                {ok, DDL} ->
+                    log_new_type_is_duplicate(Table);
+                {ok, _StoredDDL} ->
+                    %% there is a different DDL for the same table/version
+                    %% FIXME recompile anyway? That is the current behaviour
+                    %% this cannot happen because of the is_compiled check
```
gordonguthrie: hmm, should we have this clause? Seems smelly...

src/riak_kv_ts_newtype.erl
```diff
@@ -111,101 +113,133 @@ code_change(_OldVsn, State, _Extra) ->
...
-                [NewDDL, BucketType, ?DDL_RECORD_NAME, OldVsn, NewVsn]),
-    %% We don't know what to do with this new DDL, so stop
+do_new_type(Table) ->
+    case retrieve_ddl_from_metadata(Table) of
+        undefined ->
+            %% this bucket type name does not exist in the metadata!
+            log_missing_ddl_metadata(Table);
+        DDL ->
+            CurrentVersion = riak_ql_ddl:current_version(),
+            case riak_kv_compile_tab:get_ddl(Table, CurrentVersion) of
```
gordonguthrie, line 126: So in this case the DDL has been written to the meta data but for some reason it hasn't been compiled (node crashed or whatever) so we 'carry on'?

src/riak_kv_ts_newtype.erl
```diff
@@ -239,13 +272,111 @@ add_ddl_ebin_to_path() ->
...
+        true ->
+            lager:info("beam file for table ~ts exists", [Table]),
+            ok;
+        false ->
+            lager:info("beam file for table ~ts must be recompiled", [Table]),
+            actually_compile_ddl(Table)
+    end.
+
+%%
+beam_exists(Table) when is_binary(Table) ->
```
gordonguthrie, line 375: We would have a running system where some nugget deletes the `beam`s from disk but they are still in memory... What happens then?

gordonguthrie (commented):
I haven't successfully run all the tests or done enough code/running poking about, but this should be enough to get started

src/riak_kv_compile_tab.erl
andytill: I didn't use it tbh. Happy to delete it.

src/riak_kv_compile_tab.erl
andytill: Yes, but the DDL record capabilitiy will also be downgraded so none of the nodes will be able to use that table.

src/riak_kv_ts_newtype.erl
andytill: What is in the metadata might not be a DDL record type we can deal with, so prefer the one in the dets table if it exists which means the node has been downgraded. If it is not in dets then it is a truly new table.  It is safe to compile a new beam and load it as many teams as is necessary.

src/riak_kv_ts_newtype.erl
andytill: We would compile a new beam file and put it onto disk, but it wouldn't be loaded because the table helper modules are loaded naturally by the beam vm, the existing in-memory module would execute the function calls.

Since the module uses the compilers `vsn` attribute the compiler would almost certainly be exactly the same compiler that compiled the beam file that the nugget deleted.

src/riak_kv_ts_svc.erl:160: The call riak_kv_qry:submi
andytill: Deleted in b30d1a455693842756231ee88ca9f696737c387a.

src/riak_kv_qry_compiler.erl
andytill: Renamed in 7ddb56c53edf31e54b7bb4878d616534e48ab137.

src/riak_kv_qry_coverage_plan.erl
andytill: Commented in 89413be4ad76819228569425b0c86a3504a11f30

src/riak_kv_ts_newtype.erl
andytill: It should be recompiled anyway, because the table is allowed to be changed before it is activated. I will change.

andytill: This conditional branch now deletes the old DDL records and compiles the new ones, in commit 80912d4ed0e9431ed40999ed0da8c058da5ec29c.

There is a corresponding rt as well.

src/riak_kv_qry_compiler.erl
```diff
@@ -650,12 +650,12 @@ hash_timestamp_to_quanta(QField, QSize, QUnit, QIndex, MaxSubQueries, Where1) ->
             {error, {too_many_subqueries, ?E_TOO_MANY_SUBQUERIES(NoSubQueries)}}
     end.
 
-make_wheres(Where, QField, Min, Max, Boundaries) when Min > Max ->
-    make_wheres(Where, QField, Max, Min, Boundaries);
-make_wheres(Where, QField, Min, Max, Boundaries) ->
+make_wheres(Where, QField, LowerBound, UpperBound, Boundaries) when LowerBound > UpperBound ->
```
gordonguthrie, line 653: The problem here is that `LowerBound > UpperBound` looks smelly - when it is an artifact of `ASC/DESC` so a reader coming to this will think its an error - mebbies the answer isn't variable changes but a comment?

gordonguthrie:
thumbot retry

gordonguthrie:
Can't get the staff, triggering a retry while the remedial PR was unmerged, who hired this @gordonguthrie clown anyhoo?

gordonguthrie:
thumbot retry

gordonguthrie:
Third time lucky becuz I am the muppet of all muppets...

gordonguthrie:
thumbot retry

gordonguthrie:
The one where thumbot is not retrying...

gordonguthrie:
+1

gordonguthrie:
+1 b6630a3

hazen:
@borshop merge

PR #1506, opened 2016-10-19
RIAK-2949: Refactor vnode request records into requests module with factories and getters

paulhenrich:
Just opening this up so thumbs can tell us when we mess something up.

Jira: https://bashoeng.atlassian.net/browse/RIAK-2949


hazen:
@paulhenrich Feel free to look at the `kv_ts_merge` branch to see how this might play out in a KV/TS world

src/riak_kv_vnode.erl
```diff
@@ -934,36 +958,15 @@ handle_coverage(?KV_LISTBUCKETS_REQ{item_filter=ItemFilter},
...
-                              qry=Query},
-                FilterVNodes, Sender, State) ->
-    %% v2 = ack-based backpressure
-    handle_coverage_index(Bucket, ItemFilter, Query,
-                          FilterVNodes, Sender, State, fun result_fun_ack/2).
+handle_coverage_request(kv_index_request, Req, FilterVNodes, Sender, State) ->
+    Bucket = riak_kv_requests:get_bucket(Req),
+    ItemFilter = riak_kv_requests:get_item_filter(Req),
+    Query = riak_kv_requests:get_query(Req),
+    ResultFun = case riak_kv_requests:get_ack_backpressure(Req) of
```
lehoff, line 949: This is not consistent with the way it is done for `kv_listkeys_request` where the result fun is applied to the `Bucket` and `Sender` on the spot before calling the next function.

Since we have both `Bucket` and `Sender` in our hands right now we should apply them immediately.
And once the usage has been unified we should introduce a `result_fun_for_request(Req, Bucket, Sender)` function that returns the right result fun.

src/riak_kv_vnode.erl
```diff
@@ -878,41 +839,104 @@ handle_command({get_index_entries, Opts},
...
+                        _FilterVNodes,
+                        Sender,
+                        State=#state{async_folding=AsyncFolding,
+                                     bucket_buf_size=BufferSize,
+                                     mod=Mod,
+                                     modstate=ModState}) ->
+    ItemFilter = riak_kv_requests:get_item_filter(Req),
     %% Construct the filter function
     Filter = riak_kv_coverage_filter:build_filter(ItemFilter),
     BufferMod = riak_kv_fold_buffer,
```
lehoff, line 932:
```
BufferMod = riak_kv_fold_buffer,

Buffer = BufferMod:new(BufferSize, result_fun(Sender)),
FoldFun = fold_fun(buckets, BufferMod, Filter, undefined),
FinishFun = finish_fun(BufferMod, Sender),
```

while the use in the FoldFun and FinishFun is fine., it is problematic in the BufferMod:new/3 since it stops xref from spotting the call. 
Please replace that one with the actual atom.

lehoff, line 932: Further down we have

```
case AsyncFolding andalso AsyncBackend of
         true ->
             Opts = [async_fold];
         false ->
             Opts = []
     end,
```

At the very least we shouldn't export `Opts` from the case statement, but even better would be to move the case statement into a function, say, `options_for_folding_and_backend/2`

src/riak_kv_vnode.erl
```diff
@@ -934,36 +958,15 @@ handle_coverage(?KV_LISTBUCKETS_REQ{item_filter=ItemFilter},
...
+    ItemFilter = riak_kv_requests:get_item_filter(Req),
+    Query = riak_kv_requests:get_query(Req),
+    ResultFun = case riak_kv_requests:get_ack_backpressure(Req) of
+                    true -> fun result_fun_ack/2;
+                    false -> fun result_fun/2
+                end,
+    handle_coverage_index(Bucket, ItemFilter, Query, FilterVNodes, Sender, State, ResultFun).
 
 -spec prepare_index_query(?KV_INDEX_Q{}) -> ?KV_INDEX_Q{}.
 prepare_index_query(#riak_kv_index_v3{term_regex=RE} = Q) when
```
lehoff: On line 998 remove the useless `%% Update stats...` comment.

lehoff: Further down we have:

```
 %% @HACK
             %% Really this should be decided in the backend
             %% if there was a index_query fun.
             FoldType = case riak_index:return_body(Query) of
                            true -> fold_objects;
                            false -> fold_keys
                        end,
```

This should be put into a function `fold_type_for_query/1`

lehoff (changes requested):
Just minor clean-ups laying around for grabs, when done it can be merged.

include/riak_kv_vnode.hrl
```diff
@@ -1,83 +1,20 @@
...
-          options :: list()}).
-
--record(riak_kv_w1c_put_req_v1, {
-    bkey :: {binary(),binary()},
-    encoded_obj :: binary(),
-    type :: primary | fallback
-    % start_time :: non_neg_integer(), Jon to add?
-}).
 
 -record(riak_kv_w1c_put_reply_v1, {
```
hazen, line 4: Is this record not being refactored because of the same reason as `riak_kv_listkeys_req_v2`?

src/riak_kv_vnode.erl
```diff
@@ -1075,32 +1081,27 @@ handle_handoff_command(Req=?KV_PUT_REQ{}, Sender, State) ->
...
-                        object=Object,
-                        req_id=ReqId,
-                        start_time=StartTime,
-                        options=Options} = Req,
+            #state{idx = Idx} = State,
+            ReqId = riak_kv_requests:get_request_id(Req),
             StartTS = os:timestamp(),
             riak_core_vnode:reply(Sender, {w, Idx, ReqId}),
-            {Reply, UpdState} = do_put(Sender, BKey,  Object, ReqId, StartTime, Options, State),
+            {Reply, UpdState} = do_put(Sender, Req, State),
```
hazen, line 1049: Love how this cleans up the interface

src/riak_kv_vnode.erl
```diff
@@ -2787,223 +2801,5 @@ rollover_test_() ->
              ]}
     }.
 
-dummy_backend(BackendMod) ->
```
hazen, line 2809: We agreed this was not needed any more, right?

test/fsm_eqc_vnode.erl
```diff
@@ -99,9 +98,14 @@ get_postcommits() ->
 init([]) ->
     {ok, active, #state{}}.
 
-active(?VNODE_REQ{index=Idx,
-                  sender=Sender,
-                  request=?KV_GET_REQ{req_id=ReqId}}, State) ->
+active(?VNODE_REQ{index = Idx,
```
hazen, line 101: Why are we keeping the macro here?

lehoff (approved):
All previous comment addressed.

src/riak_kv_vnode.erl
jvoegele: @javajolt Our understanding is that this test code was written for an obsolete testing tool that is no longer in use, and that the test never gets executed.

@bsparrow435 Can you confirm?

test/fsm_eqc_vnode.erl
jvoegele: The `VNODE_REQ` record is defined in `riak_core_vnode.hrl` and since our scope is just the request records defined in `riak_kv_vnode.hrl` we left this one as-is. Per @lehoff this is for another round.

lehoff: I think we need to revisit how we do requests across CORE and KV - the extraction of CORE did not get a super clean separation of concepts.

Now we have the KV side of requests in order and then we can go back and look a the broader picture when we have live with this solution for a while.

gordonguthrie:
Before we get a merge on this on the KV side I would like an integration branch created and a trial merge with TS done to see if it makes sense to yolomerge this clean-up into TS now or to defer...

hazen:
@gordonguthrie You share my thoughts. I've been taking to Jason and Paul about this change already. Would like to simply extend the changes too TS when they are done.

paulhenrich:
Thanks for the feedback @gordonguthrie & @javajolt and @lehoff. I agree we should test out the merge. I am currently investigating test failures after a giddyup run and coordinating with Gordon on email.

paulhenrich:
I opened PR #1517 to see how the merge might go with TS. It is against a copy of `kv_ts_merge`.

src/riak_kv_vnode.erl
```diff
@@ -534,17 +520,23 @@ init([Index]) ->
...
-    riak_core_vnode:reply(Sender, {r, {error, overload}, Idx, ReqID});
-handle_overload_command(?KV_VNODE_STATUS_REQ{}, Sender, Idx) ->
-    riak_core_vnode:reply(Sender, {vnode_status, Idx, [{error, overload}]});
-handle_overload_command(?KV_W1C_PUT_REQ{type=Type}, Sender, _Idx) ->
+handle_overload_request(kv_get_request, Req, Sender, Idx) ->
+    ReqId = riak_kv_requests:get_request_id(Req),
+    riak_core_vnode:reply(Sender, {r, {error, overload}, Idx, ReqId});
+handle_overload_request(kv_w1c_put_request, Req, Sender, _Idx) ->
+    Type = riak_kv_requests:get_replica_type(Req),
     riak_core_vnode:reply(Sender, ?KV_W1C_PUT_REPLY{reply={error, overload}, type=Type});
```
hazen, line 535: Question previously mentioned: Why is the reply still a macro?

include/riak_kv_vnode.hrl
jvoegele: We left this one as-is because it is a _reply_ record and it would not make sense to mix it into the `riak_kv_requests` module, which (as implied by the name) is intended only for _request_ records. We could, of course, move `riak_kv_w1c_put_reply_v1` into a new module (say `riak_kv_replies`) but since there was only one reply record in existence we decided this was not worth introducing a new module.

src/riak_kv_vnode.erl
paulhenrich: We are strictly adhering to this being a refactoring of requests in the kv_vnode. It should probably be refactored to follow the same pattern at some point.

PR #1507, opened 2016-10-19
TS Insert NULL support

paegun:
depends on:
- https://github.com/basho/riak_ql/pull/144

does:
- for Riak TS insert support for NULL values, implicitly and explicitly, changed call site to riak_ql_ddl:insert_sql_columns/3 to allow the ddl module to flesh out fields and values not specified in the query.

### DIFF
--- a/src/riak_kv_qry.erl
+++ b/src/riak_kv_qry.erl
@@ -412,7 +412,7 @@ describe_table_columns_no_quantum_test() ->
                 " t timestamp not null,"
                 " w sint64    not null,"
                 " p double,"
-                " PRIMARY KEY (f, s, t))")),
+                " PRIMARY KEY ((f, s, t), f, s, t))")),
     Res = do_describe(DDL),
     ?assertMatch(
         {ok, {_, _,
--- a/src/riak_kv_qry.erl
+++ b/src/riak_kv_qry.erl
@@ -202,6 +202,8 @@ make_insert_row([], _Positions, Row, _FieldTypes) ->
     %% Out of entries in the value - row is populated with default values
     %% so if we run out of data for implicit/explicit fieldnames can just return
     {ok, Row};
+make_insert_row([{identifier, _Identifier} | _Values], _Positions, _Row, _FieldTypes) ->
+    {error, identifier_unexpected};
 make_insert_row([TypedVal | Values], [Pos | Positions], Row, FieldTypes) ->
     %% Note the Type in TypedVal = {Type, _} is what the value was
     %% parsed into, while its counterpart in FieldTypes is what DDL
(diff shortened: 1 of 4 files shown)


paegun:
wrt 3828f59
Riak TS does not support identifiers w/i the values portion of INSERT.  Ås this limitation is not a SQL language feature and may be supported, I felt it best to address this w/i Riak KV where the identifier would likely be de-referenced rather than w/i Riak QL.

src/riak_kv_qry.erl
```diff
@@ -202,6 +202,8 @@ make_insert_row([], _Positions, Row, _FieldTypes) ->
     %% Out of entries in the value - row is populated with default values
     %% so if we run out of data for implicit/explicit fieldnames can just return
     {ok, Row};
+make_insert_row([{identifier, _Identifier} | _Values], _Positions, _Row, _FieldTypes) ->
+    {error, identifier_unexpected};
```
hmmr: Make sure this error has a human-readable message. It's a matter of adding another `make_errorthing_resp` and adding a case branch here: https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_ts_svc.erl#L455L461.

paegun: added

paegun:
thumbot retry

src/riak_kv_ts_util.erl
```diff
@@ -112,25 +112,30 @@ build_sql_record_int(describe, SQL, _Cover) ->
     {ok, #riak_sql_describe_v1{'DESCRIBE' = D}};
 
 build_sql_record_int(insert, SQL, _Cover) ->
-    T = proplists:get_value(table, SQL),
-    case is_binary(T) of
+    Table = proplists:get_value(table, SQL),
```
hazen, line 115: Verbosity for the win! Never been a fan of single letter variables. Except when I was coding AppleSoft BASIC which _only_ had single character variables.

src/riak_kv_ts_svc.erl
```diff
@@ -564,6 +566,11 @@ make_undefined_field_in_insert_resp(BadFields) ->
       ?E_BAD_QUERY,
       flat_format("undefined fields: ~s", [string:join(BadFields, ", ")])).
 
+make_identifier_unexpected_resp(Identifier) ->
```
hazen, line 569: Do we test this bad response anywhere?

paegun: We do not test this bad response and the reason I did not originally provide this function to make a response for the case that the error is `identifier_unexpected` is that `xlate_insert_to_putdata` does not provide a means to report multiple errors, which excludes the ability to report on multiple error types.

hazen:
```
  riak_kv_qry: describe_table_columns_no_quantum_test...*failed*
in function riak_kv_qry:describe_table_columns_no_quantum_test/0 (src/riak_kv_qry.erl, line 406)
**error:{badmatch,{error,{0,riak_ql_parser,<<"No local key specified.">>}}}
```

andytill:
@javajolt need to pull from `riak_ts-develop`.

hazen:
:+1: 2a4a500

hazen:
@borshop merge

PR #1510, opened 2016-10-20
Replace msgpack encoding with ttb

erikleitch:
## **Overview**

This is one of two related PRs to replace msgpack with ttb as the default on-disk encoding format for TS.   The other is https://github.com/basho/eleveldb/pull/223.

Msgpack was originally chosen as the encoding format because in early TS development (before the query pipeline was even written), micro-benchmarking of decoding riak objects in C++ for secondary filtering found that msgpack was significantly faster than naive implementation of the ei decoding library for ttb.  In the context of the end-to-end query pipeline, however, that decoding in C++ represents less than 10% of the total query latency.  Far and away the largest contribution to the latency is converting encoded records back to erlang terms in riak_kv_qry:decode_results() when they are returned to riak from eleveldb.  In this context, the `binary_to_term` BIF is _significantly_ more efficient than our implementation of msgpack decoding -- up to 30% faster for queries of 100K rows, and increasing with the size of the query.  

You can find more discussion of the original micro-benchmarks, the query pipeline, and performance testing of this branch at https://github.com/basho/internal_wiki/wiki/Performance-Improvements-for-Riak#ts1.5.

The purpose of these PRs is to improve query performance by replacing msgpack with ttb encoding as the default on-disk format for TS data.
## **Changes in this repo**

`src/riak_kv_eleveldb_backend.erl` For backwards compatibility, the accompanying eleveldb PR changes the eleveldb fold to allow either msgpack or ttb-encoded data, and determines which encoding has been used on the fly for each key.  This removes the need to specify an encoding format when initiating the fold from erlang.  The `{encoding, type}` tuple has therefore been removed from the options (but will simply be ignored if it is specified, so that the eleveldb code will continue to work even if riak_kv is subsequently downgraded).

`src/riak_kv_ts_api.erl` Encode function in `put_data_to_partitions` has been changed to default to `erlang` encoding, aka ttb.

`src/riak_kv_w1c_worker.erl` Similarly, `kv_or_ts_details` has been changed to return an encoding function that uses `erlang` encoding instead of `msgpack`

`src/riak_object.erl` This was a holdover from long ago (prior to TS1.0 release), when riak_object:from_binary had to be told how to decode a msgpack-encoded record secretly masquerading as a normal riak object.  We changed the encoding flag to indicate the encoding type so that this dangerous situation was no longer needed, but left the deprecated interface lying around.  It's not used by any code today, and is confusing, and I'm therefore removing it.
## **Compatibility**

**Backwards Compatibility** `riak_object:from_binary` has always used the encoding flag written with each data record to switch on encoding type and decode accordingly.  Data written previously with the msgpack flag will therefore continue to be correctly decoded -- customers upgrading to TS1.5 will not have to do anything special to process data written in msgpack format by earlier versions of TS.  Interleaved data written in either format are seamlessly processed by eleveldb, and this is part of the coverage of unit tests included with the eleveldb PR.

**Downgrade Compatibility** In these PRs, the existing TTB encoding flag (used for normal KV objects) has been reused for TS records -- no new encoding flag has been created.  That means that if data are written in the new format (TTB encoding) and riak is subsequently downgraded to TS1.4, both msgpack and ttb-encoded data will be correctly decoded by earlier versions of TS.  _Caveat:_ If downgrading from TS1.5, a patched version of eleveldb will be required, so that records written in TTB format are not ignored by the eleveldb fold.  TS1.5 eleveldb can simply be used with downgraded versions of RiakTS.

Although the encoding flag has been removed from the streaming fold options, the TS1.5 version of eleveldb will simply ignore the parameter if specified, so downgrading to earlier versions of TS will not cause an error.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -510,8 +510,7 @@ range_scan(FoldIndexFun, Buffer, Opts, #state{fold_opts=_FoldOpts,
     Options = [
                {start_key,   StartKey2},
                {end_key,     EndKey2},
-               {fold_method, streaming},
-               {encoding,    msgpack} | range_scan_additional_options(W)
+               {fold_method, streaming} | range_scan_additional_options(W)
               ],
     KeyFolder = fun() ->
                         Vals = eleveldb:fold(Ref, FoldFun, [], Options),
--- a/src/riak_kv_ts_api.erl
+++ b/src/riak_kv_ts_api.erl
@@ -135,7 +135,7 @@ put_data_to_partitions(Data, Bucket, BucketProps, DDL, Mod) ->
     CappedBatchSize = app_helper:get_env(riak_kv, timeseries_max_batch_size,
                                          1024 * 1024),
     EncodeFn =
-        fun(O) -> riak_object:to_binary(v1, O, msgpack) end,
+        fun(O) -> riak_object:to_binary(v1, O, erlang) end,
 
     {ReqIds, FailReqs} =
         lists:foldl(
--- a/src/riak_kv_w1c_worker.erl
+++ b/src/riak_kv_w1c_worker.erl
@@ -475,7 +475,7 @@ kv_or_ts_details(RObj, {ok, LocalKey}) ->
     MD1 = dict:erase(?MD_TS_LOCAL_KEY, MD),
     RObj1 = riak_object:update_metadata(RObj, MD1),
     {RObj1, {riak_object:key(RObj), LocalKey},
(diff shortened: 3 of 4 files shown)


paegun:
+1 d3614c9

NOTE: The downgrade caveat mentioned is great. If we release Riak TS 1.4.2 w/ only eleveldb upgraded to the version supporting both msgpack and TTB, this is a low (read: near zero) risk and supports all scenarios of upgrade/downgrade between 1.4.x and 1.5.0 .

paegun:
Not supporting downgrade from TS 1.5.0 -> 1.4.x accepted, so merging.

PR #1514, opened 2016-10-28
Add Status to SQL "SHOW TABLES" result

paegun:
- added Status field to SQL "SHOW TABLES" result.

Related PRs:
- https://github.com/basho/riak_test/pull/1176
- https://github.com/basho/private_basho_docs/pull/250
- https://github.com/basho/riak_kv/pull/1530
  - !IMPORTANT! ^^ moves the function that is used to determine if a table exists, is in all the places it needs to be, ... Status.


andytill:
This PR conflicts with the changes I have made in https://github.com/basho/riak_kv/pull/1500 which changes the responsibilities of the dets compile table. It now stores versions of the DDL as different DDL record types, e.g. `ddl_v1`, `ddl_v2`.

The status value should come from the bucket type status in riak_core, so either created or activated.

andytill:
I re-read the issue (https://bashoeng.atlassian.net/browse/RTS-1404) and after PR #1500 is merged we need to do a check on the tables required capability against the cluster capability to work out if it should be disabled or not.

src/riak_kv_compile_tab.erl
```diff
@@ -151,7 +153,87 @@ get_ddl(BucketType, Version) when is_binary(BucketType), is_atom(Version) ->
...
+    riak_kv_ts_util:get_table_ddl(Table).
+
+-spec check_table_feature_supported(DDLRecCap::atom(),
+                                    DecodedReq::ts_service_req()) -> ts_service_req() |
+                                                                     {error, string()}.
+check_table_feature_supported(DDLRecCap, DecodedReq) ->
+    riak_kv_ts_util:check_table_feature_supported(DDLRecCap, DecodedReq).
+
+-else.
+expected_table_status(Table) when Table =:= <<"my_type2">> ->
```
hazen: Super nit picky, but why not just match on the table name, e.g. `expected_table_status(<<"my_type2">>)`?

src/riak_kv_compile_tab.erl
```diff
@@ -151,7 +153,87 @@ get_ddl(BucketType, Version) when is_binary(BucketType), is_atom(Version) ->
...
+    <<"Not Active">>;
+expected_table_status(Table) when Table =:= <<"my_type4">> ->
+    <<"Not Active">>;
+expected_table_status(Table) when Table =:= <<"my_type5">> ->
+    <<"Not Active">>;
+expected_table_status(_Table) ->
+    <<"Active">>.
+
+get_table_ddl(Table) when Table =:= <<"my_type4">> ->
```
hazen: Same as above

src/riak_kv_compile_tab.erl
```diff
@@ -151,7 +153,87 @@ get_ddl(BucketType, Version) when is_binary(BucketType), is_atom(Version) ->
...
+    {error, no_type};
+get_table_ddl(Table) when Table =:= <<"my_type5">> ->
+    {error, missing_helper_module};
+get_table_ddl(_Table) ->
+    Module = {}, %% not used by caller
+    DDL = {}, %% not used by caller
+    {ok, Module, DDL}.
+
+check_table_feature_supported(DDLRecCap, DecodedReq={ok, _Req, {"perm", Table}}) when
```
hazen: Is `<<"my_type2">>` the only failure? Maybe restructure so that you first catch that one case, then fall through to the passing case?

src/riak_kv_compile_tab.erl
```diff
@@ -151,7 +153,87 @@ get_ddl(BucketType, Version) when is_binary(BucketType), is_atom(Version) ->
...
+-spec get_table_ddl(binary()) ->
+    {ok, module(), ?DDL{}} | {error, term()}.
+get_table_ddl(Table) ->
+    riak_kv_ts_util:get_table_ddl(Table).
+
+-spec check_table_feature_supported(DDLRecCap::atom(),
+                                    DecodedReq::ts_service_req()) -> ts_service_req() |
+                                                                     {error, string()}.
+check_table_feature_supported(DDLRecCap, DecodedReq) ->
+    riak_kv_ts_util:check_table_feature_supported(DDLRecCap, DecodedReq).
```
hazen: What about exposing `riak_kv_ts_util:is_table_supported/2` and using it directly instead of faking a request record?  Then is returns a boolean which is a bit clearer than returning a record upon success...

hazen (commented):
Just a few observations. Let me know your thoughts.

src/riak_kv_compile_tab.erl
paegun: I did not and do not understand why the function riak_kv_ts_util:check_table_feature_supported/2 returns a record. check_... by pattern should return a boolean. So...conflate this PR purpose w/ code reviewing a previous PR, mission possible, mission accepted.

src/riak_kv_compile_tab.erl
paegun: Good point, not nit picky, will fix.

src/riak_kv_compile_tab.erl
paegun: The conditions required to be active are greater than the conditions required to not be, so I posed the function guards along those lines. Being pessimistic in this way IMHO is more than just "future proofing" but should make the coding of expected results easier to reason about.

paegun: I found it easier to reason about to list those active, but yes, I see your point, changing.

src/riak_kv_compile_tab.erl
```diff
@@ -151,7 +153,83 @@ get_ddl(BucketType, Version) when is_binary(BucketType), is_atom(Version) ->
...
+get_table_status(Table) ->
+    case get_table_ddl(Table) of
+        {error, no_type} -> <<"Not Active">>;
+        {error, missing_helper_module} -> <<"Not Active">>;
+        {ok, _Mod, _DDL} -> get_table_status_by_version(Table);
+        _ -> get_table_status_by_version(Table)
+    end.
+
+get_table_status_by_version(Table) ->
+    Req = #tsqueryreq{},
```
hazen: So I didn't have a problem with `check_table_feature_supported/2` returning a request.  It seems that part is actually used down in `riak_kv_ttb_ts:decode/2`.  My real issue was we are faking a request to essentially pass the table name down into `riak_kv_ts_util:is_ddl_version_supported/2`.  We are passing in `{ok, Req, {"perm", Table}}` which is pattern matched on `{ok, _, {_, Table}}` which just calls `is_ddl_version_supported(v2, Table)`.  I was just suggesting you expose that function to simplify the call and clarify the test condition (a version comparison.)

Less code == good.

javajolt:
@borshop merge

hazen:
:+1: dc18758

hazen:
@borshop merge

PR #1515, opened 2016-10-28
Bugfix: do not match dead process IDs in compilation table

macintux:
@hmmr discovered a scenario where frequently stopping and starting nodes while testing DDL compilation can result in an error due to lingering process IDs from the previous instantiation of the node.

Clean up PIDs after the compile is finished, and also explicitly match only `compiling` records so that users of prior versions with legacy data do not hit this bug.

Also extend commentary and add a `riak_test` intercept point to make system testing for this bug easier.

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -45,7 +45,7 @@
          S == failed orelse
          S == retrying)).
 
-%% 
+%%
 -spec new(file:name()) ->
          {ok, dets:tab_name(), dets:tab_name()} | error.
 new(FileDir) ->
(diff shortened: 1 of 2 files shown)


macintux:
Brett will be creating a test for this.

hazen:
```
riak_kv_compile_tab.erl:172: The call riak_kv_compile_tab:insert(BucketType::any(),DDLVersion::any(),DDL::any(),'undefined',State::'compiled' | 'compiling' | 'failed' | 'retrying') breaks the contract (BucketType::binary(),DDLVersion::riak_ql_component:component_version(),DDL::term(),CompilerPid::pid(),State::compiling_state()) -> 'ok' | 'error'
```

Looks like we need to expand the state definition

macintux:
Sadly enough, not only did I not run dialyzer, I didn't even try to compile it. Got too enthusiastic about the fix.

macintux:
Fix submitted

hmmr:
+1 72e17e6

macintux:
Thanks @hmmr. Let's wait until we have a test, though, before we merge this.

andytill:
This PR conflicts with the changes I have made in #1500 which changes the responsibilities of the dets compile table. It now stores versions of the DDL as different DDL record types, e.g. ddl_v1, ddl_v2.

If you anyone writes the test I will run it against the changes in my PR.

macintux:
This is irrelevant with Andy's changes.

PR #1516, opened 2016-10-30
Equality operator with quantum timestamp RTS-1286

andytill:
Rewrite queries with `WHERE mytimestamp = 1234' to 'WHERE mytimestamp >= 1234 AND mytimestamp <= 1234` when `mytimestamp` is the quantum is a partition key. This allows the the quantum time to be defined exactly, instead of having an inclusive range of the same time value.

With table `bios` defined as `CREATE TABLE bios ( ts timestamp PRIMARY KEY((id, QUANTUM(ts, 15, 'm')), ts));`.

``` sql
riak-shell(3)>DESCRIBE bios;         
+------+---------+-------+-----------+---------+--------+----+
|Column|  Type   |Is Null|Primary Key|Local Key|Interval|Unit|
+------+---------+-------+-----------+---------+--------+----+
|  ts  |timestamp| false |     1     |    1    |   15   | m  |
+------+---------+-------+-----------+---------+--------+----+

riak-shell(4)>SELECT * FROM bios WHERE ts > 1 AND ts < 200;
+------------------------+
|           ts           |
+------------------------+
|1970-01-01T00:00:00.01Z |
|1970-01-01T00:00:00.011Z|
+------------------------+

riak-shell(5)>SELECT * FROM bios WHERE ts = '1970-01-01T00:00:00.01Z';
+-----------------------+
|          ts           |
+-----------------------+
|1970-01-01T00:00:00.01Z|
+-----------------------
```

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -739,7 +739,26 @@ break_out_timeseries(Filters1, PartitionFields1, no_quanta) ->
 break_out_timeseries(Filters1, PartitionFields1, QuantumField) when is_binary(QuantumField) ->
     case find_timestamp_bounds(QuantumField, Filters1) of
         {_, {undefined, undefined}} ->
-            error({incomplete_where_clause, ?E_TSMSG_NO_BOUNDS_SPECIFIED});
+            %% if we don't have a time range then check for a time equality
+            %% filter e.g. mytime = 12345, which is rewritten as
+            %% mytime >= 12345 AND mytime <= 12345
+            {QEqFilters, OtherFilters} =
+                lists:splitwith(fun({Op, F, _}) ->
+                                    Op == '=' andalso F == QuantumField
+                                end, Filters1),
+            case QEqFilters of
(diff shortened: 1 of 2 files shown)


paegun:
If the error condition for E_CANNOT_HAVE_TWO_EQUALITY_FILTERS_ON_QUANTUM_WITHOUT_RANGE is rethought as a request for multiple ranges that are guaranteed not to overlap, hence result in surely no records being yielded, throwing an error does not match how SQL RDBMSs (common) handle such requests nor does throwing an error match how Riak TS (consistent) handles such requests. From my experience, it is common and consistent to yield no records instead of throwing an error.

andytill:
I agree that this should return no records, ideally without even sending sub queries. It is consistent with Riak TS behaviour, for example:

```
riak-shell(4)>SELECT * FROM bios WHERE ts > 100 AND ts < 10;                                                          
Error (1001): The lower time bound is greater than the upper time bound.
```

paegun:
+1 77e3d71

andytill:
@borshop merge

PR #1520, opened 2016-11-01
Pass old object to Yokozuna for more efficient sibling deletes. [JIRA: RIAK-2986]

JeetKunDoug:
`riak_kv_vnode` has all the information necessary to tell Yokozuna how to delete previous siblings, but didn't pass that information on because it would require a read-before-write. However, in omitting this information, we were required to do an incredibly expensive operation on the Solr side ("Delete By Query") to clean up sibling data. This PR plus work on the Yokozuna side will allow us to do sibling cleanup without Delete by Query, which is now reserved for AAE repairs only.
- Refactor `prepare_put` and related functions to make it easier to reason about when a read-before-write is necessary.
- Add check to see if bucket is indexed in Solr and include it in the read-before-write case
- When we have an old object, send it to Solr in the `?INDEX()` macro + `yz_kv:index` call.
- When no old object is available, send `no_old_object`

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -98,9 +98,11 @@
 %% Use values so that test compile doesn't give 'unused vars' warning.
 -define(INDEX(A,B,C), _=element(1,{A,B,C}), ok).
 -define(INDEX_BIN(A,B,C,D,E), _=element(1,{A,B,C,D,E}), ok).
+-define(IS_SEARCH_ENABLED_FOR_BUCKET(BProps), _=element(1, {BProps}), false).
 -else.
 -define(INDEX(Obj, Reason, Partition), yz_kv:index(Obj, Reason, Partition)).
 -define(INDEX_BIN(Bucket, Key, Obj, Reason, Partition), yz_kv:index_binary(Bucket, Key, Obj, Reason, Partition)).
+-define(IS_SEARCH_ENABLED_FOR_BUCKET(BProps), yz_kv:is_search_enabled_for_bucket(BProps)).
 -endif.
 
 -ifdef(TEST).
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1476,147 +1478,169 @@ prepare_put(State=#state{vnodeid=VId,
...
-                             starttime=StartTime}) ->
+                             starttime=StartTime,
+                             bprops = BProps}) ->
     %% Can we avoid reading the existing object? If this is not an
     %% index backend, and the bucket is set to last-write-wins, then
     %% no need to incur additional get. Otherwise, we need to read the
     %% old object to know how the indexes have changed.
+    IndexBackend = is_indexed_backend(Mod, Bucket, ModState),
+    IsSearchable = ?IS_SEARCH_ENABLED_FOR_BUCKET(BProps),
+    RequiresReadBeforeWrite = LWW andalso (not IndexBackend) andalso (not IsSearchable),
```
ian-mi: It seems like RequiresReadBeforeWrite is named incorrectly. It should either be renamed to something like BlindPut or the test and following case statement should be negated.

src/riak_kv_vnode.erl
```diff
@@ -1540,26 +1540,19 @@ prepare_put_existing_object(#state{idx =Idx} = State,
     {ActorId, State2} = maybe_new_key_epoch(Coord, State, OldObj, RObj),
     case put_merge(Coord, LWW, OldObj, RObj, ActorId, StartTime) of
         {oldobj, OldObj} ->
-            {{false, OldObj}, PutArgs, State2};
+            {{false, {OldObj, no_old_object}}, PutArgs, State2};
```
ian-mi, line 1543: Could replace {false, {OldObj, no_old_object}} by {false, OldObj}, changing perform_put to reflect this, as there is never a new object in the false case.

src/riak_kv_vnode.erl
```diff
@@ -1486,8 +1486,8 @@ prepare_put(State=#state{vnodeid=VId,
     %% old object to know how the indexes have changed.
     IndexBackend = is_indexed_backend(Mod, Bucket, ModState),
     IsSearchable = ?IS_SEARCH_ENABLED_FOR_BUCKET(BProps),
-    RequiresReadBeforeWrite = LWW andalso (not IndexBackend) andalso (not IsSearchable),
-    case RequiresReadBeforeWrite of
+    SkipReqdBeforeWrite = LWW andalso (not IndexBackend) andalso (not IsSearchable),
```
ian-mi: This appears to address my comment on the previous commit. Assuming this should be SkipReadBeforeWrite.

JeetKunDoug: Yea - if you look at the overall PR I actually renamed it to "SkipReqdBeforeWrite" which I realize is misspelled, so I'll fix that.

src/riak_kv_vnode.erl
JeetKunDoug: True, but I really felt like having the _same structure_ for all cases actually made the code easier to read, and also to _know_ that you've taken the old object stuff into account in each code path, which is why I pushed it through to all the cases.

src/riak_kv_vnode.erl
JeetKunDoug: As you can now see, renamed.

ian-mi:
+1

russelldb:
Looks fine to me.

Such a large refactor without first writing to tests to capture the existing behaviour is a concern. From comparing to the old code it _looks_ like you captured everything. In this case I would have just added the new branch for solr objects and refactored in a separate PR, but I also understand cleaning as you go. But without tests…at short notice…

Anyway. It looks sane, +1 from me.

Since there is no description in the PR of how to reproduce the ill effects, there is no way to verify that this PR mitigates them. I'll take your word for it.

src/riak_kv_vnode.erl
```diff
@@ -1476,147 +1478,162 @@ prepare_put(State=#state{vnodeid=VId,
...
+    {{true, {ObjToStore, OldObj}},
+     PutArgs#putargs{index_specs = IndexSpecs,
+                     is_index    = IndexBackend}, State}.
+
+maybe_prune_vclock(_PruneTime=undefined, RObj, _BProps) ->
+    RObj;
+maybe_prune_vclock(PruneTime, RObj, BProps) ->
+    riak_object:prune_vclock(RObj, PruneTime, BProps).
+
+get_index_specs(_IndexedBackend=true, CacheData, RequiresGet, NewObj, OldObj) ->
```
lehoff, line 1564: Would it make sense to provide even more pattern matching in the function head?

```
get_index_specs(_IndexedBackend=true, _CacheData=undefined, _RequiresGet, NewObj, OldObj) ->
     riak_object:diff_index_specs(NewObj, OldObj);
get_index_specs(_IndexedBackend=true, _CacheData, _RequiresGet=true, NewObj, OldObj) ->
     riak_object:diff_index_specs(NewObj, OldObj);
get_index_specs(_IndexedBackend=true, CacheData, _RequiresGet=false, NewObj, OldObj) ->
    NewData = riak_object:index_data(NewObj),
    riak_object:diff_index_data(NewData, CacheData);
get_index_specs(_IndexedBackend=false, _CacheData, _RequiresGet, _NewObj, _OldObj) ->
    [].
```

src/riak_kv_vnode.erl
```diff
@@ -1476,147 +1478,162 @@ prepare_put(State=#state{vnodeid=VId,
...
+    IsSearchable = ?IS_SEARCH_ENABLED_FOR_BUCKET(BProps),
+    SkipReadBeforeWrite = LWW andalso (not IndexBackend) andalso (not IsSearchable),
+    case SkipReadBeforeWrite of
+        true ->
+            prepare_blind_put(Coord, RObj, VId, StartTime, PutArgs, State);
+        false ->
+            prepare_read_before_write_put(State, PutArgs, IndexBackend, IsSearchable)
+    end.
+
+is_indexed_backend(Mod, Bucket, ModState) ->
```
lehoff, line 1497: Would it make sense to change the return type for `is_indexed_backed/3` to `indexed_backend | non_indexed_backend`?
And then change the name to `backend_index_type/3`

It would make the logic read nicer:

```
BackendIndexType = backend_index_type(Mod, Bucket, ModState),
...
SkipReadBeforeWrite = LWW andalso (BackendIndexType == non_indexd_backend) andalso ...
```

This would also simplify the `get_index_specs/5` function below as it would not have to name the first argument because the atom says it all.

Might make sense to do the same for `IsSearchable`.

src/riak_kv_vnode.erl
```diff
@@ -1476,147 +1478,162 @@ prepare_put(State=#state{vnodeid=VId,
...
+            not_found;
+        {ok, TheOldObj, _UpdModState} ->
+            {ok, TheOldObj}
+    end;
+get_old_object_or_fake(false, Bucket, Key, _Mod, _ModState, CacheClock) ->
+    FakeObj0 = riak_object:new(Bucket, Key, <<>>),
+    FakeObj = riak_object:set_vclock(FakeObj0, CacheClock),
+    {ok, FakeObj}.
+
+determine_requires_get(CacheClock, RObj, IsSearchable) ->
```
lehoff, line 1608: As with the backend index type, maybe change from returning `boolean()` to returning `requires_get | does_not_require_get`.

lehoff: Replace the case statement with a pattern match in the function head.

lehoff (approved):
If the small changes are not implemented a Jira issue should be created so that we do them in the next release.

lehoff:
+1

JeetKunDoug:
@russelldb agree on the testing - I leaned on the existing tests in riak_kv_vnode when I did the refactoring part of the PR and fixed things appropriately. Adding the search stuff in tests in `riak_kv` is miserable because of the circular dependencies and the way we are handling testing of `?INDEX` and related code in this module - we need to spend some time writing the long-awaited "during a put at the vnode hook" code so we can decouple these things and test more cleanly in `riak_kv` rather than leaning on tests in yokozuna.

bashopatricia:
create jira issue

PR #1521, opened 2016-11-01
Local keys must be specified, fix test data to include the local key

andytill:
Fix a failing test. Local keys must be specified, fix test data to include the local key.


macintux:
+1

PR #1526, opened 2016-11-03
Fix put FSM race condition

ian-mi:

This has been tested to fix the race condition using PULSE.


bsparrow435:
Looks good @ian-mi  👍

bsparrow435:
👍 f5386fcd742cc42a0c65a164089006316f551f47

PR #1527, opened 2016-11-03
Fix put FSM race condition (#1526) [JIRA: RIAK-3142]

ian-mi:


bsparrow435:
👍

bashopatricia:
create jira issue

Basho-JIRA:
Fixed in 2.0 branch. 

PR #1528, opened 2016-11-03
Fix put FSM race condition (#1526)

ian-mi:

### DIFF
--- a/.thumbs.yml
+++ b/.thumbs.yml
@@ -1,0 +1,11 @@
+minimum_reviewers: 2
+merge: true
+build_steps:
+  - make clean
+  - make deps
+  - make compile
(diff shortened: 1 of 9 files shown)


nickelization:
@ian-mi did you mean to make this a PR from develop-2.2 to 2.0? Given the title and description, I'm guessing maybe you meant to create this from a different branch where you just backported the race condition fix?

ian-mi:
No, this is a forward merge of the fix which was made against 2.0. More commits have been added to 2.0, however, so maybe this PR can be replaced by a forward merge of 2.0.

bsparrow435:
Closing, will be forward ported in a seperate PR.

PR #1529, opened 2016-11-04
wm_timeseries: support [] (= NULL) in /ts/put request body

hmmr:
Required by https://github.com/basho/riak_test/pull/1185.

This allows HTTP put API calls to insert rows with NULLs (if that column is declared 'not null'). Indirect test for that is in ts_simple_http_security_SUITE.

### DIFF
--- a/src/riak_kv_wm_timeseries.erl
+++ b/src/riak_kv_wm_timeseries.erl
@@ -393,10 +393,14 @@ json_struct_to_obj(FieldValueList, Fields) ->
             || Field <- Fields],
     list_to_tuple(List).
 
-extract_field_value({Name, Type}, FVList) ->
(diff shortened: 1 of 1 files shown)


paegun:
+1 a931d76

Good that there's tests at the riak_test level.

It would be better to have unit tests here that cover all of the field combinations including ddl versus (insert or put) query nullability and type against the range of values that stress each combination. Considering no such test existed for type-level checking, IMO we should pick this up in a general improvement task later.

hmmr:
@paegun There are a couple of other `@todo` items, such as remove any leaks of `?DDL` outside of riak_ql. This module should appear in the tech debt drawer. But I didn't want to go outside of this specific regression in this PR.

hmmr:
@borshop merge

PR #1530, opened 2016-11-04
Check table existence before it is used RTS-1094

andytill:
Two recent changes circumvented previous checks for table existence.

1. Checks for tables being disabled because of capabilities/table features.
2. The insert API requires the helper module to exist to create the record for it, which happens before the check for table existence.

How to reproduce, open the riak-shell and run without creating table gerp.

```sql
riak-shell(1)>INSERT INTO gerp VALUES(9);
Error (1019): gerp is not an active table
```

This changes the point at which checks for table existence are made. It is not in the decode phase of the riak_api behaviour and before records are created.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -12,7 +12,7 @@
             {d, 'TEST_FS2_BACKEND_IN_RIAK_KV'}]}.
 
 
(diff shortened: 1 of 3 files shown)


paegun:
-1 This change was developed after #1514 and moved the functions that are used therein to check the table's status.

PR #1532, opened 2016-11-10
Fix race between capabilities registration and AAE startup. [JIRA: RIAK-2985]

JeetKunDoug:
During testing, it was found to be possible for a hashtree to start
_before_ the `{riak_kv, object_hash}` capability was registered, causing
the wrong version to be chosen when the tree started up. This can cause
old trees to be loaded, ignoring new trees that have more recent data,
and the wrong hash version to be used, which can cause unnecessary
repairs.

Moving the registration of the anti-entropy-related capabilities
to `riak_kv_entropy_manager` and changing the order of startup in
`riak_kv_sup` to make sure the entropy manager is started before
the vnodes start ensures that the capability is registered and
negotiated before hashtrees try to start.

This can be tested by running the `yz_aae_test` in a loop on certain hardware - found during testing of 2.2.0, and tends to reproduce more readily on a SmartOS-hosted Centos 7 machine. Unfortunately, it's a race condition, which makes it difficult to test in isolation.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -163,10 +163,6 @@ start(_Type, _StartArgs) ->
                                            mapred_2i_pipe,
                                            [{true, true}, {false, false}]}),
 
-            riak_core_capability:register({riak_kv, anti_entropy},
-                                          [enabled_v1, disabled],
-                                          disabled),
-
             riak_core_capability:register({riak_kv, handoff_data_encoding},
                                           [encode_raw, encode_zlib],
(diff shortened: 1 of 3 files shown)


ian-mi:
+1 0c3bcb6

JeetKunDoug:
@borshop merge

bashopatricia:
create jira issue

Basho-JIRA:
Completed 

***[posted via JIRA by Douglas Rohrer]***

PR #1535, opened 2016-11-14
Merge develop-2.2 into develop

paulhenrich:


rebar.config
```diff
@@ -32,9 +32,9 @@
         {bitcask, ".*", {git, "git://github.com/basho/bitcask.git", {tag, "2.0.4"}}},
```
fadushin, line 32: Let's revert these to branch dependencies

paulhenrich: Leaving this alone for now and merging per conversation in eng_riak.

PR #1538, opened 2016-11-15
Parallel decode of query results

erikleitch:
<hr>

**Summary**

This PR implements changes to decode SQL SELECT query results at the vnodes.  Currently, SQL queries are split into subqueries by the coordinating node, each vnode in the coverage plan executes the subquery as a fold over keys in eleveldb, and the results are returned to the coordinating vnode, which then decodes the results from on-disk format to erlang terms.

The problem with this architecture is that although large queries can be parallelized across the cluster by tailoring TS quanta to split large queries into many subqueries that execute as parallel folds, the decoding is still serialized at the coordinating node.  For large queries, the decoding dominates the query latency, so we therefore lose the benefit of parallelizing the queries in the first place.

This change delegates decoding of the query results to the vnodes in the coverage plan, so that both the eleveldb fold and the decoding of the resulting data can be effectively parallelized across the cluster.  This results in a > N-fold speedup for large queries, where N is the number of physical nodes in the cluster

<hr>

**Overall Design**

Delegating decoding to the vnodes instead of the coordinator is straightforward, but to ensure functionality in mixed clusters, the implementation is a bit more complicated.  In summary, the design is:

- we register a new capability `{riak_kv, decode_query_results_at_vnode}` for riak_kv, indicating whether or not decoding can be done at the vnode.

- the coordinator sends query requests to the vnodes as before -- i.e., no modifications have been made to the request path.  This ensures that query requests can be sent from either an old or new coordinator to either an old or new vnode.

- the new vnode checks `{riak_kv, decode_query_results_at_vnode}` to determine whether results should be sent decoded or encoded. If true, the vnode sends a decoded Chunk, as `{results, {decoded, Chunk}}`. If false, the vnode sends an encoded Chunk, as `{results, Chunk}` (current behavior)

  This allows the coordinator to handle either decoded or encoded data, by pattern-matching on the message structure.

<hr>

**Details**

`src/riak_kv_app.erl` 

- Modified to register the new `{riak_kv, decode_query_results_at_vnode}` capability.

`src/riak_kv_qry_worker.erl` 
    
- Now exports `decode_results/1` because it may be used either at the vnode (in `riak_kv_vnode.erl`) or here by the coordinator.  
- Helper function `get_decoded_results/1` has been added to handle either encoded or decoded data, depending on whether the sending vnode is old or new.  

`src/riak_kv_vnode.erl` 
    
- A range-scan-specific result acknowledgement function `range_scan_result_fun_ack/2` replaces the default ack function for query requests.  
- `range_scan_result_fun_ack/2` checks cluster capabilities, and if `{riak_kv, decode_query_results_at_vnode}` capability is `true`, decodes the results before sending back to the coordinator (if `false` it sends back encoded results as before)

<hr>

**Backwards/Downgrade Compatibility**

The aim of the design is to support functionality in mixed clusters, where we may have _either_ old or new coordinators talking to _either_ old or new vnodes; in short, there should be no impact on functionality if a customer upgrades to TS1.5, and subsequently downgrades some nodes.  I summarize the interaction for each of the 4 possible scenarios below:

1. new coordinator + new vnode

   - new vnode checks capabilities, sends back _decoded_
   data, since `{riak_kv, decode_query_results_at_vnode} == true`
   - new vnode sends back decoded data as `{results, {decoded, Chunk}}`
   - new coordinator knows not to decode the result because of pattern match on `{results, {decoded, Chunk}}`

2. old coordinator + new vnode

   - new node checks capabilities, sends back _encoded_
   data, since `{riak_kv, decode_query_results_at_vnode} == false`
   - new vnode sends back encoded data as `{results, Chunk}`, as before
   - old coordinator decodes the data as before

3. new coordinator + old vnode

   - old vnode sends back encoded data, as before
   - old vnode sends back encoded data as `{results, Chunk}`
   - new coordinator knows to decode the result because of pattern match on `{results, Chunk}`

4. old coordinator + old vnode

   - as before

<hr>

**Performance**

Performance with this branch is described in detail here: https://github.com/basho/internal_wiki/wiki/Performance-Improvements-for-Riak#ts1.5.  

_Disclaimer: the following discussion presents results from TS1.4 with the modifications in this PR, except for the capabilities lookup.  I'm doing this 1. because I already have those results handy, and I'm still working on the comparison with this branch in TS1.5-alpha, and 2. because if we remove the TTB encoding as planned, these results are more representative of the performance boost we should expect._  

I reproduce two relevant plots below.  Each plot shows query latencies for a range of table schema sizes and query sizes, for queries executed monolithically against a single partition (left panels),  the same queries split into sub-queries across 3 nodes (middle panels) and the ratio of those latencies (right panels).

The first demonstrates that the change behaves as expected.  Namely, that by reducing the quantum so that large queries are parallelized across the cluster, the ratio approaches ~N for large queries (in this case N = 3):

tsquerylatency_qcmp_prototype

The second compares monolithic queries between TS1.4 and this branch, demonstrating that the additional overhead from disterl handling of decoded terms has a negligible impact on the query latency (ratio is consistent with 1):

tsquerylatency_qcmp_single


src/riak_kv_qry_worker.erl
```diff
@@ -339,6 +351,20 @@ run_select_on_chunk(SubQId, Chunk, #state{qry = Query,
...
+%%
+%%   if already decoded, simply returns the decoded data
+%%
+%%   if not, decodes and returns
+%% ------------------------------------------------------------
+
+get_decoded_results({decoded, Chunk}) ->
+    Chunk;
+get_decoded_results(Chunk) ->
+    decode_results(lists:flatten(Chunk)).
```
andytill, line 367: I tried removing both instances of `lists:flatten/1` in this PR, and of course it doesn't work! We get this error:

```
2016-11-17 09:55:15.258 [error] <0.1541.0> CRASH REPORT Process <0.1541.0> with 0 neighbours exited with reason: no function clause matching riak_kv_qry_worker:decode_results([[{<<16,0,0,0,3,12,183,128,8,16,0,0,0,2,18,179,220,173,247,91,133,210,195,98,153,0,8,18,179,...>>,...},...]]) line 202 in gen_server:terminate/6 line 744
```

It seems like in `riak_kv_fold_buffer` a chunk of rows called an `Item` is added to an accumulator but in a non-flat way, e.g. `[Row1,Row2]|Acc]`. This is interesting because when `riak_kv_fold_buffer` works out the size of the buffer it uses the number of `Item` chunks it has received, and not the number of deep rows.

If you thought the flatten was a problem we could rewrite `decode_results/1` to iterate a deep list so flattening is not necessary. Flattening basically reallocates the list, each element requires one word of memory plus one word for each element: http://erlang.org/doc/efficiency_guide/advanced.html.

src/riak_kv_vnode.erl
```diff
@@ -2041,6 +2041,60 @@ result_fun_ack(Bucket, Sender) ->
...
+%% parallel decoding), we reply with the query results already
+%% decoded, so that the work of decoding the results can be
+%% parallelized among the vnodes in the query coverage plan, instead
+%% of being serialized at the coordinating vnode.
+%%
+%% If false (i.e., we have a mixed-mode cluster), we reply with the
+%% encoded results, as before.
+%% ------------------------------------------------------------
+
+range_scan_result_fun_ack(Bucket, Sender) ->
```
andytill, line 2077: This function contains tabs and space indentation.

src/riak_kv_vnode.erl
```diff
@@ -2041,6 +2041,60 @@ result_fun_ack(Bucket, Sender) ->
...
+    fun(Items) ->
```
andytill, line 2078: A small feature request, or I can do it after if you prefer. Can this fun call a function which does the heavy lifting, instead of inside the fun makes it traceable, and so debuggable. I had to do this for my `lists:flatten/1` investigation above.

``` erlang
range_scan_result_fun_ack_fun(Bucket, Sender) ->
    fun(Items) ->
        range_scan_result_fun_ack(Bucket, Sender, Items)
    end.

range_scan_result_fun_ack(Bucket, Sender, Items) ->
```

hmmr:
Nothing to raise issues about, except a collateral change in riak_ts-deveop-1.4 (already pushed, https://github.com/basho/riak_kv/commit/d05fde70c38b908df9) to ensure this capability is known to 1.4 nodes (and therefore, can be negotiated) in a mixed-cluster scenario. A PR adding this capability to be accounted for after node transition, is https://github.com/basho/riak_test/pull/1198.

src/riak_kv_qry_worker.erl
```diff
@@ -40,7 +40,13 @@
...
          handle_info/2,
          terminate/2,
-         code_change/3
+         code_change/3,
+
+	 %% decode_results/1 is now exported because it may be used in
+	 %% riak_kv_vnode to pre-decode query results at the vnode
+	 %% rather than here
+
+	 decode_results/1
```
hmmr, line 49: Off-by-one indent size here (for lack of any essential objections :).

hmmr:
With this change to riak_ts-develop-1.4: https://github.com/basho/riak_kv/commit/9fa4fba1965b80c7c4e02814a827750b02f886a3, and this PR in riak_test: https://github.com/basho/riak_test/pull/1198, the up/downgrade tests pass.

A rebase or merge is needed, though.

hmmr:
+1

I guess @andytill needs to approve, too.

paegun:
+1 b1970a2

I agree re: `lists:flatten` suggestion, but also agree that the local performance cost is far outweighed by the overarching performance improvement.

hmmr:
+1 b1970a2

hmmr:
@borshop merge

PR #1539, opened 2016-11-16
Use query buffers path for queries with any of OFFSET/LIMIT/ORDER BY RTS-1527

hmmr:
Now that the parser emits `?SQL_SELECT` records with `'GROUP BY' = []` (and not `undefined`) when the query has no ORDER BY clause but may have LIMIT and/or OFFSET clause, we need to check for all these other clauses as well in order to decide whether to send the query execution down query buffer
path.

This fixes the problem when a query with *no* ORDER BY clause but *with* a LIMIT clause, fails to effectively apply the LIMIT restriction.

### DIFF
--- a/src/riak_kv_qry.erl
+++ b/src/riak_kv_qry.erl
@@ -277,7 +277,9 @@ do_select(SQL, ?DDL{table = BucketType} = DDL) ->
                                        {ok, {new|existing, riak_kv_qry_buffers:qbuf_ref()} |
                                              undefined} |
                                        {error, any()}.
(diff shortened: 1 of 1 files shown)


andytill:
Verified that this change fixes the issue. It could do with an rt using the SQL query literal that showed the issue.

``` sql
SELECT * FROM docstab WHERE a = 'a' AND b > 0 AND b < 10 LIMIT 1;
```

andytill:
+1 f84703f

andytill:
Test added https://github.com/basho/riak_test/pull/1196, verified that is passed on this branch and fails on `riak_ts-develop`.

andytill:
@borshop merge

PR #1540, opened 2016-11-17
Add blob support

macintux:
The new `blob` type is effectively identical to `varchar` and is treated as such for most operations. They may diverge further as we look into proper utf-8 support for character strings, but that's a future discussion.

Since `blob` is basically `varchar` (and will be represented as such for downgrade purposes) rather than implement a new type, `riak_ql` now has the concept of "external" vs "internal" type representations. Externally a blob will be represented as a blob, while internally for storage it's a varchar.

Pull Requests:
* riak_kv https://github.com/basho/riak_kv/pull/1540
* riak_ql https://github.com/basho/riak_ql/pull/156
* riak_pb https://github.com/basho/riak_pb/pull/211
* riak_shell https://github.com/basho/riak_shell/pull/61
* riak_test https://github.com/basho/riak_test/pull/1212


src/riak_kv_ts_util.erl
```diff
@@ -534,28 +534,38 @@ key_element_to_string(V) -> lists:flatten(io_lib:format("~p", [V])).
 
 %%
 %% This one quotes values which should be quoted for assignment
-element_to_quoted_string(V) when is_binary(V) -> varchar_quotes(V);
-element_to_quoted_string(V) when is_float(V) -> mochinum:digits(V);
-element_to_quoted_string(V) -> lists:flatten(io_lib:format("~p", [V])).
+element_to_quoted_string(V, blob) -> hex_as_string(hexlify(V));
```
andytill: The `mochihex` module is used in riak_ql. I used that instead of rolling my own so I didn't have to add unit tests to prove correctness.

andytill:
@borshop retry

src/riak_kv_ts_util.erl
macintux: Fixed

macintux:
This can be retried once `pb` and `ql` are merged.

macintux:
@borshop: retry

macintux:
Tests weren't running but needed another merge with upstream anyway. Running through tests manually.

andytill:
+1 6601d25

andytill:
@borshop merge

PR #1544, opened 2016-11-21
Report TS errors of underlying atom type

paegun:
* added error reporting for overload and other forms of error conditions reported as atoms.

### DIFF
--- a/src/riak_kv_ts_svc.erl
+++ b/src/riak_kv_ts_svc.erl
@@ -516,6 +516,8 @@ sub_tsqueryreq(_Mod, DDL = ?DDL{table = Table}, SQL, State) ->
         {error, {too_many_subqueries, NQuanta, MaxQueryQuanta}} ->
             {reply, make_max_query_quanta_resp(NQuanta, MaxQueryQuanta), State};
 
+        {error, AReason} when is_atom(AReason) -> %%<< i.e. overload, primary_not_available
+            {reply, make_rpberrresp(?E_SUBMIT, atom_to_list(AReason)), State};
         {error, Reason} ->
             {reply, make_rpberrresp(?E_SUBMIT, Reason), State}
     end.
--- a/src/riak_kv_ts_svc.erl
+++ b/src/riak_kv_ts_svc.erl
@@ -516,10 +516,10 @@ sub_tsqueryreq(_Mod, DDL = ?DDL{table = Table}, SQL, State) ->
         {error, {too_many_subqueries, NQuanta, MaxQueryQuanta}} ->
             {reply, make_max_query_quanta_resp(NQuanta, MaxQueryQuanta), State};
 
-        {error, AReason} when is_atom(AReason) -> %%<< i.e. overload, primary_not_available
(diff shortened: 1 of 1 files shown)


hazen:
@paegun Do you have a case which triggered this?

paegun:
@javajolt RTS-1389 (#1418)

src/riak_kv_ts_svc.erl
```diff
@@ -516,6 +516,8 @@ sub_tsqueryreq(_Mod, DDL = ?DDL{table = Table}, SQL, State) ->
         {error, {too_many_subqueries, NQuanta, MaxQueryQuanta}} ->
             {reply, make_max_query_quanta_resp(NQuanta, MaxQueryQuanta), State};
 
+        {error, AReason} when is_atom(AReason) -> %%<< i.e. overload, primary_not_available
+            {reply, make_rpberrresp(?E_SUBMIT, atom_to_list(AReason)), State};
```
hazen: I see that other calls to `make_rpberrresp` seem to use `to_string/1` instead of `atom_to_list`. Maybe we should first test for `{error, Reason} when is_list(Reason) ->` as the first clause and convert anything else using `to_string/1` as the second clause.  That would catch any other reason, like a binary or record or whatever.

src/riak_kv_ts_svc.erl
```diff
@@ -555,7 +555,7 @@ check_table_and_call(Table, Fun, TsMessage, State) ->
 
 %%
 -spec make_rpberrresp(integer(), string()) -> #rpberrorresp{}.
-make_rpberrresp(Code, Message) ->
+make_rpberrresp(Code, Message) when is_list(Message) ->
```
hazen, line 558: Nice extra guard

src/riak_kv_ts_svc.erl
```diff
@@ -675,6 +675,8 @@ make_table_created_missing_resp(Table) ->
       ?E_CREATED_GHOST,
       flat_format("Table ~s has been created but found missing", [Table])).
 
+to_string(X) when is_atom(X) ->
```
hazen, line 678: Is this clause necessary? Seems like the original function would cover it.
```
lists:flatten(io_lib:format("~p",[some_atom])).
"some_atom"
```

paegun: Necessary, no, but performance-related, so IMHO such a simple optimization should be employed.

Yes, the result in terms of testable value is equivalent. `io_lib:format/2` needs to parse the format string which is surely more costly than simply calling `atom_to_list` when the type is known to be atom. At its core, io_lib:format/2 assumedly does an equivalent to atom_to_list/1, so all of the code between the pattern match on `is_atom/1` to `atom_to_list` is waste. Additionally, needing to unpack the result using `lists:flatten/1` is a waste. These wastes should be used for the general cases that we don't explicitly know about (whether we should have a general case here is questionable since this is an internal function where we can/should know every type that we are converting).

hazen:
:+1: 8636fe2

hazen:
@borshop merge

PR #1545, opened 2016-11-23
[RTS-1094] Check table readiness on decode

paegun:
Fixes riak-shell crash on attempt to INSERT INTO a table that does not exist (is not active).

* exposed get_table_status for use in query validation path.
* added check for table readiness (active, sufficient version) for query execution.
  * the check excludes query types that do not have a table, ie SHOW TABLES.

Was originally developed as https://github.com/basho/riak_kv/pull/1530 , but due to several factors this counter proposal for solution was offered (factors follow):
* long duration before the PR was reviewed, much underlying changed, chasing dialyzer errors, test failures due to missed pattern matches due to changes to function returns, ... pragmatism called for rethinking a rewrite of the fix.
* DRY, there should be one way to determine if a table is Active.

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -27,6 +27,7 @@
          delete_table_ddls/1,
          get_all_table_names/0,
          get_table_status_pairs/0,
+         get_table_status/1,
          get_compiled_ddl_versions/1,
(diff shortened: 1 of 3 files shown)


gordonguthrie:
+1 2db8f0f

PR #1546, opened 2016-11-24
append (not mix in) new fields in #riak_select_v3

hmmr:
RTS-1561.

Also set default value for `'OFFSET'` to `[]`.


hazen:
:+1: 554aa65

hazen:
@borshop merge

PR #1548, opened 2016-11-29
Ensure a table exists before asking for its DDL

hazen:
Add `riak_kv_ts_util:is_table_active_and_supported/2` to make sure the DDL helper module exists before going on to check to see if features are supported.

Sadly `get_table_ddl` had issues when calling it against a non-existent table, so I went with `code:ensure_loaded/1`:

```
in function ets:lookup/2
  called as lookup(metadata_manager_prefixes_ets,{core,bucket_types})
in call from riak_core_metadata_manager:ets_tab/1 (src/riak_core_metadata_manager.erl, line 617)
in call from riak_core_metadata_manager:read/1 (src/riak_core_metadata_manager.erl, line 585)
in call from riak_core_metadata:get/3 (src/riak_core_metadata.erl, line 97)
in call from riak_core_claimant:get_bucket_type/3 (src/riak_core_claimant.erl, line 203)
in call from riak_core_bucket:get_bucket/1 (src/riak_core_bucket.erl, line 123)
in call from riak_kv_ts_util:get_table_ddl/1 (src/riak_kv_ts_util.erl, line 303)
in call from riak_kv_ts_util:is_table_active_and_supported/2 (src/riak_kv_ts_util.erl, line 689)
```

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -211,15 +211,11 @@ get_table_ddl(_Table) ->
     DDL = {}, %% not used by caller
     {ok, Module, DDL}.
 
-is_table_supported(DDLRecCap, Table = <<"my_type2">>) ->
-    is_table_supported_not_active(DDLRecCap, Table);
-is_table_supported(DDLRecCap, Table) ->
-    is_table_supported_active(DDLRecCap, Table).
-
-is_table_supported_active(_DDLRecCap, _Table) ->
+is_table_supported(_DDLRecCap, <<"my_type2">>) ->
+    {error, "The table is not active"};
+is_table_supported(_DDLRecCap, _Table) ->
     true.
-is_table_supported_not_active(_DDLRecCap, _Table) ->
-    {error, "The table is not active"}.
+
 -endif.
 %% / Forwards/Mocks for getting table status, isolating the interaction for testability
 
--- a/src/riak_kv_ts_util.erl
+++ b/src/riak_kv_ts_util.erl
@@ -292,6 +292,8 @@ queried_table(
 queried_table(#riak_sql_show_create_table_v1{'SHOW_CREATE_TABLE' = Table}) ->
     Table.
 
+%% Create a simple, mockable version for unit testing, too
+-ifndef(TEST).
 -spec get_table_ddl(binary()) ->
                            {ok, module(), ?DDL{}} |
                            {error, term()}.
@@ -310,6 +312,10 @@ get_table_ddl(Table) when is_binary(Table) ->
                     {ok, Mod, DDL}
             end
     end.
+-else.
+get_table_ddl(_Table) ->
+    {ok, module, ?DDL{}}.
+-endif.
 
 
 %%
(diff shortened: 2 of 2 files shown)


src/riak_kv_compile_tab.erl
```diff
@@ -211,15 +211,11 @@ get_table_ddl(_Table) ->
     DDL = {}, %% not used by caller
     {ok, Module, DDL}.
 
-is_table_supported(DDLRecCap, Table = <<"my_type2">>) ->
-    is_table_supported_not_active(DDLRecCap, Table);
-is_table_supported(DDLRecCap, Table) ->
-    is_table_supported_active(DDLRecCap, Table).
-
-is_table_supported_active(_DDLRecCap, _Table) ->
```
macintux, line 219: Removed because unused?

src/riak_kv_ts_util.erl
```diff
@@ -669,17 +669,35 @@ check_table_feature_supported(DDLRecCap, DecodedReq) ->
...
             DecodedReq
     end.
 
 %%
+-spec is_table_active_and_supported(DDLRecCap::atom(),
+                                    Table::binary()) ->
+                                        boolean() | {error, string()}.
+is_table_active_and_supported(DDLRecCap, Table) ->
+    Mod = riak_ql_ddl:make_module_name(Table),
+    case code:ensure_loaded(Mod) of
```
macintux: `riak_kv_ts_util:get_table_ddl` works for me, at least with very simplistic testing. We should use this unless there's something more subtle I'm missing.

```(dev5@127.0.0.1)1> riak_kv_ts_util:get_table_ddl(<<"foo">>).
{error,no_type}
(dev5@127.0.0.1)2> riak_core_bucket_type:create(<<"foo">>, []).
ok
(dev5@127.0.0.1)3> riak_kv_ts_util:get_table_ddl(<<"foo">>).
{error,no_type}
(dev5@127.0.0.1)5> riak_core_bucket_type:activate(<<"foo">>).
ok
(dev5@127.0.0.1)6> riak_kv_ts_util:get_table_ddl(<<"foo">>).
{error,missing_helper_module}```

File: src/riak_kv_ts_util.erl [hunk 3]
```diff
@@ -669,17 +669,35 @@ check_table_feature_supported(DDLRecCap, DecodedReq) ->
                     {error, create_table_not_supported_message(Table)}
             end;
         {ok, _, {_, Table}} ->
-            case is_table_supported(DDLRecCap, Table) of
+            case is_table_active_and_supported(DDLRecCap, Table) of
```
macintux, line 676: I'm ok with this as a temporary measure pending a more thorough reworking, but for the record, treating `false` as `unsupported` and `{error, _}` as `inactive` for return values is an undesirable pattern for an `is` function.

src/riak_kv_compile_tab.erl
hazen: Yes, these functions essentially returned constants for testing and were only used in that context.

src/riak_kv_ts_util.erl
hazen: Happy to rename it and restructure it.  It's all how much we want to refactor in this PR.

src/riak_kv_ts_util.erl
hazen: The unit tests fail with the aforementioned output: 
```
in function ets:lookup/2
  called as lookup(metadata_manager_prefixes_ets,{core,bucket_types})
```
I'll look at those tests.  Entirely possible they don't set up a real bucket which is why they fail.

paegun: Yes, the bucket doesn't exist in testing. @andytill took a different approach in https://github.com/basho/riak_kv/pull/1530, passing a function to allow for mocking. I didn't find the approach complicated, but the mock provided IMHO an overly simplified bucket properties response. Down that rabbit hole, I decided that simplifying the mocking to ensure the full range of return types up a level achieved the same goal in a more straightforward manner. The point of the mock is to isolate the functionality under unit test to just the unit under development. The fact that a TS table creation results in these invariants is tested elsewhere. So, I suggest creating a function for the code added to check if the module is loaded `is_ddl_module_loaded/1` and mock it for testing.

macintux:
Ready to give this a +1 as soon as `riak_kv` is fixed so buildbot can make it happen.

macintux:
@borshop: retry

macintux:
I hereby bequeath @javajolt my +1 as soon as his rebase onto a functional kv passes testing.

PR #1549, opened 2016-11-29
remove allow_qbuf_reuse

hmmr:
Depends on https://github.com/basho/riak_pb/pull/212. Accompanying changes in riak_test are in https://github.com/basho/riak_test/pull/1213.

After consultations, the newly introduced flag `#tsqueryreq.allow_qbuf_reuse` (#1479) to set up query buffers for reuse by follow-up queries, appears to not sufficiently cover the circumstances such a reuse could be unequivocally understood by the user. Specifically, it does not prevent a client to accidentally reuse query buffers it did not create (such as when some client B happened to issue a similar query before, and it has not expired by the time client A issues its own, expecting it to fetch fresh data).

An adequate solution would involve generating unique query buffer id (not a hash of query parts, as before), which will be passed in `#tsqueryresp` and accepted in `#tsqueryreq`. The present PR also lays some groundwork for this change.

Compared to the use of query buffers with `allow_qbuf_reuse = false`, there is no change in query execution semantics. Compared to the (unsupported) practice of flipping the flag to true, some queries will take longer to execute as every query, follow-up or not, still gets a new query buffer set up and populated for it.

This should also ease the upgrade/downgrade chores as 1.4 clients will be able to communicate with 1.5.

Additionally, this PR has a separate commit (https://github.com/basho/riak_kv/commit/d60714c282cc548d6e4342b581281b4f9a3c33fa) to ensure all query buffer ldb instances are cleaned up on riak_kv application shutdown.

### DIFF
--- a/include/riak_kv_ts.hrl
+++ b/include/riak_kv_ts.hrl
@@ -103,7 +103,8 @@
           group_by = ?GROUP_BY_DEFAULT :: [{identifier, binary()}] | [{FieldPos::integer(), FieldName::binary()}],
           %% since v3
           'OFFSET'       = []   :: [riak_kv_qry_compiler:offset()],
-          %% to be supplied in #tsqueryreq{}
+          %% to be supplied in #tsqueryreq.qbuf_id, which is expected
+          %% to appear in a future release
           qbuf_id               :: undefined | binary()  %% control reuse of existing buffers
        }).
 
--- a/include/riak_kv_ts.hrl
+++ b/include/riak_kv_ts.hrl
@@ -103,7 +103,8 @@
           group_by = ?GROUP_BY_DEFAULT :: [{identifier, binary()}] | [{FieldPos::integer(), FieldName::binary()}],
           %% since v3
           'OFFSET'       = []   :: [riak_kv_qry_compiler:offset()],
-          allow_qbuf_reuse = false :: boolean()  %% control reuse of query buffers
+          %% to be supplied in #tsqueryreq{}
+          qbuf_id               :: undefined | binary()  %% control reuse of existing buffers
        }).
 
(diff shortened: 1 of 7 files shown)


hazen:
@hmmr I'll take a look at testing the changes so we can at least build again

include/riak_kv_ts.hrl
```diff
@@ -103,7 +103,8 @@
           group_by = ?GROUP_BY_DEFAULT :: [{identifier, binary()}] | [{FieldPos::integer(), FieldName::binary()}],
           %% since v3
           'OFFSET'       = []   :: [riak_kv_qry_compiler:offset()],
-          allow_qbuf_reuse = false :: boolean()  %% control reuse of query buffers
+          %% to be supplied in #tsqueryreq{}
```
macintux: Please update the comment to indicate this is a placeholder for a future release

src/riak_kv_qry_buffers.erl
```diff
@@ -700,34 +666,29 @@ compute_chunk_size(Data) ->
...
-                #qbuf{} ->
-                    {ok, {existing, QBufRef}};
-                false ->
-                    {ok, {new, QBufRef}}
-            end;
-        {error, _} = ErrorReason ->
-            ErrorReason
-    end.
+get_qref(_SQL, _QBufs) ->
+    AlwaysUniqueId = crypto:rand_bytes(8),
```
macintux: `crypto:rand_bytes/1` is deprecated. Maybe we don't care, but if this is going to be a long-term solution I'd suggest `crypto:strong_rand_bytes/1` instead.

macintux:
:+1: e4cea37

macintux:
yolomerging

PR #1550, opened 2016-11-29
expose riak_kv.query.timeseries.max_running_fsms

hmmr:
This is to make configurable the max number of index FSM spawned to serve TS queries.

For @erikleitch to review.

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -814,7 +814,7 @@
 %% @doc Throttle the number of simultaneously running FSMs.  This number times the size of data chunks
 %% determines the max volume of data the coordinator will have to deal with.
 {mapping, "riak_kv.query.timeseries.max_running_fsms", "riak_kv.timeseries_query_max_running_fsms", [
-  {default, 5},
+  {default, 20},
   {datatype, integer},
   {validators, ["validate_timeseries_query_max_running_fsms"]}
 ]}.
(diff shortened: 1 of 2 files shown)


erikleitch:
+1 

Code changes look good.  

We should decide whether we want to change the default value.  Personally, I don't see a compelling reason _not_ to.  On the client side, if the qry worker is tied up processing data from the current query, it will simply not be available for another query, and incoming queries will be queued until the max queue depth is reached, and an overload error is thrown.  On the server side, based on my quick reading of the code (maybe incorrect?), we already have a pressure release valve encoded as the max_query_data_size, which is # fsms x size of data chunks.  So if the subqueries' data exceeds this limit, we already have a mechanism for cancelling overlarge queries.  (This seems like a better gauge of when a query is too large anyway -- not the # of subqueries)  And in the use-cases we've seen so far, customers are far more interested in infrequent large queries returning quickly than they are about getting lots of queries pushed through the system.

Code doesn't build because tsqueryreq allow_qbuf_reuse has been removed, but is still referenced in this branch.  That'll have to be fixed before we merge.

erikleitch:
+1 bad3ba9

hazen:
@borshop merge

javajolt:
@hmmr I'm having some test failures seemingly related to this merge, like `ts_simple_select`. I'll investigate more.

PR #1552, opened 2016-11-30
Added SQL DELETE for single keys

gordonguthrie:
Change adds:
* SQL statement lexer/parser
* compilation validation
* use of new DDL helper modules


include/riak_kv_ts.hrl
```diff
@@ -134,4 +134,10 @@
             'EXPLAIN' = ?SQL_SELECT{} :: ?SQL_SELECT{}
         }).
 
+-record(riak_sql_delete_query_v1,
+        {
+          'FROM'     = <<>> :: binary() | {list, [binary()]} | {regex, list()},
+          'WHERE'    = []   :: [riak_ql_ddl:filter()],
```
hmmr: `'WHERE'` _must not_ be an empty list here, no? This applies to `#riak_sql_select_*` records as well, even though it may be conceivably relaxed for those in future. But it still definitely is mandatory for `#riak_sql_delete_query`.

include/riak_kv_ts.hrl
```diff
@@ -134,4 +134,10 @@
             'EXPLAIN' = ?SQL_SELECT{} :: ?SQL_SELECT{}
         }).
 
+-record(riak_sql_delete_query_v1,
```
hmmr, line 139: To avoid an (arguably unlikely) round of introducing `SQL_DELETE` macro for the current-version record when we happen to need a `riak_sql_delete_v2`, can it be done now, in this PR?

src/riak_kv_qry.erl
```diff
@@ -370,6 +375,26 @@ do_explain(DDL, Select) ->
...
+%%
+%% Single Key DELETE statement
+%%
+
+do_delete(#riak_sql_delete_query_v1{'FROM'     = F,
+                                    'WHERE'    = W,
+                                    helper_mod = Mod}) ->
+    case Mod:get_delete_key(W) of
+        {ok, Key} -> 
+            case riak_kv_ts_api:delete_data(Key, F) of
```
hmmr, line 387: Commendable use of a `riak_kv_ts_api:function`.

src/riak_kv_ts_api.erl
```diff
@@ -78,6 +79,8 @@ api_call_to_perm(query_explain) ->
     "riak_ts.query_explain";
 api_call_to_perm(describe_table) ->
     "riak_ts.describe_table";
+api_call_to_perm(query_delete) ->
+    "riak_ts.query_delete";
```
hmmr: I would argue that this should converge onto "riak_ts.delete", as it is done for `query_insert`.

src/riak_kv_ts_svc.erl
```diff
@@ -30,34 +30,36 @@
...
+-define(E_ACTIVATE,              1017).
+-define(E_BAD_QUERY,             1018).
+-define(E_TABLE_INACTIVE,        1019).
+-define(E_PARSE_ERROR,           1020).
+-define(E_NOTFOUND,              1021).
 -define(E_SELECT_RESULT_TOO_BIG, 1022).
 -define(E_QBUF_CREATE_ERROR,     1023).
 -define(E_QBUF_LDB_ERROR,        1024).
 -define(E_QUANTA_LIMIT,          1025).
+-define(E_INVALID_DELETE,        1026).
```
hmmr: Application growing pains. We now have `E_DELETE` for the KV-style delete vs `E_INVALID_DELETE` for the TS-style delete. Not that I can suggest better names..

hmmr (changes requested):
Single-key delete should have the same ACL atom as DELETE query.

And, the issue with validation/diagnostics of DELETE queries with incomplete set of values.

src/riak_kv_ts_api.erl
gordonguthrie: Done

src/riak_kv_qry.erl
gordonguthrie: Merci ;-)

include/riak_kv_ts.hrl
gordonguthrie: Ooh, you dialyzer whisperer you, good call...

Done

include/riak_kv_ts.hrl
gordonguthrie: The style, on inspecting the code is to introduce the Macro on the first update, so I will go with that for now, if that's OK?

src/riak_kv_ts_svc.erl
gordonguthrie: Lets go with E_DELETE - same error, one name

hmmr (approved):

hmmr:
+1

paegun:
+1 07a283e

PR #1558, opened 2016-12-03
Fix for ts_simple_recompile_ddl regression

paegun:
* for fix for ts_simple_recompile_ddl regression, exposed insert_previous to encapsulate how previous compile state is expected/stored.

### DIFF
--- a/src/riak_kv_compile_tab.erl
+++ b/src/riak_kv_compile_tab.erl
@@ -34,6 +34,7 @@
          new/1,
          populate_v3_table/0
         ]).
+-export([insert_previous/3]).
 
 -include_lib("riak_pb/include/riak_ts_pb.hrl").
 -include_lib("riak_ql/include/riak_ql_ddl.hrl").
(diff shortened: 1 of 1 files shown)


src/riak_kv_compile_tab.erl
```diff
@@ -112,11 +113,24 @@ insert(BucketType, DDL) when is_binary(BucketType), is_tuple(DDL) ->
...
     CompileState = compiling,
+    insert_previous(BucketType, DDL, CompileState).
+
+convert_ddl_v2_to_ddl_v1(#ddl_v2{} = DDL) ->
+    #ddl_v1{
+       table = DDL#ddl_v2.table,
+       fields = DDL#ddl_v2.fields,
+       partition_key = DDL#ddl_v2.partition_key,
+       local_key = DDL#ddl_v2.local_key
+      }.
```
andytill: Not all tables can be downgraded and be used by TS 1.4, for example tables using the BLOB type or descending keys. If the DDL conversion is used from riak_ql_ddl then this is checked as part of the conversion. If the table cannot be downgraded then it will not be usable on this node. The `riak_ql_ddl_rec_version` capability stop it being used on 1.5 nodes too.

paegun:
@andytill much appreciated on the call to convert ddl using riak_ql.

davidx:
thumbot retry

davidx:
thumbot retry

hazen:
:+1:

andytill:
@paegun you're welcome :)

PR #1560, opened 2016-12-03
use Mod:revert_ordering_on_local_key in {get,delete}_data for DESC keys

hmmr:
Depends on https://github.com/basho/riak_ql/pull/162.

Single-key primitive operations in `riak_kv_ts_api` need to provide to `riak_client:get` and `:delete` keys adjusted for fields declared DESC in the DDL, as is done for list_keys in https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_qry_worker.erl#L190.

This fixes regressions currently seen in `ts_simple_single_key_ops` (and possibly others).

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -187,10 +187,8 @@ make_key_conversion_fun(Table) ->
     Mod = riak_ql_ddl:make_module_name(Table),
     DDL = Mod:get_ddl(),
     fun(Key) when is_binary(Key) ->
-            {ok, DescAdjustedKey} =
-                riak_ql_ddl:negate_if_desc(
-                  tuple_to_list(sext:decode(Key)), Mod, DDL),
-            {ok, PK} = riak_ql_ddl:lk_to_pk(DescAdjustedKey, Mod, DDL),
+            {ok, PK} = riak_ql_ddl:lk_to_pk(
+                         Mod:revert_ordering_on_local_key(sext:decode(Key)), Mod, DDL),
             list_to_tuple(PK);
        (Key) ->
             lager:error("Key conversion function "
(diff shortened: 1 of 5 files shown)


davidx:
thumbot retry

davidx:
thumbot retry

src/riak_kv_qry_worker.erl
```diff
@@ -187,10 +187,8 @@ make_key_conversion_fun(Table) ->
     Mod = riak_ql_ddl:make_module_name(Table),
     DDL = Mod:get_ddl(),
     fun(Key) when is_binary(Key) ->
-            {ok, DescAdjustedKey} =
-                riak_ql_ddl:negate_if_desc(
-                  tuple_to_list(sext:decode(Key)), Mod, DDL),
-            {ok, PK} = riak_ql_ddl:lk_to_pk(DescAdjustedKey, Mod, DDL),
+            {ok, PK} = riak_ql_ddl:lk_to_pk(
```
macintux: Where will the `bad_compound_key` exception be handled in this case?

hmmr: Over there: https://github.com/basho/riak_kv/pull/1560/commits/b6f142e66433d844fc01d3e982546fa2bca1c63f#diff-47d6b445c0ee52daa420d1d48d98f394R638

macintux: This function is passed from a `riak_kv_index_fsm` process to a `riak_kv_vnode` to be invoked in a fold. I don't see any exception handling in a quick scan of that code path, which leads me to believe that whatever provokes that exception could bring down a vnode.

hmmr: In contrast to its use in `riak_kv_ts_api:{get,delete}_data` where the key comes from user input, the folding functions here will have the key that's guaranteed to be valid (it is read directly from the backend).

macintux: Ok. Please add a comment here, though, so someone in the future doesn't have the same panic attack.

hazen:
:+1:

andytill:
@thumbot retry

andytill:
+1

PR #1562, opened 2016-12-06
Merge riak_kv sweeper code to develop

nickelization:
There are still some things left to do before we release this code, but it's solid enough to merge it to develop.

The original sweeper code was written by a contractor, Mikael Lixenstrand. His development work still exist in various repos under branches name 'object_ttl'. This builds off of his work, adding lots of additional test coverage, fixing a bunch of bugs, and generally improving the quality of the code.

This PR depends on three other pull requests being merged first:

- [x] https://github.com/basho/riak_repl/pull/765
- [x] https://github.com/basho/riak_pb/pull/216
- [x] https://github.com/basho/riak_core/pull/882


src/riak_kv_delete.erl
```diff
@@ -189,11 +199,61 @@ get_w_options(Bucket, Options) ->
...
+
+obj_outside_grace_period(RObj) ->
+    obj_outside_grace_period(RObj, app_helper:get_env(riak_kv, tombstone_grace_period)).
+obj_outside_grace_period(_RObj, false) ->
+	false;
+obj_outside_grace_period(RObj, TombstoneGracePeriod) ->
+	Now = os:timestamp(),
+
+	Contents = riak_object:get_contents(RObj),
+	case [{M, V} || {M, V} <- Contents,
```
lehoff: Running through the entire list should only be done when all the content is in_grace.

Leverage `lists:any/2` by writing something like this:

```erlang
obj_outside_grace_period(RObj, TombstoneGracePeriod) ->
    Now = os:timestamp(),

    Contents = riak_object:get_contents(RObj),
    list:any( fun(Content) -> not content_in_grace(Content, Now, TombstoneGracePeriod) end,
                 Contents).

content_in_grace({Metadata, Value}, Now, TombstoneGracePeriod) ->
    in_grace(Metadata, Now , TombstoneGracePeriod).
```

This way the code reads exactly what the logic is and it will stop as soon as it finds anything outside the grace period.

src/riak_kv_entropy_manager.erl
```diff
@@ -507,47 +537,64 @@ add_hashtree_pid(true, Index, Pid, State=#state{trees=Trees, trees_version=VTree
...
+    case NumLocks >= Limit of
+        true ->
+            build_limit_reached;
+        false ->
+            ok
+    end;
+check_lock_type(_Type, _Locks) ->
+    ok.
+
+get_nr_of_locks(Type, Locks) ->
```
lehoff, line 568: Unfortunately, we don't have Elixir's `Enum.count/2` function available, which would fit right in.

Consider not building the list only to measure its length:
```erlang
get_nr_of_locks(Type, Locks) ->
    lists:foldl( fun(#lock{type=T}, Count) when Type==T -> Count+1; 
                            (_, Count) -> Count end,
                     0,
                     Locks ).
```

There ought to be a `lists:count/2` function in OTP, but, alas, no such luck.

src/riak_kv_get_core.erl
```diff
@@ -195,21 +195,28 @@ final_action(GetCore = #getcore{n = N, merged = Merged0, results = Results,
...
-                          %% dominated (in need of repair) indexes and
-                          %% the list of not_found (in need of repair)
-                          %% indexes.
-                          [{Idx, outofdate} || {Idx, {ok, RObj}} <- Results,
-                                  riak_object:strict_descendant(MObj, RObj)] ++
-                              [{Idx, notfound} || {Idx, {error, notfound}} <- Results]
-                  end,
+
+    ReadRepairs =
+        case ObjState of
```
lehoff: This case should be its own function, e.g., `object_read_repairs(ObjState, MObj)`.

src/riak_kv_get_core.erl
```diff
@@ -195,21 +195,28 @@ final_action(GetCore = #getcore{n = N, merged = Merged0, results = Results,
...
+                %% The user need to use delete_mode keep to
+                %% use the sweeper as reaper
+                case riak_kv_delete:obj_outside_grace_period(MObj) of
+                    false ->
+                        get_read_repairs(Results, MObj);
+                    true ->
+                        []
+                end
+        end,
     Action = case ReadRepairs of
```
lehoff: This case should also be its own function.
And the `length([xx || {_Idx, {ok, _RObj}} <- Results])` should be a function that uses the non-existant `lists:count/2` function (we might want to add that function in a util module).
But, regardless, a function called `count_sucessful_results(Results)` should be introduced, no matter how it is implemented.

src/riak_kv_get_core.erl
```diff
@@ -233,6 +240,18 @@ final_action(GetCore = #getcore{n = N, merged = Merged0, results = Results,
...
     {Action, GetCore#getcore{merged = Merged}}.
 
+%% Any object that is strictly descended by
+%% the merge result must be read-repaired,
+%% this ensures even tombstones get repaired
+%% so reap will work. We join the list of
+%% dominated (in need of repair) indexes and
+%% the list of not_found (in need of repair)
+%% indexes.
+get_read_repairs(Results, MObj) ->
```
lehoff, line 257: Instead of iterating twice over the `Results` list, consider using `lists:filtermap/2` along these lines:

```erlang
get_read_repairs(Results, MObj) ->
    lists:filtermap( fun(Res) -> index_read_repair(Res, MObj) end, Results ).

index_read_repair({Idx, {ok, RObj}}, MObj) ->
    case riak_object:strict_descendant(MObj, RObj) of
         true ->
            {Idx, outofdate}
        false ->
            false
   end;
index_read_repair({Idx, {error,notfound}}, _MObj) ->
    {Idx, notfound}.
```

src/riak_kv_object_ttl.erl
```diff
@@ -0,0 +1,60 @@
...
+-export([participate_in_sweep/2,
+         successful_sweep/2,
+         failed_sweep/2]).
+
+%% riak_kv_sweeper callbacks
+participate_in_sweep(_Index, _Pid) ->
+    InitialAcc = 0,
+    {ok, ttl_fun(), InitialAcc}.
+
+ttl_fun() ->
```
lehoff: This is a level of unnecessary indirection and it blurs the fact that the function that the `participate_in_sweep/2` has to return has arity 3. 

Change to something like this:
```erlang
participate_in_sweep(_Index, _Pid) ->
    InitialAcc = 0,
    {ok, ttl_fun/3, InitialAcc}.

ttl_fun({{_Bucket, _Key}, RObj}, Acc, [{bucket_props, BucketProps}]) ->
        case riak_kv_util:is_x_expired(RObj, BucketProps) of
            true ->
                 {deleted, Acc};
            _ ->
                {ok, Acc}
        end.
```

Furthermore, this also gives the opportunity to add a spec on the ttl_fun/3, if one so desires.

src/riak_kv_sweeper.erl
```diff
@@ -0,0 +1,369 @@
...
+
+handle_cast(_Msg, State) ->
+    {noreply, State}.
+
+handle_info({test_tick, Ref, From}, State) ->
+    {noreply, State1} = handle_info(sweep_tick, State),
+    From ! {ok, Ref},
+    {noreply, State1};
+
+handle_info(sweep_tick, State) ->
```
lehoff: It is a good idea to remove internal timers as much as possible to make testing easier.
See https://github.com/lehoff/chronos for a deeper explanation.

In this case adding a `trigger_sweep/0` function to the API will remove the need for `handle_info({test_tick...`.

And `schedule_sweep_tick/0` should be changed to use `timer:apply_after/4` or create a module to handle this, alternatively use `chronos`.

src/riak_kv_sweeper.erl
```diff
@@ -0,0 +1,369 @@
...
+    EstimatedNrKeys;
+maybe_estimate_keys(_Index, false, _) ->
+    false.
+
+-spec get_estimtate(riak_kv_sweeper:index()) ->
+    EstimatedKeys :: non_neg_integer().
+get_estimtate(Index) ->
+    case riak_kv_index_hashtree:get_lock(Index, estimate) of
+        ok ->
+            Result = case riak_kv_index_hashtree:estimate_keys(Index) of
```
lehoff: Put the case in a function, e.g, `hashtree_estimate_keys(Index)`.

src/riak_kv_sweeper_cli.erl
```diff
@@ -0,0 +1,180 @@
...
+    SweepsTable = format_sweeps(Sweeps, IndexedParticipants),
+    ActiveSweeps = format_active_sweeps(Sweeps, IndexedParticipants),
+    ParticipantsTable ++ SweepsTable ++ ActiveSweeps.
+
+format_sweeps(Sweeps, IndexedParticipants) ->
+    Now = os:timestamp(),
+    Header = io_lib:format("~s", ["Sweep results:"]),
+    SortedSweeps = lists:keysort(#sweep.index, Sweeps),
+    Rows =
+        [begin
```
lehoff: Make the `begin ... end` a separate function.

src/riak_kv_sweeper_cli.erl
```diff
@@ -0,0 +1,180 @@
...
+            [clique_status:text(Header),
+             clique_status:table(Rows)]
+    end.
+
+format_progress(#sweep{index = Index,
+                       active_participants = AcitvePart,
+                       estimated_keys = {EstimatedKeys, _TS},
+                       swept_keys = SweptKeys},
+                IndexedParticipants) ->
+    case EstimatedKeys of
```
lehoff: Create a helper function to extract the estimated keys without a case:
```erlang
format_progress(#sweep{index = Index,
                       active_participants = AcitvePart,
                       estimated_keys = {EstimatedKeys, _TS},
                       swept_keys = SweptKeys},
                IndexedParticipants) ->
   EstimatedKeysFormat = format_estimated_keys(EstimatedKeys),
   [{"Index", Index},
    {"Swept keys", SweptKeys}] ++
   EstimatedKeysFormat ++
   format_active_participants(AcitvePart, IndexedParticipants);

format_progress(_, _) ->
    "".

format_estimated_keys(EstimatedKeys) when is_integer(EstimatedKeys)
                                                                             andalso EstimatedKeys > 0 ->
    [{"Estimated Keys", EstimatedKeys}];
format_estimated_keys(_) ->
    [].
```

src/riak_kv_sweeper_cli.erl
```diff
@@ -0,0 +1,180 @@
...
+          lists:keymember(Mod, #sweep_participant.module, AcitvePart)}
+     end
+     || {Index, #sweep_participant{module = Mod}} <- IndexedParticipants].
+
+format_results(Now, Results, IndexedParticipants) ->
+    lists:flatten([get_result(Index, Participant, Results, Now)
+                     || {Index, Participant} <- IndexedParticipants]).
+
+get_result(Index, Participant, Results, Now) ->
+    case dict:find(Participant#sweep_participant.module, Results) of
```
lehoff: Do not export variables from the case statement.

```erlang
{LastRun, OutcomeString} =
    case ...
```

include/riak_kv_sweeper.hrl
```diff
@@ -0,0 +1,65 @@
...
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+
+%% Used as configuration for sweep participants.
+-record(sweep_participant,
```
lehoff: `#sweep_participant{}` should be made into its own module with suitable
  accesssors/setters  or be defined in `riak_kv_sweeper_fold`

include/riak_kv_sweeper.hrl
```diff
@@ -0,0 +1,65 @@
...
+         fun_type :: riak_kv_sweeper:fun_type(), %% delete_fun | modify_fun | observe_fun
+         sweep_fun :: fun(),        %%
+         run_interval :: integer() | fun(), %% Defines how often participant wants to run.
+         acc :: any(),
+         options = [] :: list(),    %% optional values that will be added during sweep
+         errors = 0 :: integer(),
+         fail_reason
+        }).
+
+-record(sweep,
```
lehoff: `#sweep{}` should be made local to the `riak_kv_sweeper_state` module as it is mostly
  used there and it should be exported as `riak_kv_sweeper_state.t()`.

include/riak_kv_sweeper.hrl
```diff
@@ -0,0 +1,65 @@
...
+         active_participants,  %% Active in current run
+         start_time :: erlang:timestamp(),
+         end_time :: erlang:timestamp(),
+         queue_time :: erlang:timestamp(),
+         estimated_keys :: {non_neg_integer(), erlang:timestamp()},
+         swept_keys :: non_neg_integer() | undefined
+        }).
+
+%% Sweep accumulator
+-record(sa,
```
lehoff: `#sa{}` should be made local to the `riak_kv_sweeper_fold` module as it is only used there.

src/riak_kv_stat.erl
```diff
@@ -184,6 +184,12 @@ do_update({vnode_put, Idx, USecs}) ->
     ok = exometer:update([P, ?APP, vnode, puts], 1),
     ok = create_or_update([P, ?APP, vnode, puts, time], USecs, histogram),
     do_per_index(puts, Idx, USecs);
+do_update({reap_tombstone, _Index}) ->
+    P = ?PFX,
```
lehoff: Inline all the `?PFX` - no need to define `P` like that just to save a few letters of typing.

src/riak_kv_sweeper.erl
```diff
@@ -0,0 +1,369 @@
...
+                           FinalAcc :: sweep_fun_acc()) -> _.
+
+-callback failed_sweep(Index :: index(), Reason :: term()) -> _.
+
+%% ====================================================================
+%% API functions
+%% ====================================================================
+-spec start_link() -> {ok, pid()}.
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
```
lehoff: Best to create a `?SERVER` macro (that may call ?MODULE in its definition).
`?MODULE` is for designating callbacks and debug things.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+fun_type_rank(delete_fun) -> 1;
+fun_type_rank(modify_fun) -> 2;
+fun_type_rank(observe_fun) -> 3.
+
+-spec inform_participants(#sa{}, riak_kv_sweeper:index()) -> [].
+inform_participants(#sa{active_p = Succ, failed_p = Failed}, Index) ->
+    successful_sweep(Succ, Index),
+    failed_sweep(Failed, Index).
+
+-spec successful_sweep([#sweep_participant{}], riak_kv_sweeper:index()) -> [].
```
lehoff: Change the spec so that the return type is `ok` and use `lists:foreach/2` as the list is not used for anything.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+successful_sweep(Succ, Index) ->
+    [Module:successful_sweep(Index, FinalAcc) ||
+       #sweep_participant{module = Module, acc = FinalAcc} <- Succ].
+
+-spec failed_sweep([#sweep_participant{}], riak_kv_sweeper:index()) -> [].
```
lehoff: As above: Change the spec so that the return type is ok and use lists:foreach/2 as the list is not used for anything.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+failed_sweep(Failed, Index) ->
+    [Module:failed_sweep(Index, Reason) ||
+       #sweep_participant{module = Module, fail_reason = Reason } <- Failed].
+
+failed_sweep(Failed, Index, Reason) ->
+    [Module:failed_sweep(Index, Reason) ||
```
lehoff: Use `lists:apply/2`.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+     format_result(succ, Succ) ++ format_result(fail, Failed)}.
+
+-spec format_result('succ'|'fail', [#sweep_participant{}]) ->
+    [{riak_kv_sweeper:participant_module(), 'succ' | 'fail'}].
+format_result(SuccFail, Results) ->
+    [{Module, SuccFail} || #sweep_participant{module = Module} <- Results].
+
+%% Throttle depending on swept keys.
+-spec maybe_throttle_sweep(binary(), #sa{}) -> #sa{}.
+maybe_throttle_sweep(_RObjBin, #sa{throttle = {pace, _, _}} = Acc0) ->
```
lehoff: Move the case into the function clause.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+    end;
+
+%% Throttle depending on total obj_size.
+maybe_throttle_sweep(RObjBin, #sa{throttle = {obj_size, _, _}} = Acc0) ->
+    #sa{throttle = {obj_size, Limit, Wait},
+        total_obj_size = TotalObjSize,
+        throttle_total_wait_msecs = TotalWait} = Acc0,
+    ObjSize = byte_size(RObjBin),
+    TotalObjSize1 = ObjSize + TotalObjSize,
+    case (Limit =/= 0) andalso (TotalObjSize1 > Limit) of
```
lehoff: Move the `Limit` into the function clause since it will point to the false case faster.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+                           throttle_total_wait_msecs = TotalWait + Wait},
+            timer:sleep(Wait),
+            maybe_extra_throttle(Acc1#sa{total_obj_size = 0});
+        _ ->
+            maybe_extra_throttle(Acc0#sa{total_obj_size = TotalObjSize1})
+    end.
+
+%% Throttle depending on how many objects the sweep modify.
+-spec maybe_extra_throttle(#sa{}) -> #sa{}.
+maybe_extra_throttle(Acc) ->
```
lehoff, line 279: Extract `TotalWait` in the function clause and create a helper function to calculate `Limit` and `Wait`.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+    end.
+
+%% Throttle depending on how many objects the sweep modify.
+-spec get_sweep_throttle() -> {'obj_size'|'pace',
+                               Limit :: non_neg_integer(),
+                               Sleep :: non_neg_integer()}.
+get_sweep_throttle() ->
+    {Type, Limit, Sleep} =
+        app_helper:get_env(riak_kv, sweep_throttle, ?DEFAULT_SWEEP_THROTTLE),
+    case riak_kv_sweeper:in_sweep_window() of
```
lehoff, line 309: Create an `adjust_sleep(Sleep)` function and simplify this to:

```erlang
get_sweep_throttle() ->
    {Type, Limit, Sleep} =
        app_helper:get_env(riak_kv, sweep_throttle, ?DEFAULT_SWEEP_THROTTLE),
    AdjustedSleep = adjust_sleep(Sleep),
    {Type, Limit, Sleep}.

adjust_sleep(Sleep) ->
    case riak_kv_sweeper:in_sweep_window() of
        true ->
            Div = app_helper:get_env(riak_kv, sweep_window_throttle_div,
                                     ?SWEEP_WINDOW_THROTTLE_DIV),
           Sleep div Div;
        false ->
            Sleep
    end.
```

Maybe make the result of `riak_kv_sweeper:in_sweep_window()` the 2nd argument to `adjust_sleep`.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+
+send_to_sweep_worker(Msg, #sweep{index = Index}) ->
+    lager:info("no pid ~p to ~p " , [Msg, Index]),
+    no_pid.
+
+-spec maybe_receive_request(#sa{}) -> #sa{}.
+maybe_receive_request(#sa{active_p = Active, failed_p = Fail } = Acc) ->
+    receive
+        stop ->
+            Active1 =
```
lehoff: Make a function for this. Say, `set_fail_reason_for_participants(Active, sweep_stop)`

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+                [ActiveSP#sweep_participant{fail_reason = sweep_stop } ||
+                 #sweep_participant{} = ActiveSP <- Active],
+            Acc#sa{active_p = [],  failed_p = Active1 ++ Fail};
+        {disable, Module} ->
+            case lists:keytake(Module, #sweep_participant.module, Active) of
```
lehoff: Make this a function, e.g., `maybe_disable_module`.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+-spec run_participant_fun(#sweep_participant{},
+                          riak_kv_sweeper:index(),
+                          {riak_object:bucket(), riak_object:key()},
+                          {riak_object:riak_object(),
+                           [#sweep_participant{}],
+                           [#sweep_participant{}],
+                           NumMutated :: non_neg_integer(),
+                           NumDeleted :: non_neg_integer(),
+                           BucketProps :: []}) ->
+    {riak_object:riak_object(),
```
lehoff: Create a local record for this accumulator. It is too complicated to have a 6-ary tuple.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+        true ->
+            {BucketProps, BucketPropsDict1} = get_bucket_props(Bucket, BucketPropsDict),
+            {[{bucket_props, BucketProps}], BucketPropsDict1};
+        false ->
+            {[], BucketPropsDict}
+    end.
+
+-spec get_bucket_props(riak_object:bucket(),
+                       BucketProps :: dict()) ->
+    {BucketProps :: [], BucketPropsDict :: dict()}.
```
lehoff: Do we have a more telling type for `BucketProps`?  Just a list is rather vague.

src/riak_kv_sweeper_fold.erl
```diff
@@ -0,0 +1,641 @@
...
+
+    Acc#sa{stat_mutated_objects_counter = StatNumMutated + NumMutated,
+           stat_deleted_objects_counter = StatNumDeleted + NumDeleted,
+           stat_obj_size_counter = StatObjSizeCounter + RObjBinSize,
+           stat_swept_keys_counter = StatSweptKeysCounter + 1}.
+
+-spec stat_send(riak_kv_sweeper:index(),
+                {'final'| 'in_progress', #sa{}}) -> #sa{}.
+stat_send(Index, {final, Acc}) ->
+    [ riak_kv_stat:update({sweeper, Index, failed, Module, 1})
```
lehoff: Use `lists:foreach/2` here - the lists are not used.

lehoff (changes requested):
Mostly structural issues. Should be easy to fix.

nickelization:
Thanks for all the review comments, @lehoff! Will address them today or tomorrow once I finish debugging some problems with overload that I dug up yesterday.

src/riak_kv_delete.erl
nickelization: Good catch. Fixed! (Although I used `lists:all` instead of `lists:any` since I believe we only want to reap the object if all the tombstones are outside the grace period, not just one.)

src/riak_kv_entropy_manager.erl
nickelization: This feels like a premature optimization to me. This function shouldn't be called that often (or on a critical path), and I do feel that the simple one-line implementation we have now is much clearer and more readable than a fold would be.

If we had lists:count I would gladly use that, though. Perhaps we should write our own implementation of it and add it to a utility module somewhere. Thoughts?

src/riak_kv_get_core.erl
nickelization: Agreed! Done.

src/riak_kv_entropy_manager.erl
lehoff: Having our own `lists:count/2` would be preferred.

nickelization: `riak_core_util:count/2` now merged - see basho/riak_core#888. Will update sweeper code to make use of it tomorrow.

nickelization: Done.

src/riak_kv_get_core.erl
nickelization: Agreed, and now done using the new `riak_core_util:count/2` func.

src/riak_kv_get_core.erl
nickelization: Good call. Done.

src/riak_kv_object_ttl.erl
nickelization: Nice catch. Fixed.

include/riak_kv_sweeper.hrl
nickelization: So a couple weeks ago, I already tried to put it into its own module...but my first attempt really made the code a lot uglier and less readable, because then we can no longer pattern match on record fields in function heads. In order to fix this in a nice clean way, we're going to have to rewrite a pretty substantial amount of code.

If I had written this code, I would not have used shared records all over the place, but I am questioning whether it is worth the effort and trouble to rewrite the sweeper at this point, given that it works and we are under a deadline crunch. I can do it, but it will not be a trivial change.

Also consider that these are not records like e.g. riak_object, where they are passed around all over the system. All these shared records are mostly just used by a few tightly-coupled sweeper modules. They are also not passed across node boundaries via RPC, so there should be no risk of upgrade problems involving mixed clusters with differing record definitions.

src/riak_kv_sweeper.erl
nickelization: Done.

src/riak_kv_sweeper.erl
nickelization: Just pushed some commits in which I switched the sweeper to use `chronos`, and updated all the CT tests to manually execute sweep ticks instead of waiting for the sweeper to tick itself.

itest/riak_kv_sweeper_SUITE.erl
```diff
@@ -433,19 +438,13 @@ scheduler_remove_participant_test(Config) ->
...
 
+    tick_per_index(Indices),
     receive {From, WaitIndex} ->
             riak_kv_sweeper:remove_sweep_participant(sweep_observer_1),
             From ! {TestCasePid, continue}
     end,
 
-    ok = receive_msg({ok, failed_sweep, sweep_observer_1, WaitIndex}, min_scheduler_response_time_msecs()),
-    #sweep{results = Result} = get_sweep_on_index(WaitIndex),
-    {ok,{_, fail}} = dict:find(sweep_observer_1, Result),
```
raghavkarol: @nickelization - Think we shouldn't remove the `{ok, {_, fail}}` match. The intentions is to capture the return and document the return value. It can actually be made stricter:  `{ok,{{_, _, _}, fail}}`.

The matching makes this test case unstable, but I think that can be addressed with doing a `match_retry(...)` instead.

itest/riak_kv_sweeper_SUITE.erl
```diff
@@ -639,34 +662,17 @@ scheduler_estimated_keys_lock_fail_test(Config) ->
...
-    [ok = receive_msg({ok, successful_sweep, sweep_observer_1, I}, min_scheduler_response_time_msecs()) || I <- Indices],
+
+    sweep_all_indices(Indices),
+
     [true = meck:called(riak_kv_index_hashtree, get_lock, [I, estimate]) || I <- Indices],
     [false = meck:called(riak_kv_index_hashtree, estimate_keys, [I]) || I <- Indices],
     [false = meck:called(riak_kv_index_hashtree, release_lock, [I]) || I <- Indices],
     ok.
 
-scheduler_sweep_concurrency_test(_Config) ->
```
raghavkarol: @nickelization - not sure why you've decided to kill this testcase? Is running concurrent sweeps being tested elsewhere that I'm missing?

include/riak_kv_sweeper.hrl
nickelization: Done. Had to first rework some test code that was peering directly into `#sa{}`, but the resulting tests are better now anyway.

src/riak_kv_stat.erl
nickelization: @lehoff Technically it's not just saving typing, because ?PFX is not a constant: it expands to a call to `riak_core_stat:prefix()` which does an `app_helper:get_env`. Caching the result of this call in `P` should theoretically offer a tiny performance boost. That said, it's probably negligible, and I'd rather have simpler code, so I went ahead and made your suggested changes anyway. But, just wanted to point that out to you in case it changes your mind at all.

include/riak_kv_sweeper.hrl
nickelization: @lehoff Any thoughts/opinions on the above? Do you think it's worth the time and effort to make the change, or should we leave it as-is?

include/riak_kv_sweeper.hrl
nickelization: @lehoff Similar to the case with `#sweep_participant{}`, I wish Mikael had written the code differently, but I'm questioning if it's worth the effort to change it now. Thoughts?

itest/riak_kv_sweeper_SUITE.erl
nickelization: A lot of the tests now run sweeps concurrently, since we bump up the sweep concurrency and spam a bunch of ticks in a row. I figured explicitly testing it is redundant now that it's being implicitly tested everywhere.

src/riak_kv_sweeper_fold.erl
nickelization: Done.

src/riak_kv_sweeper_fold.erl
nickelization: Done.

src/riak_kv_sweeper_fold.erl
nickelization: And done!

src/riak_kv_sweeper_fold.erl
nickelization: I assume you mean `lists:foreach/2`. Done.

src/riak_kv_sweeper_fold.erl
nickelization: ✅

src/riak_kv_sweeper_fold.erl
nickelization: Done.

src/riak_kv_sweeper_fold.erl
nickelization: @lehoff I'm not sure what you're referring to with the `TotalWait` comment. Clarification?

nickelization: I did go ahead and factor out the code to get `Limit` and `Wait` - that part made sense 👍

src/riak_kv_sweeper_fold.erl
nickelization: Done.

src/riak_kv_sweeper_fold.erl
nickelization: Done.

src/riak_kv_sweeper_fold.erl
nickelization: Done.

src/riak_kv_sweeper_fold.erl
nickelization: Agreed. It started as a smaller tuple when I was first refactoring this code into a fold, and it crept up in size without my really paying attention to it. Switched to a record.

src/riak_kv_sweeper_fold.erl
nickelization: Did some digging, and found that it is actually `[{atom(), any()}]`. Fixed.

src/riak_kv_sweeper_fold.erl
```diff
@@ -200,34 +200,31 @@ format_result(SuccFail, Results) ->
...
-            %% This way we can respond on requests while throttling
-            Acc1 = Acc0#sa{throttle = get_sweep_throttle()},
-            riak_kv_sweeper:sleep_for_throttle(Wait),
-            maybe_extra_throttle(Acc1);
-        _ ->
-            maybe_extra_throttle(Acc0)
-    end;
-
+maybe_throttle_sweep(_RObjBin, #sa{throttle = {pace, Limit, Wait}, swept_keys = SweepKeys} = Acc)
+  when SweepKeys rem Limit =:= 0 ->
```
raghavkarol, line 252: The comment 
``` 
%% We use recieve ...
%% This way ...
``` 
Is no longer correct  and should be deleted

src/riak_kv_sweeper_fold.erl
```diff
@@ -465,10 +465,14 @@ update_stat_counters(NumMutated, NumDeleted, RObjBinSize, Acc) ->
 -spec stat_send(riak_kv_sweeper:index(),
                 {'final'| 'in_progress', #sa{}}) -> #sa{}.
 stat_send(Index, {final, Acc}) ->
```
raghavkarol, line 510: Just a note, list comprehensions [do not generate a list](http://erlang.org/doc/efficiency_guide/listHandling.html#id68285), when the compiler can detect that the return value is not used.

itest/riak_kv_sweeper_SUITE.erl
raghavkarol: Agreed that a lot of tests are running sweeps concurrently, but we do not validate that the sweeps are indeed running concurrently. I changed to code to hardcode sweep concurrency of 1 and all the tests still passed, while the same will fail with #1580

src/riak_kv_sweeper_fold.erl
lehoff: Perfect.

src/riak_kv_sweeper_fold.erl
lehoff: I meant that instead of pattern matching on the first line of the function move that pattern match into the function head.

src/riak_kv_sweeper_fold.erl
lehoff: Thx.

include/riak_kv_sweeper.hrl
lehoff: For `#sweep{}` the changes are not that big, so I think we should do it now rather than later.

Outside `riak_kv_sweeper_state` `#sweep{}` is only used here (disregarding the test):
```
src/riak_kv_sweeper.erl
139:-spec status() -> {[#sweep_participant{}], [#sweep{}]}.

src/riak_kv_sweeper_cli.erl
100:              || #sweep{state = State} = Sweep <- Sweeps, State == running],
110:format_progress(#sweep{index = Index,

src/riak_kv_sweeper_fold.erl
284:                           #sweep{}) -> 'ok' | 'no_pid'.
285:send_to_sweep_worker(Msg, #sweep{pid = Pid}) when is_pid(Pid) ->
289:send_to_sweep_worker(Msg, #sweep{index = Index}) ->
```

`riak_kv_sweeper` is just a type definition.

`riak_kv_sweeper_cli`: first usage can be fixed by adding a function to extract the state.
2nd usage would require moving the formatting into the `riak_kv_sweeper_state` module, which is not unfair given that it is relating to a type - quite a common pattern in OCaml.

`riak_kv_sweeper_fold`: accessor functions to get the pid and index out.

For this record it makes sense to hide it.

nickelization: Fair enough! I will address this today 👍

src/riak_kv_sweeper_fold.erl
nickelization: Ooh wow! I didn't know that, that's super cool.

@lehoff I already made lots of list comprehension to `lists:foreach` changes at your suggestion, to avoid generating an intermediate list. I'm now wondering if that is a good idea, since Raghav has helpfully pointed out that basically this optimizes down to the same thing in the end. Any objections if I revert some of my commits in cases where a list comprehension is more readable/succinct?

src/riak_kv_sweeper_fold.erl
nickelization: Ah nice catch, thanks! Fixed.

src/riak_kv_sweeper.erl
nickelization: Done.

include/riak_kv_sweeper.hrl
lehoff: Revisiting its usage today... it should not be in its own module. It belongs in `riak_kv_sweeper_fold`. 

Outside `riak_kv_sweeper_fold` the usage is very limited and can be fixed with a few setters/getters - tests have not been considered.
```
src/riak_kv_delete_sup.erl
70:      #sweep_participant{ description = "Object TTL",
89:      #sweep_participant{ description = "Reap tombstones",

src/riak_kv_entropy_manager.erl
244:      #sweep_participant{

src/riak_kv_sweeper.erl
25:%% schedules sweeps.  Anyone can register a #sweep_participant{} with
34:%% #sweep_participant{}.
123:-spec add_sweep_participant(#sweep_participant{}) -> ok.
139:-spec status() -> {[#sweep_participant{}], [#sweep{}]}.
161:                           [#sweep_participant{}],

src/riak_kv_sweeper_cli.erl
136:     || {Index, #sweep_participant{module = Mod}} <- IndexedParticipants].
162:format_sweep_participant(Index, #sweep_participant{module = Mod,

src/riak_kv_sweeper_state.erl
100:                            Participant:: #sweep_participant{}) -> state().
135:                          {Participants :: [#sweep_participant{}],
153:    {'ok', Participants :: [#sweep_participant{}], state()}.
162:        [Participant#sweep_participant{sweep_fun = Fun, acc = InitialAcc} ||
189:                           [#sweep_participant{}],
344:     {Module, #sweep_participant{run_interval = RunInterval}}
525:        {ok, #sweep_participant{run_interval = RunInterval}} ->
676:    #sweep_participant{module = Module,
744:    #sweep_participant{
```

`riak_kv_delete_sup`: a `riak_kv_sweeper_fold:new_participant/5` function to create the record.

`riak_kv_entropy_manager`: use the function above.
 
`riak_kv_sweeper`: use the `riak_kv_sweeper_fold:participant()` type instead of the record in the type declarations.

`riak_kv_sweeper_cli`:  move the formatting to `riak_kv_sweeper_fold`.

`riak_kv_sweeper_state`: 
* Add a function to get the module out. Not sure what the name should be, maybe `participant_module/1`?
* Add a function to set the fun and initial accumulator. Say, `set_sweep_parameters/3`.
* Add a function to set the run interval.

This is doable. It will make the testing in `riak_kv_sweeper_state` a bit harder, but maybe the generators for the `#sweep_participant{}` should be defined alongside the record definition?

include/riak_kv_sweeper.hrl
nickelization: Fixed. Had to make a bunch of additional changes to the CT tests, but it all works now.

src/riak_kv_sweeper_cli.erl
nickelization: Done (and the code was moved into the riak_kv_sweeper_state module as part of the encapsulation of `#sweep{}`).

include/riak_kv_sweeper.hrl
lehoff: Perfect!

src/riak_kv_sweeper.erl
```diff
@@ -0,0 +1,369 @@
...
+    %% riak_kv_index_tree clears the trees if they are expired.
+    AAEEnabled = riak_kv_entropy_manager:enabled(),
+    EstimatedKeys = get_estimate_keys(Index, AAEEnabled, State),
+
+    State2 =
+        case riak_kv_sweeper_state:get_active_participants(State, Index) of
+            {ok, [], State1} ->
+                State1;
+            {ok, ActiveParticipants, State1} ->
+                ?MODULE:update_started_sweep(Index,
```
lehoff: Doing a `?MODULE:some_fun` call is dubious. It triggers code upgrade, which does not seem to be the wish here.

I think the `?MODULE:` should be dropped.

src/riak_kv_sweeper_state.erl
```diff
@@ -0,0 +1,1076 @@
...
+
+-spec expired_or_missing(erlang:timestamp(),
+                         #sweep{},
+                         dict()) ->
+    {MissingSum :: non_neg_integer(),
+     ExpiredSum :: non_neg_integer()}.
+expired_or_missing(Now, #sweep{results = Results}, Participants) ->
+    ResultsList = dict:to_list(Results),
+    Missing = missing(run_interval, Participants, ResultsList),
+    Expired =
```
lehoff: Why is this not in a function like `missing/3`?

src/riak_kv_sweeper_state.erl
```diff
@@ -0,0 +1,1076 @@
...
+    {MissingSum, ExpiredSum}.
+
+-spec expired(erlang:timestamp(),
+              erlang:timestamp(),
+              RunInterval :: non_neg_integer() | disabled) ->
+    non_neg_integer().
+expired(_Now, _TS, disabled) ->
+    0;
+expired(Now, TS, RunInterval) ->
+    case elapsed_secs(Now, TS) - RunInterval of
```
lehoff: Replace the case statement with:
`max(0, elapesed_secs(Now, TS) - RunInterval)`

src/riak_kv_sweeper_state.erl
```diff
@@ -0,0 +1,1076 @@
...
+
+-spec random_sweep([#sweep{}]) -> #sweep{}.
+random_sweep(Sweeps) ->
+    Index = random:uniform(length(Sweeps)),
+    lists:nth(Index, Sweeps).
+
+-spec scheduler_enabled() -> boolean().
+scheduler_enabled() ->
+    case app_helper:get_env(riak_kv, sweeper_scheduler, true) of
+        E when E == true; E == false ->
```
lehoff: Change the guard to `when is_boolean(E)`

src/riak_kv_sweeper_state.erl
```diff
@@ -0,0 +1,1076 @@
...
+    'always'
+    | 'never'
+    | {StartHour :: non_neg_integer(), EndHour :: non_neg_integer()}.
+sweep_window() ->
+    case application:get_env(riak_kv, sweep_window) of
+        {ok, always} ->
+            always;
+        {ok, never} ->
+            never;
+        {ok, {StartHour, EndHour}} when StartHour >= 0, StartHour =< 23,
```
lehoff: Make a macro:
```erlang
-define(IS_HOUR(H), is_integer(H) andalso H >=0 andalso H <=0).

...
        {ok, {StartHour, EndHour}} when ?IS_HOUR(StartHour), ?IS_HOUR(EndHour) -> 
        ...
```
And perhaps `StartHour =< EndHour`??

src/riak_kv_sweeper_state.erl
```diff
@@ -0,0 +1,1076 @@
...
+            {StartHour, EndHour};
+        Other ->
+            lager:warning("Invalid riak_kv_sweep window specified: ~p. "
+                          "Defaulting to 'never'.", [Other]),
+            never
+    end.
+
+-spec get_concurrency_limit() -> non_neg_integer().
+get_concurrency_limit() ->
+    case app_helper:get_env(riak_kv, sweep_concurrency,
```
lehoff: Are we okay with this failing if `sweep_concurrency` is not an integer in the environment?

src/riak_kv_sweeper.erl
nickelization: Yeah, no idea why that was there. Dropped.

src/riak_kv_sweeper_state.erl
nickelization: Not sure why not, but it's in its own function now! Fixed.

src/riak_kv_sweeper_state.erl
nickelization: Good call. Done!

src/riak_kv_sweeper_state.erl
nickelization: And done.

src/riak_kv_sweeper_state.erl
nickelization: Macro created. However, it is valid to have `StartHour` > `EndHour`. This allows you to have a sweep window from, say, 10PM to 2AM. The existing sweeper code handles this correctly.

src/riak_kv_sweeper_state.erl
nickelization: Normally Cuttlefish should guard against non-integer values here, so it's not a huge risk, but it is possible a user might go screwing around with advanced.config and put a bad value in there. Better safe than sorry, I suppose.

Now that I look at this function, I'm also not sure why we have that `case` statement there at all, since `?DEFAULT_SWEEP_CONCURRENCY` is already an integer. Will fix that as well.

nickelization:
thumbot retry

src/riak_kv_sweeper_cli.erl
nickelization: Thanks, did this a while ago, and just noticed I did not reply to the review comment until now. Done.

src/riak_kv_sweeper_cli.erl
nickelization: Missed this comment earlier; just pushed a commit that fixes this.

include/riak_kv_sweeper.hrl
nickelization: Done and pushed. All the sweeper records are now encapsulated and `riak_kv_sweeper.hrl` has been deleted 💥

nickelization:
thumbot retry

davidx:
thumbot retry

nickelization:
Discussed with @lehoff this morning, and we've decided there's been enough review and refinement of this work, and it's time to merge it. The dialyzer step is still failing for Thumbs, but that's due to an issue in the develop branch.

PR #1566, opened 2016-12-07
Some tidy up in preparation for new docs:

gordonguthrie:
* moved diagrams into their own directory so that the casual browser
  is clear as to what documentation exists
* the diagrams themselves were being referenced from a development
  branch of riak which means they were likely to be delinked accidently


hazen:
+1

hmmr:
+1

PR #1570, opened 2016-12-08
better handle and report errors from unresponsive qry_buffer gen_server

hmmr:
Another attempt to fix `ts_simple_query_buffers_SUITE`.

Also, consume and drop messages riak_kv_qry_N workers are not supposed to handle.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -136,6 +136,10 @@ handle_info({{SubQId, QId}, {error, Reason} = Error},
 handle_info({{_SubQId, QId1}, _}, State = #state{qid = QId2}) when QId1 =/= QId2 ->
     %% catches late results or errors such getting results for invalid QIds.
     lager:debug("Bad query id ~p (expected ~p)", [QId1, QId2]),
+    {noreply, State};
+
+handle_info(_Unexpected, State) ->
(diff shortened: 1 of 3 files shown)


macintux:
+1

src/riak_kv_qry_worker.erl
```diff
@@ -136,6 +136,10 @@ handle_info({{SubQId, QId}, {error, Reason} = Error},
 handle_info({{_SubQId, QId1}, _}, State = #state{qid = QId2}) when QId1 =/= QId2 ->
     %% catches late results or errors such getting results for invalid QIds.
     lager:debug("Bad query id ~p (expected ~p)", [QId1, QId2]),
+    {noreply, State};
+
+handle_info(_Unexpected, State) ->
+    lager:info("Not handling unexpected message \"~p\" in state ~p", [_Unexpected, State]),
```
andytill, line 142: Formatting the unexpected message could result in a very large log message, for this gen_server in particular that handles subquery result chunks with an unbounded number of rows.

macintux: `lager` is designed to truncate large messages.

hmmr: lager has safety checks in place covering this, I hear.

andytill:
+1

PR #1571, opened 2016-12-12
Removed hard-coded dependency on yokozuna

fadushin:
This PR removed a hard-coded dependency on yokozuna by indirecting calls into yokozuna through a module that is loaded via configuration, using the new `index_module` (internally defined) configuration variable.

By default (for the purposes of riak_kv, by itself), the `index_module` configuration variable is undefined, which will result in a no-op during puts, deletes, and handoff operations from the vnode.

Downstream repos (notably, yokozuna), configure `index_module` to be a module the `riak_kv_vnode` will call as part of the above set of operations.

The `index_module` configuration variable is not intended to be shared with users, as its only function is to indirect calls into an indexing subsystem.

Dependent PRs: -none-
Downstream PRs: https://github.com/basho/yokozuna/pull/716

## Checklist

- [x] Observability
This PR has no impact on a running system, at least as we deploy riak today.  Any new configuration we add is hidden and should get picked up automatically as part of an upgrade.
- [x] Performance
In general this change should have no impact on performance.  In the case where yokozuna is not deployed with riak_kv (no known deployments), this change should eliminate calls into yokozuna, theoretically resulting in fewer reductions
- [x] Fault-tolerance
There are no remote processes or IPC added as part of this PR
- [x] Testing
This PR does not affect code paths in any existing tests.  No specific tests have been added for this PR, as there is no new functionality being added.
- [x] Code Clarity/Quality
- [x] Code Documentation
The rationale for the index_module configuration documentation is explained in the source code documentation.
- [x] External Documentation
n/a
- [x] Release Notes
n/a

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -2698,23 +2698,17 @@ highest_actor(ActorBase, Obj) ->
     {Actor, Epoch, riak_object:actor_counter(Actor, Obj)}.
 
 %%
-%% Technical note:  The index_module configuration parameter should contain
-%% a module name which must implement the following functions:
-%%
-%%     - index(object_pair(), write_reason(), p()) -> ok.
-%%     - index_binary(bucket(), key(), binary(), write_reason(), p()) -> ok.
-%%     - is_searchable(riak_kv_bucket:props()) -> boolean().
-%%
-%% The indexing module will be called on puts, deletes, handoff, and
-%% anti-entropy activity.  In the case of puts, if an object is being over-written,
-%% the old object will be passed as the second parameter in the object pair.
-%% The indexing module may use this old object to optimize the update (e.g.,
-%% to handle the special case of sibling writes, which may not map directly to
-%% Riak puts).
-%%
-%% NB. Currently, yokozuna is the only repository that currently
-%% implements this behavior.  C.f., Yokozuna cuttlefish schema, to see
-%% where this configuration is implicitly set.
+%% Maintenance note:  The riak_kv.update_hook configuration parameter should
+%% contain a module name which must implement riak_kv_update_hook behavior.
+%% Currently, this behavior is completely internal and is not intended for
+%% use outside of Riak; Yokozuna is the only repository that currently
+%% implements this behavior.  (C.f., Yokozuna cuttlefish schema, to see
+%% where this configuration is set.)  In the future, we may want
+%% make the riak_kv.update_hook configuration parameter a list of hooks, and
+%% call a sequence of hooks on updates, but since we currently only have one
+%% use-case, and since this is a purely internal API (and local to the riak node),
+%% we can safely make that change in the future without an impact on
+%% backwards compatibility.
 %%
 
 -spec update_hook()-> update_hook().
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -92,12 +92,6 @@
 -export([put_merge/6]). %% For fsm_eqc_vnode
 -endif.
 
--ifdef(TEST).
--define(YZ_SHOULD_HANDOFF(X), true).
--else.
--define(YZ_SHOULD_HANDOFF(X), yz_kv:should_handoff(X)).
--endif.
-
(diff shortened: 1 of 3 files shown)


src/riak_kv_vnode.erl
```diff
@@ -92,19 +92,6 @@
...
--define(INDEX_BIN(A,B,C,D,E), _=element(1,{A,B,C,D,E}), ok).
--define(IS_SEARCH_ENABLED_FOR_BUCKET(BProps), _=element(1, {BProps}), false).
--else.
--define(INDEX(Objects, Reason, Partition), yz_kv:index(Objects, Reason, Partition)).
--define(INDEX_BIN(Bucket, Key, Obj, Reason, Partition), yz_kv:index_binary(Bucket, Key, Obj, Reason, Partition)).
--define(IS_SEARCH_ENABLED_FOR_BUCKET(BProps), yz_kv:is_search_enabled_for_bucket(BProps)).
--endif.
-
 -ifdef(TEST).
 -define(YZ_SHOULD_HANDOFF(X), true).
```
JeetKunDoug: This is still calling `yz_kv` directly in the non-test case - of course Github won't let me comment on _that_ line, but this is close enough.

src/riak_kv_vnode.erl
```diff
@@ -2704,6 +2697,45 @@ highest_actor(ActorBase, Obj) ->
...
     {Actor, Epoch, riak_object:actor_counter(Actor, Obj)}.
 
+%%
+%% Technical note:  The index_module configuration parameter should contain
+%% a module name which must implement the following functions:
+%%
+%%     - index(object_pair(), write_reason(), p()) -> ok.
+%%     - index_binary(bucket(), key(), binary(), write_reason(), p()) -> ok.
+%%     - is_searchable(riak_kv_bucket:props()) -> boolean().
+%%
```
jvoegele: Would it be feasible to define these functions as callbacks in a behavior module? If so, the `yokozuna.erl` module (and any other future indexing modules that we might create) could use the behavior to get some extra compile-time checking.

src/riak_kv_vnode.erl
```diff
@@ -2704,6 +2697,45 @@ highest_actor(ActorBase, Obj) ->
     %% get the greatest event for the highest/latest actor
     {Actor, Epoch, riak_object:actor_counter(Actor, Obj)}.
 
+%%
+%% Technical note:  The index_module configuration parameter should contain
+%% a module name which must implement the following functions:
+%%
+%%     - index(object_pair(), write_reason(), p()) -> ok.
+%%     - index_binary(bucket(), key(), binary(), write_reason(), p()) -> ok.
+%%     - is_searchable(riak_kv_bucket:props()) -> boolean().
```
JeetKunDoug: As per above - need to add `should_handoff` with appropriate stuff.

src/riak_kv_vnode.erl
JeetKunDoug: +1 - good idea.

src/riak_kv_vnode.erl
```diff
@@ -134,6 +115,8 @@
           leasing = false :: boolean()
          }).
 
+-type update_hook() :: module() | undefined.
```
JeetKunDoug: Thoughts on defaulting to a `riak_kv_noop_update_hook` module, rather than `undefined` here? Implement the no-op versions of all the functions and we can get rid of all of the function heads that check for `undefined` and just call the noop version... and change all the `maybe_*` functions to just `*`.

src/riak_kv_vnode.erl
JeetKunDoug: Still need to actually add `should_handoff` here, or maybe just remove the function list here and refer to the behaviour now that we have it?

JeetKunDoug (commented):
2 suggestions below, but otherwise +1 - I like it.

src/riak_kv_vnode.erl
jvoegele: Seems like a win to me.

hazen:
Just excited to see modules decoupling!

nickelization:
+1

PR #1572, opened 2016-12-13
Silently take blob data over HTTP as base64

macintux:
@christophermancini identified a key issue: JSON does *not* like binary data. Since JSON is our sole interface for inserting and retrieving records via HTTP, that's a bit of a problem.

This set of changes allows HTTP clients such as our PHP library to use the industry-standard base64 encoding to upload and retrieve blob data for insert, query, get, and list_keys.

Must consider branching the logic to allow `0x`-prefaced strings to be sent instead of base64 to maintain consistency with the SQL command line. (This will be the only way to do SQL queries with blob as key values). Outstanding question: which format should we use for output?

And need tests, badly.

### DIFF
--- a/src/riak_kv_wm_timeseries.erl
+++ b/src/riak_kv_wm_timeseries.erl
@@ -61,7 +61,7 @@
         ]).
 
 %% webmachine body-producing functions
--export([to_json/2]).
+-export([to_json/2, value_to_json_compat/1]).
 
 -include_lib("webmachine/include/webmachine.hrl").
 -include_lib("riak_ql/include/riak_ql_ddl.hrl").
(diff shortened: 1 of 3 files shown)


macintux:
The key `riak_test` test for HTTP has been updated to use a blob as a primary key element (something which isn't actually tested elsewhere, I don't believe, oops) and all tests pass. See linked PR.

I'm inclined to not worry about supporting `0x`-prefaced values via the HTTP interface other than requiring them for select statements. We can add that later if it seems useful, but `base64` is the vital standard to support.

Removing "WIP" from this PR, it's ready for review.

gordonguthrie:
+1 - lovely piece of work produced under pressure...

christophermancini:
Great work John!

hazen:
:+1:

PR #1573, opened 2016-12-13
qry_worker: kill outstanding FSMs on cancelling query

hmmr:
When a query is cancelled (because of div_by_zero or if its projected size is found to exceed configured limit), there may be index FSMs on vnodes still busy collecting chunks.  Those chunks will be sent to the
coordinator when they are ready, where the query worker instances to receive them will be gone.

It makes sense to terminate the FSMs early, once it is known the data they are collecting will be discarded.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -68,7 +68,7 @@
           %% 1. Maintain a limited number of running fsms (i.e.,
           %%    process at most that many subqueries at a time):
           fsm_queue          = []                   :: [list()],
-          n_running_fsms     = 0                    :: non_neg_integer(),
+          running_fsms       = []                   :: [{non_neg_integer(), pid()}],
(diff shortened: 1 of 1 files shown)


hazen:
@hmmr Is there a Jira story associated with this?

hmmr:
It's RTS-1604.

I would like @andytill to have a look, though. This is a more accurate fix for an earlier crash that was already fixed #1570.

It helps to improve load by stopping vnodes doing unneeded work when a query is cancelled. But it will not prevent the dreaded timeouts to qry_buffers gen_server resulting from slow IO.

andytill:
Not sure how this can be tested.

macintux:
thumbot wait

hmmr:
@andytill It is tested on the juncture of query_orderby_max_data_size_error and the next following test case in https://github.com/basho/riak_test/pull/1257.

hmmr:
Closing, to be reopened against riak_ts-integration in https://github.com/basho/riak_kv/pull/1597.

PR #1576, opened 2016-12-13
Handle empty exchange history

bsparrow435:
When compute_exchange_info is called on an index with no previous
exchanges return undefined for the Recent and LastAll timestamps.


ian-mi:
+1

PR #1577, opened 2016-12-14
Unify how CREATE TABLE is serviced and ensure table is query-ready [JIRA: RIAK-3173]

paegun:
* fix for CREATE TABLE removing the delta between how the request is serviced between HTTP and PB and RPC.
* fix for CREATE TABLE, removing the delta between the post-condition of CREATE TABLE (was active bucket-type only) and "query-ready" (is bucket-type active, DDL module compiled and min capability met).

Depends on: https://github.com/basho/riak_core/pull/890
Tested by: https://github.com/basho/riak_test/pull/1254
Relates to: riak_test ts_cluster_riak_shell_regression_log which otherwise exposes a race condition between the CREATE TABLE and subsequent INSERT queries.

### DIFF
--- a/src/riak_kv_ts_api.erl
+++ b/src/riak_kv_ts_api.erl
@@ -126,24 +126,53 @@ create_table(SvcMod, ?DDL{table = Table}=DDL1, WithProps) ->
                                                        flat_format(
                                                          "Invalid chash mod or fun in bucket type properties: ~p:~p\n", [M, F]))};
         Props2 ->
-            create_table1(SvcMod, Table, Props2, DDLRecCap, ?TABLE_ACTIVATE_WAIT)
+            create_table1(SvcMod, Table, Props2, DDL2, DDLRecCap, ?TABLE_ACTIVATE_WAIT)
     end.
 
-create_table1(SvcMod, Table, Props, DDLRecCap, Seconds) ->
+create_table1(SvcMod, Table, Props, DDL, DDLRecCap, Seconds) ->
     case riak_core_bucket_type:create(Table, Props) of
-        ok -> wait_until_active(SvcMod, Table, DDLRecCap, Seconds);
+        ok -> wait_until_active_and_supported(SvcMod, Table, DDL, DDLRecCap, Seconds);
         {error, Reason} -> {error, SvcMod:make_table_create_fail_resp(Table, Reason)}
     end.
 
-wait_until_active(SvcMod, Table, _DDLRecCap, _Seconds=0) ->
+wait_until_supported(SvcMod, Table, _DDL, _DDLRecCap, _Seconds=0) ->
     {error, SvcMod:make_table_activate_error_timeout_resp(Table)};
-wait_until_active(SvcMod, Table, DDLRecCap, Seconds) ->
+wait_until_supported(SvcMod, Table, DDL, DDLRecCap, Seconds) ->
+    case get_active_peer_nodes() of
+        Nodes when is_list(Nodes) ->
+            wait_until_supported1(Nodes, SvcMod, Table, DDL, DDLRecCap, Seconds);
+        {error, _Reason} ->
+            timer:sleep(1000),
+            wait_until_supported(SvcMod, Table, DDL, DDLRecCap, Seconds - 1)
+    end.
+
+wait_until_supported1(_Nodes=[], _SvcMod, _Table, _DDL, _DDLRecCap, _Seconds) -> ok;
+wait_until_supported1([Node|NodesT], SvcMod, Table, DDL, DDLRecCap, Seconds) when Node =:= node() ->
+    wait_until_supported1(NodesT, SvcMod, Table, DDL, DDLRecCap, Seconds);
+wait_until_supported1(Nodes=[Node|NodesT], SvcMod, Table, DDL, DDLRecCap, Seconds) ->
+    case rpc:call(Node, riak_kv_ts_newtype, is_compiled, [Table, DDL]) of
+        true -> wait_until_supported1(NodesT, SvcMod, Table, DDL, DDLRecCap, Seconds);
+        _ ->
+            timer:sleep(1000),
+            lager:info("Waiting for table ~ts to be compiled on ~s", [Table, Node]),
+            wait_until_supported1(Nodes, SvcMod, Table, DDL, DDLRecCap, Seconds - 1)
+    end.
+
+get_active_peer_nodes() ->
+    case riak_core_ring_manager:get_my_ring() of
+        {ok, Ring} -> riak_core_ring:active_members(Ring);
+        _ -> {error, retry}
+    end.
+
+wait_until_active_and_supported(SvcMod, Table, _DDL, _DDLRecCap, _Seconds=0) ->
+    {error, SvcMod:make_table_activate_error_timeout_resp(Table)};
+wait_until_active_and_supported(SvcMod, Table, DDL, DDLRecCap, Seconds) ->
     case riak_core_bucket_type:activate(Table) of
-        ok -> ok;
+        ok -> wait_until_supported(SvcMod, Table, DDL, DDLRecCap, Seconds);
         {error, not_ready} ->
             timer:sleep(1000),
             lager:info("Waiting for table ~ts to be ready for activation", [Table]),
-            wait_until_active(SvcMod, Table, DDLRecCap, Seconds - 1);
+            wait_until_active_and_supported(SvcMod, Table, DDL, DDLRecCap, Seconds - 1);
         {error, undefined} ->
             {error, SvcMod:make_table_created_missing_resp(Table)}
     end.
--- a/src/riak_kv_ts_api.erl
+++ b/src/riak_kv_ts_api.erl
@@ -146,17 +146,23 @@ wait_until_supported(SvcMod, Table, DDL, DDLRecCap, Seconds) ->
             wait_until_supported(SvcMod, Table, DDL, DDLRecCap, Seconds - 1)
     end.
 
-wait_until_supported1(_Nodes=[], _SvcMod, _Table, _DDL, _DDLRecCap, _Seconds) -> ok;
-wait_until_supported1([Node|NodesT], SvcMod, Table, DDL, DDLRecCap, Seconds) when Node =:= node() ->
-    wait_until_supported1(NodesT, SvcMod, Table, DDL, DDLRecCap, Seconds);
-wait_until_supported1(Nodes=[Node|NodesT], SvcMod, Table, DDL, DDLRecCap, Seconds) ->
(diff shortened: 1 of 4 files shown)


macintux:
I'm not sure this does what's intended, at least not completely.

`riak_core_bucket_type:activate` returns `ok` after the claimant twiddles the active bit in CMD, so adding an extra layer of tests to make sure that has propagated makes sense.

However, `is_table_supported` just checks to make sure the module is compiled and there are no capabilities conflicts.

What we really need, if we want to be careful, is to do an rpc across the cluster to see if everyone agrees that the table is active.

paegun:
@macintux I agree wrt the code not appearing to do what it expects as a postcondition, but afaik this is due to "action at a distance". The capabilities exchange is what the wait_until_supported/4 is waiting for.

If nothing else, and if my understanding is correct, I will add a comment to wait_until_supported/4.

andytill:
Unifying create table for all APIs is very good, the other requests are call into `riak_kv_ts_api` though for writes, single key ops, queries etc.

src/riak_kv_ts_util.erl
```diff
@@ -705,6 +708,67 @@ check_table_feature_supported(DDLRecCap, DecodedReq) ->
...
+        ok -> wait_until_active_and_supported(SvcMod, Table, DDLRecCap, Seconds);
+        {error, Reason} -> {error, SvcMod:make_table_create_fail_resp(Table, Reason)}
+    end.
+
+wait_until_supported(SvcMod, Table, _DDLRecCap, _Seconds=0) ->
+    {error, SvcMod:make_table_activate_error_timeout_resp(Table)};
+wait_until_supported(SvcMod, Table, DDLRecCap, Seconds) ->
+    case is_table_supported(DDLRecCap, Table) of
+        true -> ok;
+        _ -> wait_until_supported(SvcMod, Table, DDLRecCap, Seconds - 1)
```
andytill: `Seconds` is being decremented here on each recursion but there is no sleep.

src/riak_kv_ts_util.erl
```diff
@@ -705,6 +708,67 @@ check_table_feature_supported(DDLRecCap, DecodedReq) ->
...
+            create_table1(SvcMod, Table, Props2, DDLRecCap, ?TABLE_ACTIVATE_WAIT)
+    end.
+
+create_table1(SvcMod, Table, Props, DDLRecCap, Seconds) ->
+    case riak_core_bucket_type:create(Table, Props) of
+        ok -> wait_until_active_and_supported(SvcMod, Table, DDLRecCap, Seconds);
+        {error, Reason} -> {error, SvcMod:make_table_create_fail_resp(Table, Reason)}
+    end.
+
+wait_until_supported(SvcMod, Table, _DDLRecCap, _Seconds=0) ->
```
andytill: As I understand it, this function waits for a required capabilities to be met before replying `ok`, after the table has been activated. I think this is the wrong way around. The table should be validated that it's requirements are met before it can be created, I think this was part of the table existence checks, or we can make it that way. There is no point in adding it to the metadata even if it cannot be used.

Capabilities will change when nodes come in and out of the cluster, that is unlikely to happen within a 30 second timeout. If un-negotiated capabilities is the cause of the sql log regression failures, there is a util function in rt that verifies capability values.

paegun: You are right @andytill, `create_table/3` through `convert_ddl_to_cluster_supported_version/2` is ensuring that the DDL version is the supported version for this node, so removing `wait_until_supported/4` and changing `wait_until_active_and_supported/4` to `wait_until_active/4`.

src/riak_kv_ts_util.erl
paegun: Good catch. I had determined that `is_table_supported/2` is determinate, but left the retry, but should not have.

paegun:
@andytill wrt "Unifying create table for all APIs is very good, the other requests are call into riak_kv_ts_api though for writes, single key ops, queries etc."

If I understand correctly, this PR addresses unifying the CREATE TABLE path.

So I understand the comment to mean we should similarly unify the paths for other operations and queries. If so, I agree, and that's a separate issue.

andytill:
Apologies, I wasn't very clear.  I meant we should put the create table logic into `riak_kv_ts_api` because puts, gets, deletes and queries are already shared between APIs from that module.

paegun:
@andytill you were clear and right, riak_kv_ts_api is the single proper channel to direct these commands and queries through. I've made the change suggested.

I was confused by not finding riak_kv_ts_api to be the single proper channel for all commands over HTTP and PB, but on re-running commands w/ redbug tracing on, idk how I came to that conclusion :face_palm:

hmmr:
+1

hazen:
I thought this didn't actually necessarily solve the problem (sadly)

macintux:
+1

macintux:
I'm not convinced these PRs solve the flappiness that has us concerned, but they're still valuable improvements.

macintux:
thumbot wait

macintux:
I’m now thinking that non-obvious changes without a glaring need on release day probably aren’t ideal. Asked Thumbs to wait.

paegun:
After a good amount of digging through the call flow, this change, assuring the CREATE TABLE postcondition is met in a synchronous manner is necessarily employed on the front-end of servicing the request.

The riak_core metadata exchange is asynchronous, yielding synchronous control flow to the caller upon broadcasting the bucket-type metadata. The alternatives considered follow:
1. Spawn a a gen_fsm awaiting callbacks following the broadcast for each node to signal that they have reached "ready for query" state.
1.a. Since there is a maximum amount of time, have the gen_fsm have a timeout signal and terminal state, at which point the caller would be notified that the CREATE TABLE did not succeed due to timeout.
2. Poll from the service.

Both alternatives would work, but polling from the service is more straightforward. The call flow for CREATE TABLE is already complex enough, depending upon a new trigger mechanism `sync_notify` w/i the metadata exchange, callbacks from riak_core -> riak_kv, and duplication of code that determines whether a table is compiled (*).

The reason for the deep dive through the riak_core integration is that it was thought that the riak_core integration guaranteed the CREATE TABLE postcondition *_and_* did so in a synchronous manner, meaning control flow would not yield to the caller until the postcondition was met. This is not the case, though, as following broadcast to the eager peers (plumtree), the riak_core_claimant returns control flow to the caller. This means that broadcast to the lazy peers through the plumtree or fallback to endemic broadcast (gossip) is fulfilled following the return of control flow to the caller. This is what creates the race condition window.

After analyzing the call flow in full, I believe the solution employed here is the appropriate solution. The race condition window is closed and the code responsible for servicing the CREATE TABLE request was easily modified to perform the checks as well as report back to the client in the event that the postcondition is not met w/i the allowable timeframe.


*) wrt the duplication of code that determines whether a table is compiled, this is not ideal, but is a red herring. The code in riak_kv and riak_core both properly determine that from each system view, the table is compiled, properly.

DSomogyi:
create jira issue

src/riak_kv_ts_api.erl
```diff
@@ -97,6 +104,93 @@ api_calls() ->
...
+        _ -> {error, retry}
+    end.
+
+wait_until_active_and_supported(SvcMod, Table, _DDL, _DDLRecCap, _Seconds=0) ->
+    {error, SvcMod:make_table_activate_error_timeout_resp(Table)};
+wait_until_active_and_supported(SvcMod, Table, DDL, DDLRecCap, Seconds) ->
+    case riak_core_bucket_type:activate(Table) of
+        ok -> wait_until_supported(SvcMod, Table, DDL, DDLRecCap, Seconds);
+        {error, not_ready} ->
+            timer:sleep(1000),
```
hmmr: I believe the expectation here is that the table gets activated quickly -- but not immediately -- under normal circumstances. It is likely, in other words, that the first time `riak_core_bucket_type:activate(Table)` will return `{error, not_ready}` and a second attempt made in 10 msec would succeed. Introducing a small delay on the initial iteration would, then, give you a fair chance of going through it in (much) less than a second.

src/riak_kv_wm_timeseries_query.erl
```diff
@@ -265,35 +270,23 @@ compile_query(QueryStr) ->
...
+    lists:flatten(io_lib:format(Format, Args)).
+
+make_table_create_fail_resp(Table, Reason) ->
+    flat_format("Failed to create table ~s: ~p", [Table, Reason]).
+
+make_table_activate_error_timeout_resp(Table) ->
+    flat_format("Timed out while attempting to activate table ~ts", [Table]).
+
+make_table_created_missing_resp(Table) ->
+    flat_format("Table ~s has been created but found missing", [Table]).
```
hmmr: Make it `"~ts"` here, too, to prevent the unthinkable (also in `make_table_create_fail_resp` above).

paegun:
thumbot retry

davidx:
thumbot retry

hmmr:
+1

andytill:
Is it the table activation that is async?

This call in Riak Core already checks that the table helper module is compiled before activation: https://github.com/basho/riak_kv/pull/1577

paegun:
Yes, @andytill the table activation is async, as I commented the metadata exchange uses the plumtree protocol which is async. The compilation is hooked into the metadata listener, so is async.

I assume you intended to link the compilation, so linking here:
*Registration of Trigger*
https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_ts_newtype.erl#L99

*Handling of Triggered Event*
1. https://github.com/basho/riak_core/blob/riak_ts-develop/src/riak_core_metadata_evt_sup.erl#L67
2. https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_metadata_store_listener.erl#L53

andytill:
Ah, I meant to link to the compile check in `riak_core_claimant`, https://github.com/basho/riak_core/blob/riak_ts-develop/src/riak_core_claimant.erl#L707.

paegun:
Yes, @andytill, but that *check* for whether the ddl is compiled is in response to a bucket type status check, see:
1. https://github.com/basho/riak_core/blob/riak_ts-develop/src/riak_core_bucket_type.erl#L205
2. https://github.com/basho/riak_core/blob/riak_ts-develop/src/riak_core_claimant.erl#L302
3. *is type compiled*, not compile type, https://github.com/basho/riak_core/blob/riak_ts-develop/src/riak_core_claimant.erl#L707
4. https://github.com/basho/riak_core/blob/riak_ts-develop/src/riak_core_metadata_evt_sup.erl#L78

paegun:
See https://github.com/basho/riak_kv/pull/1602 for documentation requested.

Basho-JIRA:
PR: 
* https://github.com/basho/riak_kv/pull/1577 

***[posted via JIRA by Derek Somogyi]***

hazen:
:+1:

hmmr:
+1

PR #1579, opened 2016-12-17
convert timestamps in WHERE for DELETE

hmmr:
RTS-1629

A fix for DELETEing keys containing timestamps.


hazen:
Seems to work fine

hazen:
:+1:

paegun:
+1

would be nice to have a test for the ISO 8601 formatted timestamp, but this should be picked up in a unification of test SQLs using generators.

PR #1580, opened 2016-12-20
Add test for concurrent sweeps

raghavkarol:
Reintroduce and fix the test removed in 054b62c. This testcase
validates sweep concurrency functionality asserting that an expected
number of sweeps are running in concurrently.

### DIFF
--- a/itest/riak_kv_sweeper_SUITE.erl
+++ b/itest/riak_kv_sweeper_SUITE.erl
@@ -80,7 +80,7 @@ meck_new_riak_core_modules(Partitions) ->
     VNodeIndices.
 
 meck_new_riak_core_ring(Partitions) ->
-    VNodeIndices = [N || N <- lists:seq(0, Partitions)],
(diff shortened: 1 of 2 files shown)


itest/riak_kv_sweeper_SUITE.erl
```diff
@@ -985,25 +985,34 @@ exometer_deleted_object_count(Config) ->
                     {ok, [{count, DeleteKeys}, {one, DeleteKeys}]}),
     ok.
 
+scheduler_sweep_concurrency_test(_Config) ->
```
raghavkarol, line 956: I will add checks to make sure the scheduler doesn't go over the limit also

nickelization:
Test looks good, but I think we need to rebase this on top of the latest sweeper-2.3-integration changes.

nickelization:
:+1:

nickelization:
Going to just go ahead and merge this. It's a PR on top of a PR to add a test case, Raghav and I have both looked at it, and the test case is passing.

PR #1581, opened 2016-12-21
Fix sorting filters on quantum column and = operator RTS-1286

andytill:
Call `lists:partition/2` instead of `lists:splitwith\2` to split the list into two lists of matching and not matching the split criteria, independent of the order of filters in the list.


hmmr:
+1

paegun:
+1

src/riak_kv_qry_compiler.erl
```diff
@@ -2457,6 +2457,30 @@ eqality_filter_on_quantum_specifies_start_and_end_range_test() ->
         Select?SQL_SELECT.'WHERE'
     ).
 
+eqality_filter_on_quantum_specifies_start_and_end_range_2_test() ->
```
hazen, line 2460: Typo?

PR #1583, opened 2016-12-21
Use the correct key to get the configured max quanta value [JIRA: RIAK-3143]

andytill:
The call to get the max quanta value was using the wrong key, so always defaulting to 1000. This PR fixes the call to use the correct key and get the configured value, from the shell:

```sql
riak-shell(33)>SELECT * FROM WeatherStationData WHERE StationId = 'Station-1001' AND ReadingTimeStamp > 0 AND ReadingTimeStamp < 100000000000000000;;
Error (1025): Query spans too many quanta (1157407407, max 5000)
```


hmmr:
+1

DSomogyi:
create jira issue

paegun:
+1

hazen:
+1

PR #1584, opened 2016-12-21
if timeseries_query_timeout_ms is 0 (new default), delete qbuf directly in fetch_limit

hmmr:
Because query buffers are not reusable at the moment, it makes sense to delete them immediately.


hmmr:
Closing, reopened against riak_ts-integration in https://github.com/basho/riak_kv/pull/1598.

PR #1585, opened 2016-12-21
qry_buffers_ldb: use async eleveldb:put

hmmr:
This is likely to improve the latencies of queries involving query buffers.


hmmr:
This change is incorporated in https://github.com/basho/riak_test/pull/1257.

PR #1586, opened 2016-12-21
place qbufs in a unique dir under timeseries_query_buffers_root_path

hmmr:
With associated changes to ensure that it is deleted on shutdown.

This may be relevant for setups where `timeseries_query_buffers_root_path` is set to some directory outside of `"$(platform_data_dir)`, which can potentially be shared between other nodes.


hmmr:
Retracting this PR for now because of anticipated move to single-instance leveldb for query buffers.

PR #1587, opened 2016-12-24
deferred eleveldb ops for query buffers (in-mem ORDER BY/LIMIT)

hmmr:
RTS-1641

Associated tests in https://github.com/basho/riak_test/pull/1257.

The underlying idea is to keep the accumulated data in memory until memory shortage occurs in the qry_buffers gen_server process. The two costliest operations, `eleveldb:open` and `:put` (as well as sext en-/decoding) are thus deferred, and for short queries and under light load, may never actually take place.

The heuristic to determine when to start dumping accumulated query results to leveldb is, in this initial implementation, simply a check of

      (heap_size - stack_size) < HeapSizeLimit

with the limit set in riak.conf by parameter `riak_kv.timeseries_query_buffers_inmem_max` (10M by default).


priv/riak_kv.schema
```diff
@@ -929,6 +929,20 @@
      (_) -> false
   end}.
 
+%% @doc Max heap size before dumping query buffers to ldb
```
macintux, line 932: Using the `bytesize` data type would be more user-friendly. E.g.,

```
{mapping, "anti_entropy.write_buffer_size", "riak_kv.anti_entropy_leveldb_opts.write_buffer_size", [
  {default, "4MB"},
  {datatype, bytesize},
  hidden
]}.
```

src/riak_kv_qry_buffers.erl
```diff
@@ -169,8 +172,11 @@ offset_to_scalar([A]) when is_integer(A), A >= 0 -> A.
           %% State; can be overridden)
           expire_msec :: non_neg_integer(),
 
-          %% leveldb handle for the temp storage
-          ldb_ref :: eleveldb:db_ref(),
+          %% leveldb handle for the temp storage (or, initially, a function to obtain one)
+          ldb_ref :: eleveldb:db_ref() | fun(() -> {ok, eleveldb:db_ref()} | {error, term()}),
```
macintux: I am unenthused by overloading this field in such a way.

src/riak_kv_qry_buffers.erl
```diff
@@ -420,55 +427,97 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
+          Data,
+          TotalSize, HardWatermark, InmemMax) ->
     ChunkSize = compute_chunk_size(Data),
     case TotalSize + ChunkSize > HardWatermark of
         true ->
             {error, total_qbuf_size_limit_reached};
         false ->
+            %% update stats/counters
+            ChunkId = ChunksGot0,
+            ChunksGot = ChunksGot0 + 1,
```
macintux, line 458: The comment after this line is a few lines down from where it should be. Also the tone is odd: does it mean that currently the ID is *not* the serial number of the subquery? Please make it more definitive.

src/riak_kv_qry_buffers.erl
```diff
@@ -420,55 +427,97 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
             %% ChunkId will be used to construct a new and unique key
             %% for each record. Ideally, this should be the serial
             %% number of the subquery.
-            ChunkId = ChunksGot0,
-            ChunksGot = ChunksGot0 + 1,
-            lager:debug("adding chunk ~b of ~b", [ChunksGot, ChunksNeed]),
+            IsReady = (ChunksNeed == ChunksGot),
```
macintux: I really don't like "is ready" (both as binding and field name) for this value. `all_chunks_received` would be better, if I'm understanding its role correctly.

I'm also a bit confused why `do_delete_qbuf` refuses to delete an incomplete query buffer. Is there a workflow where premature deletions are avoided this way, or are we just introducing resource leaks?

src/riak_kv_qry_buffers.erl
```diff
@@ -420,55 +427,97 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
-                                      last_accessed = os:timestamp()},
+            KeyedData = enkey(Data, KeyFieldPositions, OrdByFieldQualifiers, ChunkId),
+
+            QBuf1 = QBuf0#qbuf{size          = Size + ChunkSize,
+                               chunks_got    = ChunksGot,
+                               total_records = TotalRecords0 + length(Data),
+                               is_ready      = IsReady,
+                               last_accessed = os:timestamp()},
+
+            HaveLdbOpened = not is_function(LdbRefOrDelayedCreateFun),
```
macintux: Suggest `IsBackendOpened`: we use `is` not `have` for booleans, and the fact that it's leveldb is an implementation detail (that admittedly we're unlikely to ever change, but `Ldb` isn't quite as intuitive as `Backend`).

src/riak_kv_qry_buffers.erl
```diff
@@ -420,55 +427,97 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
+                            ok ->
+                                QBuf = QBuf1#qbuf{inmem_buffer = [],
+                                                  ldb_ref      = LdbRef},
+                                {ok, QBuf};
+                            {error, _} = ErrorReason ->
+                                ErrorReason
+                        end
+                end,
+
+            case (HaveLdbOpened orelse not can_afford_inmem(InmemMax)) of
```
macintux: This function is much too long, and this seems like one obvious place to introduce a helper function with multiple heads. Even with that I suspect there are clean ways to make yet shorter functions out of this.

src/riak_kv_qry_buffers.erl
```diff
@@ -420,55 +427,97 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
+                    case LdbRefOrDelayedCreateFun() of
+                        {ok, LdbRef} ->  %% now it's a real ref
+                            EleveldbPutF(LdbRef);
+                        {error, _} = ErrorReason ->
+                            ErrorReason
+                    end
             end
     end.
 
+can_afford_inmem(Threshold) ->
```
macintux, line 520: Perhaps I'm overlooking the obvious, but why are we not tracking the data size as chunks arrive and comparing against that, instead of incurring the overhead of checking the process information?

src/riak_kv_qry_buffers.erl
```diff
@@ -660,6 +719,44 @@ tstamp() ->
 
 
 %% data ops
+
```
macintux, line 746: I would really appreciate a spec attached to this function. The inputs and logic are complex enough, I feel, to warrant it.

src/riak_kv_qry_buffers.erl
```diff
@@ -420,55 +427,97 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
+                                                  ldb_ref      = LdbRef},
+                                {ok, QBuf};
+                            {error, _} = ErrorReason ->
+                                ErrorReason
+                        end
+                end,
+
+            case (HaveLdbOpened orelse not can_afford_inmem(InmemMax)) of
+                false ->
+                    InmemBuffer =
```
macintux: We should consider using an `ordered_set` ETS table instead of storing this directly in the process. Unclear what the performance difference would be (but hey, we have people for that) but we'd gain process crash survivability and perhaps garbage collection would be a bit less onerous since we wouldn't be bloating the process to such a large degree.

(/cc @JeetKunDoug who pointed out that ETS is worth considering. All errors in assumptions here are mine alone.)

src/riak_kv_qry_buffers.erl
macintux: We *are* tracking, so it's just a question of using that data for comparison.

That would also allow us to use ETS.

priv/riak_kv.schema
hmmr: Done in 75b04e7.

src/riak_kv_qry_buffers.erl
hmmr: Unoverloaded in 7766904.

src/riak_kv_qry_buffers.erl
hmmr: Clarified in 532a408.

src/riak_kv_qry_buffers.erl
hmmr: Renamed identifiers in 73e0697 .

Incomplete query buffers are deleted when a query is cancelled (https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_kv_qry_worker.erl#L422), so indeed, you are spot on. Fixed in 3658c00.

src/riak_kv_qry_buffers.erl
hmmr: Renamed in 3ce3bf00.

src/riak_kv_qry_buffers.erl
hmmr: Split that function in two in 4351a2a2.

src/riak_kv_qry_buffers.erl
hmmr: Added spec in a9f5f335.

src/riak_kv_qry_buffers.erl
hmmr: There is no regard for crash recovery for query buffers by design.

Query buffer manager runs on the same node where coordinator runs; if it goes down, the client will lose the query, and when the same query is reissued, the recovered buffer will be of no use because all query ids are unique (they are constructed with `term_to_binary(make_ref())`).

If the riak_kv_qry_buffers application were to crash on its own, it will purge any existing query buffers on subsequent init.

And if an IO error occurs, sending the client an "internal qbuf error" is the lesser evil than blocking the query buffer manager gen_server while recovery is in progress.

Whether ETS can lessen the burden on gc is an open question, but any gains here would have to be evaluated against (1) more code to execute (`lists:sublist(Acc, Offset, Limit)` vs lots of `ets:lookup`s or a `:fold` with counter), (2) increased code complexity.

priv/riak_kv.schema
```diff
@@ -929,6 +929,12 @@
...
+{mapping, "riak_kv.query.timeseries.qbuf_inmem_max", "riak_kv.timeseries_query_buffers_inmem_max", [
+  {default, "10MB"},
+  {datatype, bytesize}
```
hazen, line 935: Agree with John. Nice addition.

src/riak_kv_qry_buffers.erl
```diff
@@ -387,27 +396,28 @@ do_delete_qbuf(QBufRef, #state{qbufs = QBufs0,
     case get_qbuf_record(QBufRef, QBufs0) of
         false ->
             {reply, {error, bad_qbuf_ref}, State0};
-        #qbuf{is_ready = false} ->
-            {reply, {error, qbuf_not_ready}, State0};
+        #qbuf{ldb_ref = undefined} ->
+            {reply, ok, State0#state{qbufs = lists:keydelete(QBufRef, 1, QBufs0)}};
```
hazen, line 400: This is just to clean up any old references which might be sticking around?  Just making sure this is a possible case.

src/riak_kv_qry_buffers.erl
```diff
@@ -420,49 +430,107 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
             {error, total_qbuf_size_limit_reached};
         false ->
-            %% ChunkId will be used to construct a new and unique key
-            %% for each record. Ideally, this should be the serial
-            %% number of the subquery.
+            %% update stats/counters: ChunkId will be used to
+            %% construct a new and unique key for each record (because
+            %% there may be multiple records with the same key
+            %% constructed from columns BY which to ORDER).  This is
+            %% not the subquery id, but it's ok as long as we entuple
```
hazen: Since `entuple` isn't a real word, perhaps something like "construct a tuple from the new key first.  This encapsulation happens in..."

src/riak_kv_qry_buffers.erl
```diff
@@ -420,49 +430,107 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
+                                          ldb_ref      = Ref},
+                        {ok, QBuf};
+                    {error, _} = ErrorReason ->
+                        ErrorReason
+                end
+        end,
+
+    IsBackendOpened = (LdbRef /= undefined),
+    case (IsBackendOpened                %% already storing to backend: keep doing so
+          orelse not CanAffordInMem) of  %% can't keep in memory: switch to backend
```
hazen: This `case` statement is rather confusing.  So in the `false` case the backend is not open and we can afford in-memory, right?  The `true` case is when the backend is not opened, but we cannot afford to keep it in memory.  The middle case with the `when` clause if relatively straightforward.  I'd suggest either breaking this up, or at least adding more comments to the 3 cases to clarify the actions.

src/riak_kv_qry_buffers.erl
```diff
@@ -420,49 +430,107 @@ do_batch_put(QBufRef, Data, #state{qbufs          = QBufs0,
...
+    PInfo = process_info(self(), [heap_size, stack_size]),
+    HeapSize = proplists:get_value(heap_size, PInfo),
+    StackSize = proplists:get_value(stack_size, PInfo),
+    Allocated = HeapSize - StackSize,
+    Allocated < Threshold.
+
 get_ordby_field_qualifiers({_, AscDesc, Nulls}) ->
     {AscDesc, Nulls}.
 
+maybe_only_rows(KeyedRows, _StripKeysUsedForSorting = false) ->
```
hazen, line 530: Like the clarifying argument names

src/riak_kv_qry_buffers.erl
```diff
@@ -660,6 +738,57 @@ tstamp() ->
...
+
+-spec enkey([data_row()],
+            KeyFieldPositions::[non_neg_integer()],
+            OrdByFieldQualifiers::[{asc|desc, nulls_first|nulls_last}],
+            ChunkId::non_neg_integer()) ->
+                   [{{KeyOrd::riak_pb_ts_codec:ldbvalue(),
+                      ChunkId::non_neg_integer(),
+                      Idx::non_neg_integer()},
+                     data_row()}].
+enkey(Rows, KeyFieldPositions, OrdByFieldQualifiers, ChunkId)
```
hazen, line 755: And it passes dialyzer. Impressive!

hmmr:
Closing and reopening as https://github.com/basho/riak_kv/pull/1599.

hazen (commented):
See about moving these comments to the new PR

src/riak_kv_qry_buffers.erl
hmmr: It's for incomplete qbufs (which have not received the complete set of chunks, and are not ready for `fetch_limit`), This actually happens when a query started to be executed but got cancelled. John has hit a bug.

src/riak_kv_qry_buffers.erl
hmmr: You won't believe it, but it passed on the first run!

src/riak_kv_qry_buffers.erl
hmmr: Reworded comment in 29d95943.

src/riak_kv_qry_buffers.erl
hmmr: Better comments, now at each `case` branch, in ed8a4c6b.

src/riak_kv_qry_buffers.erl
hmmr: A good means of expression indeed, if used sparingly.

macintux:
This branch will need to be updated against integration to fix protobuffs issues.

priv/riak_kv.schema
macintux: One more comment, sorry this didn't occur to me sooner. Ideally I think this would be a per-node setting, rather than per-query, but I think that can be tabled for a future release. However, the fact that this setting is per-query needs to be explicit in the comment in the schema file and in the docs.

macintux:
I believe we've indicated that the public docs update must be done before this can be considered mergeable; are there docs somewhere?

src/riak_kv_qry_buffers.erl
```diff
@@ -660,6 +743,57 @@ tstamp() ->
...
+         %%    whereas the chunk it refers to may certainly not be
+         %%    increasing because of out-of-order arrivals from the
+         %%    vnodes.
+         Key = {KeyOrd, ChunkId, Idx},
+         %% d. Encode the record (don't bother constructing a
+         %%    riak_object with metadata, vclocks):
+         {Key, Row}
+     end || {Row, Idx} <- RowsIndexed].
+
+make_sorted_keys([], desc, nulls_first) ->
```
macintux, line 780: The first 4 function heads can be collapsed to two; `nulls_first` (or `_last`) is the only parameter that matters on an empty list.

Having said that, I'm also curious whether that's the intent. It feels like `nulls_last` should put them first if also `desc`.

Please either collapse the heads or fix the logic.

src/riak_kv_qry_buffers.erl
```diff
@@ -688,11 +822,14 @@ touch_qbuf(QBufRef, State0 = #state{qbufs = QBufs0}) ->
     State0#state{qbufs = lists:keyreplace(
                            QBufRef, 1, QBufs0, {QBufRef, QBuf9})}.
 
-kill_ldb(RootPath, Table, LdbRef) ->
+kill_ldb(RootPath, Table, LdbRef) when LdbRef /= undefined ->
     ok = riak_kv_qry_buffers_ldb:delete_table(Table, LdbRef, RootPath),
+    ok;
+kill_ldb(_RootPath, _Table, _DelayedAndNeverCalledFun) ->
```
macintux, line 828: Very minor nitpick and you're free to ignore, but it feels more common to put the `undefined` head first instead of doing a guard.

hmmr:
Changes to the documentation are in https://github.com/basho/basho_docs/pull/2409.

src/riak_kv_qry_buffers.erl
hmmr: Simplified and *corrected* in 0544675.

Now nulls are sorted as described in https://drive.google.com/file/d/0B-2H_9GJ40JpMDAwY0VyWjkwSWc (page 617 in the pdf version of it). Specifically:

ASC NULLS LAST yields
```1, 2, 3, null, null```
ASC NULLS FIRST yields
```null, null, 1, 2, 3```
DESC NULLS LAST yields
```null, null, 3, 2, 1```
DESC NULLS FIRST yields
```3, 2, 1, null, null```

The test also needed to be updated: done in https://github.com/basho/riak_test/pull/1257/commits/c595fe7e0f5a8d8d53c18f34383bc499f5bae73b.

src/riak_kv_qry_buffers.erl
hmmr: This function will be gone in https://github.com/basho/riak_kv/pull/1601, so I'll spare a commit. Will take note of the guard style guide item.

macintux:
+1

hazen:
:+1:

PR #1588, opened 2016-12-30
[RFC] Conditional CRDT read requests

lucafavatella:
(I am currently designing an application that is kind of document store, where documents are stored in Riak KV as CRDTs, updated in Riak quite infrequently and cached in mobile devices for offline reading. I am currently evaluating cacheability of CRDTs in mobile devices - hence this ticket.)

**Problem statement**

From [RFC 7232 section 1](https://tools.ietf.org/html/rfc7232#section-1) i.e. HTTP/1.1 “Conditional Requests”:
> Conditional GET requests are the most efficient mechanism for HTTP cache updates ...

Riak KV has conditional read requests for plain objects - both via the PB and HTTP APIs:
* In PB API, see request parameters [`if_modified`](http://docs.basho.com/riak/kv/2.2.0/developing/api/protocol-buffers/fetch-object/) that is meant to be the Riak object vclock of a previous read response;
* In HTTP API, see request parameter [`If-None-Match`](http://docs.basho.com/riak/kv/2.2.0/developing/api/http/fetch-object/) that is meant to be `ETag` header of a previous read response. Notice that the HTTP read response has both the Riak object vclock `X-Riak-Vclock` and the entity tag `ETag`, and notice also the usage of Riak object vclock and entity tag in the [write request](http://docs.basho.com/riak/kv/2.2.0/developing/api/http/store-object/).
  * KV generates entity tag specifying webmachine callback function [`generate_etag`](https://github.com/webmachine/webmachine.github.io/blob/44d032b6da614ce0e5728b0cabf65b6f66ad6ed9/resources.md) implemented [in KV](https://github.com/basho/riak_kv/blob/2.1.6/src/riak_kv_wm_object.erl#L1033-L1048) and also called [inside other callbacks](https://github.com/basho/riak_kv/blob/2.1.6/src/riak_kv_wm_object.erl#L793) by KV;
  * (Riak KV also has timestamp-based conditional requests but I am ignoring that on purpose as not relevant for this ticket.)

Riak KV has no conditional read requests for CRDTs:
* PB read API has no conditional read - see [KV doc](http://docs.basho.com/riak/kv/2.2.0/developing/api/protocol-buffers/dt-fetch/) and Erlang [client](https://github.com/basho/riak-erlang-client/blob/2.5.1/src/riakc_pb_socket.erl#L1227) [code](https://github.com/basho/riak_pb/blob/2.2.0.2/src/riak_pb_dt_codec.erl#L226-L256);
* HTTP read API has [no](https://github.com/basho/riak_kv/blob/2.1.6/src/riak_kv_wm_crdt.erl#L149-L161) conditional read.

**Proposal**

I propose to enhance the Riak KV HTTP API wrt CRDT with conditional read requests.

I put together an initial *untested rough* patch and the code seems simple - the problem I have is getting the full understanding of the impact of the change. The impact I see is:
* (A) Clients issuing GET or HEAD requests to a CRDT will get a new header `ETag`. This shall be acceptable;
* (B) Unknown impact on HTTP/webmachine behaviour different from GET/HEAD requests;
* (C) Unclear which tests would be needed for accepting this change.

For (B) it appears the change I made shall not impact anything else except GET/HEAD requests.
* The patch if no Etag can be generated returns `undefined` that is the same as the [default return value](https://github.com/basho/webmachine/blob/1.10.8-basho1/src/webmachine_resource.erl#L103-L104);
* Webmachine appears not to make distinction on whether the callback `generate_etag` is unexported [vs.](https://github.com/basho/webmachine/blob/1.10.8-basho1/src/webmachine_resource.erl#L160) is exported returning `undefined` - apart from log messages.
I based my investigation on Riak KV 2.2.0 that has [webmachine 1.10.8-basho1](https://github.com/basho/webmachine/tree/1.10.8-basho1) (checked via chain or rebar.config fails, terminating in https://github.com/basho/riak_api/blob/2.1.5/rebar.config#L6). In webmachine, the string `generate_etag` is only in [`webmachine_resource`](https://github.com/basho/webmachine/blob/1.10.8-basho1/src/webmachine_resource.erl) and [`webmachine_decision_core`](https://github.com/basho/webmachine/blob/1.10.8-basho1/src/webmachine_decision_core.erl).

### DIFF
--- a/src/riak_kv_wm_crdt.erl
+++ b/src/riak_kv_wm_crdt.erl
@@ -157,7 +157,8 @@
          encodings_provided/2,
          resource_exists/2,
          process_post/2,           %% POST handler
-         produce_json/2           %% GET/HEAD handler
+         produce_json/2,           %% GET/HEAD handler
+         generate_etag/2
         ]).
 
 -include_lib("webmachine/include/webmachine.hrl").
@@ -402,6 +403,22 @@ produce_json(RD, Ctx=#ctx{module=Mod, data=RObj, include_context=I}) ->
     Body = riak_kv_crdt_json:fetch_response_to_json(
                      Type, Value, get_context(RespCtx,I), ModMap),
     {mochijson2:encode(Body), RD, Ctx}.
+
+generate_etag(RD, Ctx=#ctx{module=Mod, data=RObj, method=Method})
+  when Method == 'GET'; Method == 'HEAD' ->
+    {{RespCtx, _}, _} = riak_kv_crdt:value(RObj, Mod),
+    ETag =
+        case RespCtx of
+            undefined ->
+                undefined;
+            <<>> ->
+                undefined;
+            _ when is_binary(RespCtx) ->
+                base64:encode(RespCtx) %% HTTP/1.1 ETag (https://tools.ietf.org/html/rfc7232#section-2.3) is valid Base64 (https://tools.ietf.org/html/rfc2045#section-6.8).
+        end,
+    {ETag, RD, Ctx};
+generate_etag(RD, Ctx) ->
+    {undefined, RD, Ctx}.
 
 %% Internal functions
 


lucafavatella:
Not sure why Dialyzer fails but looks unrelated to this PR. Maybe you shall make [PLT target](https://github.com/basho/riak_kv/blob/46372dc6a83fc0fbfc1da414bd2da22d7817ed85/tools.mk#L57-L63) as PHONY?

lucafavatella:
Superseded by #1668

PR #1589, opened 2017-01-03
In-memory SQL `LIMIT`

andytill:
When a query contains a `LIMIT` clause and optionally an `OFFSET` but no `ORDER BY` then Riak TS no longer creates a temporary table.

It now collects the results as it would for a normal query and then uses the `OFFSET` and `LIMIT` if provided to take a slice of the query results. For queries containing `LIMIT` but not `ORDER BY`, latency is reduced from 30x to 50x to 10%. The increase is measured from a query not containing a `LIMIT` clause.

Performance tests with eleveldb `LIMIT`:

image

Performance tests with in-memory `LIMIT`:

image

This does not affect queries that have `ORDER BY` and `LIMIT` clauses.

The function head `riak_kv_qry_worker:estimate_query_size/1` which checks for a `LIMIT` has been removed. This has an existing bug where it is multiplying `LIMIT` by the chunk size, meaning that the estimated result size would be much larger than what it would otherwise calculate and making it more likely to cancel the query.


hmmr:
There is a more comprehensive solution in https://github.com/basho/riak_kv/pull/1587, which also supports ORDER BY and will work with arbitrary ranges.

That said, I am going to +1 this PR on the assumption that we will find ways before 1.6 release to (1) efficiently stop the workers after the required LIMIT records have been collected, and (2) drop records before we reach OFFSET.

One big merit I see going for this simple implementation is that it will go really well in conjunction with descending keys, in situations where the ingest rate is stable and constant (LIMIT records are guaranteed to be found in a range that is small enough to fit in memory), and the prevailing queries are LAST-like,

hmmr:
+1

macintux:
I'd be a bit nervous about adding this if we weren't guaranteed to add some overload protections (apparently in Andrei's pending changes) before the next release. Should we delay merging until we're sure that will be incorporated?

hmmr:
@macintux The overload protection remains there -- it's just not taking into account the LIMIT value. The projected query result size is still checked against the configured limit (in the other head of `estimate_query_size`). The upshot is that we will be conservatively cancelling queries based on the entire WHERE range even where LIMIT is small and manageable -- but again, this is fine because in this implementation, the entire WHERE range will have to be kept in memory.

src/riak_kv_qry_worker.erl
```diff
@@ -506,6 +487,26 @@ prepare_final_results(#state{
...
     end.
 
+maybe_apply_offset_limit(?SQL_SELECT{'OFFSET' = [], 'LIMIT' = []}, Rows) ->
+    Rows;
+maybe_apply_offset_limit(?SQL_SELECT{'OFFSET' = [Offset], 'LIMIT' = []}, Rows) when is_integer(Offset), Offset >= 0 ->
+    safe_nthtail(Offset, Rows);
+maybe_apply_offset_limit(?SQL_SELECT{'OFFSET' = [], 'LIMIT' = [Limit]}, Rows) when is_integer(Limit), Limit >= 0 ->
+    lists:sublist(Rows, Limit);
+maybe_apply_offset_limit(?SQL_SELECT{'OFFSET' = [Offset], 'LIMIT' = [Limit]}, Rows) when is_integer(Offset), is_integer(Limit), Offset >= 0, Limit >= 0 ->
+    lists:sublist(safe_nthtail(Offset, Rows), Limit);
```
hmmr, line 497: Why not use `liists:sublist/3`, like so:
```
1> lists:sublist(_Result = [a,s,d,f], _Offset = 2, _Limit = 10).
[s,d,f]
```

src/riak_kv_qry_worker.erl
```diff
@@ -447,11 +429,11 @@ subqueries_done(QId, #state{qid          = QId,
                                     [[riak_pb_ts_codec:ldbvalue()]]}.
 prepare_final_results(#state{qbuf_ref = undefined,
                              result = IndexedChunks,
-                             qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = rows} = Select}}) ->
+                             qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = rows} = Select} = Query}) ->
     %% sort by index, to reassemble according to coverage plan
     {_, R2} = lists:unzip(lists:sort(IndexedChunks)),
```
hmmr: How does the naive sorting performed here play with the DESC qualifier in the PK definition? Consider a query with WHERE range that fetches 20 records, and consider it has a LIMIT value of 10. If the interesting column has been declared DESC in the DDL, then is the expectation to retrieve the top ten records, or the bottom ten? It appears it will be the latter.

Incidentally, this is also the behaviour in https://github.com/basho/riak_kv/pull/1587, but it has the resulting order specified separately, explicitly or implied via ASC/DESC qualifier in the ORDER BY clause. I am wondering if this should be clarified in the official docs (https://github.com/basho/private_basho_docs/pull/299).

Still, it would probably be great if,_ in absence of ORDER BY_, the DESC in the DDL did take effect. It would enable retrieving the actual last records, and thus easily emulate LAST.

src/riak_kv_qry_worker.erl
hmmr: The first paragraph in the comment above is me being inattentive. Sorting is not done on records within chunks, which I managed managed to miss.

andytill:
Closing cus.

PR #1590, opened 2017-01-05
Queries on descending keys tables sometimes do not return results RTS-1678

andytill:
There was a check on the keys to work out if the start/end keys should be swapped, sometimes this was returning false (do not swap) when it should have returned true. Date for descending keys is stored like.

```
(1,2,-3)
(1,2,-2)
(1,2,-1)
```

Where the last column is descending and the inserted values were un-negated e.g. `(1,2,3)` etc. And the query start/end keys need to be `c >= -3 AND C <= -1`. If the query start/end keys are **not** swapped then the range will not be possible e.g. `c >= -1 AND c <= -3`.

Symptoms are:
1. Depending on the column names, results may not be returned. Column names were mistakenly used in the data used for comparison.
2. When the `DESC` keyword is used a local key column that is not also in the partition key e.g. `PRIMARY KEY((a,b,QUANTUM(c,1,'s'),a,b,c,d))` , then no results can be returned.

This PR fixes both issues above.

### DIFF
--- a/src/riak_kv_qry_compiler.erl
+++ b/src/riak_kv_qry_compiler.erl
@@ -147,19 +147,18 @@ compile_where_clause(?DDL{} = DDL,
 %% now break out the query on quantum boundaries
 -spec expand_query(?DDL{}, ?SQL_SELECT{}, proplists:proplist()) ->
                           {ok, [?SQL_SELECT{}]} | {error, term()}.
-expand_query(?DDL{table = Table, local_key = LK, partition_key = PK},
-             ?SQL_SELECT{} = Q1, Where1) ->
+expand_query(?DDL{local_key = LK, partition_key = PK},
+             ?SQL_SELECT{helper_mod = Mod} = Q1, Where1) ->
     case expand_where(Where1, PK) of
         {error, E} ->
             {error, E};
         {ok, Where2} ->
-            Mod = riak_ql_ddl:make_module_name(Table),
-            IsDescending = lists:member(descending, Mod:field_orders()),
+            IsDescending = is_partition_key_cols_descending(Mod:field_orders(), PK),
             SubQueries1 =
                 [Q1?SQL_SELECT{
                        is_executable = true,
                        type          = timeseries,
-                       'WHERE'       = maybe_fix_start_order(IsDescending, X, Mod, LK),
+                       'WHERE'       = maybe_fix_start_order(IsDescending, X),
                        local_key     = LK,
                        partition_key = PK} || X <- Where2],
             SubQueries2 =
@@ -171,6 +170,14 @@ expand_query(?DDL{table = Table, local_key = LK, partition_key = PK},
                 end,
             {ok, SubQueries2}
     end.
+
+%% Check if any columns in the partition key have the DESC keyword applied.
+%%
+%% Only check the partition key, otherwise the start/end key does not need to be
+%% flipped. The data is not stored in a different order to the start/end key.
+is_partition_key_cols_descending(FieldOrders, #key_v1{ast = AST}) ->
+    PKOrders = lists:sublist(FieldOrders, length(AST)),
+    lists:member(descending, PKOrders).
 
 %% Calulate the final result for an aggregate.
 -spec finalise_aggregate(#riak_sel_clause_v1{}, [any()]) -> [any()].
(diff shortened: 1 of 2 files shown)


src/riak_kv_qry_compiler.erl
```diff
@@ -482,40 +489,27 @@ extract_stateful_functions2({{window_agg_fn, FnName}, _} = Function, FinaliserLe
...
-    StartKey = riak_ql_ddl:make_key(Mod, LK,  StartVals),
-    EndKey = riak_ql_ddl:make_key(Mod, LK,  EndVals),
-    (StartKey < EndKey).
+    %% Swap the start/end keys so that the backends will
+    %% scan over them correctly.  Likely cause is a descending
+    %% timestamp field.
+    W1 = lists:keystore(startkey, 1, W, {startkey, EndKey0}),
+    W2 = lists:keystore(endkey, 1, W1, {endkey, StartKey0}),
+    W1 = lists:keystore(startkey, 1, W, {startkey, EndKey0}),
+    W2 = lists:keystore(endkey, 1, W1, {endkey, StartKey0}),
```
hmmr: Looks like an echo in code.

andytill: My excuse is that my keyboard's delete button is broken ;) (it actually is). Fixed in a64c5a50a4a0393d958eeecc34210d95fe5324ae.

src/riak_kv_qry_compiler.erl
```diff
@@ -147,19 +147,18 @@ compile_where_clause(?DDL{} = DDL,
...
         {ok, Where2} ->
-            Mod = riak_ql_ddl:make_module_name(Table),
-            IsDescending = lists:member(descending, Mod:field_orders()),
+            IsDescending = is_partition_key_cols_descending(Mod:field_orders(), PK),
             SubQueries1 =
                 [Q1?SQL_SELECT{
                        is_executable = true,
                        type          = timeseries,
-                       'WHERE'       = maybe_fix_start_order(IsDescending, X, Mod, LK),
+                       'WHERE'       = maybe_fix_start_order(IsDescending, X),
```
hmmr, line 161: It seems you are applying `maybe_fix_start_order` for columns are not declared DESC in the DDL. Is it correct?

hmmr:
+1

src/riak_kv_qry_compiler.erl
andytill: A "maybe" function is equivalent to a case expression, it does not change the start order if the local key does not have a DESC key.

hmmr: I was under assumption that the key can contain multiple fields on which a SELECT can have ranges in WHERE, but it's not the case. Apologies.

hmmr:
+1

macintux:
Looking at the newly-added tests, I don't see one for a `DESC` non-timestamp field in the partition key. Shouldn't there be one?

andytill:
@macintux Can you give an example `CREATE TABLE` statement you would like to have tested?

macintux:
I expressed that poorly; a column that exists in the partition key but flagged as `DESC` in the local key.

E.g.:

```
create table desc1 (a varchar not null, time timestamp not null, primary key ((a, quantum(time, 15, 'm')), a desc, time));
```

andytill:
@macintux thanks, I will add a test for that table.

andytill:
@macintux that was a bug, thanks!

src/riak_kv_qry_compiler.erl
```diff
@@ -172,6 +171,19 @@ expand_query(?DDL{table = Table, local_key = LK, partition_key = PK},
             {ok, SubQueries2}
     end.
 
+%% Check if any columns in the partition key have the DESC keyword applied.
```
macintux: The intro to this comment is no longer accurate.

Also the function name itself is no longer accurate, and is a bit tough to parse. What about `is_last_partition_column_descending`?

src/riak_kv_qry_compiler.erl
```diff
@@ -172,6 +171,19 @@ expand_query(?DDL{table = Table, local_key = LK, partition_key = PK},
...
+%%
+%% Only check the last column in the partition key, otherwise the start/end key
+%% does not need to be flipped. The data is not stored in a different order to
+%% the start/end key.
+%%
+%% Checking the last column in the partition key is safe while the other columns
+%% must be exactly specified e.g. anything but the QUANTUM column must be
+%% covered with an equals operator in the where clause.
+is_partition_key_cols_descending(FieldOrders, #key_v1{ast = AST}) ->
+    PKOrders = lists:sublist(FieldOrders, length(AST)),
```
macintux: `length(AST)` is a non-obvious way to determine the number of columns in the partition key (and without looking at a trace I have no idea why it works). Suggest wrapping that in a short utility function.

macintux: Ok, looking at the invocation I can figure out why it works, but would still recommend a short wrapper to give a more intuitive name to the concept.

One other minor nitpick: `lists:nth` seems like a potentially faster way of turning these two lines into 1, potentially faster.

andytill: Just to clarify, you would like a function to return the partition key length?

macintux: Untested, possibly buggy. I could live with embedding `length(AST)` with a comment as well, I just like this approach a bit better.

```
is_partition_key_cols_descending(FieldOrders, PK) ->
   lists:nth(num_cols(PK), FieldOrders) == descending.

%% The abstract syntax tree for the key record contains one element per column
num_cols(#key_v1{ast = AST}) ->
    length(AST).
```

src/riak_kv_qry_compiler.erl
andytill: Fixed in 1b54b43

src/riak_kv_qry_compiler.erl
andytill: Changes in 1b54b43

macintux:
+1

hmmr:
+1

PR #1592, opened 2017-01-09
Improve sweeper tests

raghavkarol:
- 10x speedup in running time from ~100 secs to ~10 secs.

- removed calls `match_retry(fun riak_kv_sweeper:status/0, [{}, {}]),`
  Expected match `[{}, {}]` was never returned by the fun in arg 1, so
  they delayed the tests ~15 seconds.

- Moved sweeper tests to test directory. This commit reverts 3bcd65d
  but avoids the problem of auto-discovery by `rebar eunit` using
  `test_` as prefix for all tests instead of `_test` suffix
  auto-discovered by eunit.

- `meck:new()` calls are slow (~500 msecs on the development mac) and
  the tests were calling meck:new in `init_per_testcase`. All calls to
  `meck:new` moved to a separate process started in
  `init_per_suite`. `meck:new()` can't be called directly in
  `init_per_suite` because `init_per_suite` is run in a separate
  process that terminates, killing all mocked modules, when
  `init_per_suite` completes.

- removed `after` clause in `receive_msg` and changed `timetrap` to 5
  seconds. If an individual testcase needs to run longer, the timetrap
  for it should be increased.

- Used real modules instead of mocked ones for test sweep
  participants.

- Application environment variable `sweep_participant_testing_expire`
  when set to `immediate` overrides ignores the `run_interval` so we
  avoid the need to wait for `now - last_run > run_interval` to be
  true.

- Use timetrap of 10 seconds.

- Use exponetial slow down for match_retry function. On the
  development macs very few retries are needed whereas more and longer
  retries are needed on the buildbots.

- Change receive_msg to poll and using `match_retry`. On buildbot we
  often timeout waiting to receive messages which are in the process
  queue.


nickelization:
Great work! One small suggestion I would have for stuff like this in the future: if you're going to move a file, try to do it in a separate commit, because otherwise it's really hard to see what other changes were made to that file. I wouldn't waste time fixing it at this point for this PR, just a tip for the future.

test/sweeper_meck_helper.erl
```diff
@@ -0,0 +1,198 @@
...
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
```
nickelization: If you hadn't already explained to me why this is needed, I think I might be looking at this file and wondering why we need a separate process for setting up meck. I'd suggest just adding a small comment here to describe the purpose of this module.

test/riak_kv_sweeper_SUITE.erl
```diff
@@ -0,0 +1,892 @@
...
+    sweeper_callback_wait:before_visit(any, fun(_) ->
+                                                    Running = riak_kv_sweeper:stop_all_sweeps(),
+                                                    true = Running >= 1
+                                            end),
+    Running = [riak_kv_sweeper_state:sweep_index(Sweep) ||
+                  Sweep <- Sweeps,
+                  riak_kv_sweeper_state:sweep_state(Sweep) == running],
+    [ok = receive_msg({ok, failed_sweep, sweeper_callback_wait, I}) || I <- Running],
+    ok = check_sweeps_idle(Indices),
+    ok.
```
nickelization: We could eliminate this extra line, since the previous expression already evaluates to `ok`.

test/riak_kv_sweeper_SUITE.erl
```diff
@@ -0,0 +1,892 @@
...
+assert_all_indices_idle(Indices) ->
+    {_, Sweeps} = riak_kv_sweeper:status(),
+    SweepData = [{riak_kv_sweeper_state:sweep_state(S),
+                  riak_kv_sweeper_state:sweep_index(S)} ||
+                    S <- Sweeps],
+    SweepIndices = lists:sort([I || {idle, I} <- SweepData]),
+    SweepIndices = Indices.
+
+receive_msg(Msg) ->
+    ok = match_retry(
```
nickelization, line 863: This change I don't quite understand. Why do a `match_retry`, and not just a `receive` with an appropriate timeout? I see the note you put about this in the PR, but I don't think I understand what kind of issue this change is designed to fix.

test/riak_kv_sweeper_SUITE.erl
```diff
@@ -0,0 +1,892 @@
...
+match_retry(Max, N, _Sleep, _Fun, {'_', _Result}) when N > Max ->
+    max_retries;
+
+match_retry(Max, N, Sleep, Fun, {'_', Result}) ->
+    case Fun() of
+        Result ->
+            Result;
+        _ ->
+            timer:sleep(trunc(Sleep)),
+            match_retry(Max, N+1, max(1000, Sleep*1.1), Fun, {'_', Result})
```
nickelization, line 897: What benefit do we get from the exponential backoff? It seems like unnecessary complexity, and if we slow down on the retries then that may slow down the test a little. Could we stick with a constant sleep, and just increase the number of retries instead?

test/sweeper_meck_helper.erl
raghavkarol: Good idea, adding a comment

test/riak_kv_sweeper_SUITE.erl
raghavkarol: When running tests on `buildbot` I have observed that sometimes messages are  not received by the waiting process but are in the message queue, causing the test case to fail. Adding debug printing, `ct:pal(...)`,  to the `receive` loop causes the messages to be delivered. So, I suspect it's a scheduling issue and adding the `match_retry` and sleeping is to facilitate the other process being scheduled. 

Tests ran much more stable on `buildbot` after this change, but that might be co-incidental and not causal. I agree it's very confusing.

test/riak_kv_sweeper_SUITE.erl
raghavkarol: Exponentially increasing the delay by 10% gives minimal wait time on both development macs and `buildbot`.

Using a constant value e.g. , 500 ms slows all tests using this function by at least 500 ms and ads adds ~20 secs to the total time for tests to run on our development macs. The development mac's need very few (1-2), short waits for the test to execute correctly.

test/riak_kv_sweeper_SUITE.erl
nickelization: Boy, that's pretty weird, and definitely makes me uneasy. I guess if it works around the issue then I'm okay with putting this code in, but I would definitely add some comments around it explaining why we're doing this, otherwise this code looks very strange and confusing to me.

test/riak_kv_sweeper_SUITE.erl
nickelization: Sure, I get that, but is it a problem to just keep checking very quickly? The code you wrote starts with a 100 ms delay, then backs off from there. Why not just keep checking every 100 ms until it succeeds? It would chew a little more CPU, but if it makes the test finish faster and simplifies our test code, that seems like it might be a worthwhile tradeoff.

test/riak_kv_sweeper_SUITE.erl
raghavkarol: I've added a comment, and I agree it's super confusing.

test/riak_kv_sweeper_SUITE.erl
raghavkarol: It's a problem on `buildbot`, where a 100 msec delay and 100 wait's still timeout.

nickelization:
+1

PR #1593, opened 2017-01-10
Fix dialyzer error on develop branch

raghavkarol:


nickelization:
I'm never happy to be ignoring dialyzer warnings, but if we were already ignoring this and we just need to update the spec for the ignore, I guess that's okay.

nickelization:
+1

PR #1594, opened 2017-01-11
Use query filters in start/end key RTS-1510

andytill:
Depends on https://github.com/basho/riak_ql/pull/163

Given a primary key with an additional column on the local key:

```sql
PRIMARY KEY((QUANTUM(a,1,'s')),a,b);
```

With where clause:

```sql
WHERE a = 1 AND b >= 2 AND b <= 5
```

And row keys:

```erlang
{1,1}
{1,2}
{1,3}
{1,4}
{1,5}
{1,6}
```

Then the start key passed to eleveldb would be `{1,_}` and the end key would be `{1,_}`. All of the rows would be loaded, msgpack decoded and have the filter run on it.

With this change, filters on the local key are put into the start/end keys to reduce the range that eleveldb has to cover. Now the start key for this query is `{1,2}` and the end key is `{1,5}`. In a much larger key set this could save many objects being loaded from disk. The more filters that a user can give on columns in the local key, the better the less work the system has to do to respond.

The original aim was to remove filters that could be put into the start/end keys but this is not possible, some values may fall into the range that would otherwise be filtered. This is similar to the requirement of the quantum being on the last column of the partition key.

To pick which filter should be used for the start and end keys, the filters are reduced to a maximum of two per column. One for the start and one for the end key. With these new filter checks, it is easy to check for queries that are not possible e.g. checking that a cell is null when it is defined in table schema as `NOT NULL`. First described in https://github.com/basho/rfc/pull/33.


src/riak_kv_qry_compiler.erl
```diff
@@ -1097,9 +1110,378 @@ modify_where_key(TupleList, Field, NewVal) ->
...
+        #filtercheck{'=' = F} ->
+            %% this filter has been specified twice so remove the second occurence
+            {eliminate, Acc};
+        #filtercheck{'>' = {_,_,{const,GtVal}} = GtF} = Check1 when GtVal < EqVal ->
+            %% query like `a = 10 AND a > 8` we can eliminate the greater than
+            %% clause because if a to be 10 it must also be greater than 8
+            Acc2 = append_to_eliminate_later(GtF, Acc),
+            Check2 = Check1#filtercheck{'>' = undefined, '=' = F},
+            {F, lists:keystore(FieldName, #filtercheck.name, Acc2, Check2)};
+        #filtercheck{'>=' = {_,_,{const,GteVal}} = GteF} = Check1 when GteVal < EqVal ->
```
paegun: '>' and '>=' and '<' and '<=' cases that differ only in the operator's inclusion of '=' should be DRY'd up using guards b/c there's more than a simple line of code w/i the case's statement(s) list.

src/riak_kv_qry_compiler.erl
```diff
@@ -1097,9 +1110,378 @@ modify_where_key(TupleList, Field, NewVal) ->
...
+            %% clause because if a to be 10 it must also be greater than 8
+            Acc2 = append_to_eliminate_later(GteF, Acc),
+            Check2 = Check1#filtercheck{'>=' = undefined, '=' = F},
+            {F, lists:keystore(FieldName, #filtercheck.name, Acc2, Check2)};
+        #filtercheck{'<' = {_,_,{const,LtVal}}} when LtVal =< EqVal ->
+            %% query requires a column to be equal to a value AND less than that
+            %% value which is impossible
+            throw({error, {impossible_where_clause, << >>}});
+        #filtercheck{'<=' = {_,_,{const,LteVal}} = LteF} = Check1 when LteVal >= EqVal ->
+            %% query like `a = 10 AND a <= 10` the less than or equals can be
```
paegun: There are more redundant cases, such as `a > 8 AND a > 10`. These redundancies are not truly part of the implied semantic of filtercheck. A different way of handling such redundancies is by having a phase following filtercheck that normalizing the filter ranges, eliminating fully contained filter ranges, i.e. {a, (10, infinity)} is fully contained by {a, (8, infinity)}.

While this change is not included in the scope of the work, despite the lack of total coverage of eliminating redundant filter cases, I do feel like it advances the product's UX and performance, reducing a good amount of redundant filter cases.

paegun:
+1

The use of throw to short circuit when an impossible filter condition is encountered is great.

check_where_clause_is_possible and underlying _fold is a decently large chunk of functionality that may be better lifted into a separate module.

check_where_clause_is_possible mapping to a `filtercheck` does make things clear, but seems to cause some not DRY code as '>=' and '>' as well as '<=' and '<' cases necessarily duplicate the list of statements w/i the case. The code overall is clear, not overly abstract. This IMO just doubles up the suggestion to lift the check_where_clause_is_possible functionality into a separate module.

Despite minor comments here, I believe this code is a great improvement and clear, so +1 .

paegun:
thumbot wait

src/riak_kv_qry_compiler.erl
```diff
@@ -1097,9 +1110,378 @@ modify_where_key(TupleList, Field, NewVal) ->
...
+%% -------------------------------------------------------------------
+
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 
+-define(
+    assertPropsEqual(Expected1, Actual1),
+    Expected2 = lists:sort(Expected1),
+    Actual2 = lists:sort(Actual1),
+    ?assertEqual(lists:sort(Expected2), lists:sort(Actual2))
```
hmmr, line 1482: Looks like you are sorting twice here.

src/riak_kv_ts_api.erl
```diff
@@ -101,8 +101,12 @@ api_calls() ->
                    {ok, riak_kv_qry:query_tabular_result()} |
                    {error, term()}.
 query(QueryStringOrSQL, DDL) ->
-    riak_kv_qry:submit(QueryStringOrSQL, DDL).
-
+    case riak_kv_qry:submit(QueryStringOrSQL, DDL) of
+        {error, {impossible_where_clause, _}} ->
+            {ok, {[],[], []}};
```
hmmr, line 106: Shouldn't this be an error?

andytill: We now have three types of where clause:
1. A valid where clause which covers the partition key
2. An invalid clause that does not cover the partition key
3. A valid clause that covers the partition key but we can pre-determine that it won't return results.

2 and 3 are quite different conditions. 2 is valid SQL which would return results in a relational DB but Riak TS cannot because of its requirements to cover the partition key, we have to return an error or an incorrect result because we can't access the actual data.

3 is valid, and in TS 1.5 it would be run but we can pre-determine that it will never return any results, so as an optimisation, we return the known result without hitting the vnodes.

src/riak_kv_qry_compiler.erl
andytill: Different bindings.

hmmr:
I can see what this PR tries to achieve and how (including why conditions are sometimes dropped outright during filter rewriting, and some other times, put on "exclude later" list). It would probably be useful to the client if combinations of conditions that the filter checks determine to be impossible were reported as an error, but then again, we have an EXPLAIN statement for that.

Will +1 as soon as it turns green (we need to approve the riak_ql dependee first).

andytill:
thumbot retry

hmmr:
+1

paegun:
+1

paegun:
@thumbot retry

PR #1595, opened 2017-01-11
Remove redundant eunit tests

raghavkarol:
These were the original tests writen by lixen and all of these cases
are covered in the ../test/riak_kv_sweeper_SUITE.erl These tests, do
not run stable on `buildbot`; at least two separate and obscure
failures have been observed, the meck process dies or eunit times out
trying to run the tests.


JeetKunDoug:
+1 - I'd like to see some additional comments in the replacement SUITE as to what the tests are testing, and how, exactly, they prove what they test though.

raghavkarol:
@JeetKunDoug updated the replacement suite as requested.

test/riak_kv_sweeper_SUITE.erl
```diff
@@ -234,10 +382,21 @@ test_scheduler_remove_participant(Config) ->
...
+%%
+%% The test cases uses the sweeper_callback_wait:before_visit to make
+%% a sweep request on an idle index when we have reached the max
+%% concurrency limit, and asserts that the requested sweep is queued.
+%%
 test_scheduler_queue(Config) ->
     Indices = ?config(vnode_indices, Config),
     new_sweep_participant(sweeper_callback_wait),
-    sweeper_meck_helper:backend_modules(async, _NumKeys=5000, _ObjecSizeBytes=0),
+    sweeper_meck_helper:backend_modules(async, _NumKeys=1000, _ObjecSizeBytes=0),
```
nickelization: I just noticed we misspell "Object" as "Objec" in a bunch of places all over this test suite.

test/riak_kv_sweeper_SUITE.erl
```diff
@@ -207,16 +309,62 @@ test_scheduler_worker_process_crashed(Config) ->
...
+%%
 test_scheduler_run_interval(Config) ->
     Indices = ?config(vnode_indices, Config),
-    new_sweep_participant(sweeper_callback_1, 1),
+    RunInterval = 1,
+    new_sweep_participant(sweeper_callback_1, RunInterval),
+    sweeper_meck_helper:backend_modules(async, _NumKeys=1000, _ObjecSizeBytes=0),
+
+    sweep_all_indices(Indices),
+    timer:sleep(timer:seconds(RunInterval + 1)),
```
nickelization: I know it's a short sleep, but since we're just adding this as a new test, is there a way we could rework the test to eliminate it? Partly I'd like the test to run as fast as possible, but I'm also slightly concerned as to whether this will pass reliably - it seems like it should, but we've seen a lot of wonky timing-related failures on the builders lately...

test/riak_kv_sweeper_SUITE.erl
```diff
@@ -207,16 +309,62 @@ test_scheduler_worker_process_crashed(Config) ->
...
+%%
+test_scheduler_run_interval_as_fun(Config) ->
+    Indices = ?config(vnode_indices, Config),
+    RunIntervalFun = fun() -> 1 end,
+    new_sweep_participant(sweeper_callback_1, RunIntervalFun),
     sweeper_meck_helper:backend_modules(async, _NumKeys=1000, _ObjecSizeBytes=0),
 
     sweep_all_indices(Indices),
-    timer:sleep(min_scheduler_response_time_msecs()),
+    timer:sleep(timer:seconds(RunIntervalFun() + 1)),
```
nickelization: Something that just occurred to me: we have a couple tests now that verify that sweeps run after the run interval has expired, but do we have any tests that verify the inverse? We may also want to test that sweeps _don't_ run again until the run interval has elapsed. Perhaps if we could rework the tests to get rid of the sleep, then maybe that's a test we could add at the same time (e.g. if we use something like wait_until, we could track how long it had to wait, and make sure it had to wait at least as long as the run_interval).

nickelization:
@raghavkarol Added a few review comments, but they're not necessarily super related to this PR. If you want to address them here, great, but if not let me know and I can +1 this so that we can fix the other stuff in separate PRs.

test/riak_kv_sweeper_SUITE.erl
raghavkarol: This is not a new test, just looks like that in the diff. I don't think stability is a concern; the sleep is not for asynchronous activity to complete but to allow for the passage of time

Making the test faster definitely a good idea and a mocking functions that the sweeper scheduler functions uses to get time would be the way to go.

test/riak_kv_sweeper_SUITE.erl
raghavkarol: That's a good point, I'll add a test for it.

raghavkarol:
@nickelization - removed the timer sleep from the tests and added a cases for checking that sweeps are not scheduled before their `run_interval`, please review.

raghavkarol:
@nickelization - removed the timer sleep from the tests and added a cases for checking that sweeps are not scheduled before their run_interval, please review.

nickelization:
Awesome, looks great @raghavkarol!

nickelization:
+1

PR #1596, opened 2017-01-12
Don't let Thumbs auto-squash

nickelization:
Discussed with the team. We would rather not have Thumbs auto-merge, and we're not a fan of squashing by default either. Editing the config to match our preferences.

### DIFF
--- a/.thumbs.yml
+++ b/.thumbs.yml
@@ -1,5 +1,6 @@
 minimum_reviewers: 2
-merge: true
+merge: false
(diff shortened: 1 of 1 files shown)


JeetKunDoug:
+1

paulhenrich:
+1

JeetKunDoug:
@thumbot merge

nickelization:
@thumbot merge

PR #1597, opened 2017-01-13
qry_worker: kill outstanding FSMs on cancelling query

hmmr:
(See https://github.com/basho/riak_kv/pull/1573 for previous comments.)

When a query is cancelled (because of div_by_zero or if its projected size is found to exceed configured limit), there may be index FSMs on vnodesstill busy collecting chunks.  Those chunks will be sent to the
coordinator when they are ready, where the query worker instances to receive them will be gone.

It makes sense to terminate the FSMs early, once it is known the data they are collecting will be discarded.


yorkshire-steve:

PR #1601, opened 2017-01-16
[WIP] single instance leveldb for all qbufs

hmmr:
https://bashoeng.atlassian.net/browse/RTS-1690

In 1.5.0, each query buffer is kept in a separate leveldb table. This involves many a file and dir creation/deletion ops, with the only benefit that dropping operation is as easy as `eleveldb:close() andalso rmdir()`.

With automatic bucket expiry becoming available in leveldb, we can keep a single leveldb instance always open (never having to call `eleveldb:open` ourself) and, now storing query results in `{Bucket, Key}`, offload deletion of per-query buckets to leveldb expiry mechanism.

The PR is currently WIP: need to add code to properly enable the expiry.


hmmr:
Work mostly done; putting it on hold pending leveldb changes from @matthewvon.

src/riak_kv_qry_buffers.erl
```diff
@@ -185,13 +219,10 @@ offset_to_scalar([A]) when is_integer(A), A >= 0 -> A.
...
           chunks_need :: integer(),
 
-          %% flipped to true when chunks_need == chunks_got
-          all_chunks_received = false :: boolean(),
-
           %% total records stored (when complete)
           total_records = 0 :: non_neg_integer(),
 
-          %% %% iterator cache (need a working iterator support in eleveldb)
+          %% TODO: iterator cache
```
matthewvon, line 212: Not sure what you think an "iterator cache" will be, but it could be a horrible thing ... leaving you in a position of never finding data.  Iterators lock in a "snapshot" of the database.  You likely want a fresh snapshot AFTER the query data finishes posting.

hmmr: The lifecycle of each qbuf table is such that after writing to it is completed, it becomes a read-only. table (effectively, the last snapshot by your definition?). This is when I intended to open some number of iterators, with the idea to have a map of `{Pos, Iter}`, so that when I need to fulfill a query with OFFSET N, I could pick the iterator whose associated `Pos` is closest to `N` and use it instead of iterating from the beginning.

However, there is no pressing need to develop on this further as query buffers are currently not reusable.

matthewvon:
FYI:  I do have the first half of the leveldb/eleveldb code posted and ready.  I cannot start the second half until all the reviewers finish with the bucket expiry branches.  The code needed to request query id status is still being debated by the Erlangers.  The request mechanism is common to both bucket expiry and query buffers.

You will NOT directly use this option.  But just letting you know I really have done a little of the work:

https://github.com/basho/leveldb/wiki/mv-disable-recovery-log

I will use this option during the generation of a special ExpiryModule for the query buffers.

yorkshire-steve:

PR #1602, opened 2017-01-19
Documented CREATE TABLE flow and postcondition [JIRA: RTS-1739]

paegun:
* added documentation of the table creation flow w/i a general outline for explaining tables and records to distinguish how Riak TS stores these inherently different types of objects.


paegun:
changed base branch due to merge of #1577 .

docs/tabular_buckets.md
```diff
@@ -0,0 +1,84 @@
+# Tabular Bucket Types
+## Abstract
+Riak TS is built upon Riak KV which is built upon Riak Core in so much as
+Riak Core was factored out of a monolithic Riak KV and in so much as Riak TS
+utilizes several function call paths of Riak KV but fundamentally differs
+as neither the table nor a record within a table is a Riak Object in the sense
+that consumers as well as developers of Riak KV have committed to memory, the
+rules to reason about to guarantee eventual consistency in an AP system. To
```
hazen: I suggest that the previous run-on sentence should be refactored.  Perhaps have one sentence about the code origin KV->core->KV->TS, then the need to reason about eventual consistency.

docs/tabular_buckets.md
```diff
@@ -0,0 +1,84 @@
...
+create a coherent understanding of Riak TS Table and Record entities, the
+operations that act upon these entities with specific callouts to differences
+between the TS entities and Riak KV ~entities Bucket Type, Bucket, Key, and
+Object.
```
hazen: I feel we're missing a verb in the previous sentence.

docs/tabular_buckets.md
```diff
@@ -0,0 +1,84 @@
...
+
+Subsets of Riak TS operations are exposed via native Erlang, riak-admin, HTTP,
+and protobuf (PB). However, the full set of operations described herein are
+exposed via the SQL subset supported by Riak TS. The SQL language implementation
+is exposed via the HTTP and PB interfaces, so available to all clients in a
```
hazen: s/so/so is/

docs/tabular_buckets.md
```diff
@@ -0,0 +1,84 @@
...
+To ensure the bucket type operation is consistent, the bucket type activation
+request is focused through the Riak Core Claimant. The Claimant orchestrates
+the activity of creating and activating the bucket type which culminates in the
+plumtree broadcast of the bucket type metadata around the cluster.
+
+Since LevelDB and Riak KV are unaware of the Riak TS record format, translation
+to and from a raw Riak Object format are performed by Riak TS. While a general
+translation could be performed, this would require traversing lists to access
+fields. Therefore table-specific mapping is preferred and provided by compiling
+such modules, generally termed DDL module within Riak TS.
```
hazen: Might want to note that this is in `riak_ql` and perhaps even mention the module

docs/tabular_buckets.md
```diff
@@ -0,0 +1,84 @@
...
+
+The compilation of the table module on each node is handled by a Riak TS specific
+metadata store listener which forwards metadata store events to the Riak TS
+new type module which translates the DDL into a compiled module, storing the
+beam in the ddl_ebin directory. These modules are also responsible for providing
```
hazen: I'd break that up into two sentences: The compilation/metadata store events as the first then the generated type module as the second. You might also want to add a sentence about how these QL-generated modules are created in the first place.

docs/tabular_buckets.md
```diff
@@ -0,0 +1,84 @@
...
+the status of the table-specific compiled module on each node.
+
+While the SQL `CREATE TABLE <table_name> ...` operation could be labeled
+eventually consistent, like most Riak KV storage operations, this would put the
+burden of retrying on failure due to a "not yet active" statefor subsequent
+queries. Such a lack of certainly consistent postcondition, especially for the
```
hazen: "this would put the burden on..." on whom?  Where is this burden placed?

docs/tabular_buckets_create_table.msc
```diff
@@ -0,0 +1,128 @@
+msc {
```
hazen, line 1: Awesome, but very big image of the data flows

hazen (changes requested):
Great to have this information captured somewhere. Maybe we could just update and rebase this existing PR?

DSomogyi:
create jira issue

hazen (approved):

hazen:
+1

docs/tabular_buckets.md
macintux: Whew, yeah. That first sentence is a doozy.

docs/tabular_buckets.md
```diff
@@ -0,0 +1,86 @@
...
+## Abstract
+Riak TS is built upon Riak KV which is built upon Riak Core in so much as
+Riak Core was factored out of a monolithic Riak KV and in so much as Riak TS
+utilizes several function call paths of Riak KV but fundamentally differs
+as neither the table nor a record within a table is a Riak Object in the sense
+that consumers as well as developers of Riak KV have committed to memory, the
+rules to reason about to guarantee eventual consistency in an AP system. To
+create a coherent understanding of Riak TS Table and Record entities, we will
+explore the operations that act upon these entities with specific callouts to
+differences between the TS entities and Riak KV ~entities Bucket Type, Bucket,
```
macintux: Tilde seems out of place

docs/tabular_buckets.md
```diff
@@ -0,0 +1,86 @@
...
+cluster.
+
+## Table Operations
+### Create
+When a client issues a SQL `CREATE TABLE <table_name> ...` statement, Riak TS
+parses and transforms the SQL into a data-definition language (DDL) format of
+the table and requests Riak Core to store the DDL as a new bucket type property.
+
+To ensure the bucket type operation is consistent, the bucket type activation
+request is focused through the Riak Core Claimant. The Claimant orchestrates
```
macintux: s/focused/serialized/

docs/tabular_buckets.md
```diff
@@ -0,0 +1,86 @@
...
+the activity of creating and activating the bucket type which culminates in the
+plumtree broadcast of the bucket type metadata around the cluster.
```
macintux: Not sure it's accurate to say "culminates". It begins with the broadcast of the metadata, culminates with the toggling of the active flag (which admittedly is another broadcast of metadata, but this still seems a bit misleading).

docs/tabular_buckets.md
```diff
@@ -0,0 +1,86 @@
...
+
+Since LevelDB and Riak KV are unaware of the Riak TS record format, translation
+to and from a raw Riak Object format are performed by Riak TS. While a general
+translation could be performed, this would require traversing lists to access
+fields. Therefore table-specific mapping is preferred and provided by compiling
+such modules, generally termed DDL module (compiled and beamed around the
```
macintux, line 44: "beamed" <- I see what you did there

docs/tabular_buckets.md
```diff
@@ -0,0 +1,86 @@
...
+cluster for each table) within Riak TS.
+
+The compilation of the table module on each node is handled by a Riak TS specific
+metadata store listener which forwards metadata store events to the Riak TS
```
macintux, line 48: While currently accurate, I wouldn't characterize the listener as TS-specific. It's certainly intended to be a general-purpose feature, just hasn't (to my knowledge) migrated to the KV development stream yet.

docs/tabular_buckets.md
```diff
@@ -0,0 +1,86 @@
...
+new type module which translates the DDL into a compiled module. Compilation
+results in storing the beam in the ddl_ebin directory. The presence of these
+modules is used in determining the status of the table-specific compiled module
+on each node.
+
+While the SQL `CREATE TABLE <table_name> ...` operation could be labeled
```
macintux: "could be labeled eventually consistent" implies that it could accurately be described as eventually consistent. I would say "could be implemented as an eventually consistent operation".

docs/tabular_buckets_create_table.msc
```diff
@@ -0,0 +1,128 @@
...
+
+    C -> KVA [label="create table"];
+    KVA -> CBT [label="create"];
+    CBT -> CC [label="create_bucket_type"];
+    CC -> CC [label="claimant"];
+    CC -> CC [label="create_bucket_type"];
+    CC -> CC [label="get_bucket_type"];
+    CC -> CM [label="get"];
+    --- [label="bucket DNE"];
+    CC <- CM [label="<- get, undefined"];
```
macintux, line 20: I'm confused about lines 20 and 21. 21 has two different arrow directions (perhaps it's showing both the call and the return?), and it's not clear why CC calls both CM:get and CM:get_bucket_type.

docs/tabular_buckets.md
paegun: Afaik ~ in documentation means "not quite" or approximately, which in this case should be understood as "not quite" b/c neither a Riak KV Bucket Type nor a Bucket is an entity.

docs/tabular_buckets_create_table.msc
paegun: Yes, <- means return, either as a literal function return or a return message (cast).

CC call get_bucket_type on itself which contains a call from CC to get  on CM. This is the flow of things. Including the self call to get_bucket_type could be elided, but IMO helps in the understanding of what CM's get is for.

macintux:
Will need to change the target branch.

macintux:
+1

hazen:
+1

PR #1604, opened 2017-01-20
Update the riak_kv_backend behavior to use -callback specs

nickelization:
This updates the `riak_kv_backend` behavior to use `-callback` directives instead of the old-style `behaviour_info` system. I've also improved some related type specifications in a few places. These changes allow us to kill off another line in the dialyzer.ignore-warnings file.

### DIFF
--- a/src/riak_kv_backend.erl
+++ b/src/riak_kv_backend.erl
@@ -22,7 +22,6 @@
 
 -module(riak_kv_backend).
 
--export([behaviour_info/1]).
 -export([callback_after/3]).
 
 -type fold_buckets_fun() :: fun((binary(), any()) -> any() | no_return()).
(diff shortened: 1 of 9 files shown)


src/riak_kv_backend.erl
```diff
@@ -31,33 +30,52 @@
 -type fold_objects_fun() :: fun((binary(), binary(), term(), any()) ->
                                        any() |
                                        no_return()).
+
```
raghavkarol: What do the `Index` and `SecondaryIndex` mean in this type spec? I could not find any information about them here http://erlang.org/documentation/doc-5.10.3/doc/reference_manual/typespec.html

src/riak_kv_backend.erl
```diff
@@ -31,33 +30,52 @@
...
+-type index_spec() :: {add, Index, SecondaryKey} | {remove, Index, SecondaryKey}.
```
raghavkarol: What do the `Index` and `SecondaryIndex` mean in this type spec? I could not find any information about them here http://erlang.org/documentation/doc-5.10.3/doc/reference_manual/typespec.html

raghavkarol (commented):
Looks good, I have one question about a type construct that I couldn't find documentation about. A comment or a pointer to some documentation would be helpful.

src/riak_kv_backend.erl
nickelization: If you just use a variable name like that (i.e. capitalized with no type arguments), it's the same as not specifying any type. If we wanted to be explicit, we could edit it to look like this, which would be functionally identical:
```
-type index_spec() :: {add, Index :: any(), SecondaryKey :: any()} | {remove, Index :: any(), SecondaryKey :: any()}.
```
If I were writing the type spec from scratch, I probably would have gone in and figured out what the actual types are here, instead of just using untyped placeholders like that. However this type spec already existed prior to this PR, but before this, it was copied and pasted manually to every single backend's source file. I didn't take the time to improve the existing spec, but I did at least get rid of all the identical copies scattered around the source tree.

raghavkarol: After discussing with @nickelization the spec we arrived at the conclusion that the spec 

`
-type index_spec() :: {add, Index, SecondaryKey} | {remove, Index, SecondaryKey}.
`

may not be functionally identical to

`-type index_spec() :: {add, Index :: any(), SecondaryKey :: any()} | {remove, Index :: any(), SecondaryKey :: any()}.`

And is likely a polymorphic type declaration interpreted as, a union of two three tuples where
 - the first element is the atom `add` or `remove`
 - the second element must have the same type
 - the third element must have the same type 

Example
- VALID `{add, one, <<1>>}` and `{remove, one, <<2>>}` 
- INVALID `{add, [one], <<>>}` and `{remove, {one}, <<>>}`
- ????  `{add, {one, 1}, <<1>>}` and `{remove, {one, 2}, <<2>>}`

raghavkarol:
+1

PR #1607, opened 2017-01-20
[RTS-1707, RTS-1709] Extend Grafana webmachine and add CORS support, for add-ons

paegun:
depends on:
* https://github.com/basho/riak_api/pull/121

supports:
* https://github.com/basho/grafana-riakts-datasource/pull/12

provides:
* added cross-origin resource sharing (CORS) support as an advanced.config setting.
  * CORS to be used for add-ons, such as Grafana.
    * NOTE: e11150b (multiline) contains preferable UX to f5954a5 (CSV)
* added Grafana integration webmachine
  * w/o support for:
    * tag lookup tables
    * where filter value type coercion
  * w/ support for:
    * list tables
    * list table fields
    * query, including expansion of $time template field.


paegun:
thumbot wait

macintux:
Verified that I see the same dialyzer errors locally:

```
riak_kv_wm_utils.erl:398: The pattern "*" can never match the type {atom(),_,_}
riak_kv_wm_utils.erl:399: The pattern {Scheme, Host} can never match the type {atom(),_,_}
```

I don't see any way for "\*" to be an argument; line 353 can send the atom '\*' but not the string.

And there's no way for a 2-tuple to be sent.

priv/riak_kv.schema
```diff
@@ -967,3 +967,45 @@
   fun(Value) when is_integer(Value) andalso Value > 0 -> true;
      (_) -> false
   end}.
+
+%% @doc Register allowable origins (for CORS), specified as a list of ip
```
macintux: Perhaps a question for @angrycub: should we make this `riak_kv.cors.allowable_origin` instead of `riak_kv.allowable_origin` to make its purpose explicit?

priv/riak_kv.schema
```diff
@@ -967,3 +967,45 @@
...
+%%   }}
+%% ]}
+%%
+%% The resulting allowable origins are "http://127.0.0.1:300",
+%% "http://10.10.1.240:*", "https://127.0.0.1:4430".
+%%
+%% The special case of "Allow All" is achieved by specifying the scheme as '*'
+%% (wildcard), i.e.
+%%
+%% {riak_kv, [
```
macintux: It appears there is no way to specify a wildcard in `riak.conf`:

```
05:47:30.261 [error] "*" cannot be converted to a(n) IP
```

Did you look for a way to allow that, or is there some option I'm overlooking?

src/riak_kv_wm_utils.erl
```diff
@@ -335,10 +338,71 @@ is_valid_referer(RD) ->
         OriginTuple ->
             true;
         RefererTuple ->
-            lager:debug("WM referrer not origin.  Origin ~p != Referer ~p\n", [OriginTuple, RefererTuple]),
-            false
+            allowable_origin(OriginTuple, RefererTuple)
```
macintux: This is a minor style concern, but I would find `allowable_origin/2` a bit easier to read if, instead of handling `OriginTuple == OriginTuple` in `is_valid_referer/1` directly, it was a separate function head.

So: is_valid_referer's case statement would be undefined/invalid/Referer, and then the first function head for `allowable_origin/2` would be `(Origin, Origin) -> true`.

src/riak_kv_wm_utils.erl
```diff
@@ -335,10 +338,71 @@ is_valid_referer(RD) ->
...
+        false ->
+            lager:debug("WM referrer not origin. Origin ~p != Referer ~p" ++
+                        " and Referer not in allowable origins.\n",
+                        [OriginTuple, RefererTuple]),
+            false;
+        _ -> true
+    end.
+
+normalize_referer('*') -> {"*", "*", "*"};
+normalize_referer("*") -> normalize_referer('*');
```
macintux: As noted in the main PR comments, this function head is irrelevant and giving dialyzer heartburn. Same with the one that follows.

src/riak_kv_wm_graph.erl
```diff
@@ -0,0 +1,726 @@
...
+        {ok, [{table, Table},
+              {field, Field}]} ->
+            show_tag_values(Table, Field, RD, Ctx);
+        {ok, [{table, Table},
+              {field, Field},
+              {values, _Values}]} ->
+            show_tag_values(Table, Field, RD, Ctx)
+    end;
+query_result(Query = <<"SELECT ", _T/binary>>, RD, Ctx) ->
+    %% TODO: test/fix GROUP BY ___ fill(null)
```
macintux: Are these `TODO` comments overlooked or documented in another ticket?

macintux:
The Grafana error if Riak is not configured to allow requests is unhelpful:

"RiakTS Error: undefined"

src/riak_kv_wm_utils.erl
macintux: would also make the function more reusable

priv/riak_kv.schema
paegun: Yes, the wildcard atom '*' is able to be used like `http` and `https` at the scheme level of the configuration schema employed here.

priv/riak_kv.schema
paegun: I've requested @angrycub to weigh in on the name here.

paegun:
Better reporting of errors attempting to connect and query Riak TS w/i the datasource setup will be handled in the Riak TS Grafana Datasource plugin.

priv/riak_kv.schema
macintux: I mean, is there any way to use `*` in `riak.conf`? (Maybe I'm misunderstanding your answer.)

lukebakken: @macintux - what exactly did you put into `riak.conf`?

macintux: 3 weeks ago, dunno now

macintux: `allowable_origin.foo = *` (any name should be same)

lukebakken: OK, yeah, that's what I would expect it to be. Must be a bug.

macintux: It's an `ip` data type, and cuttlefish afaik doesn't allow for wildcards in that

priv/riak_kv.schema
angrycub: Talked with @paegun about the schema today.  I'm cool with `cors_allowable_origin` since there aren't any other CORS settings (and don't appear to be the possibility of more)

priv/riak_kv.schema
angrycub: @paegun and I discussed how we might be able to approach this differently, but also that using the left-hand side for anything other than grouping is a violation of the "cuttlefish way"

priv/riak_kv.schema
```diff
@@ -968,44 +968,48 @@
...
-%% This is equivalent to specifying the port using the wildcard.
-{mapping, "allowable_origin.$name", "riak_kv.allowable_origin.http", [
-    {datatype, ip},
-    {default, {}},
-    {level, advanced}
+%% Allowable schemes include http and https, and all where the wildcard (all)
+%% is used to allow all origins to share resources with Riak.
+{mapping, "listener.$scheme.cors_allowable_origin.$origin", "riak_api.cors_allowable_origins",
+[
+ {datatype, [ip, flag]}
```
macintux: What do `on` and `off` mean as values here? 

Tried setting this in `riak.conf`:

```
listener.http.cors_allowable_origin.foo = off
```

Got this failure:

```09:31:11.691 [error] CRASH REPORT Process <0.364.0> with 0 neighbours exited with reason: {{{badmatch,false},[{riak_api_web,binding_config,2,[{file,"src/riak_api_web.erl"},{line,48}]},{riak_api_sup,'-web_processes/1-lc$^0/1-0-',1,[{file,"src/riak_api_sup.erl"},{line,63}]},{riak_api_sup,web_processes,1,[{file,"src/riak_api_sup.erl"},{line,63}]},{riak_api_sup,init,1,[{file,"src/riak_api_sup.erl"},{line,53}]},{supervisor,init,1,[{file,"supervisor.erl"},{line,239}]},{gen_server,init_it,6,[{file,"gen_server.erl"},{line,304}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,239}]}]},...} in application_master:init/4 line 133
09:31:11.691 [info] Application riak_api exited with reason: {{{badmatch,false},[{riak_api_web,binding_config,2,[{file,"src/riak_api_web.erl"},{line,48}]},{riak_api_sup,'-web_processes/1-lc$^0/1-0-',1,[{file,"src/riak_api_sup.erl"},{line,63}]},{riak_api_sup,web_processes,1,[{file,"src/riak_api_sup.erl"},{line,63}]},{riak_api_sup,init,1,[{file,"src/riak_api_sup.erl"},{line,53}]},{supervisor,init,1,[{file,"supervisor.erl"},{line,239}]},{gen_server,init_it,6,[{file,"gen_server.erl"},{line,304}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,239}]}]},...}
```

angrycub: Feels like these are better expressed as 2 whole items since they are enumerable. one mapping for "listener.http.cors_allowable_origin.$origin" and "listener.https.cors_allowable_origin.$origin"
then make the datatype {datatype, string).  This removes the InvalidSchemes test. You might want to write a single validator for both cors settings (to deal with the possibilities of IP and wildcards) and a single translation function that both mappings target to get the tuple you are looking for.  You can target multiple mappings to the same translation function to build complex configuration tuples from more than one simeple config

Cuttlefish IP datatype is going to be working against you unless you have some PR work going on over there to support the ip addresses with non-integer required ports.

paegun: Oy, thank you @macintux . Testing of the flag, which is useful for `listener.all.cors_allowable_origin.all = on`, was performed, but I didn't test flag for http and https schemes. I have added a filter to the http and https listener config schema bits here: https://github.com/basho/riak_api/pull/121

macintux: @angrycub I thought some time ago we needed a data type for IP + port information, but I've never made the time to dig deeper.

paegun: @angrycub 
wrt splitting out `all`since it is the only flag datatype does have a indisputable benefit, so I will make that change.

I disagree wrt splitting out `http` and `https`, this design results in duplicated code as in listener.http.$name. IME the runtime cost incurred at startup to find and throw if any invalid schemes are present is cheap, especially when compared to the need to hotfix when someone fails to keep the code in sync between the repeated http and https config code AND config code does not lend well to testing, so compounded. If you feel that I am wrong, let me know and I'll make the change.

angrycub: First, a mapping entry is not code.  Secondly, both the validation and translation functions (code) are reusable and can be DRY.  If they don't use one of the schemes that you have enumerated, it literally gets thrown out for you.  You are coding for a problem cuttlefish solves when a user supplies something without a mapping.

angrycub: @macintux the [ip](https://github.com/basho/cuttlefish/wiki/Datatypes#ip) datatype _requires_ a valid IP and a port, so it would not be able to cope with the port wildcard or the absence of a port.  It would be good if there was a datatype that just did the ip address part by itself, but for now there is not.

macintux:
Oops, yep, I forgot which gap I was upset about. Now I remember why I was unhappy we used the term ip already, since that makes it harder to come up with a name for just an address with no port.

-John

paegun:
Changes to use cuttlefish schema instead of code to enforce the scheme as well as only using datatype: flag for `all` is in, see a59f94a .

Also wrt `ip` being poorly named as it contains not only ip address or hostname in ipv4 or ipv6 format, but also the port, this is not the first time and place the exact same thing has happened. `hostname` or `ipaddress` are both decent names for the port-less variety. I prefer `hostname` as it enables ops to centralize changes in DNS instead of all over the configurations of the applications in use AND if the configuration needs to have a different field for fully-qualified domain name (`fqdn`) and/or `public_hostname`, these are all names and scale globally accordingly.

macintux:
To make sure I understand the current schema:

To allow grafana access, the operator may do one (or more) of three things to grant CORS access:
* Allow a specific IP/port combo to access http (actually, multiple such combos)
* Allow a specific IP/port combo to access https (ditto)
* Allow the world to access both http and https

Is that correct?

src/riak_kv_wm_utils.erl
```diff
@@ -332,13 +335,74 @@ is_valid_referer(RD) ->
...
+        false ->
+            lager:debug("WM referrer not origin. Origin ~p != Referer ~p" ++
+                        " and Referer not in allowable origins.\n",
+                        [OriginTuple, RefererTuple]),
+            false;
+        _ -> true
+    end.
+
+normalize_referer('*') -> {"*", "*", "*"};
+normalize_referer({Scheme, "127.0.0.1", Port}) ->
```
macintux: Pedants'R'Us.

Dredging up distant memories, this should really be 127.*. The entire `127` class A network is equivalent to localhost.

macintux:
Ok, I've assured myself my assessment of the schema is accurate. I have one nitpick that I'm unconvinced is important, but if you see any easy fix, please make it so. Will approve.

src/riak_kv_wm_utils.erl
macintux: Also: do we need to keep an eye out for ipv6 localhost?

paegun:
Yes, @macintux, 127.*.*.* is correct, so I'll expand to 127.* for that subclass of A. Great nitpick. Thank you.

macintux:
I would dispute that, actually. 0.0.0.0 means listen on everything, which is a different semantic than we’re going for here.

paegun:
Yes, I agree that 0.0.0.0 means that, but in practice, "0.0.0.0" is what is sent in the Referer header when the local web application is requested by that address. Semantically and network addressing wise, 0.0.0.0 is assuredly localhost.

macintux:
+1

paegun:
suggest closing, so will close unless requested otherwise by 2017-05-31 EOD UTC.

PR #1609, opened 2017-01-23
fix projected vs max query size check for queries w/ LIMIT clause [JIRA: RIAK-3195]

hmmr:
There was a bug in 1.5.0 where, for queries with a LIMIT clause, the projected query size was incorrectly computed. The LIMIT value was used as the number of _chunks_ in the calculations and not as the number of _rows_ as it ought to be.

Separately, this PR fixes the logic of the max query data size checks for the case where LIMIT is big but the WHERE range is small. Such queries should not be cancelled.


DSomogyi:
create jira issue

hazen:
:+1:

hazen:
@paegun This is a duplicate of #1619 which you had previously reviewed. Mind +1ing this one, too?

paegun:
+1

hazen:
thumbs retry

hazen:
+1

PR #1610, opened 2017-01-23
Remove code for the legacy reformat-indexes command [JIRA: RIAK-3238]

nickelization:
This has not been needed since Riak 1.3; we're well past the point where we can delete the code and move on. Just to be sure though, I ran this by Client Services, and they said the last time a ticket mentioned a customer using this command was in 2014.

The actual riak-admin command is removed in basho/riak_ee#418 (a separate PR for OSS Riak will be opened when that PR is merged). This PR removes all the riak_kv code under the hood that was used to execute that command.


nickelization:
Note, this PR mostly removes old code, but I also had to rework the eleveldb startup checks a bit to ensure safety on upgrades and downgrades. I thoroughly tested this by hand on my local machine and everything appears to be in good working shape, but I'd encourage any reviewers to poke at this a bit and make sure I haven't missed anything.

bashopatricia:
create jira issue

fadushin:
+1

fadushin:
+1

PR #1617, opened 2017-01-26
Remove all warning suppressions for callback specs

nickelization:
We now have callback specs defined in riak_core for these behaviors, so we no longer need to suppress any of these warnings related to them.


paulhenrich:
👍

PR #1619, opened 2017-01-26
fix projected query size calculation for queries with LIMIT [JIRA: RIAK-3210]

hmmr:
This is a backport of the fix in https://github.com/basho/riak_kv/pull/1609.

There was a bug in 1.5.0 where, for queries with a LIMIT clause, the projected query size was incorrectly computed. The LIMIT value was used as the number of chunks in the calculations and not as the number of rows as it ought to be.

Separately, this PR fixes the logic of the max query data size checks for the case where LIMIT is big but the WHERE range is small. Such queries should not be cancelled.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -72,6 +72,7 @@
           max_running_fsms   = ?MAX_RUNNING_FSMS    :: pos_integer(),
           %% 2. Estimate query size (wget-style):
           n_subqueries_done  = 0                    :: non_neg_integer(),
+          total_query_rows   = 0                    :: non_neg_integer(),
           total_query_data   = 0                    :: non_neg_integer(),
           max_query_data                            :: non_neg_integer(),
           %% For queries not backed by query buffers, results are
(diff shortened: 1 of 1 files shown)


DSomogyi:
create jira issue

hazen:
:+1:

src/riak_kv_qry_worker.erl
```diff
@@ -261,43 +262,54 @@ throttling_spawn_index_fsms(State) ->
...
-                       [Limit, ProjectedLimitData, MaxQueryData, NSubqueriesDone, length(SubQrys), OrigQry]),
+    EstLimitData = round(Limit * (TotalQueryData / TotalQueryRows)),
+    IsLimitTooBig = EstLimitData > MaxQueryData,
+
+    %% but also check the grand total, for the case when LIMIT is big
+    %% but WHERE range is still tiny
+    EstTotalData = round(TotalQueryData + (TotalQueryData / NSubqueriesDone) * length(SubQrys)),
+    IsWhereTooBig = EstTotalData > MaxQueryData,
+
+    case IsLimitTooBig and IsWhereTooBig of
```
paegun, line 284: The case here is anding, but the lager message says or, and logically either condition that is estimated to be too large should cancel the query.

hmmr: Lager message fixed in 3c3f352.

paegun:
+1

src/riak_kv_qry_worker.erl
paegun: Thank you. The match in logic between the code and log message clearly means that we are being conservative about cancelling queries since both conditions must be true.

javajolt:
thumbs retry

hazen:
:+1:

PR #1620, opened 2017-01-26
List group keys

paulhenrich:
Add support for grouped listing of keys and object metadata.

The entry point into this functionality is the `riak_client: list_group_keys/4` function, which accepts a `GroupParams` argument to control behavior, such as the maximum number of keys to return, the prefix to which keys are constrained, etc.

The functionality is implemented by the `riak_kv_group_keys_fsm`, which is a `riak_core_coverage_fsm` implementation, and which allows for gathering results from the various nodes using a coverage query. At the vnode level, the grouped key listing is performed by the `riak_kv_group_keys` module, which currently works only for eleveldb.


src/riak_kv_group_keys_response.erl
```diff
@@ -0,0 +1,70 @@
...
+         new_response/3,
+         get_metadatas/1,
+         get_common_prefixes/1,
+         is_truncated/1,
+         get_next_continuation_token/1
+        ]).
+
+-export_type([response/0]).
+
+-record(response, {
```
jvoegele: Use a more descriptive name for this record to aid debugging.

src/riak_kv_continuation.erl
```diff
@@ -0,0 +1,118 @@
...
+make_token(Results, MaxResults) when is_list(Results),
+                                     is_integer(MaxResults),
+                                     MaxResults >= 0 ->
+    make_token(Results, MaxResults, fun(X) -> X end).
+
+%% @doc Make a continuation token if and only if the length of the list of
+%% `Results' is greater than `MaxResults'. If so, the item in the list at
+%% position `MaxResults' is used as the term to make the continuation token,
+%% after first being passed to the given `Fun' to allow for transforming the
+%% value. If not, `undefined' is returned.
```
nickelization, line 58: I've been trying to get away from this kind of use of `undefined` lately; I feel like `undefined` is overloaded to the point where its use ends up being kind of ambiguous and error-prone. If we could change it to return `'no_continuation'` or something along those lines, I think that would be much nicer.

src/riak_kv_continuation.erl
```diff
@@ -0,0 +1,118 @@
...
+         make_token/2,
+         make_token/3,
+         decode_token/1]).
+-export_type([token/0]).
+
+-type token() :: binary() | undefined.
+
+%% @doc Make a continuation token for the given `Term'.
+-spec make_token(term()) -> token().
+make_token(Term) ->
```
nickelization, line 37: I wonder if it might be worth adding an extra byte to the beginning of the token to denote format/version, in case we decide to change to something else later. If nothing else, it'd be nice for efficiency if we could base64-encode raw binaries directly instead of having to wrap them in the extra layer of `term_to_binary`.

On the flip side, this might be overkill though, since I'm not sure if tokens are something we'll be using heavily enough to warrant this level of optimization. I guess use your own judgement :)

src/riak_kv_continuation.erl
```diff
@@ -0,0 +1,118 @@
...
+-spec decode_token(token()) -> term().
+decode_token(undefined) ->
+    undefined;
+decode_token(Token) ->
+    binary_to_term(base64:decode(Token)).
+
+%% ====================
+%% TESTS
+%% ====================
+
```
nickelization, line 85: Ooh man, perfect opportunity for a little QuickCheck, just to verify that encode/decode round trip :). The tests here look like they're probably sufficient already though, so totally up to you if you want to add in additional testing with EQC.

src/riak_kv_group_keys.erl
```diff
@@ -0,0 +1,547 @@
...
+-spec set_max_keys(group_params(), MaxKeys::pos_integer()) -> group_params().
+set_max_keys(GroupParams, MaxKeys) ->
+    GroupParams#group_params{max_keys = MaxKeys}.
+
+-spec set_continuation_token(group_params(), ContinuationToken::binary()) ->
+    group_params().
+set_continuation_token(GroupParams, ContinuationToken) ->
+    GroupParams#group_params{continuation_token = ContinuationToken}.
+
+%% NOTE: This  type spec with `fun(() -> no_return())' as the return type is
```
nickelization: Why is the `no_return()` type necessary? That doesn't quite smell right to me....

Even if there's no other better solution, it'd be nice to have some extra context in the comments as to what's going on here.

src/riak_kv_group_keys.erl
```diff
@@ -0,0 +1,547 @@
...
+    FoldOpts1 = [{first_key, BackendMod:to_first_key({bucket, Bucket})} | FoldOpts],
+    try iterator_open(DbRef, FoldOpts1) of
+        {ok, Itr} ->
+            iterate(Bucket, GroupParams, Itr, FoldFun, Acc)
+    catch Error ->
+              lager:debug("Could not open iterator: ~p", [Error]),
+              throw(Error)
+    end.
+
+iterator_open(DbRef, FoldOpts) ->
```
nickelization: Could we maybe get rid of this extra layer of indirection for `iterator_open`, `iterator_close`, and `iterator_move`? Feels kind of like overkill just to avoid having to type "eleveldb:" in a few places.

src/riak_kv_group_keys.erl
```diff
@@ -0,0 +1,547 @@
...
+            riak_object:get_metadata(RObj);
+        _ ->
+            %% TODO
+            {error, riak_object_has_siblings}
+    end.
+
+%% 0xFF is the biggest byte for the purposes of sorting in LevelDB,
+%% used for seeking past all keys starting with the prefix.
+-spec append_max_byte(binary()) -> binary().
+append_max_byte(Binary) when is_binary(Binary) ->
```
nickelization: But...is it possible someone could write a key starting with `<<Binary/binary, 16#FF>>` into Riak? Then instead of seeking past the prefix, it would seek to the first key starting with the specified prefix followed by 0xff.

Maybe there is something that prevents you from using bytes like this in keys when doing a put. I don't have time to check now, so maybe this is totally safe, but just wanted to raise this possibility.

paulhenrich: I think we decided that in this case it just breaks the optimization a bit of scanning past keys, so in the unlikely case a user writes a key like that we accept it. (it doesn't change the behavior of the consumer, just the efficiency of part of the scan)

src/riak_kv_group_keys.erl
paulhenrich: Our goal was to start isolating this so it can be used with the memory backend as well, though I agree it looks a little odd as-is.

paulhenrich:
👏🏼

PR #1621, opened 2017-01-30
properly estimate_query_size for aggregation and grouping queries [JIRA: RIAK-3223]

hmmr:
~Aggregation and grouping queries don't accumulate data and hence, don't need their size checked continuously against the configured `max_returned_data_size`. Rather, it can, and should, be done once the full set of groups (each to be returned as a separate row) is known.~

Edit: Grouping queries will grow its result set to the number of unique values in the selection.  In the extreme case of grouping by a column of all-unique values, its size will be the size of the entire selection.

Also filed to riak_ts-integration-1.5.2 in https://github.com/basho/riak_kv/pull/1622.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -260,9 +260,38 @@ throttling_spawn_index_fsms(State) ->
 
 
 estimate_query_size(#state{n_subqueries_done = NSubqueriesDone} = State)
-  when NSubqueriesDone < 2  ->
+  when NSubqueriesDone < 2 ->
+    %% If not enough chunks are received, defer checks
     State;
-
+estimate_query_size(#state{sub_qrys = []} = State) ->
+    %% If all chunks are here (no more left in sub_qrys), consider it
+    %% is safe to proceed.
+    State;
+estimate_query_size(#state{qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = aggregate}}} =
+                        State) ->
+    %% Aggregation alone does not increase the size of result (there
+    %% is only one row returned)
+    State;
+estimate_query_size(#state{n_subqueries_done = NSubqueriesDone,
+                           max_query_data    = MaxQueryData,
+                           sub_qrys          = SubQrys,
+                           result            = Result,
+                           qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = group_by}} =
+                               OrigQry} = State) ->
+    %% Grouping queries will grow its result set to the number of
+    %% unique values in the selection.  In the extreme case of
+    %% grouping by a column of all-unique values, its size will be the
+    %% size of the entire selection.
+    CurrentTotalSize = erlang:external_size(Result),
+    BytesPerChunk = CurrentTotalSize / NSubqueriesDone,
+    ProjectedGrandTotal = round(CurrentTotalSize + (BytesPerChunk * length(SubQrys))),
+    if ProjectedGrandTotal > MaxQueryData ->
+            lager:info("Cancelling aggregating query because projected result size exceeds limit (~b > ~b, subqueries ~b of ~b done, query ~p)",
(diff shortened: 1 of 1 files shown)


src/riak_kv_qry_worker.erl
```diff
@@ -261,6 +261,11 @@ throttling_spawn_index_fsms(State) ->
 estimate_query_size(#state{n_subqueries_done = NSubqueriesDone} = State)
   when NSubqueriesDone < 2  ->
     State;
+estimate_query_size(#state{qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = Type}}} = State)
+  when Type /= rows ->
```
andytill: Any reason to not estimate grouped rows? The result can be as big as the full rows.

andytill: Needs unit tests.

andytill (changes requested):

DSomogyi:
create jira issue

src/riak_kv_qry_worker.erl
hmmr: Added one in bb4cd7b.

hmmr: Done in 7512632, tests fixed in bb4cd7b.

src/riak_kv_qry_worker.erl
```diff
@@ -258,6 +258,28 @@ throttling_spawn_index_fsms(State) ->
...
+                           qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = CalcType}} =
+                               OrigQry} = State)
+  when CalcType /= rows,
+       NSubqueriesDone == 1 ->
+    %% For aggregate queries, the set of columns to return does not
+    %% grow: it is determined at query compilation time.  We should
+    %% check it once, for the extreme case where grouping is done on a
+    %% column with values all unique.
+    CurrentTotalSize = erlang:external_size(AggregateResult),
+    BytesPerChunk = CurrentTotalSize / NSubqueriesDone,
```
hazen: Why divide by `NSubqueriesDone` is it's always 1 in the guard?

src/riak_kv_qry_worker.erl
```diff
@@ -258,6 +258,28 @@ throttling_spawn_index_fsms(State) ->
...
 
 
+estimate_query_size(#state{n_subqueries_done = NSubqueriesDone,
+                           max_query_data    = MaxQueryData,
+                           sub_qrys          = SubQrys,
+                           result            = AggregateResult,
+                           qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = CalcType}} =
+                               OrigQry} = State)
+  when CalcType /= rows,
+       NSubqueriesDone == 1 ->
```
hazen: Not that familiar with the code, but why are we constraining on `NSubqueriesDone == 1`?  Is that particular to aggregates?

hazen (commented):
Tests pass, I'll let you decide if a change is necessary. Probably just a 1-liner and an answer to my question.

src/riak_kv_qry_worker.erl
hmmr: For `aggregate` queries, the structure in `#state.result` does not grow in size. It is set up initially, at query compilation, with initial states of all groups (columns like `AVG(x)`) which `riak_kv_qry_worker:add_subquery_result`, eventually calling `riak_kv_qry_compiler:run_select`, updates with every collected row.

For `group_by` queries, the result set _does_ grow. A query such as `AVG(x) GROUP BY y` will grow to the number of distinct values in column `y`, which in extreme cases can represent a situation where OOM protection is very much applicable.

d53e0d0 is a properer version of the size checks.

src/riak_kv_qry_worker.erl
hmmr: That was me being, you know, too much methodical.

hmmr:
Docs amended in https://github.com/basho/basho_docs/pull/2415.

andytill (approved):

andytill:
+1

hazen:
+1

PR #1622, opened 2017-01-30
properly estimate query size for aggregate queries [JIRA: RIAK-3225]

hmmr:
~Aggregation and grouping queries don't accumulate data and hence, don't need their size checked continuously against the configured `max_returned_data_size`. Rather, it can, and should, be done once the full set of groups (each to be returned as a separate row) is known.~

Edit: Grouping queries will grow its result set to the number of unique values in the selection.  In the extreme case of grouping by a column of all-unique values, its size will be the size of the entire selection.

This fix is also filed to mainline riak_ts-integration in https://github.com/basho/riak_kv/pull/1621.

### DIFF
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -260,7 +260,7 @@ throttling_spawn_index_fsms(State) ->
 
 
 estimate_query_size(#state{n_subqueries_done = NSubqueriesDone} = State)
-  when NSubqueriesDone =< 2  ->
+  when NSubqueriesDone < 2  ->
     State;
 estimate_query_size(#state{qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = aggregate}}} =
                         State) ->
--- a/src/riak_kv_qry_worker.erl
+++ b/src/riak_kv_qry_worker.erl
@@ -279,7 +279,7 @@ estimate_query_size(#state{n_subqueries_done = NSubqueriesDone,
     %% size of the entire selection.
     CurrentTotalSize = erlang:external_size(Result),
     BytesPerChunk = CurrentTotalSize / NSubqueriesDone,
-    ProjectedGrandTotal = round(CurrentTotalSize + BytesPerChunk * length(SubQrys)),
+    ProjectedGrandTotal = round(CurrentTotalSize + (BytesPerChunk * length(SubQrys))),
     if ProjectedGrandTotal > MaxQueryData ->
             lager:info("Cancelling aggregating query because projected result size exceeds limit (~b > ~b, subqueries ~b of ~b done, query ~p)",
                        [ProjectedGrandTotal, MaxQueryData, NSubqueriesDone, length(SubQrys), OrigQry]),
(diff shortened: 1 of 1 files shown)


DSomogyi:
create jira issue

src/riak_kv_qry_worker.erl
```diff
@@ -262,7 +262,31 @@ throttling_spawn_index_fsms(State) ->
 estimate_query_size(#state{n_subqueries_done = NSubqueriesDone} = State)
   when NSubqueriesDone < 2  ->
     State;
-
```
andytill: Is the projected estimation checked when the last chunk has been received, and so could be cancelled?

src/riak_kv_qry_worker.erl
```diff
@@ -262,7 +262,31 @@ throttling_spawn_index_fsms(State) ->
...
+                           result            = Result,
+                           qry = ?SQL_SELECT{'SELECT' = #riak_sel_clause_v1{calc_type = group_by}} =
+                               OrigQry} = State) ->
+    %% Grouping queries will grow its result set to the number of
+    %% unique values in the selection.  In the extreme case of
+    %% grouping by a column of all-unique values, its size will be the
+    %% size of the entire selection.
+    CurrentTotalSize = erlang:external_size(Result),
+    BytesPerChunk = CurrentTotalSize / NSubqueriesDone,
+    ProjectedGrandTotal = round(CurrentTotalSize + BytesPerChunk * length(SubQrys)),
```
andytill: Probably for my benefit only, but could you please put brackets around the multiplication instead of relying on precedence?

src/riak_kv_qry_worker.erl
```diff
@@ -566,4 +590,46 @@ prepare_final_results_test() ->
...
+    ok = check_error_receiver_self_destructs(ErrorReceiverPid, 50).
+
+check_error_receiver_self_destructs(_Pid, 0) ->
+    didnt_receive_error_message;
+check_error_receiver_self_destructs(Pid, N) ->
+    case is_process_alive(Pid) of
+        false ->
+            ok;
+        true ->
+            timer:sleep(100),
```
andytill: This `time:sleep/1` is making the test very slow.

Preferably the side effect could be moved out of the function being tested.

Alternatively, the test process should receive a message instead of waiting on this process to die.

andytill (changes requested):

src/riak_kv_qry_worker.erl
hmmr: Added () in 0596954.

src/riak_kv_qry_worker.erl
hmmr: Simplified in 0596954.

src/riak_kv_qry_worker.erl
hmmr: Yes, checks are performed on every chunk, starting from chunk 2 all the way to the last.

(For the last chunk to cause the query to be cancelled, (1) it needs to be a very large one tipping the average chunk size over the value that was (2) just below `MaxQuerySize / length(SubQrys)` all along.)

andytill: It has already been loaded into memory though, so we know it is ok

hmmr: Even if it is received, the fact remains that we have exceeded the configured limit.

The simplicity of the size checks here is both the strength and the weakness. Ideally, we should be proactively monitoring the amount of memory used, by this query as well as by others served by us, and evaluate chances of successful delivery of the result to the client relative to that. We could, additionally, throttle the vnodes as a means to maintain safe levels of available memory. We should also keep track of the delivery times, to estimate our chances of getting the data in and out before the client request times out.

Instead, we shove responsibility to the user, and only promise the simple checks will be performed and limits observed as configured.

andytill: We just need to skip the estimate check if this is the last chunk. It is already in memory so anything bad that could happen will have happened already.

hmmr: Done in 3584655.

hmmr:
Docs amended in https://github.com/basho/basho_docs/pull/2415.

andytill (approved):

andytill:
+1

hazen:
+1

PR #1628, opened 2017-02-20
RTS-759: Make riak_kv.allow_async_put=false be the default

hazen:
From @erikleitch's original PR:

> Performance improvements for TS1.0 left this code in a weird state. The code had defaulted to `allow_async_put=true`, and we modified the async put path to do fast synchronous writes instead.
>
> This created a situation where `allow_async_put=true actually` called down to `eleveldb:sync_put` to do synchronous writes, and allow_async_put=false called down to `eleveldb:put`, which is a sort-of synchronous put.
>
> This PR attempts to correct the logic controlling synchronous puts, so that:
>
> the version of `handle_request(kv_put_request,...)` called when `allow_async_put=false` causes the code to call `eleveldb:sync_put`, which does the most performant synchronous writes
>
> the version of `handle_request(kv_put_request,...)` called when `allow_async_put=true` causes the code to call `eleveldb:async_put` (the old default behavior), which does less-performant asynchronous writes
>
> `allow_async_put` defaults to false, so that the most performant behavior is now the default

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -479,7 +479,7 @@ init([Index]) ->
     case catch Mod:start(Index, Configuration) of
         {ok, ModState} ->
             %% Get the backend capabilities
-            DoAsyncPut =  case app_helper:get_env(riak_kv, allow_async_put, true) of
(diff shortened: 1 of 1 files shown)


yorkshire-steve:

PR #1629, opened 2017-02-22
Fix some issues with run interval funs for existing sweeper plugins

nickelization:
We had a couple of sweeper funs that weren't defaulting to the correct value of `'disabled'` when there was no defined run interval. This could've caused issues if the value was unset on a running node.

Type specs have also been updated to correctly define what sorts of return values are allowed from run interval funs. Unfortunately, Dialyzer failed to flag any problems with the existing run interval funs, even after the type specs were updated. But at least things are documented correctly now.


raghavkarol:
+1

JeetKunDoug:
+1

nickelization:
@thumbot merge

PR #1630, opened 2017-02-24
Optimize W1C DEL path to remove read-before-write

erikleitch:
---
**Summary**

Key deletes on the normal path happen as follows:

- if no vclock is passed with the key to be deleted, a quorum read of the current value is forced
- a tombstone object is then written with this vclock to all nodes in the preflist
- a get_fsm is initiated to read back the values, ultimately deleting the keys if only tombstones are returned

For the write-once path, the vclock is _never_ used to resolve potential conflicts during a read, so the vclock is irrelevant.  This means that the initial read step above, to determine the vclock, is unnecessary.

This PR eliminates the unnecessary vclock read for both KV and TS versions of the write-once delete path, by substituting a valid empty vclock (i.e., `vclock:fresh()`) for `undefined` on initiation of the delete process.

---
**Files Modified**

There are three paths to write-once deletes.  I discuss each of these in turn, along with the corresponding file modifications:

1. **riak_client.erl** KV deletes are initiated via a call to `riak_client:delete`.  I have modified this to have the same structure as `riak_client:put`, i.e., instead of calling `normal_delete` it calls `maybe_normal_delete`, which checks if the bucket is a write-once bucket or a normal bucket.  If write-once, it calls `write_once_delete`, which initiates the `riak_kv_delete` process with an empty vclock (`vclock:fresh()`), bypassing the read-before-write (see `riak_kv_delete:delete/8`, which switches on the vclock).  If a normal bucket, the process is initiated with an `undefined` vclock, which does the usual read-before-write.

2. **riak_kv_ts_api.erl** TS deletes via the client `delete` API call are initiated via a call to `riak_kv_ts_api:delete_data`.   Previously, `delete_data/5` included a VClock0 arg which if undefined led to the read-before-write behavior discussed above, or allowed a vclock to be passed in if it pre-existed the call.    Because all TS buckets are write-once buckets, this vclock is _always_ irrelevant, and the argument and the logic to parse it and use it in the `riak_kv_delete` initialization are irrelevant.  I have removed the argument from `delete_data` and the logic to check the value.  The call to `riak_client:delete_vclock` has been replaced by a call to an explicitly exported function `riak_client:write_once_delete` to make clear what is going on.

   **riak_kv_ts_svc.erl** Because any vclock passed in with the `tsdelreq` request is ultimately unused, and because the vclock argument to `riak_kv_ts_api:delete_data/5` has been removed,  I have removed the code which extracts it from the `tsdelreq` structure here.

3. TS deletes via the query interface, e.g., `delete from Table where partitionKey='value' and time = 100` is the third path to the write-once delete.  This is initiated via a call to `riak_kv_ts_api:delete_data` from `riak_kv_qry.erl`, so the chain discussed above applies equally to query-deletes.

---
**Performance Tests**

To verify the changes, I've run latency-vs-object-size tests for deletes via all three paths discussed above.  For the KV path, I create a write-once bucket and measure the mean latency for deletes of 100 keys at each object size.  For both of the TS paths, I create a table schema with a single varchar column, and measure the mean latency for deletes of 100 keys at each column size.  Below, I compare delete latencies for the riak_ts-develop branch (dev) and this branch (opt).

Results indicate a similar improvement on all three delete paths.  For object sizes where the latency is dominated by _operational_ latency (ie, communication hops between client and server, and among server threads, < 100KB), the write-once latency drops by a factor of ~2 (1.7).  For object sizes where the latency is dominated by the object size (> 100KB), the latency drops dramatically, since removing the read largely eliminates any handling of the object contents in riak.

screen shot 2017-02-23 at 2 33 38 pm

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -425,12 +425,21 @@ maybe_normal_delete(Bucket, Key, Options, Timeout, {?MODULE, [Node, ClientId]}) 
 write_once_delete(Bucket, Key, Options, {?MODULE, [Node, ClientId]}) ->
     write_once_delete(Bucket, Key, Options, ?DEFAULT_TIMEOUT, {?MODULE, [Node, ClientId]}).
 
-%% Write-once get path ignores vclocks.  We pass in an empty vclock
-%% (vclock:fresh()) so that no read-before-delete is triggered in
-%% riak_kv_delete:delete/8
+%% Write-once get path ignores vclock ancestry, instead comparing the
+%% vclock timestamps when merging write-once objects.
+%%
+%% We therefore pass in a write-once vclock initialized with a
+%% timestamp, as vclock:fresh(<<0:8>>, 1) (see
+%% riak_kv_w1c_worker:w1c_vclock/1) so that:
+%%
+%% 1) No read-before-delete is triggered in riak_kv_delete:delete/8
+%% (because vlock is not undefined), and
+%%
+%% 2) so that riak_object:merge_write_once/2 does the correct thing
+%% (compares vclock timestamps stored under Node <<0:8>>)
 
 write_once_delete(Bucket, Key, Options, Timeout, {?MODULE, [Node, ClientId]}) ->
-    normal_delete_vclock(Bucket, Key, vclock:fresh(), Options, Timeout, {?MODULE, [Node, ClientId]}).
+    normal_delete_vclock(Bucket, Key, vclock:fresh(<<0:8>>, 1), Options, Timeout, {?MODULE, [Node, ClientId]}).
 
 %% Normal path uses vclocks to resolve object merges.  We pass in
 %% undefined to ensure that read-before-delete is triggered in
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -420,30 +420,31 @@ maybe_normal_delete(Bucket, Key, Options, Timeout, {?MODULE, [Node, ClientId]}) 
     end.
 
 %% This version is exported, since it is called directly from
-%% riak_kv_ts_api:delete_data/4
+%% riak_kv_ts_api:delete_data/4.  It simply calls the internal version
+%% with the default timeout.
 
 write_once_delete(Bucket, Key, Options, {?MODULE, [Node, ClientId]}) ->
     write_once_delete(Bucket, Key, Options, ?DEFAULT_TIMEOUT, {?MODULE, [Node, ClientId]}).
 
-%% Write-once get path ignores vclock ancestry, instead comparing the
-%% vclock timestamps when merging write-once objects.
+%% The write-once get path ignores vclock ancestry, instead comparing
+%% just the vclock timestamps when merging write-once objects.
 %%
 %% We therefore pass in a write-once vclock initialized with a
-%% timestamp, as vclock:fresh(<<0:8>>, 1) (see
-%% riak_kv_w1c_worker:w1c_vclock/1) so that:
+%% timestamp (riak_object:new_w1c_vclock/0), so that:
 %%
 %% 1) No read-before-delete is triggered in riak_kv_delete:delete/8
-%% (because vlock is not undefined), and
+%% (because vclock is initialized, and not undefined), and
 %%
-%% 2) so that riak_object:merge_write_once/2 does the correct thing
-%% (compares vclock timestamps stored under Node <<0:8>>)
+%% 2) riak_object:merge_write_once/2 does the correct thing (has a
+%% real vclock time to compare when implementing time-based LWW logic)
 
 write_once_delete(Bucket, Key, Options, Timeout, {?MODULE, [Node, ClientId]}) ->
-    normal_delete_vclock(Bucket, Key, vclock:fresh(<<0:8>>, 1), Options, Timeout, {?MODULE, [Node, ClientId]}).
+    normal_delete_vclock(Bucket, Key, riak_object:new_w1c_vclock(), Options, Timeout, {?MODULE, [Node, ClientId]}).
 
 %% Normal path uses vclocks to resolve object merges.  We pass in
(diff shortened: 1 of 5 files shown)


ph07:
Looks great, thanks for the detailed analysis. Seems it will us 1.6-2x performance improvement for writes, impressive.
Question - what is the purpose of the read back operation after the tombstone is written? Why not let LevelDB perform the compaction?

src/riak_client.erl
```diff
@@ -398,20 +399,45 @@ delete(Bucket,Key,Options,Timeout,{?MODULE, [Node, _ClientId]}=THIS) when is_lis
...
+
+write_once_delete(Bucket, Key, Options, {?MODULE, [Node, ClientId]}) ->
+    write_once_delete(Bucket, Key, Options, ?DEFAULT_TIMEOUT, {?MODULE, [Node, ClientId]}).
+
+%% Write-once get path ignores vclocks.  We pass in an empty vclock
+%% (vclock:fresh()) so that no read-before-delete is triggered in
+%% riak_kv_delete:delete/8
+
+write_once_delete(Bucket, Key, Options, Timeout, {?MODULE, [Node, ClientId]}) ->
+    normal_delete_vclock(Bucket, Key, vclock:fresh(), Options, Timeout, {?MODULE, [Node, ClientId]}).
```
JeetKunDoug: I'm a bit concerned about `vclock:fresh()` here - given other concurrent writes using the logic in [`w1c_vclock`](https://github.com/basho/riak_kv/blob/f8fbc77af4e27395124aaae07715de4b74a90855/src/riak_kv_w1c_worker.erl#L217), and the new "take the last timestamp" logic when resolving values, I think we may end up with deletes _always losing_ against a concurrent write (and remember concurrent doesn't mean "at the same time" - the delete could be later, but end up being overwritten by a node that comes back online and ends up repairing the value).

JeetKunDoug (changes requested):
On first pass, I have some concerns about the "the vclock is never used" note in the PR... that isn't true any more - on the `riak_ts-develop` branch, see [`riak_object:merge_write_once/2`](https://github.com/basho/riak_kv/blob/riak_ts-develop/src/riak_object.erl#L325), which clearly uses vclocks. I left a comment in the appropriate spot to perhaps use similar logic to the `w1c_vclock` function in riak_kv_w1ce_worker to generate the vclock, which I _think_ should take care of the issue,but more thought is needed.

JeetKunDoug:
I need to take another pass at this after lunch, but wanted to get my initial comments in here for you to look at @erikleitch .

src/riak_client.erl
erikleitch: You're absolutely right -- I missed that we had already changed this logic in the TS branches, from: 

```
case crypto:hash(sha, term_to_binary(OldObject)) =< crypto:hash(sha, term_to_binary(NewObject)) of
        true ->
            OldObject;
        _ ->
            NewObject
    end.
```

to 

``` 
pick_last_w1c_timestamp({OldObject, vclock:get_timestamp(<<0:8>>, OldObject#r_object.vclock)},
                            {NewObject, vclock:get_timestamp(<<0:8>>, NewObject#r_object.vclock)}).
```
(I thought that change had only been made in the KV branches).

Given that, I think that if we use `riak_kv_w1c_worker:w1c_vclock/1` in place of `vclock:fresh()`, it should have the correct behavior: a tombstone with a later timestamp than a repair will win, but a near-simultaneous update of the value with a later timestamp than the tombstone will win over the tombstone.

macintux: Don't forget that vclocks have fairly low timestamp resolution (1 second), so the window of opportunity for collisions is quite large. I don't think that really impacts the decision-making process, but worth noting.

I agree, exporting and using `w1c_vclock/1` seems logical.

erikleitch:
In the last commit, I've replaced the vclock:fresh() initialization with vclock:fresh(<<0:8>>, 1), which initializes the vclock with a timestamp as-used by the riak_object:merge_write_once/2 logic.  I include the change in this comment too.

It turns out we can't just export riak_kv_w1c_worker:w1c_vclock/1 because it operates on an object which in this case doesn't exist.  What I would ideally do is modify vclock.erl to include an exported fresh_w1c or fresh(w1c) method, which would return vclock:fresh(<<0:8>>, 1), and modify both riak_kv_w1c_worker and riak_client to use that method, but I hesitated to expand the scope of this PR to yet another repo...

```
%% Write-once get path ignores vclock ancestry, instead comparing the                                                           
%% vclock timestamps when merging write-once objects.                                                                           
%%                                                                                                                              
%% We therefore pass in a write-once vclock initialized with a                                                                  
%% timestamp, as vclock:fresh(<<0:8>>, 1) (see                                                                                  
%% riak_kv_w1c_worker:w1c_vclock/1) so that:                                                                                    
%%                                                                                                                              
%% 1) No read-before-delete is triggered in riak_kv_delete:delete/8                                                             
%% (because vlock is not undefined), and                                                                                        
%%                                                                                                                              
%% 2) so that riak_object:merge_write_once/2 does the correct thing                                                             
%% (compares vclock timestamps stored under Node <<0:8>>)                                                                       

write_once_delete(Bucket, Key, Options, Timeout, {?MODULE, [Node, ClientId]}) ->
    normal_delete_vclock(Bucket, Key, vclock:fresh(<<0:8>>, 1), Options, Timeout, {?MODULE, [Node, ClientId]}).
```

erikleitch:
Last commit refactors the code to remove all explicit calls to `vclock:fresh(<<0:8>>, 1)`, and replace them with calls to `riak_object:new_w1c_vclock()` so that it is clear what is being done.

Affected calls are in `riak_w1c_worker.erl` and `riak_client.erl`, and of course `riak_object` now exports `new_w1c_vclock`

macintux:
+1

JeetKunDoug (approved):

JeetKunDoug:
+1

PR #1631, opened 2017-02-27
Support the `now()` function in query select and where clause

andytill:
riak_ql: https://github.com/basho/riak_ql/pull/170
riak_kv: https://github.com/basho/riak_kv/pull/1631
rfc: https://github.com/basho/rfc/pull/51

Support the `now()` function in the select and where clause. The function returns the local time of the coordinating node.

```sql
SELECT now(), a
FROM mytable
WHERE a = now()
```


src/riak_kv_qry_compiler.erl
```diff
@@ -4195,4 +4262,263 @@ query_with_desc_last_local_key_column_no_quantum_test() ->
...
+now_function_in_range_with_subtraction_test() ->
+    ?assertQueryCompilesToFilter(
+        [{now_milliseconds,10000}],
+        "SELECT * FROM table1 WHERE a = 'hi' AND b = 4000 AND c > now() AND c < now() - 1s",
+        {and_,{'>',{field,<<"c">>,timestamp},{const,10000}},{'<',{field,<<"c">>,timestamp},{const,9000}}}
+    ).
+now_function_in_range_with_multiplication_test() ->
+    ?assertQueryCompilesToFilter(
+        [{now_milliseconds,10000}],
+        "SELECT * FROM table1 WHERE a = 'hi' AND b = 4000 AND c > now() AND c < now() * 2",
```
hmmr, line 4479: I wonder if multiplication and division should be prohibited with `now()`?

src/riak_kv_qry_compiler.erl
```diff
@@ -4195,4 +4262,263 @@ query_with_desc_last_local_key_column_no_quantum_test() ->
...
+now_function_in_where_test() ->
+    ?assertQueryCompilesToFilter(
+        [{now_milliseconds,1979}],
+        "SELECT * FROM table1 WHERE a = 'hi' AND b = 4000 AND c > now()",
+        {'>',{field,<<"c">>,timestamp},{const,1979}}
+    ).
+now_function_in_where_with_addition_test() ->
+    ?assertQueryCompilesToFilter(
+        [{now_milliseconds,1979}],
+        "SELECT * FROM table1 WHERE a = 'hi' AND b = 4000 AND c > now()+1",
```
hmmr, line 4449: How about `now() + 1 * 10`?

src/riak_kv_qry_compiler.erl
```diff
@@ -1159,6 +1174,56 @@ modify_where_key(TupleList, Field, NewVal) ->
...
+                "Function ~p has arity of ~p but was called with ~p arguments.",
+                [FnName, ExpectedArity, ActualArity]),
+            throw({error,{invalid_query,  Msg}})
+    end.
+
+format_binary(Fmt, Args) ->
+    iolist_to_binary(io_lib:format(Fmt, Args)).
+
+cast_value_to_ast(T, V) when (T == integer orelse T == timestamp), is_integer(V) -> {T, V};
+cast_value_to_ast(T, V) when (T == integer orelse T == timestamp), is_float(V)   -> {T, erlang:round(V)};
```
hmmr: With expressions rounded to column type, will conditions like `WHERE a = 1 + 0.1` still evaluate to `true` if the column `a` is of type `sint64` and fetch the row with `a = 1`?

src/riak_kv_qry_compiler.erl
```diff
@@ -58,28 +57,43 @@
...
         {ok, Q2} ->
-            compile_where_clause(DDL, Q2);
+            compile_where_clause(DDL, Q2, Options);
         {error, _} = Error ->
             Error
     end.
 
+%% create a proplist of options for query compilation e.g. configured or
+%% transient values.
+options() ->
```
hmmr, line 86: In a module as big as this, simple a function called simply `options` is not expressive enough. Suggest `compile_options`.

src/riak_kv_qry_compiler.erl
andytill: We can't have a composable language with arbitrary restrictions.

src/riak_kv_qry_compiler.erl
andytill: It might be large but it is single responsibility, a compiler with options is understandable I think. For example the `erl_parse` module just calls `Options`, `Options`. 

http://erlang.org/doc/man/erl_parse.html#abstract-2

src/riak_kv_qry_compiler.erl
andytill: Yes.

src/riak_kv_qry_compiler.erl
andytill: Can you elaborate with the full query that has the issue?

hmmr: `"SELECT * FROM table1 WHERE a = 'hi' AND b = 4000 AND c > now()+1*2"`. I don't contend there _is_ an issue; I just suggest you have it covered in the unit tests.

andytill: Added the unit test in commit 3251a13

gordonguthrie:
+1

hmmr:
+1

andytill:
@thumbot retry

hazen:
:+1:

hazen:
thumbs retry

PR #1634, opened 2017-03-07
Optimize list_buckets in leveldb

fadushin:
This PR includes an optimization of the list_buckets path in the leveldb back end.

In the past, the leveldb back-end would delegate listing buckets to the eleveldb layer.

With this change, we leverage key sorting in leveldb to skip over all keys in a bucket until we get to the next BKey starting with the "next" bucket in lexical order.  This allows us to traverse the back-end data in O(k), where k is the number of buckets, instead of O(n), where n is the number of keys.

This feature requires leveldb, but there are no API changes that alter the behavior of other back ends.

Note: At some point, we should consider implementing the same
strategy in the memory back end.

### DIFF
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -265,28 +265,92 @@ delete(Bucket, PrimaryKey, IndexSpecs, #state{ref=Ref, write_opts=WriteOpts}=Sta
                    any(),
                    [],
                    state()) -> {ok, any()} | {async, fun()}.
+%% TODO: What fold opts are really supported here?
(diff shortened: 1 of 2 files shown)


yorkshire-steve:

fadushin:
Unfortunate this work is being dropped, but maybe some day it will get picked up.

martincox:
Deleted the branch here but archived out in a separate org. I don't foresee a time where all of the open PRs and branches are going to be vetted and taken up - which is unfortunate really.

If anyone shouts up and feels that the work should be taken forward and is wanting to progress it, we can always pull it back in.

fadushin:
Understood.

PR #1636, opened 2017-03-08
Merge develop-2.2 into develop

bsparrow435:
Pick up AAE hash tree upgrade fixes for a few deadlocks during node
join operations.  Pick up fix which allows riak_ensemble puts to
correctly function during handoff.

### DIFF
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -45,6 +45,7 @@
          get_version/0,
          get_partition_version/1,
          get_pending_version/0,
+         get_upgraded/0,
          get_trees_version/0]).
 
 -export([all_pairwise_exchanges/2]).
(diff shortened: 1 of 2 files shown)


src/riak_kv_vnode.erl
```diff
@@ -1176,7 +1176,7 @@ handle_info({ensemble_put, Key, Obj, From}, State=#state{handoff_target=HOTarget
                         {fail, _Idx, _ReqID} ->
                             failed
                     end,
-            ((Reply =/= failed) and (HOTarget =/= undefined)) andalso raw_put(HOTarget, Key, Obj),
+            ((Reply =/= failed) and (HOTarget =/= undefined)) andalso raw_put({Idx, HOTarget}, Key, Obj),
```
fadushin, line 1179: Yes, I remember this.  +1 to fixing that.

src/riak_kv_entropy_manager.erl
```diff
@@ -45,6 +45,7 @@
          get_version/0,
          get_partition_version/1,
          get_pending_version/0,
+         get_upgraded/0,
```
fadushin, line 48: Out of curiosity, is anyone (else) calling get_upgraded/0, or is this a convenience/util function for field diagnosis?

Ah, I see, it's used in check_remote_upgraded.  Still, I like the idea of the function being exposed, too :)

fadushin: (And yes, I think having that call is a good idea)

fadushin:
+1

src/riak_kv_entropy_manager.erl
bsparrow435: This is called remotely as well from other nodes when we're checking if an upgrade has already started in the cluster.

PR #1638, opened 2017-03-13
correct formula for allocated heap size in qbuf dump trigger & reap temp tables sooner

hmmr:
RTS-1809 (in part)

The `heap_size` and `stack_size` items of `process_info` are values in words, which need to be multiplied by `system_info(wordsize)` before they can be used in comparison with values in bytes.

Also, reduce the factory setting of `riak_kv.query.timeseries.qbuf_expire_ms` to 0. It only makes sense with reusable query buffers, abandoned months ago.


andytill:
I didn't notice before, but why is stack size subtracted from the allocated size?

hmmr:
From the docs re `heap_size`:

> Size is the size in words of youngest heap generation of the process. This generation currently include the stack of the process.

If  *includes* the stack size, I gather it should then be subtracted.

andytill:
What is it about the stack size that shouldn't be included in the allocation size?

hmmr:
Arrival of a data chunk does not contribute to stack growth.

andytill:
+1

macintux:
Please update the comment on `riak_kv.query.timeseries.qbuf_expire_ms`.

Help me understand: if cuttlefish supplies a "please ignore this setting entirely" null mapping that would ignore that setting instead of convert it into `riak_kv.timeseries_query_buffers_expire_ms`, would we want to take advantage of that? That value is entirely ignored now?

macintux:
+1

hazen:
:+1:

PR #1639, opened 2017-03-14
Disable tombstone reaping by default

nickelization:
We discussed this in a meeting this afternoon and decided we're not going to have this plugin enabled by default in the next release.


angrycub:
+1

PR #1642, opened 2017-03-16
Make Sweeper config variable names consistent

nickelization:
We have one setting name prefixed with "sweeper." and some other ones starting with "riak_kv.sweep." All the sweeper options should probably start with the same prefix.

Additionally, since "Riak KV" means "Riak, the KV store database" to the general public, we shouldn't be prefixing settings with "riak_kv" (especially since we could merge the sweeper over to Riak TS at some point ;-)

Using the "sweeper." prefix everywhere is succinct and helpful, since Cuttlefish is meant to be user-friendly, and customers will think of the Sweeper as its own distinct component. Whichever application it happens to be located in is probably irrelevant to them, hence my decision to kill the "riak_kv." prefix.

### DIFF
--- a/priv/riak_kv.schema
+++ b/priv/riak_kv.schema
@@ -646,27 +646,27 @@
 %% * `always` (default) No restrictions
 %% * `never` Sweeps will never be attempted
 %% * `window` Hours during which sweeps is permitted, where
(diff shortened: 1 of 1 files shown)


angrycub:
:+1:

fadushin:
+1

PR #1643, opened 2017-03-21
Fix for outstanding riak_kv/679 dataloss cases

russelldb:
These commits fix the remaining kv679 dataloss edge cases. They include a simple quickcheck model/simulator. The riak-tests for these changes can be found here https://github.com/basho/riak_test/pull/1296.

The fix is a slight change to the per-actor-epoch code that shipped in 2.1, whereas before we only updated the epoch on coordinating writes, now we update it on non-coordinating writes where we detect some local amnesia. The fix is to add a no-op/no-dot entry to the version vector with a new actor-epoch and a counter of one. This ensures that the next write coordinated by the vnode uses the new epoch, and is not instead a repeated event.


JeetKunDoug:
@russelldb if you can start by addressing the dialyzer warnings while Nick and I are reviewing I'd appreciate it. On first pass, makes sense to me and generally looks good, but I want to think a bit more about it. There are a few stylistic things we're trying to avoid in here, but nothing we can't clean up post-merge.

russelldb:
Happy to address the stylistic concerns if you air them.

Did I introduce new dialyzer warnings?

russelldb:
Ah, I see it, thumbs is pretty useful/helpful.

JeetKunDoug:
Yea - we ❤️ Thumbs these days... only real issues are where we may have set some of the settings in a way that we later figured out we didn't like...

I was just running dialyzer on `develop` to make sure there isn't anything outstanding there before replying to you but looking at the errors I'm fairly certain they are from this commit (and can now confirm that is true).

src/riak_kv_vnode.erl
```diff
@@ -1681,29 +1688,23 @@ select_newest_content(Mult) ->
...
-    %% @TODO Check the clock of the incoming object, if it is more advanced
-    %% for our actor than we are then something is amiss, and we need
-    %% to mark the actor as dirty for this key
     ResObj = riak_object:syntactic_merge(CurObj, UpdObj),
-    case riak_object:equal(ResObj, CurObj) of
-        true ->
-            {oldobj, CurObj};
-        false ->
-            {newobj, ResObj}
+    if NewEpoch ->
```
JeetKunDoug: please replace this with a `case NewEpoch` - trying to kill `if` statements with 🔥 these days, and would like to get rid of the ?ELSE macro as well (but understand that the other usage isn't currently part of this PR, so it's OK to leave it for now).

src/riak_kv_vnode.erl
```diff
@@ -1512,7 +1512,7 @@ prepare_put_new_object(#state{idx =Idx} = State,
                      false ->
                          []
                  end,
-    {EpochId, State2} = new_key_epoch(State),
+    {EpochId, State2} = maybe_new_key_epoch(Coord, State, undefined, RObj),
```
nickelization: Instead of using `undefined` here, could we use a more distinctive atom like `no_object` or something along those lines? Would make the code a bit clearer, and also help guard against any future bugs where an unrelated use of `undefined` might leak into this code. Admittedly this is somewhat of a personal style preference though, so I won't argue if you'd strongly prefer leave it as-is. Looking through the code, we'd have to update `maybe_new_key_epoch` and also `highest_actor`, but I think it should be a safe and easy change.

(Commenting on the first commit where this appears, but it does look like this code later got moved into `maybe_update_vclock`.)

src/riak_kv_vnode.erl
```diff
@@ -2598,47 +2598,40 @@ non_neg_env(App, EnvVar, Default) when is_integer(Default),
...
-                                 {binary(), #state{}}.
-maybe_new_key_epoch(false, State, _, _) ->
-    %% Never add a new key epoch when not coordinating
-    %% @TODO (rdb) need to mark actor as dirty though.
-    {State#state.vnodeid, State};
-maybe_new_key_epoch(true, State=#state{counter=#counter_state{use=false}, vnodeid=VId}, _, _) ->
+-spec maybe_new_key_epoch(Coord::boolean(), State::#state{},
+                          LocalObject::riak_object:riak_object(),
+                          IncomingObject::riak_object:riak_object()) ->
+                                 {ActorId:: binary(), #state{}}.
```
nickelization: Thanks for adding the variable names to the type specs - definitely improves readability in many places.

In the case of `ActorId` I see we specify `ActorId :: binary()` in a couple places. Maybe worth defining a local `actor_id()` type in this module to avoid redundancy?

src/riak_kv_vnode.erl
```diff
@@ -2656,17 +2649,19 @@ key_epoch_actor(ActorBin, Cntr) ->
...
-%% never acted before. `KeyEpoch' is the highest epoch for the
-%% `ActorBase'. `Counter' is the greatest event seen by the `VnodeId'.
+%% key. It is the actor we want to increment for a new event, if we
+%% are not starting a new epoch for the key.  Must work with
+%% non-epochal and epochal actors.  The return tuple is `{ActorId,
+%% Epoch, Counter}' where `ActorId' is the highest ID starting with
+%% `ActorBase' that has acted on this key. `KeyEpoch' is the highest
+%% epoch for the `ActorBase'. `Counter' is the greatest event seen by
+%% the `VnodeId'.
 -spec highest_actor(binary(), riak_object:riak_object()) ->
```
nickelization, line 2716: Since we are now allowed to pass `undefined` into this function, we should add it to the type spec.

nickelization:
@russelldb what happened? I was about to continue my review, and was excited to get this stuff merged.

russelldb:
I have no idea, I didn't do anything!

nickelization:
Wow, weird! Well, glad it was nothing then, haha.

russelldb:
Figured it out, I had a script running that did a push remote --mirror but the local repo did not have the branch so it deleted it, sorry for the confusion

nickelization:
Ahh gotcha, no worries, makes sense.

russelldb:
in the past all my PRs where from a basho branch, not a fork :)

src/riak_kv_vnode.erl
```diff
@@ -2598,75 +2629,95 @@ non_neg_env(App, EnvVar, Default) when is_integer(Default),
...
-                          riak_object:riak_object(),
-                          riak_object:riak_object()) ->
-                                 {binary(), #state{}}.
-maybe_new_key_epoch(false, State, _, _) ->
-    %% Never add a new key epoch when not coordinating
-    %% @TODO (rdb) need to mark actor as dirty though.
-    {State#state.vnodeid, State};
-maybe_new_key_epoch(true, State=#state{counter=#counter_state{use=false}, vnodeid=VId}, _, _) ->
+-spec maybe_new_key_epoch(Coord::boolean(), State::#state{},
+                          LocalObject::riak_object:riak_object(),
```
nickelization: This type spec should include `undefined` as a possible value for `LocalObject` (or whatever other placeholder we end up using to mean "no local object" if we end up changing the code to use a different atom).

test/kv679_eqc.erl
```diff
@@ -0,0 +1,493 @@
...
+
+-define(NUMTESTS, 100).
+-define(QC_OUT(P),
+        eqc:on_output(fun(Str, Args) ->
+                              io:format(user, Str, Args) end, P)).
+
+eqc_test_() ->
+    {setup,
+     fun() ->
+             meck:new(riak_core_bucket),
```
nickelization: We could just call `setup()` and `teardown()` here instead of repeating the bodies of the functions.

test/kv679_eqc.erl
```diff
@@ -0,0 +1,493 @@
...
+    %%         %% locally there is no entry for VId but the incoming
+    %%         %% object has an entry
+    %%         O2 = riak_object:syntactic_merge(LocalObj, IncomingObject),
+    %%         {N, riak_object:new_actor_epoch(O2, A)};
+    %%     {Epoch, _A} ->
+    %%         {Epoch, riak_object:syntactic_merge(LocalObj, IncomingObject)}
+    %% end.
+
+%% like vclock:timestamp() but monotinically increasing
+timestamp() ->
```
nickelization: This function will not return monotonically increasing values (or even necessarily unique ones) since we're adding the microseconds field directly to the seconds field without any scaling. I'm not sure if this is critical to the correctness of this test though, will have to think about it a bit more.

russelldb: It's my mistake, I was meant to do something smarter but never finished. Will address. So far the test has worked, but that might be good luck.

test/kv679_eqc.erl
russelldb: Will do, I added `setup` and `teardown` after the `eqc_test` function, didn't refactor.

test/kv679_eqc.erl
```diff
@@ -0,0 +1,493 @@
...
+
+gen_pl(Replica) ->
+    ?LET(R2, ?SUCHTHAT(X, elements(?REPLICAS), X /= Replica), {Replica, R2}).
+
+%% Helper functions
+
+set_up_replicas() ->
+    [ ets:insert(?MODULE, #replica{id=Id}) || Id <- ?REPLICAS ++ [?P1] ].
+
+%% two dicts come in, a true or a dict comes out, you can't explain
```
nickelization, line 314: 😆

test/kv679_eqc.erl
```diff
@@ -0,0 +1,493 @@
...
+
+-compile(export_all).
+
+-define(P1, p1).
+-define(REPLICAS, [p2, p3, f1, f2, f3]).
+
+-record(state, {
+          pfhead=?P1,
+          %% Actor Ids for replicas in the system
+          replicas= ?REPLICAS,
```
nickelization: Seems like this is never modified, right? Would be slightly simpler to just use the `?REPLICAS` constant directly in the code and kill this state field. Though if there are future plans to make this field mutable then it might be worth keeping it around.

EDIT - Same with `pfhead` too.

nickelization:
I've picked over all this code with a fine-tooth comb, and other than the few minor things I noted in my comments, it looks great. Solid stuff! @JeetKunDoug is setting up a Giddyup run with this code to check for any regressions, but assuming that all passes, this gets a +1 from me.

test/kv679_eqc.erl
russelldb: yeah, these are left over from earlier iterations.

russelldb:
Many thanks for the review. I'm working on it. One dialyzer warning remains, I have had to post to the erlang ML for help.

src/riak_kv_vnode.erl
russelldb: Ah, I found an existing type `vnodeid()` but I'm not sure if it muddies the water since in some places that means the one generated by the vnode_status_mgr, and some cases it means the actor_id generated by smushing together the vnodeid and epoch. Maybe we need both, using the bitstring <<_:Length>> type spec?

russelldb:
Re-running the eqc I've found some new counter examples. Hold off on this until I figure them out, please.

nickelization:
@russelldb Was chatting with @JeetKunDoug and he mentioned a test failure during the Giddyup run for `partition_repair`. I checked it out locally, and looks like it fails consistently (and passes on develop). Haven't done any real investigation into the failure, and I'm about to sign off for the day, but I'll try and take a look tomorrow morning if you don't get a chance to check it out yourself before then.

russelldb:
@nickelization looking at it. So far a bit baffled. All the `norfound` objects are from the same partition, partition `0`. Hopefully I'll have a better idea by the time you're online

nickelization:
@russelldb been a busy day and I haven't had much chance to look into this any further yet (I probably should've mentioned earlier, but today is my last day working for Basho, so I've had some loose ends to tie up). The one potential lead I have for you via Doug is that he noticed those new "amnesia" log messages showing up during repairs. I guess that might not be unexpected, due to the nature of the test...but again, I haven't had a chance to look into it any further yet, and might not get to it before my time here is up :(

russelldb:
@nickelization wow. Well thanks for all the time you put into this PR up to now. All the best in whatever you're doing next. Stay in touch, please! Hope we can work together again in the future.

russelldb:
@JeetKunDoug Finally figured it out. https://github.com/basho/riak_test/blob/develop/tests/partition_repair.erl#L223 is the problem

The data in the vnode partition no longer matches the data in the stash since it now has the new epoch actor entry added to the vclock. I guess I need to makes this `verify` function smarter, in that each key in the partition after repair must have the same value, and _at least_ the vclock of the stash. Sound about right?

JeetKunDoug:
@russelldb yea - that looks about right to me.. a more intelligent comparison should do it. LMK if you need anything else, or when you've got changes done, and I'll take another look.

russelldb:
@JeetKunDoug I pushed a fix to the riak-test at https://github.com/basho/riak_test/pull/1296

JeetKunDoug:
@russelldb thanks - it seemed like yo had solutions for the dialyzer errors - any clues on the eunit failure in the Thumbs output? Looks potentially unrelated but curious if you have taken a look yet? If not I can try to see if it's reproducible locally on develop or not and we can go from there.

nickelization:
@russelldb you're welcome, and thanks! Same to you, on all counts :)

russelldb:
@JeetKunDoug I used your solutions for the dialyzer errors! Will look at the eunit

russelldb:
@JeetKunDoug Fixed the failing eunit. `fsm_qc_vnode` uses `riak_kv_vnode:put_merge`. That 

> %% TODO: The riak_kv_vnode code should be refactored to expose this function
> %%       so we are close to testing the real thing.

is looking more and more pertinent as time goes on :D

JeetKunDoug:
Now that all the unit test/dialyzer issues are taken care of, going to kick off another integration test run... should have results in the morning.

JeetKunDoug:
+1 - integration tests run and triaged. Looks good. Thanks @russelldb for the great work, as always.

angrycub:
+1

JeetKunDoug:
@thumbot merge

russelldb:
Wooo! Thanks @JeetKunDoug and @nickelization for the review.

davidx:
@thumbot settings

PR #1657, opened 2017-12-07
Intermittent test failure fixes

russelldb:
Also some deps changes that will need looking at again as we move from tags to branches.

Still for discussion - the JMX stuff in the schema. It should probably be complete for all jmx properties, and the same for SNMP, and an error logged at start up if they're configured `true` or `on`.

The other commits are test fixes for `hll` and `stats`, and some dirty path munging for running tests either in riak_kv or at the top level riak project.

### DIFF
--- a/src/riak_kv_stat.erl
+++ b/src/riak_kv_stat.erl
@@ -516,7 +516,7 @@ stats() ->
      %% node stats: gets
      {[node, gets], spiral, [], [{one  , node_gets},
                                  {count, node_gets_total}]},
-     {[node, gets, fsm, active], counter, [], [{value, node_get_fsm_active}]},
+     {[node, gets, fsm, active], counter},
      {[node, gets, fsm, errors], spiral, [], [{one, node_get_fsm_errors},
                                               {count, node_get_fsm_errors_total}]},
      {[node, gets, objsize], histogram, [], [{mean  , node_get_fsm_objsize_mean},
--- a/src/riak_kv_test_util.erl
+++ b/src/riak_kv_test_util.erl
@@ -170,7 +170,6 @@ wait_for_children(PPid) ->
 setup(TestName, SetupFun) ->
     %% Cleanup in case a previous test did not
     cleanup(TestName, SetupFun, setup),
-
     %% Load application environments
     Deps = dep_apps(TestName, SetupFun),
     do_dep_apps(load, Deps),
@@ -209,6 +208,7 @@ cleanup(Test, CleanupFun, StartedApps) ->
                                            not lists:member(A, Deps)
                                    end, lists:reverse(StartedApps)),
 
+
     %% Stop the applications in reverse order.
     do_dep_apps(stop, Apps),
 
@@ -221,7 +221,6 @@ cleanup(Test, CleanupFun, StartedApps) ->
     %% where riak_kv_stat is already started by another supervisor from a
     %% previous test.
     wait_for_unregister(riak_kv_stat),
-
     %% Stop distributed Erlang
     net_kernel:stop(),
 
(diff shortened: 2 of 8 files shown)


src/riak_kv_stat.erl
```diff
@@ -516,7 +516,7 @@ stats() ->
      %% node stats: gets
      {[node, gets], spiral, [], [{one  , node_gets},
                                  {count, node_gets_total}]},
-     {[node, gets, fsm, active], counter, [], [{value, node_get_fsm_active}]},
+     {[node, gets, fsm, active], counter},
```
ghost, line 519: Looks good, just a question really, how come the other args are no longer required?

src/riak_kv_test_util.erl
```diff
@@ -287,18 +287,19 @@ dep_apps(Test, Extra) ->
...
+                %%     	{ok, Started} = start_app_and_deps(A, Acc),
+                %%             Started;
+                %%         _ ->
+                %%             Acc
+                %%     end;
+                %% (F, Acc) ->
+                %%    F(start),
+                %%    Acc
+                %% end,
+                [], Apps);
```
ghost: Good call - nested function definition inside fold was ugly.

src/riak_kv_test_util.erl
```diff
@@ -265,8 +264,9 @@ dep_apps(Test, Extra) ->
                 application:set_env(riak_core, ring_state_dir, Test ++ "/ring"),
                 application:set_env(riak_core, platform_data_dir, Test ++ "/data"),
                 application:set_env(riak_core, handoff_port, 0), %% pick a random handoff port
-                {ok, Dir} = file:get_cwd(),
-                Dirs = [Dir ++ "/../deps/*/priv"],
+                %% @TODO this is wrong still
```
ghost: some indication of why it's wrong might be helpful

src/riak_kv_test_util.erl
```diff
@@ -326,6 +339,7 @@ start_app_and_deps(Application, Started) ->
         true ->
             {ok, Started};
         false ->
+            _Apps = application:which_applications(),
```
ghost: I'm curious as to what this does

ghost (changes requested):

src/riak_kv_test_util.erl
russelldb: should probably fix the commenting out though

src/riak_kv_test_util.erl
russelldb: I have no idea…I'll have to have a read of the whole file again. This is what happens with so much time between commit and review, eh?

russelldb: I think it just needs to be taken out, probably a different idea, and then underbarred the binding just to make it compile.

russelldb: Oh yeah, it was for some printf debugging

src/riak_kv_test_util.erl
russelldb: …erm…you got me. It's "wrong" in that it is a hack maybe?

src/riak_kv_stat.erl
russelldb: Hmmmm…

russelldb: No idea, it looks wrong though, eh? `node_get_fsm_active` vs `node_gets` everywhere else. Damn, I wish I could remember october 2017 a little better.

russelldb: Hmmm…yeah…really…no idea…will have to go back and read through it all again

russelldb: I honestly don't understand stats since they were changed to exometer.

russelldb: Oh the commit message says it is "an unused duplicate alias", that answer your question?

ghost: that's fine, just curious

russelldb: Look at https://github.com/basho/riak_kv/blob/35b6e56e827895f068bb74cfdbf80eda945fad5a/src/riak_kv_stat.erl#L622 it's not "args no longer required?" but that they were never required.

ghost: Got it, thanks.

russelldb:
Pushed some commits to address @bryanhuntesl comments

src/riak_kv_test_util.erl
ghost: Adding a comment to draw attention to it doesn't detract from the P/R - don't pay this one any more attention on my account.

ghost (approved):

ghost (approved):

ghost (approved):

ghost (approved):

russelldb:
Same question, squash first?

ghost:
+1 squash

PR #1547, opened 2016-11-28
Delete macro unused since 17242c9

lucafavatella:

PR #1667, opened 2018-03-23
Delete macro unused since 17242c9

ghost:
Originally

https://github.com/basho/riak_kv/pull/1547

But mistakenly targeted at develop


russelldb:
lgtm. No error in eunit tests. Grep suggest that macro is unused.

+1 from me…BUT I have to ask, what led the original PR author to this change, particularly? Is it a test of the community process?

I'm +1, but in general a PR needs a _reason_ in my opinion, a why.

ghost:
Original author was investigating the use of etag to reduce cost of repeated http requests. As follow on from that, I'm evaluating/testing Luca's changes to riak_kv (small change) to expose CRDT context and use that as Etag atm .

russelldb:
Interesting. What would having the context (and not the body) tell the client?

I ask because the context might not change, but the value might change considerably for a Set or a Map.

yorkshire-steve:

PR #1668, opened 2018-03-26
Expose CRDT context as HTTP ETag header on JSON HEAD/GET request 

ghost:
OK so this originally came from [Luca Favatella](https://github.com/lucafavatella) (ESL) #1588 - but:

* targeted at 'develop' branch rather than develop2-2
* slight error - putting a binary into the return tuple/3 (list expected)

There's a [last modified](https://github.com/webmachine/webmachine/wiki/Resource-Functions#last_modified) webmachine callback too, but that's outside the scope of this pull request.

Example: 

```
./rel/riak/bin/riak-admin bucket-type create maps '{"props":{"datatype":"map"}}'
```
```
./rel/riak/bin/riak-admin bucket-type activate maps
```
```
curl -XPOST http://localhost:8098/types/maps/buckets/foo/datatypes/bar \
   -d ' { "update": { "first_name_register": "Dave", "phone_number_register": "555567" } }'
```
```
curl -vvv --connect-timeout 2 -XHEAD -H "Content-Type: application/json"  http://localhost:8098/types/maps/buckets/foo/datatypes/bar
```
```
Warning: Setting custom HTTP method to HEAD with -X/--request may not work the
Warning: way you want. Consider using -I/--head instead.
*   Trying ::1...      
* TCP_NODELAY set                                        
* Connection failed                               
* connect to ::1 port 8098 failed: Connection refused
*   Trying 127.0.0.1...         
* TCP_NODELAY set    
* Connected to localhost (127.0.0.1) port 8098 (#0)
> HEAD /types/maps/buckets/foo/datatypes/bar HTTP/1.1
> Host: localhost:8098
> User-Agent: curl/7.54.0                   
> Accept: */*
> Content-Type: application/json
>
< HTTP/1.1 200 OK
< Vary: Accept-Encoding
< Server: MochiWeb/1.1 WebMachine/1.10.9 (cafe not found)
< ETag: "g2wAAAABaAJtAAAADCMJ/vmrKPL5AAAAAWEBag=="
< Date: Mon, 26 Mar 2018 20:04:07 GMT
< Content-Length: 144
<
^C

```

### DIFF
--- a/src/riak_kv_wm_crdt.erl
+++ b/src/riak_kv_wm_crdt.erl
@@ -157,7 +157,8 @@
          encodings_provided/2,
          resource_exists/2,
          process_post/2,           %% POST handler
-         produce_json/2           %% GET/HEAD handler
+         produce_json/2,           %% GET/HEAD handler
+         generate_etag/2
         ]).
 
 -include_lib("webmachine/include/webmachine.hrl").
@@ -402,6 +403,22 @@ produce_json(RD, Ctx=#ctx{module=Mod, data=RObj, include_context=I}) ->
     Body = riak_kv_crdt_json:fetch_response_to_json(
                      Type, Value, get_context(RespCtx,I), ModMap),
     {mochijson2:encode(Body), RD, Ctx}.
+
+generate_etag(RD, Ctx=#ctx{module=Mod, data=RObj, method=Method})
+  when Method == 'GET'; Method == 'HEAD' ->
+    {{RespCtx, _}, _} = riak_kv_crdt:value(RObj, Mod),
+    ETag =
+        case RespCtx of
+            undefined ->
+                undefined;
+            <<>> ->
+                undefined;
+            _ when is_binary(RespCtx) ->
+                base64:encode(RespCtx) %% HTTP/1.1 ETag (https://tools.ietf.org/html/rfc7232#section-2.3) is valid Base64 (https://tools.ietf.org/html/rfc2045#section-6.8).
+        end,
+    {ETag, RD, Ctx};
+generate_etag(RD, Ctx) ->
+    {undefined, RD, Ctx}.
 
 %% Internal functions
 
--- a/src/riak_kv_wm_crdt.erl
+++ b/src/riak_kv_wm_crdt.erl
@@ -414,7 +414,7 @@ generate_etag(RD, Ctx=#ctx{module=Mod, data=RObj, method=Method})
             <<>> ->
                 undefined;
             _ when is_binary(RespCtx) ->
-                base64:encode(RespCtx) %% HTTP/1.1 ETag (https://tools.ietf.org/html/rfc7232#section-2.3) is valid Base64 (https://tools.ietf.org/html/rfc2045#section-6.8).
+                binary_to_list(base64:encode(RespCtx)) %% HTTP/1.1 ETag (https://tools.ietf.org/html/rfc7232#section-2.3) is valid Base64 (https://tools.ietf.org/html/rfc2045#section-6.8).
         end,
     {ETag, RD, Ctx};
 generate_etag(RD, Ctx) ->


martinsumner:
Conditional read requests are a bit odd in Riak - because 99% of the work has to be done to discover you can send back a 304 response (when compared to sending a 200), i.e. n=3 reads (from disk) across the network and object merging etc.  The only bit it avoids is sending the last network hop to the client.

This last network hop might be significant if the objects are large, but CRDTs are currently constrained so that they shouldn't be large.  So I guess I'm struggling with the need, beyond the consistency with other parts of Riak.

As a PR, would this be better waiting for consideration with 3.0 work on larger CRDTs?  As then it would potentially have a bit more purpose?

martinsumner:
Not a major change I know, I just have a bit of fear about why it was left out in the first place.

binarytemple:
Thanks @martinsumner. Ideally, this would be used when traversing a list of crdt in order to audit an installation over an extremely expensive or slow network link. 

It's pretty low risk, would you accept it if it only matched on head requests? That would increase the complexity in the client though as they'd have to use two different paths when fetching objects via GET. 

The original pr was last year, Luca has a load of them outstanding (big Riak fan) for a bunch of different repositories. 

Would be nice to see two crdt improvements into the next build (this and gsets) and start testing them before the public get involved. 

My only concern would be whether it would be backwards compatible to earlier Riak 2.0 versions. 

Perhaps @russelldb would have some insight into this (compatibility) ?

martinsumner:
The target contents of this release have been agreed for a number of months though (back to StokeCon 2017), and it has taken 3 months to get from PRs to the point where we are ready to tag the release.  The PR may have lay around for ages, but I don't remember it being mentioned as a need for this release before.  Perhaps I've got this wrong?

I've never know anyone working on a release react well to post deadline drops with no tests (however small the change).  Unless it is fixing a production issue, can we not just put it back on the backlog for now?  For nothing more than reasons of etiquette?  It is been a hard enough 3 months for some grinding towards a fixed target, let alone making it a moving one.

src/riak_kv_wm_crdt.erl
```diff
@@ -403,6 +404,22 @@ produce_json(RD, Ctx=#ctx{module=Mod, data=RObj, include_context=I}) ->
                      Type, Value, get_context(RespCtx,I), ModMap),
     {mochijson2:encode(Body), RD, Ctx}.
 
+generate_etag(RD, Ctx=#ctx{module=Mod, data=RObj, method=Method})
+  when Method == 'GET'; Method == 'HEAD' ->
+    {{RespCtx, _}, _} = riak_kv_crdt:value(RObj, Mod),
```
russelldb, line 409: We need to talk about this. As I commented elsewhere, this CRDT context is not suitable as an etag. Imagine a Set CRDT that has had two items added, its context may be `[{actor1, 2}]` and it has a value of `[alice, bob]`. Now imagine I remove `bob`. The context is _still_ `[{actor1, 2}]` but the value is now `[bob]`. You can't use the CRDT context (and opaque binary term) as an ETag since the underlying value can change without the context changing.

russelldb:
I remember leaving a comment about this elsewhere, confusing. As I said in the review I just started, and in the "elsewhere" comment, CRDT contexts (as implemented in Riak) won't work as ETags. Happy to discuss more.

Martin is right also, was going to tag on Thursday/Friday for release. But equally, I understand that it would be helpful and generally good to be flexible and include whatever we can for the betterment of the community and Riak

Given the scope of the change I have no problem including it in 2.2.5, except, as stated above, I don't think it does what the author thinks it does.

However, I don't think this code is ready, until we discuss the issue with contexts and etags. I don't see why there can't be a 2.2.6 in a month or two if we can't get this sorted swiftly.

russelldb (changes requested):
Why not just use the riak_object version vector to generate the etag, as riak does with regular objects?

ghost:
Great feedback, special thanks to @russelldb for pointing out the (critical) implementation flaw, I'm pretty sure that wasn't the desired behavior 😂

I agree that this should be targeted at develop-2.2.6, there's no particular push to get it in this week but I think it could add some value.

For the future PR I suggest :  

* Expose the riak_object version vector as suggested by Russell. 
* Expose the last modified attribute for good measure (also useful for auditing)
* Provide a test - implemented in the same fashion as the other tests on wm/riakc .

If I went ahead and implemented these suggestions, should we be returning both (Etag and Last-Modified) for both GET and HEAD requests? I understand to be correct but haven't dug out any RFCs.

Does that sound like a good idea?

russelldb:
Sounds good to me @binarytemple . I'd have a peek at what is done for conditional requests for regular objects if the behaviour needs clarifying.

If you need help with riak_test give me a shout.

PR #1675, opened 2019-01-15
Develop 2.9

martinsumner:
Merge develop-2.9 changes into main basho repo for riak_kv.

The changes include:

- Changes to GET and PUT FSM (and riak_kv_vnode) to support use of HEAD capability within backends;

- Integration of leveled backend;

- Addition of support for kv_index_tictactree based anti-entropy;

- Support for aae_folds to provide inter-cluster anti-entropy building blocks based on kv_index_tictactree;

- Integration with core node worker pool and ability to queue async work based on previously snapped snapshots;

- Vnode soft limit monitoring when selecting PUT co-ordinators.


martinsumner:
@martincox - this gets all the work on 2.9.0 into the develop-2.9 branch on riak_kv.  Can you +1?

martincox:
This is a behemoth of a PR!

Am I right in seeing that the additional features introduced here, such as HEAD requests are exposed exclusively through the HTTP API? Therefore, we should maybe raise an issue for these to be exposed through PB (if needed / required)?

martinsumner:
The only API features are the aae_fold API and the new riak_repl feature which allows an object to be touched (prompting re-replication without modification).  These two features are only available via HTTP - so I'll create an issue for the PB follow-up (although I'm not sure with aae_fold if in the long run the right answer is exposure via API or access direct using riak_client).

The use of HEAD request is internal only at the moment (and only changes behaviour when the backend supports the `head` capability, and when the user has made a GET request).  There is no change to the API involved in this.  I will add an issue with changing the behaviour if the user submits a HEAD request (currently this is still a use of GET under the hood with the body stripped at the end, and this could be improved)

PR #1682, opened 2019-02-27
Leveled - Always v1

martinsumner:
Change the object format check on PUT to always use v1 if the backend is leveled.


martinsumner:
Impacts leveled only, so merging in

PR #1683, opened 2019-02-28
Make pause on delete

martinsumner:
Dumb mistake detected in code review - the pause on delete had actually been removed


martinsumner:
Following code review with @ramensen

PR #1686, opened 2019-03-07
Updated tests and eqc to OTP20

ThomasArts:


martinsumner:
Removing locks across the board.  This is causing continuous issues with getting a stable build as we change.  We will not .gitignore the rebar.lock in the root riak repo.

PR #1689, opened 2019-03-18
RFC: [riak 3.0] replace eper with (newer) redbug + recon

llelf:
eper is now [pretty much dead](https://github.com/massemanet/eper#readme).


martinsumner:
https://github.com/basho/riak_kv/blob/develop-3.0/rebar.config

This change happened via the develop-2.9 merge, and is now in 3.0

PR #1702, opened 2019-06-19
Develop 2.9

martinsumner:
Changes for riak kv 2.9.0 patch.

After discovery of another memory-related issue with leveled, this change:

- rolls forward the leveled version to include fix of memory management (using binary copy for sub-parts to prevent the retention of large binary objects)
- adds identification of leveled backends, and additional logging related to memory, to try and ease troubleshooting of memory issues going forward


martinsumner:
@ramensen this is the riak_kv uplift which needs to be merged in so that a new riak_kv tag can be produced to patch riak for the leveled memory issues

ranisen:
Agree this needs to go in. Thanks

PR #1708, opened 2019-08-22
i1707 merge fake object - log don't crash

martinsumner:
Unable to pinpoint the cause of this issue.  This change logs some further details at the point the issue occurs, and protects from the vnode crash - causing only a PUT failure instead.


martinsumner:
https://github.com/basho/riak_test/pull/1321

martinsumner:
One question is - is it safe to not let it crash?  In that, in theory, the problem may be related to vnode state and so the crash resets the issue.  Without a crash/reset could more PUTs fail?

Also, in not crashing, we now only have the log information, and not the crash.log, to work with to understand the issue.

It is likely only future hindsight will be able to tell us the right answer.  On balance, it is determined that failing the PUT not the vnode is currently the best approach - as the cause of the issue is unknown, its impact should be limited.

PR #1710, opened 2019-08-23
Mas i1677 head

martinsumner:
As part of the investigation into the implementation of a genuine HEAD API to Riak, the current GET code and its use of HEADs was reviewed.  The outcome of this review was:

- The code was hard to follow and lacking in both comments and dialyzer specs;
- The code was lazy, it would give sub-optimal answers in some cases because the extra fetch workload was probably not a problem.  However, this variance increases the scope for there to be incorrect behaviour that is not detected by system tests.

This PR adds comments and specs.  It also refactors the code in attempt to both improve readability and consistency of behaviour.  The unit tests have then been expanded to reflect this.

The change also removes an is_x_deleted check, as deleted tombstones (even when fetched by HEAD requests) should already not require fetching as they have already been converted into actual objects (not just HEAD objects) - https://github.com/basho/riak_kv/blob/riak_kv-2.9.0p4/src/riak_kv_get_core.erl#L109-L119.

### DIFF
--- a/src/riak_kv_get_core.erl
+++ b/src/riak_kv_get_core.erl
@@ -374,6 +374,10 @@ merge(Replies, AllowMult) ->
 %% a backend not supporting HEAD was called, or the operation was an UPDATE),
 %% or body-less objects.
 %%
+-spec merge_heads(list(result()), boolean()) ->
+        {notfound, undefined}|
+        {tombstone, riak_object:riak_object()}|{ok, riak_object:riak_object()}|
+        {fetch, list(non_neg_integer())}.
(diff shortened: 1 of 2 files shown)


martinsumner:
This has been run through a standard volume test, as well as riak_test tests, as well as the extended unit tests.  

Without the fix the extended unit tests fail.  This is not obviously critical, however, it would be safer  and more reliable to have the fixes in place

yorkshire-steve:
+1 from me. Code reviewed and discussed with Martin. Agreed including this fix in next 2.9.0 patch is appropriate.

PR #1716, opened 2019-09-25
3.0 dialyzer fixes

martinsumner:
This PR requires updates to riak_pipe and riak_core to complete dialyzer fixes (found in branch mas-3.0-dialyzer1)


src/riak_kv_hll.erl
```diff
@@ -189,7 +189,10 @@ to_version(_Version, HllSet) ->
 stats(HllSet) ->
     [{S, stat(S, HllSet)} || S <- ?STATS].
 
--spec stat(stat(), hllset()) -> byte() | precision() | undefined.
+-spec stat(stat(), hllset()) -> number() | undefined.
```
ThomasArts: Actually, if I try to make the type:
`-spec stat(stat(), hllset()) -> 1..255 | 4..16 | undefined.` it passes dialyzer

src/riak_kv_index_hashtree.erl
```diff
@@ -304,7 +304,8 @@ init([Index, VNPid, Opts]) ->
                            built=false,
                            use_2i=Use2i,
                            path=Path,
-                           version=Version},
+                           version=Version,
+                           expired = false},
```
ThomasArts, line 308: seems unrelated to dialyzer, but fine

src/riak_kv_mrc_map.erl
```diff
@@ -140,12 +140,6 @@ process(Input, _Last,
         {ok, _NonListResults} ->
             ?T(_FittingDetails, [map, error],
                {error, {non_list_result, Input}}),
-            {ok, State};
-        {forward_preflist, Reason} ->
```
ThomasArts, line 144: Here strange things seem to happen... you remove the forward_preflist alternative, but also make it crash in case the client would return `{error, _}`

weird that Dialyzer does not pick it up that map actually can return an error tuple in its type, but not in reality.
```
-spec map(riak_kv_mrc_pipe:map_query_fun(), term(), term())
         -> {ok, term()}
          | {forward_preflist, Reason :: term()}
          | {error, Reason :: term()}.
```
I advise to change the type.

src/riak_kv_util.erl
```diff
@@ -306,11 +306,10 @@ fix_incorrect_index_entries(Idx, FixFun, Acc0, FixOpts) ->
         {_,_,ErrorCount}=Res ->
             MarkRes = mark_indexes_reformatted(Idx, ErrorCount, ForUpgrade),
             case MarkRes of
-                error ->
-                    %% there was an error marking the partition as reformatted. treat this like
-                    %% any other error (indicating the need to re-run reformatting)
-                    {element(1, Res), element(2, Res), 1};
-                _ -> Res
+                ok ->
```
ThomasArts: This reads a bit weird. Why not just always return Res
```
{_,_,ErrorCount}=Res ->
            mark_indexes_reformatted(Idx, ErrorCount, ForUpgrade),
            Res
```
Showing that this index formatting is purely there for it's side effect

src/riak_kv_vnode.erl
```diff
@@ -3359,45 +3362,18 @@ object_info({Bucket, _Key}=BKey) ->
 handoff_data_encoding_method() ->
     riak_core_capability:get({riak_kv, handoff_data_encoding}, encode_zlib).
 
-%% Decode a binary object. We first try to interpret the data as a "new format" object which indicates
-%% its encoding method, but if that fails we use the legacy zlib and protocol buffer decoding:
+%% Decode a binary object. Assumes data is in new format, legacy no longer
```
ThomasArts, line 3365: gives a nice cleanup!

src/riak_kv_wm_index.erl
```diff
@@ -181,15 +180,15 @@ malformed_request(RD, Ctx) ->
...
+    Continuation = wrq:get_qs_value(?Q_2I_CONTINUATION, RD),
     PgSort0 = wrq:get_qs_value(?Q_2I_PAGINATION_SORT, RD),
     PgSort = case PgSort0 of
         undefined -> undefined;
         _ -> normalize_boolean(string:to_lower(PgSort0))
     end,
-    MaxResults0 = wrq:get_qs_value(?Q_2I_MAX_RESULTS, ?ALL_2I_RESULTS, RD),
-    TermRegex = wrq:get_qs_value(?Q_2I_TERM_REGEX, undefined, RD),
-    Timeout0 =  wrq:get_qs_value("timeout", undefined, RD),
+    MaxResults0 = wrq:get_qs_value(?Q_2I_MAX_RESULTS, RD),
```
ThomasArts: This works, because you use validate_max later on. For clarity, it would probably be nicer to call validate_max directly on this value.

```
MaxVal = validate_max(wry:get_qs_value(...))
```

src/riak_kv_wm_index.erl
```diff
@@ -316,7 +315,7 @@ validate_timeout(Str) ->
             {error, Str}
     end.
 
-validate_max(all) ->
+validate_max(undefined) ->
```
ThomasArts, line 317: slightly concerned about the mismatch LessThanZero below. Can be equal to zero.
But this has nothing to do with this refactoring for dialyzer

ThomasArts (approved):

src/riak_kv_hll.erl
martinsumner: Interesting, this is strange behaviour.  I will amend and use that definition

src/riak_kv_index_hashtree.erl
martinsumner: Dialyzer needs it, as otherwise expired would be undefined (there is no default value for state.expired)

src/riak_kv_mrc_map.erl
martinsumner: Agreed, will change

src/riak_kv_util.erl
martinsumner: Agreed.  I think the whole index reformatting logic is as a result of a legacy upgrade, and so no longer relevant.

I will make your change for now, and then think about stripping the whole index reformat logic in a later branch

src/riak_kv_wm_index.erl
martinsumner: I've changed the variable name for now.  Zero would make no sense

src/riak_kv_wm_index.erl
martinsumner: Agreed

Issue #1714, opened 2019-09-13
Node failure - whether or not to "let it crash"


martinsumner:
In order to correctly recover from a failure, `riak_core_node_watcher` must detect that the service is down on a given node.  Without such an event, fallback nodes will not be started, and the vnodes used in requests will not change to reflect those fallback vnodes.

My understanding is that there are two triggers for discovering a failed node in `riak_core_node_watcher`:

- erlang node monitoring through net_kernel
- registering a health_check to test the availability of a particular service on a node

The supervision tree for riak_kv will normally attempt to restart individual services that fail, but if they continue to fail with sufficient intensity the failure will ripple up the supervision tree and bring the node down.  This will trigger a `node_down` message for other the riak_core_node_watcher to detect on other nodes.

As of, a long time ago, the use of health checks (the second method) was disabled by default in riak_kv: https://github.com/basho/riak_kv/commit/5860d68a8ac7eed3eb33e9346d90b946864a488f.  The actual health check used seems to overlap with overload handling, and it is easy to see how it could have a negative impact with node flapping under load.

This then leads to the question:

- What if a failure can render a riak-kv node inoperable, but there are no crashes in the supervision tree?

There appears to have been a case of this with a riak customer, where it appears the file system got switched into a read-only mode on at least one node.  The vnode backend was eleveldb, and the impacted vnodes were responding to PUTs with a [db_write error](https://github.com/basho/eleveldb/blob/develop-2.9/c_src/workitems.cc#L205), but this was returned as an error - it did not cause a crash of either the backend or the vnode.

As the node was still "up", there had been no crash, the vnodes on the broken node were still being selected to coordinate PUTs.  Any PUT coordinated by a vnode on the failed node would then fail (a failure on a coordinator is immediately sent back to the application as a failure, there is no attempt to try the other n -1 vnodes).  So the application saw an ongoing series of intermittent failures.

[...]

PR #1717, opened 2019-09-27
Mas i1714 readonlyfs

martinsumner:
This is related to issue - https://github.com/basho/riak_kv/issues/1714

This is tested by - https://github.com/basho/riak_test/blob/mas-i1714-readonlyfs/tests/verify_readonly.erl

If a file-system is unusable at startup, a node won't start.  However, if a node is up and running and then the file-system cannot be used (for example because permissions have been changed, or the a container has remounted the drive as read-only), then a bitcask or eleveldb backend will not crash.  It will instead return errors for every PUT.  Any PUT for which that vnode is chosen as the co-ordinator will then error back to the client (with a report of the file system error).

This error situation may continue for a long time without the cluster recovering.  Without a node going down, the cluster will not start fallback vnodes and change the ring.  To crash the node the vnodes must crash (to ripple the crash up the supervision tree).

This change detects certain low-level errors, and decides that they should cause the vnode to crash.  This is any `db_write` error in leveldb, or certain posix file system errors in bitcask (eacces, erofs, enodev, enospc).

With these changes the test now passes, if we remove write access permissions from a node, that node will fail and the cluster will recover with almost no PUT failures (worst case o(1)).

### DIFF
--- a/src/riak_kv_bitcask_backend.erl
+++ b/src/riak_kv_bitcask_backend.erl
@@ -57,7 +57,7 @@
 -define(VERSION_FILE, "version.txt").
 -define(API_VERSION, 1).
 -define(CAPABILITIES, [async_fold,size]).
--define(TERMINAL_POSIX_ERRORS, [eacces, erofs, enodev]).
+-define(TERMINAL_POSIX_ERRORS, [eacces, erofs, enodev, enospc]).
 
 %% must not be 131, otherwise will match t2b in error
 %% yes, I know that this is horrible.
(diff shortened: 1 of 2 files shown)


martinsumner:
@martincox This would change behaviour for you guys, so it needs your review.  There's no urgency, and we could also make the change to eleveldb only if necessary.

martincox:
Wondering if it might be better to stop gracefully (`riak:stop/1`), given that we're catching explicit known errors? Would save a bunch of noise from crashes, maybe.

martinsumner:
A few components and backends will always write some state to disk when they are called to stop, so I don't think you will avoid noise from crashes (although there would be less noise) as this will result in crashes anyway.

It may be that the issue is very temporary (especially possible in cloud environments), so stopping not crashing would stop this being handled through automatic restart and being controllable through restart intensity.  I think forcing a crash and putting the outcome in the hands of the supervisors is preferable.

martincox:
Yep, that does make sense. Didn't think about trying to shut down gracefully and not being able to actually persist anything.

Looks good to me. :+1:

PR #1718, opened 2019-10-01
Mas i1711 eper

martinsumner:
Redbug and recon are known to be used tools, and continue to be supported, so request them directly (rather than redbug being pulled in via eper).

This should resolve the intermittent unit test failure in eper.

The change requires the following change to be made in riak - https://github.com/basho/riak/pull/989

The prove the change there is the following update to riak_test - https://github.com/basho/riak_test/pull/1325


martinsumner:
@martincox is this OK with you?
@llelf tis is to bring [this proposed change](https://github.com/basho/riak_kv/pull/1689) into the 2.9 stack, it will then get rolled through into 3.0

ranisen:
+1 - Happy with that change

PR #1720, opened 2019-10-02
Mas i1719 multihead

martinsumner:
Adds support for head/3 request to the multi_backend if the underlying backend mapped to that bucket supports it.

This is tested in https://github.com/basho/riak_test/pull/1326

The test covers backend assignment direct to bucket properties and via types.


ranisen:
+1 Happy with this change

PR #1726, opened 2019-10-18
QuickCheck tests update

ThomasArts:
Using `os:timestamp()` in the QuickCheck property made us move from monotonic to non-monotonic time. In get_fsm_eqc we test with `allow_mult == false` and assumed monotonic time. Therefore, failures popped up that are well known to appear if time is none monotonic and therefore there is a `allow_mult == true` option.

We changed the property to always use monotonically increasing time.

Other changes are mostly cosmetic. Note that we move eunit-formatter from a production dependency to a test dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: compile rel cover test dialyzer
+.PHONY: compile rel cover test dialyzer eqc
 REBAR=./rebar3
 
 compile:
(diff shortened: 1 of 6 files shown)


martinsumner:
Suggest description for get_fsm eqc test:

%% Description
%%
%% The test is based on a fixed set of generated objects.  These objects are
%% then used as the randomised answers from vnodes to the get_fsm.  The 
%% objects have a deterministic merge order (based on the assumption the 
%% timestamps generated are strictly monotonically increasing) - and so the
%% test can compare the get_fsm result to the expected result.
%%
%% As well as varying the object responses, the GET options and parameters are
%% mutated to create further scenarios.
%%
%% TODO:
%% The test does not use dotted values at present, or HEAD responses
%%

PR #1729, opened 2019-10-31
Develop 3.0 lastgaspring

martinsumner:
See https://github.com/basho/riak_core/pull/945

This depends on this riak_core PR.

Last gasp monitoring added in all 3 places where ensemble tries to react to ring changes.  Also blocks vnodes starting on a lastgasp ring.


martinsumner:
@ThomasArts - this makes use of the lastgasp on the ring to avoid some of the log noise issues.  Also has the necessary additional changes to stop riak_ensemble from firing off the last gasp ring.

ThomasArts (approved):
Only after 20th tick the old logic kicks in

PR #1738, opened 2019-12-16
Include option for security on nextgen_repl

martinsumner:
Can add security on protocol buffers interface only.  It is required to be certificate-authenticated SSL - no other security options are supported - so client as well as server certificates are required.

Tested within the test [nextgenrepl_rtq_pbsecurity](https://github.com/basho/riak_test/blob/mas-i1691-ttaaefullsync/tests/nextgenrepl_rtq_pbsecurity.erl)

Note that enabling replication security for nextgen_repl requires security to be enabled for all applications of the cluster.  This is a change form riak_repl where security was an independent option to security on the general API.

Not all failure scenarios are tested here - the pc_security riak_test provides a more general test of the API security features.

### DIFF
--- a/src/riak_kv_ttaaefs_manager.erl
+++ b/src/riak_kv_ttaaefs_manager.erl
@@ -500,7 +500,7 @@ generate_sendfun(SendClient, NVal) ->
 %% @doc
 %% Make a remote client for connecting to the remote cluster
 -spec init_client(client_protocol(), client_ip(), client_port(),
-                    string()|undefined)
+                    ssl_credentials()|undefined)
                     -> {rhc:rhc()|no_client, rhc}|
                         {pid()|no_client, riakc_pb_socket}.
 init_client(http, IP, Port, _Cert) ->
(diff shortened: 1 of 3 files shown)


martinsumner:
@ThomasArts this PR enables the option for SSL security for the new repl features (with mutual certificate authentication).  riak_test test for this is reference in the PR above.

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -459,10 +499,11 @@ generate_sendfun(SendClient, NVal) ->
 
 %% @doc
 %% Make a remote client for connecting to the remote cluster
--spec init_client(client_protocol(), client_ip(), client_port())
+-spec init_client(client_protocol(), client_ip(), client_port(),
+                    string()|undefined)
```
ThomasArts: This seems wrong... wonder why dialyzer does not catch it.
The last argument is used specified to be string() | undefined but below it is used to pass on Credentials, which is a 4-tuple.

ThomasArts (commented):
Please check the type of string() in init_client.

src/riak_kv_ttaaefs_manager.erl
martinsumner: Yes good catch.  I don't know why dialyzer is missing it either

PR #1739, opened 2020-01-10
Change sink worker to allow for multiple queues in config

martinsumner:
Peer configuration can now contain a queue-name, and hence we can have mutliple queues in the configuration.

Also allows for the worker count to be changed for a given queue.

There is now an iteration number kept - to make it easier when switching configurations to discard in-flight work items that relate to old configuration.

This change means that protocol no longer defaults to 'http' - it must be explicity set.

### DIFF
--- a/src/riak_kv_replrtq_snk.erl
+++ b/src/riak_kv_replrtq_snk.erl
@@ -373,7 +373,7 @@ code_change(_OldVsn, State, _Extra) ->
 %%% Internal functions
 %%%============================================================================
 
-% @doc calculate the mean avoiding divide by 0
+%% @doc calculate the mean avoiding divide by 0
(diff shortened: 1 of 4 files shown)


src/riak_kv_replrtq_snk.erl
```diff
@@ -339,25 +381,39 @@ calc_mean(Time, Count) ->
 % usable peer information.
 % tokenised string expected to be of form:
 % "192.168.10.1:8098|192.168.10.2:8098 etc"
```
ThomasArts: should we not have a name here in the comment as well?

ThomasArts (approved):

src/riak_kv_replrtq_snk.erl
martinsumner: Updated comment

PR #1743, opened 2020-01-20
Mas i1691 pauseonbacklog

martinsumner:
The legacy AAE implementation had defence against the vnode backend being able to handle PUT load at a much faster pace than the AAE backend.  This would work by having a counter of PUTs, and every N PUTs rather than sending the PUT async, it would do so sync - and the vnode would be delayed to allow the AAE system to catch-up.

This is the token-based mechanism used on the old AAE implementation - https://github.com/basho/riak_kv/blob/develop-2.9/src/riak_kv_vnode.erl#L3317-L3325.
 
There were two problems with this in production - the delay could be very long (especially when rebuilds had been initiated), and there was no monitoring logs/stats to inform the operator that the delay was occurring.

Originally, the potential for an AAE backlog was ignored when implementing Tictac AAE - there was no mechanism to pause the vnode.  This PR provides this extra defence, but is designed so that:

- vnode is only delayed when the backlog reaches 1s
- the vnode is not delayed indefinitely
- the operator is informed at each stage.

The PR also rolls in two fixes that have been discovered during load testing.

- updating the count of sink workers doesn't have the required prompt_work() built in;
- there is a process leak in the riak_kv_ttaaefs_manager whereby it keeps initiating new PBC clients, but doesn't stop them.

### DIFF
--- a/docs/NextGenREPL.md
+++ b/docs/NextGenREPL.md
@@ -130,27 +130,29 @@ Each node in `riak_kv` starts three processes that manage the inter-cluster repl
 
 ### Real-time Replication - Step by Step
 
-Previous replication implementations initiate replication through a post-commit hook.  Post-commit hooks are fired from the `riak_kv_put_fsm` after "enough" responses have been received from other vnodes (based on n, w, dw and pw values for the PUT).  Without enough responses, the replication hook is not fired, although the client should receive an error and retry, a process of retrying that eventually may fire the hook.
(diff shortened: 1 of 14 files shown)


martinsumner:
Short queue time used to invoke this in test - https://github.com/basho/riak_test/blob/mas-i1691-ttaaefullsync/tests/verify_tictac_aae.erl#L80 - and this confirms AAE still operates.

martinsumner:
Pause element and aae_ping reviewed with @ThomasArts

ThomasArts (approved):

PR #1747, opened 2020-01-31
Mas i1691 compress option

martinsumner:
Add option to compress objects before sending them via repl.

### DIFF
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -68,14 +68,16 @@
 -record(state, {client,    % local client
                 req,       % current request (for multi-message requests like list keys)
                 req_ctx,   % context to go along with request (partial results, request ids etc)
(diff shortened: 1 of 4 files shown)


src/riak_kv_pb_object.erl
```diff
@@ -189,7 +190,9 @@ process(#rpbfetchreq{queuename = QueueName}, #state{client=C} = State) ->
                                 deleted_vclock = pbify_rpbvc(Vclock)},
                 State};
         {ok, RObj} ->
-            EncObj = riak_object:nextgenrepl_encode(repl_v1, RObj),
+            ToCompress =
```
ThomasArts: If you are not going to change the value other than via configuration, it is much faster to save the value ToCompress in the state in the init function and match on the state in this branch.

ThomasArts (approved):

src/riak_kv_pb_object.erl
martinsumner: change made on state

PR #1748, opened 2020-02-04
Log not crash on remote connection/timeout failure

martinsumner:
It would be expected to have issues reaching remote clusters - they need to be logged, but not fill the crash.log


ThomasArts (approved):

martinsumner:
Depends on https://github.com/martinsumner/kv_index_tictactree/pull/68

PR #1752, opened 2020-03-24
Explicitly filter out non-primary exchanges

martincox:
Explicitly filter out non-primary exchanges from exchange_queue as opposed to flipping indexes and allowing it to fail. Prevents logging of not_responsible messages when these errorneous exchanges are attempted.

Relates to https://github.com/basho/riak/issues/1011

### DIFF
--- a/src/riak_kv_entropy_manager.erl
+++ b/src/riak_kv_entropy_manager.erl
@@ -956,11 +956,7 @@ add_index_exchanges(Index, State) ->
 -spec prune_exchanges([exchange()])
                      -> [exchange()].
 prune_exchanges(Exchanges) ->
-    L = [if A < B ->
-                 {A, B, IndexN};
-            true ->
-                 {B, A, IndexN}
-         end || {A, B, IndexN} <- Exchanges],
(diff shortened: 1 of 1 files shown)


src/riak_kv_entropy_manager.erl
```diff
@@ -956,11 +956,7 @@ add_index_exchanges(Index, State) ->
 -spec prune_exchanges([exchange()])
                      -> [exchange()].
 prune_exchanges(Exchanges) ->
-    L = [if A < B ->
-                 {A, B, IndexN};
-            true ->
-                 {B, A, IndexN}
-         end || {A, B, IndexN} <- Exchanges],
+    L = [Exchange || {A, B, _IndexN} = Exchange <- Exchanges, A < B],
```
ThomasArts, line 959: I have a hard time to see that in all cases in which prune_exchages is used, this will give the desired result. The original returns all tripples `{A, B, I}` and ensures that A =< B for all of them. Your solution just removes all `{A, B, I}` if A >= B.
Apparently you want this, because that's what the commit comment states. But how is function used by the rest of riak?

Issue #1754, opened 2020-03-26
Recent writes loss in the presence of heavy handoff activity


keynslug:
Riak KV may lost recent writes in the presence of heavy handoff activity.

I've [made a test][1] with the help of Jepsen framework specifically to investigate this kind of buggy behaviour because it hit us several times already in our production environment. This test observes an occurence of this bug almost every run to be honest. An ordinary test run follows roughly these steps:
1. Set up 5 riak kv nodes in a docker-compose environment.
2. Join them together in a single cluster.
3. Start pouring reads and writes over some small keyspace (10 keys was sufficient most of the time) concurrently through some small number of clients (again, 40 was sufficient), each putting as little as 1 rps. Note that this process would continue until the very end. Note also that clients take effort to ensure that writes are _causally consistent_ (using past reads' vclocks) and performed with n = 3, r/w/pr/pw = quorum, lww = false, allow_mult = false.
4. Give the cluster 40 seconds to stabilise. Most of the time the cluster would still have a number of ongoing transfers by this time, but that's ok.
5. Then, set up 2 extra nodes and join them in the cluster.
6. Again, give the cluster another 120 seconds to stabilise, after that order those 2 extra nodes to _leave_ the cluster.
7. Wait for them to leave cleanly. This is the most time consuming phase, but usually 200 seconds was sufficient for it to finish.
8. Repeat the process starting from step 4, until the test run time limit passes. Usually I ran it for 15 minutes, it was enough for 3 such cycles.

I have attached a [Jepsen report archive][2] of one such test run. Most notable observations are in `results.edn`, here's an excerpt, redacted for brewity:
```
...
"jepsen638479067-6" {:valid? false,
    :configs (...),
    :final-paths ([{:op {:process 25,
                        :type :ok,
                        :f :read,
                        :value 20,
                        :index 16396,
                        :time 207095405146,
                        :vclock "a85hYGD...JZAA=="},
                    :model #knossos.model.CASRegister{:value 20}}
                    {:op {:process 25,
                        :type :ok,
                        :f :read,
                        :value 31,
                        :index 16408,
                        :time 207247072654,
                        :vclock "a85hYGD...SyAA=="},
                    :model #knossos.model.Inconsistent{:msg "can't read 31 from register 20"}}]),
    :previous-ok {...},
    :last-op {...},
    :op {...}},
...
```

[...]

PR #1755, opened 2020-04-12
Fix spurious lost put coordinator acks

keynslug:
Put FSM did not account for the possibility of multiple coordinator forwards, hence failing to acknowledge request originator in such cases.  Multiple forwards are presumably possible under high transfer activity in the cluster, when participating nodes may end up disagreeing on the ring state for a short amount of time.

More details: https://github.com/basho/riak_kv/issues/1754

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -215,7 +215,14 @@ monitor_remote_coordinator(false = _UseAckP, _MiddleMan, _CoordNode, StateData) 
     {stop, normal, StateData};
 monitor_remote_coordinator(true = _UseAckP, MiddleMan, CoordNode, StateData) ->
     receive
-        {ack, CoordNode, now_executing} ->
+        {ack, CoordNodeFinal, now_executing} ->
+            case CoordNodeFinal of
+                CoordNode ->
+                    ok;
+                _ ->
+                    lager:warning("unexpected forward-ack from ~p, expected from ~p",
+                                  [CoordNodeFinal, CoordNode])
+            end,
             {stop, normal, StateData}
     after StateData#state.coordinator_timeout ->
             exit(MiddleMan, kill),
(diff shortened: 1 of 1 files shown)


keynslug:
Ok I finally managed to run some tests.

1. I did not observed a single instance of lost write after running Jepsen test from #1754 five times in a row with this patch applied.

   There was several instances of `unexpected forward-ack from riak@<some-node>, expected from riak@<another-node>` in logs though, which I introduced in this patch.

2. It (presumably) did not break anything, given the results of running `riak_test` groups `kv_all` and `smoke.

    There are 4 failing testcases, most of which due to discrepancies between my test environment (docker container, vanilla Erlang/OTP 17) and test authors' expectations. Details on each of them:
    * `pb_cipher_suites-leveled` fails with ciphersuite mismatch,
    * `pb_security-leveled` fails with unexpected success in opening a pb connection,
    * `verify_riak_stats-leveled` fails to start up nodes with yokozuna enabled, supposedly because there was no usable JRE in a container,
    * `kv679_dataloss_fb2-leveled` fails with timeout waiting for 4 nodes to become connected, though looking at test code it seems to me that there was no attempt to bring one of those nodes back to life.

That said I believe the PR is ready for review.

martinsumner:
Thanks for all the effort,  I will look at this within the next two days.

martinsumner:
riak_test tests need to be run with OTP16 (basho flavour).  I think the test failures are understandable as environment issues - but I'm going to double-check.

eunit tests and eqc tests on riak_kv all pass as expected (eqc tests were checked on merge into `develop-3.0` to pick-up 3.0 improvements and fixes).

martinsumner (approved):

martinsumner:
All riak_test tests have passed in standardised environment with this patch.

martinsumner:
Although there exists a Jepsen test to prove this, running Jepsen tests isn't part of the release process - so there exists risk of future regression.

I'm concerned that the effort to re-produce in riak_test of property-based test will be disproportionate to the value gained.

@ThomasArts @martincox any bright ideas on an ongoing test for this to check we don't regress in the future - or are you happy that the one-off validation via Jepsen is sufficient in this case.

I'm leaning towards accepting without further testing, but I might just be being a bit lazy.  A riak_test should be possible, although there might be the need for quite a bit of intercepting, so the riak_test might be inefficient to maintain.

PR #1757, opened 2020-05-10
Allow the mbox_check to be disabled

martincox:
Where it may be necessary to disable this option, inspect boolean environment variable `mbox_check_enabled` flag to allow it to be toggled without passing additional client options.

Default behaviour leaves the option as originally intended / configured, with the explicit override disabling it, if it has been set.

### DIFF
--- a/src/riak_kv_put_fsm.erl
+++ b/src/riak_kv_put_fsm.erl
@@ -1245,7 +1245,7 @@ get_soft_limit_option(Options) ->
     %% turn off soft-limit checking. However, by default, we should
     %% use them (if supported) - unless it is explicitly disabled by toggling the
     %% mbox_check_enabled environment flag.
-    SoftLimitedWanted = app_helper:get_env(riak_kv, mbox_check_enabled, true) andalso
+    SoftLimitedWanted = app_helper:get_env(riak_kv, mbox_check_enabled, SoftLimitSupported) andalso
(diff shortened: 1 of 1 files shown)


martinsumner (approved):

martinsumner:
There's a test for this that could be extended:

https://github.com/basho/riak_test/blob/develop-3.0-292/tests/verify_vnode_polling.erl

However, this has been missed out of the test groups at present.  I'll have a look at sorting the test before 3.0 release.

martincox:
Didn't realise. I'll have a look at the test 👍

martinsumner:
Did you have a look at the test?  Want to make sure I don't miss this for 3.0

martinsumner:
It was a simple change in the end to test this - https://github.com/basho/riak_test/blob/b1b16240b842baa2e2835aefd0545b1e420f47da/tests/verify_vnode_polling.erl

martinsumner:
@martincox I created a related PR (https://github.com/basho/riak_kv/pull/1763) to bring this into develop-2.9 as well.

However, in doing that I considered the situation whereby mbox_check_enabled could be specifically set to true, but SoftLimit is not supported in the cluster - this should resolve to no mbox_check, but doesn't necessarily so.  Or have I misunderstood.

martincox:
Yeah, think you're right. In both instances (client options and environment config), you could set it to true without the cluster supporting it? I guess the better way might be to constrain it via schema.

Issue #1759, opened 2020-06-08
Vnode clearing does not clear tictac_aae state


martinsumner:
https://github.com/basho/riak_kv/blob/develop-2.9/src/riak_kv_vnode.erl#L2208-L2225

This does not destroy any parallel tictac_aae store, or tictac_aae hashtree caches.  This can lead to problems when a node leaves, then is re-joined without manually clearing the disk.

PR #1760, opened 2020-06-12
Force clearing of Tictac AAE on vnode delete

martinsumner:
To prevent an issue whereby state can be left if a node leaves then re-joins.  Clear tictac_aae just like standard aae

https://github.com/basho/riak_kv/issues/1759


martinsumner:
https://github.com/basho/riak_test/pull/1345

ThomasArts (approved):

Issue #1758, opened 2020-06-06
Unhandled false return from riak_index:object_key_in_range


pma:
This case statement https://github.com/basho/riak_kv/blob/develop-2.9/src/riak_kv_leveled_backend.erl#L405 should handle a possible return of 'false' from riak_index:object_key_in_range, due to this line here: 
https://github.com/basho/riak_kv/blob/develop-2.9/src/riak_index.erl#L469

I got the pattern match error while playing with RiakCS using leveled as backend. Listing directories via the s3 API would trigger the error under some situations I couldn't  reproduce reliably.

martinsumner:
Any clues at all about the conditions?  It will help me generate a riak_test test.

I can match the riak_kv_eleveldb_backend code to resolve this I assume - https://github.com/basho/riak_kv/blob/develop-2.9/src/riak_kv_eleveldb_backend.erl#L854-L855

pma:
@martinsumner I still have to figure out how RiakCS  maps paths to keys to come up with a test case that does not depend on running RiakCS. This was the simpler case to reproduce via the S3 API:

**When using leveldb backend**
```
~ ❯❯❯ s3cmd mb s3://mybucket                                                                                                                                                                                    ⏎
Bucket 's3://mybucket/' created
~ ❯❯❯ s3cmd put test.txt s3://mybucket/foo/bar/baz/test.txt                                                                                                                                                     ⏎
upload: 'test.txt' -> 's3://mybucket/foo/bar/baz/test.txt'  [1 of 1]
 4 of 4   100% in    0s    85.65 B/s  done
~ ❯❯❯ s3cmd put test.txt s3://mybucket/
upload: 'test.txt' -> 's3://mybucket/test.txt'  [1 of 1]
 4 of 4   100% in    0s    91.32 B/s  done
~ ❯❯❯ s3cmd ls s3://mybucket/foo/
                       DIR   s3://mybucket/foo/bar/
```

**When using leveled backend**
```
~ ❯❯❯ s3cmd mb s3://mybucket                                                                                                                                                                                    ⏎
Bucket 's3://mybucket/' created
~ ❯❯❯ s3cmd put test.txt s3://mybucket/foo/bar/baz/test.txt
upload: 'test.txt' -> 's3://mybucket/foo/bar/baz/test.txt'  [1 of 1]
 4 of 4   100% in    0s    78.07 B/s  done
~ ❯❯❯ s3cmd put test.txt s3://mybucket/
[...]

PR #1761, opened 2020-06-15
Mas i1758 objectoutofrange

martinsumner:
Riak CS object folds were not correctly supported.  The folds were not starting on start_key and not stopping on out of range (crashing instead).

The "skip" state has never been tested, but this was incorrectly returning the Acc not the InnerAcc - so this is now resolved.  It is not clear how the skip state can be encountered.

### DIFF
--- a/src/riak_kv_leveled_backend.erl
+++ b/src/riak_kv_leveled_backend.erl
@@ -447,13 +447,15 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{bookie=Bookie}) ->
                                 throw(InnerAcc)
                         end
                     end,
+                EndKey = null,
                 % StartKey and StartInclusive based on query, but the EndKey
                 % and EndInclusive should be handled by the passed in fold
-                % function, so null is used for EndKey
+                % function (by the riak_index range checker), so null is used
+                % for EndKey
                 leveled_bookie:book_objectfold(Bookie, 
                                                 ?RIAK_TAG,
                                                 FilterBucket, 
-                                                {StartKey, null}, 
+                                                {StartKey, EndKey},
                                                 {SpecialFoldFun, Acc}, 
                                                 false);
             {false, false} ->
--- a/src/riak_kv_leveled_backend.erl
+++ b/src/riak_kv_leveled_backend.erl
@@ -418,7 +418,10 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{bookie=Bookie}) ->
                                                         InnerAcc)
                                 end;
                             {skip, _BK} ->
-                                Acc
+                                Acc;
+                            _ ->
+                                % Required for riak_test/tests/verify_cs_bucket
+                                throw({break, Acc})
                         end
                     end,
                 leveled_bookie:book_objectfold(Bookie, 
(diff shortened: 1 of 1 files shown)


martinsumner:
https://github.com/basho/riak_kv/issues/1758

src/riak_kv_leveled_backend.erl
```diff
@@ -400,31 +408,41 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{bookie=Bookie}) ->
...
+                                        FoldObjectsFun(ObjB,
+                                                        {o, ObjK, Obj}, 
                                                         InnerAcc)
                                 end;
                             {skip, _BK} ->
-                                Acc
+                                InnerAcc;
+                            _ ->
+                                % Required for riak_test/tests/verify_cs_bucket
+                                throw(InnerAcc)
```
ThomasArts, line 447: probably `throw({key_out_of_range, InnerAcc})` if it needs a throw at all?

src/riak_kv_leveled_backend.erl
```diff
@@ -400,31 +408,41 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{bookie=Bookie}) ->
...
                         end
                     end,
+                % StartKey and StartInclusive based on query, but the EndKey
+                % and EndInclusive should be handled by the passed in fold
+                % function, so null is used for EndKey
                 leveled_bookie:book_objectfold(Bookie, 
                                                 ?RIAK_TAG,
                                                 FilterBucket, 
-                                                all, 
+                                                {StartKey, null},
```
ThomasArts: The commend talks about EndKey, but here StartKey is `null` that is confusing

ThomasArts: I guess this is an interval from StartKey to`null`

src/riak_kv_leveled_backend.erl
```diff
@@ -400,31 +408,41 @@ fold_objects(FoldObjectsFun, Acc, Opts, #state{bookie=Bookie}) ->
...
-                                Acc
+                                InnerAcc;
+                            _ ->
+                                % Required for riak_test/tests/verify_cs_bucket
+                                throw(InnerAcc)
                         end
                     end,
+                % StartKey and StartInclusive based on query, but the EndKey
+                % and EndInclusive should be handled by the passed in fold
+                % function, so null is used for EndKey
```
ThomasArts:
```suggestion
                % function, so the interval is denoted {StartKey, null} where null is the EndKey
```

src/riak_kv_leveled_backend.erl
martinsumner: Yes, null means there's no end to the range - so it will stop on the first object out of range (or the end of the bucket).

There is perhaps a more general problem here about end_inclusive.  riak_kv_eleveldb_backend uses the riak_index range checkers and a throw for most queries, so that end_inclusive and start_inclusive is controlled via the those range checkers (and at the end of the range this means handling the state false).

In other queries the leveled backend hasn't used these checks, and relied on an end key in the range.  However, this then misses the control about end_inclusive.  It is only in this CS fold we have discovered the end_inclusive being used, so it is only here we rely on the range checker not setting an end_key to finish the fold.

In other folds, leveled backends will ignore end_inclusive, but leveldb backends will respect end_inclusive - but we don't know if that is actually a problem (if other folds don't set end_inclusive).

I will raise an issue to investigate and potentially correct this anomaly, but for now, as this is intended to patch a release I'm focusing on the minimal change to correct the known usage of end_inclusive - rather than making. amore general change to use riak_index range checkers.

ThomasArts (approved):

PR #1769, opened 2020-07-09
Mas i1765 parallel

martinsumner:
Reduce the risks of tictacaae running out of control and causing crashes.

1 - If exchanges are taking longer than expected, then automatically slow down exchange pokes
2 - Don't take snapshots for rebuilds prior to queueing a rebuild, as if many rebuilds are triggered then the snapshot may be enqueued for a long time causing write amplification, and may also lead to snapshot timeouts.


martinsumner:
https://github.com/basho/riak_test/pull/1350

PR #1770, opened 2020-08-05
Mas i1765 tokenbucket

martinsumner:
There are a number of general changes here to address issues with tictacaae performance in parallel mode:

- Update to kv_index_tictactree process that enabled some back-pressure controls, by not being entirely async.  Also support for queue controls on fold_clocks queries to prevent unbounded numbers of queries to be queued with snapshots (which hold memory whilst enqueued).  Moves the snapshot of the keystore for aae tree rebuild inside the fold function, so that the fold function can be queue without first needing to take a snapshot.

- Alignment of repair functionality between kv_exchange_fsm and tictacaae, and spacing of repairs to reduce the risk of overloading vnodes with rehash requests.

- The use of the AAE token bucket on tictacaae as well as legacy aae (although on tictacaae it triggers the async ping process to throttle, not a direct sync call).

- Additional controls for tuning and logging of AAE, to allow for better debugging of AAE problems.

- Tick stepping/skipping in the riak_kv_vnode to back-off from exchanges when previous exchanges have not completed, and avoid initial rebuilds contending for resource with handoffs when nodes are restarted.

- Alignment with new functionality in the kv_index_tictactree to avoid queueing snapshots for later work (riaking memory leaks and snapshot timeouts).

- Updated leveled with reduced logging to avoid issues of erlang.log rolling before log indexing can capture contents.


martinsumner:
Replaces https://github.com/basho/riak_kv/pull/1769

src/riak_kv_vnode.erl
```diff
@@ -400,27 +412,71 @@ tictac_rebuild(B, K, V) ->
...
             end,
-        Pool = select_queue(be_pool, State),
-        riak_core_vnode:queue_work(Pool, 
-                                    {fold, FoldFun, FinishPlusReturnFun},
-                                    Sender,
-                                    State#state.vnode_pool_pid)
+            ok
+        end,
+    JustReturnFun =
+        fun(ok) ->
```
ThomasArts, line 448: How sure can we be that this JustReturnFun is always called with 'ok' as an argument?
The way I read it is that this is the FinishFun on a foldfun... possibly read it wrong.

But if I read this right and the StoreFun is going to return something else then 'ok', then this finish fun crashes.

ThomasArts, line 448:
```suggestion
        fun(_) ->
```

ThomasArts: aaecontroller_rebuildtress can return 'skipped' as well.

src/riak_kv_vnode.erl
```diff
@@ -1013,41 +1079,47 @@ handle_command({fold_indexes, FoldIndexFun, Acc}, Sender, State=#state{mod=Mod,
     end;
 
 
-handle_command({rebuild_complete, store}, _Sender, State) ->
+handle_command({rebuild_complete, store, ST}, _Sender, State) ->
```
ThomasArts, line 1082: This is an exported funtion, used by riak_core_vnode... I trust that all places that it needs 3 instead of 2 tuple are checked.

src/riak_kv_vnode.erl
```diff
@@ -1130,41 +1205,78 @@ handle_command(tictacaae_exchangepoke, _Sender, State) ->
...
+                                    [Local, Remote, {DocIdx, N}]),
+                        0
             end,
             ok = aae_controller:aae_ping(State#state.aae_controller,
                                             os:timestamp(),
                                             self()),
-            {noreply, State#state{tictac_exchangequeue = Rest}}
+            {noreply, State#state{tictac_exchangequeue = Rest,
+                                    tictac_skiptick = SkipCount}};
+        {_, SkipCount} ->
```
ThomasArts, line 1252: Unlikely, but potentially the Skipcount can get below zero.
This would happen when the exchange state is a non-empty list with an element of the wrong format. So, it would skip the case above. Most likely you then have bigger problems then a negative skipcount.

But you could do your max trick here.

src/riak_kv_vnode.erl
```diff
@@ -1130,41 +1205,78 @@ handle_command(tictacaae_exchangepoke, _Sender, State) ->
...
+            lager:warning("Skipping a tick due to non_zero " ++
+                            "skip_count=~w", [SkipCount]),
+            {noreply, State#state{tictac_skiptick = SkipCount - 1}}
```
ThomasArts:
```suggestion
            {noreply, State#state{tictac_skiptick = max(0, SkipCount - 1)}}
```

ThomasArts (approved):
Little concern about JustReturnFun, otherwise fine

src/riak_kv_vnode.erl
martinsumner: https://github.com/basho/riak_kv/blob/develop-2.9/src/riak_kv_worker.erl#L46 

So JustReturnFun will always be passed the outcome of FoldFun.  But in this case the FoldFun must always return ok:

https://github.com/basho/riak_kv/blob/mas-i1765-tokenbucket/src/riak_kv_vnode.erl#L445

martinsumner: But that still returns ok from the FoldFun:

https://github.com/basho/riak_kv/blob/mas-i1765-tokenbucket/src/riak_kv_vnode.erl#L441-L445

ThomasArts (approved):

PR #1776, opened 2020-11-12
Mas i1775 ttaaesafety

martinsumner:
Bug fixes and improvements to TTAAE full-sync


martinsumner:
https://github.com/basho/riak_test/compare/develop-3.0...mas-i1775-ttaaesafety

priv/riak_kv.schema
```diff
@@ -90,14 +90,14 @@
 
 %% @doc Frequency to prompt exchange per vnode
 %% The number of milliseconds which the vnode must wait between self-pokes to
-%% maybe prompt the next exchange. Default is 4 minutes - check all partitions
-%% when n=3 every 30 minutes.
+%% maybe prompt the next exchange. Default is 8 minutes - check all partitions
+%% when n=3 once every hour (in each direction).
```
ThomasArts: probably add the formula on how to get from 8 and 3 to once every hour. That is not straigthforward

priv/riak_kv.schema
```diff
@@ -852,6 +852,33 @@
   {default, q1_ttaaefs}
 ]}.
 
+%% @doc For tictac full-sync what is the maximum number of AAE segments to be
+%% compared per exchange.  Reducing this will speed up clock compare queries,
+%% but will increase the number of exchanges required to complete a repair.
+%% If using range_check to speed-up repairs, this can be reduced as the
+%% range_check maxresults will be boosted by the ttaaefs_rangeboost  When using
+%% range_check a value of 64 is recommended, reducing to 32 or 16 if the cluster
+%% has very a large volume of keys and limited capacity.  Only reduce below
```
ThomasArts:
```suggestion
%% has a very large number of keys and limited capacity.  Only reduce below
```

priv/riak_kv.schema
```diff
@@ -852,6 +852,33 @@
...
+%% impact on KV performance when repairs are required.
+{mapping, "ttaaefs_maxresults", "riak_kv.ttaaefs_maxresults", [
+  {datatype, integer},
+  {default, 64}
+]}.
+
+%% @doc For tictac full-sync what is the maximum number of AAE segments to be
+%% compared per exchange.  When running a range_check query this will be the
+%% ttaaefs_max results * ttaaefs_rangeboost.
+%% When using range_check, a small maxresults can be used, ineffecting making
```
ThomasArts: Don't understand `infecting making`

src/riak_client.erl
```diff
@@ -878,7 +879,35 @@ ttaaefs_fullsync(WorkItem, SecsTimeout, Now) ->
     wait_for_reqid(ReqId, SecsTimeout * 1000).
 
 
-
+-spec remove_node_from_coverage() -> ok.
+remove_node_from_coverage() ->
```
ThomasArts: If you give an argument to this function ParticipateInCoverage, then the second one can call the first with the default value. Less code, more consistency

src/riak_kv_delete.erl
```diff
@@ -94,6 +97,7 @@ delete(ReqId,Bucket,Key,Options,Timeout,Client,ClientId,VClock) ->
                     ?DTRACE(?C_DELETE_INIT2, [1], [<<"reap">>]),
                     {ok, C2} = riak:local_client(),
                     AsyncTimeout = 60*1000,     % Avoid client-specified value
+                    timer:sleep(TombPause),
```
ThomasArts, line 101: Explain, why is 10 milliseconds a good time to resolve the issues you have?

src/riak_kv_vnode.erl
```diff
@@ -1686,15 +1702,44 @@ handle_aaefold({merge_branch_nval, Nval, BranchIDs},
...
+                                            ReturnFun, 
+                                            fun preflistfun/2),
+            
+            {noreply, State};
+        false ->
+            %% Using fetch_clocks_range will mean that the AF3_QUEUE will be
+            %% used for scheduling the work not the aae_runner.  Also, the 
+            %% fetch clock query will not attempt to rebuild segments of the
+            %% tree cache - this is left to internal anti-entropy
+            handle_aaefold({fetch_clocks_range,
```
ThomasArts, line 1722: kind of too many arguments to be able to really get what's going on here.

ThomasArts: you could equally well have called with fetch_clocks_nvals, since that directly returns in this new call.

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -345,53 +385,146 @@ handle_cast({all_sync, ReqID, From, _Now}, State) ->
...
         bucket ->
             [H|T] = State#state.bucket_list,
-            {MegaSecs, Secs, _MicroSecs} = Now,
-            UpperTime = MegaSecs * 1000000  + Secs,
-            LowerTime = UpperTime - 60 * 60,
             % Note that the tree size is amended as well as the time range.
             % The bigger the time range, the bigger the tree.  Bigger trees
             % are less efficient when there is little change, but can more
-            % accurately reflect bigger changes (with less false positives).
+            % accurately reflect bigger changes (with less fasle positives).
```
ThomasArts:
```suggestion
            % accurately reflect bigger changes (with less false positives).
```

src/riak_kv_vnode.erl
```diff
@@ -1111,14 +1111,23 @@ handle_command({rebuild_complete, trees, _ST}, _Sender, State) ->
...
     XT = State#state.tictac_exchangetime + timer:now_diff(os:timestamp(), ST),
+    case EndState of
+        PositiveState
+            when PositiveState == root_compare;
+                 PositiveState == branch_compare ->
+            ok;
+        UnwelcomeState ->
+            lager:info("Tictac AAE exchange for partition=~w pending_state=~w",
+                        [State#state.idx, UnwelcomeState])
+    end,
```
ThomasArts, line 1130: Not sure you like list comprehensions, but for such a side-effect: do this if Condition, they are really nice.
The above code would become a one-liner:
```
     [ lager:info("Tictac AAE exchange for partition=~w pending_state=~w",
       [State#state.idx, EndState]) || EndState /= root_compare andalso EndState /= branch_compare].
```
Instead of `andalso` you can use `,` or `and`.
But you may also absolutely dislike it :-).

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -52,11 +56,17 @@
 -define(LOOP_TIMEOUT, 15000).
     % Always wait at least 15s after completing an action before
     % prompting another
--define(CRASH_TIMEOUT, 3600 * 1000).
+-define(CRASH_TIMEOUT, 7200 * 1000).
     % Assume that an exchange has crashed if not response received in this
```
ThomasArts:
```suggestion
    % Assume that an exchange has crashed if no response received in this
```

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -345,53 +385,146 @@ handle_cast({all_sync, ReqID, From, _Now}, State) ->
...
+                case State#state.previous_success of
+                    false ->
+                        lager:info("No range sync as no range set"),
+                        none;
+                    {PrevMega, PrevSecs, _PrevMS} ->
+                        {MegaSecs, Secs, _MicroSecs} = Now,
+                        NowSecs = MegaSecs * ?MEGA  + Secs,
+                        {all,
+                            all,
+                            PrevMega * ?MEGA + PrevSecs - 60,
```
ThomasArts: This confuses me. In the hourly and daily, the Now argument is reduced by nr of seconds in day/hour respectively.
But here we subtract seconds from the previous successful sync. Should we not have something like, between max(PreviousSuccess , Now - 60), Now ?

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -345,53 +385,146 @@ handle_cast({all_sync, ReqID, From, _Now}, State) ->
...
+                            Bucket ->
+                                {Bucket, State#state.bucket_list}
+                        end,
+                    Filter =
+                        {filter, B, KeyRange, small, all,
+                        {LowerTime, UpperTime}, pre_hash},
+                    {State0, Timeout} =
+                        sync_clusters(From, ReqID, range, range, Filter,
+                                        NextBucketList, partial, State,
+                                        hour_check),
```
ThomasArts: hour_check or range_check here?

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -617,8 +835,25 @@ remote_sender({fetch_clocks, SegmentIDs}, Client, Mod, ReturnFun, NVal) ->
     fun() ->
         case Mod:aae_fetch_clocks(Client, NVal, SegmentIDs) of
             {ok, {keysclocks, KeysClocks}} ->
-                ReturnFun(lists:map(fun({{B, K}, VC}) -> {B, K, VC} end,
-                            KeysClocks));
+                MapFun =
```
ThomasArts: you use this MapFun at a lot of places. Would it not be better to write a function
`decode_keyclocks` that takes a list and does exactly this.... less code duplication

src/riak_kv_ttaaefs_manager.erl
```diff
@@ -725,13 +962,45 @@ generate_repairfun(ExchangeID, QueueName) ->
...
+                % otherwise query over all buckets on the given modified
+                % range
+                case PBDL of
+                    [{B, KC, LowDT, HighDT}] ->
+                        lager:info("Setting range to bucket=~p ~w ~w " ++
+                                    " last_count=~w set by work_item=~w",
+                                    [B, LowDT, HighDT, KC, WorkType]),
+                        set_range(B, all, LowDT, HighDT);
+                    _ ->
+                        MapFun = fun({_B, _KC, LD, HD}) -> {LD, HD} end,
```
ThomasArts, line 984: if you only want a LoDT and HiDT, then this code can be replaced by a fold.
```
[{_,_, LD1, HD1}|_] ->
   {LoDT, HiDT} = 
      lists:foldl(fun({_B, _KC, LD, HD}, {Min, Max}) ->
                               {min(LD, Min),
                               {max(HD,Max)}
                 end, {LD1,HD1}, PBDL),

Comment (ThomasArts) [reply src/riak_kv_ttaaefs_manager.erl:984]: I this way yo don't need the case either, because if you have one element it's just a special case

File: src/riak_kv_ttaaefs_manager.erl [hunk 14]
```diff
@@ -807,41 +1110,57 @@ take_next_workitem([NextAlloc|T], Wants,
...
     Allocations = [],
-    lists:sort(choose_schedule(Slices,
-                                Allocations,
-                                {NoSync, AllSync, DaySync, HourSync})).
-
-choose_schedule([], Allocations, {0, 0, 0, 0}) ->
+    lists:sort(
+        choose_schedule(Slices,
+                        Allocations,
+                        {NoSync, AllSync, DaySync, HourSync, RangeSync})).
```
ThomasArts: naming may be a bit weird here, since you renamed other to NoCheck, AllCheck, etc.

ThomasArts (approved):

src/riak_kv_ttaaefs_manager.erl
martinsumner: Updated

priv/riak_kv.schema
martinsumner: Corrected

src/riak_kv_delete.erl
martinsumner: The delay is harmless, in the sense the client has already had a response (although it will slow down deletes queued by riak_kv_eraser).

I suspect 1ms will make it go away, there'a network hop involved in the race condition.  However, other environments might be different.  

It is configurable if it proves to be a problem, so will leave at 10ms until more data comes back from deployments.

src/riak_kv_vnode.erl
martinsumner: I'm more concerned that I've leaked some knowledge of an atom in kv_index_tictactree into riak_kv.  I think really I should have a `aae_exchange:positive_outcomes/0` exported, so that riak_kv doesn't need to know about what these mean, and is decoupled from new positive outcomes being added.

I will look at this.

src/riak_kv_ttaaefs_manager.erl
martinsumner: The 60 is a bit of a fuzz factor, that part-fixes a problem that the last success was prompted potentially a bit before this timestamp.  I'll either get the correct timestamp passed through in some way, or make it replace 60 with ?FUZZ with a comment on the definition of FUZZ to explain

martinsumner (commented):
There is at least a switch to disable this {riak_kv, tombstone_timestamp} can be set to false - so anyone having an issue can simply flick that switch.

priv/riak_kv.schema
martinsumner: Added - it is 1 + (n - 1) * n ticks per cycle.  there are n - 1 exchanges for each partition, and n partitions ... plus 1 tick to prompt the refill of the cycle

src/riak_kv_ttaaefs_manager.erl
martinsumner: Changed to *Check

src/riak_kv_ttaaefs_manager.erl
martinsumner: Actually, no, I was wrong.  The last success time is the time the last successful query was prompted.  So I think the fuzz factor is unnecessary

src/riak_kv_ttaaefs_manager.erl
martinsumner: should be range_check

src/riak_kv_ttaaefs_manager.erl
martinsumner: I'm OK with how it reads at the moment.  I'm minded to leave as-is for now.

src/riak_kv_vnode.erl
martinsumner: Having fiddled, I'm only making matters worse -swapping one bit of special knowledge for another.  will leave as-is

Bob-The-Marauder:
Not sure if this comment is entirely relevant but including it just in case.

As part of system tuning, it's recommended to set the disks to `noatime` mounts https://www.tiot.jp/riak-docs/riak/kv/2.2.6/using/performance/#mounts

I know the question in https://postriak.slack.com/archives/C6R0LPH4N/p1607107910139900 uses "last_modified_date" and that `noatime` is only relevent for "last accessed" but thought I should mention it in case there is any potential for people to  tune Riak and, in doing so, inadvertently break functions.

martinsumner:
@Bob-The-Marauder - last modified time is the time on an individual object (which comes from the OS system time), and not related to times on files etc.  So this won't be an issue.

src/riak_kv_vnode.erl
martinsumner: This way just helps with concurrently managing the update to 2.9 and 3.0, as 2.9 won't initially have the new fetch_clocks_nval option

PR #1781, opened 2021-01-13
Assume no props if no props passed

martinsumner:
The parsing of command lines doesn't pass the emtpy string in now as the second parameter when creating bucket types - so this is done explicitly now.

Also uplift to leveled to fix perf issue.


Bob-The-Marauder (approved):
Looks good to me. I especially like the simplicity of the riak_kv_console.erl fix

PR #1783, opened 2021-02-03
Updated PR for SecondaryIndexQueries

ThomasArts:


martinsumner (commented):
I'm not sure it works to split out the process of breaking up a fixed width index by first adding a delimiter - as the index being split may contain the delimiter.

Perhaps supporting fixed width is an unnecessary complication.  Perhaps delimiters should always be used, and all binary parts on the index should be expected to be base64 encoded.  The indexes are compressed before storage, so being able to fine tune there size is perhaps a false saving.

PR #1785, opened 2021-03-19
Mas ghafix

martinsumner:
It appears the there is an issue with support for `chmod` within our Github Actions setup - in that it seems to return status of 0 to any chmod request, but doesn't change permissions as expected.

To progress, tests can now be marked with `ifndef(GITHUBEXCLUDE).` and they will not be run on gtihub actions, but will continue to be run within local `rebar3 eunit` test runs.

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -14,6 +14,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - shell: bash
     - name: Compile
       run: ./rebar3 compile
     - name: Run tests
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -18,6 +18,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Compile
-      run: rebar3 compile
+      run: ./rebar3 compile
     - name: Run tests
-      run: rebar3 do xref, dialyzer, eunit
+      run: ./rebar3 do xref, dialyzer, eunit
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -1,7 +1,11 @@
 name: Erlang CI
 
 on:
-  pull_request
+  push:
+    branches: [ develop-3.0 ]
+  pull_request:
+    branches: [ develop-3.0 ]
+
 
 jobs:
 
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -1,7 +1,7 @@
 name: Erlang CI
 
 on:
-  push
+  pull_request
 
 jobs:
 
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Compile
         run: ./rebar3 compile
-      - name: Run tests
-        shell: bash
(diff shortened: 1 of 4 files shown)


rebar.config
```diff
@@ -43,7 +43,10 @@
 {profiles, [
     {test, [{deps, [meck,
                     {eunit_formatters, ".*", {git, "git://github.com/seancribbs/eunit_formatters", {tag, "v0.5.0"}}}]}]},
-    {eqc, [{deps, [meck]}]}
+    {eqc, [{deps, [meck]}]},
+    {gha, [{deps, [meck,
```
ThomasArts: I think you should not need the meck stuff. Unit will by default run the test profile and hence will use all in test as well as gha profile in above definition of "Run unit".

ThomasArts:
```suggestion
    {gha, [{deps, [
```

rebar.config
```diff
@@ -43,7 +43,10 @@
...
+                    {eunit_formatters, ".*", {git, "git://github.com/seancribbs/eunit_formatters", {tag, "v0.5.0"}}}]},
```
ThomasArts:
```suggestion
                    
```

.github/workflows/erlang.yml
```diff
@@ -1,10 +1,7 @@
 name: Erlang CI
 
 on:
-  push:
-    branches: [ develop-3.0 ]
-  pull_request:
-    branches: [ develop-3.0 ]
+  pull_request
```
ThomasArts: As comment: I guess that this works, since only branches that have this .yml file in it do act on the pull requests and branches that are based on old rebar version simply don't have this file.

.github/workflows/erlang.yml
```diff
@@ -16,8 +13,10 @@ jobs:
...
-    - uses: actions/checkout@v2
-    - name: Compile
-      run: rebar3 compile
-    - name: Run tests
-      run: rebar3 do xref, dialyzer, eunit
+      - uses: actions/checkout@v2
+      - name: Compile
+        run: ./rebar3 compile
+      - name: Run xref and dialyzer
+        run: rebar3 do xref, dialyzer
```
ThomasArts:
```suggestion
        run: ./rebar3 do xref, dialyzer
```

.github/workflows/erlang.yml
```diff
@@ -16,8 +13,10 @@ jobs:
...
+      - name: Run eunit
+        run: rebar3 as gha do eunit
```
ThomasArts:
```suggestion
        run: ./rebar3 as gha do eunit
```

I assume you want to run the rebar version supplied not to get surprised by incompatibilities in rebar?

ThomasArts (approved):

ThomasArts:
I advice to squash the commits before merging

rebar.config
martinsumner: Removed deps from the gha profile

.github/workflows/erlang.yml
martinsumner: I'm not sure what the right configuration is, but I've reverted to make consistent with the other repos for now.

.github/workflows/erlang.yml
martinsumner: Done

.github/workflows/erlang.yml
martinsumner: Done

PR #1077, opened 2015-01-29
add wm module for GETing (and encoding) active preflist

zeeshanlakhani:
**Note**: **Bors will fail** until https://github.com/basho/riak_core/pull/705 and co is reviewed/merged.

Update for webmachine.
### Description

Part of [RIAK-1481](https://bashoeng.atlassian.net/browse/RIAK-1481).

The active preflist will return primary/fallback partitions and nodes for the available nodes at the time of query. Primary/fallback will be annotated.
This involves updates to our RiakKV WB code, RiakPB, RiakCore, RiakAPI, Riak-Erlang-Http-Client, Riak-Erlang-Client.

*The impetus for adding this to the API started with a mailing list question answered by Charlie Voiselle, http://lists.basho.com/pipermail/riak-users_lists.basho.com/2015-January/016527.html, which involved a snippet of code that we've given clients multiple times. This was then discussed with Russell Brown on HipChat, https://basho.hipchat.com/history/room/867200/2015/01/13?q=enterprising&t=rid-867200#12:23:22.

Other PRs in the series:
- https://github.com/basho/riak_pb/pull/105
- https://github.com/basho/riak_api/pull/73
- https://github.com/basho/riak_core/pull/697
- https://github.com/basho/riak-erlang-http-client/pull/50
- https://github.com/basho/riak-erlang-client/pull/204

PR #1790, opened 2021-06-21
try code:load_file if loading with code:ensure_loaded fails

hmmr:
`code:ensure_loaded` fails if our code server is in embedded mode, causing riak_cs to fail to configure riak with a custom backend storage:

``` {riak_kv,
     [{add_paths,
          ["/home/hmmr/X/titokyo/g/riak_cs/dev/dev1/riak-cs/lib/riak_cs-3.0.0pre3+build.3435.ref2aaece2f/ebin"]},
      {storage_backend,riak_cs_kv_multi_backend},
      ...
```

Loading the module with `code:load_file/1` actually makes it happen.

### DIFF
--- a/src/riak_kv_app.erl
+++ b/src/riak_kv_app.erl
@@ -126,11 +126,20 @@ start(_Type, _StartArgs) ->
     %% Check the storage backend
     StorageBackend = app_helper:get_env(riak_kv, storage_backend),
     case code:ensure_loaded(StorageBackend) of
-        {error,nofile} ->
-            lager:critical("storage_backend ~p is non-loadable.",
-                           [StorageBackend]),
+        {error, embedded} ->
+            case code:load_file(StorageBackend) of
+                {error, LoadReason} ->
+                    lager:critical("storage_backend ~p is non-loadable: ~p.",
+                                   [StorageBackend, LoadReason]),
+                    throw({error, invalid_storage_backend});
+                {module, _} ->
+                    ok
+            end;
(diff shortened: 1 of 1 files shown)


martinsumner (approved):

martinsumner:
Another Riak user raised the issue of code in additional paths not working in riak_kv >= 3.0 (e.g. map/reduce functions), and returning {error, embedded} instead.

As part of the switch to OTP 20 with 3.0, the switch was made to relx - and this changes the default mode from interactive to embedded:

https://github.com/erlware/relx/blob/master/priv/templates/extended_bin#L923-L924

So it appears the root cause here is we're now running embedded.  But this wasn't a deliberate change.  So perhaps the best answer, to solve this and other dynamic code loading issues, is to switch back to interactive.

For packaged deployments - I think we could do this by setting the CODE_LOADING_MODE in the riak script https://github.com/basho/riak/blob/develop-3.0/rel/files/riak.  This wouldn't work though when running direct from `make rel` and `make devrel`.

Are there any objections to switching back to interactive mode? @martincox

martinsumner:
See https://github.com/basho/riak/pull/1077

This changes the default back to interactive by using a pre_start script (as with changing the max log size).  This seems better than changing rel/fiiles/riak.

hmmr:
As far as #1077 fixes the loading regression, I'm totally fine with the proposed change to interactive mode.

hmmr:
Closing as basho/riak#1077 has been merged.

PR #1792, opened 2021-07-14
Mas i1788 reconcileapi

martinsumner:
Complete fetch API for PB as well as http.  Add Push API, to allow for discovered differences to be re-queued.

The intention is that this can be used to assist when reconciling a Riak store against an external data store.  Some testing outside of riak_test has been done with dynamodb.  The reconciliation is originally intended to be one-way (e.g. assuming the  non-Riak data store is a read-only copy).

The fetch API is extended to support a different encoding of the fetched object, `internal_aaehash`.  This encoding provides the SegmentId and SegmentHash for this version of the object, to allow that to be indexed within the external db (e.g. to provide merkle trees for comparison).

Testing - https://github.com/basho/riak_test/pull/1356

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -60,5 +60,5 @@
     {hyper, {git, "git://github.com/basho/hyper", {tag, "1.1.0"}}},
     {leveled, {git, "https://github.com/martinsumner/leveled.git", {tag, "1.0.4"}}},
     {kv_index_tictactree, {git, "https://github.com/martinsumner/kv_index_tictactree.git", {tag, "0.9.18"}}},
-    {riakhttpc, {git, "git://github.com/basho/riak-erlang-http-client", {tag, "3.0.4"}}}
+    {riakhttpc, {git, "git://github.com/basho/riak-erlang-http-client", {branch, "mas-i1788-reconcileapi"}}}
        ]}.
--- a/rebar.config
+++ b/rebar.config
@@ -56,7 +56,7 @@
     {sext, {git, "git://github.com/uwiger/sext.git", {tag, "1.4.1"}}},
     {riak_pipe, {git, "git://github.com/basho/riak_pipe.git", {tag, "riak_kv-3.0.5"}}},
     {riak_dt, {git, "git://github.com/basho/riak_dt.git", {tag, "riak_kv-3.0.0"}}},
-    {riak_api, {git, "git://github.com/basho/riak_api.git", {tag, "riak_kv-3.0.5"}}},
+    {riak_api, {git, "git://github.com/basho/riak_api.git", {branch, "mas-i1788-reconcileapi"}}},
     {hyper, {git, "git://github.com/basho/hyper", {tag, "1.1.0"}}},
     {leveled, {git, "https://github.com/martinsumner/leveled.git", {tag, "1.0.4"}}},
     {kv_index_tictactree, {git, "https://github.com/martinsumner/kv_index_tictactree.git", {tag, "0.9.18"}}},
--- a/rebar.config
+++ b/rebar.config
@@ -56,9 +56,9 @@
     {sext, {git, "git://github.com/uwiger/sext.git", {tag, "1.4.1"}}},
     {riak_pipe, {git, "git://github.com/basho/riak_pipe.git", {tag, "riak_kv-3.0.5"}}},
     {riak_dt, {git, "git://github.com/basho/riak_dt.git", {tag, "riak_kv-3.0.0"}}},
-    {riak_api, {git, "git://github.com/basho/riak_api.git", {branch, "mas-i1788-reconcileapi"}}},
+    {riak_api, {git, "git://github.com/basho/riak_api.git", {branch, "develop-3.0"}}},
     {hyper, {git, "git://github.com/basho/hyper", {tag, "1.1.0"}}},
     {leveled, {git, "https://github.com/martinsumner/leveled.git", {tag, "1.0.4"}}},
     {kv_index_tictactree, {git, "https://github.com/martinsumner/kv_index_tictactree.git", {tag, "0.9.18"}}},
-    {riakhttpc, {git, "git://github.com/basho/riak-erlang-http-client", {branch, "mas-i1788-reconcileapi"}}}
+    {riakhttpc, {git, "git://github.com/basho/riak-erlang-http-client", {branch, "develop-3.0"}}}
        ]}.
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -445,12 +445,14 @@ unpack_keyclock_fun(RpbKeysClock) ->
         undefined ->
             {RpbKeysClock#rpbkeysvalue.bucket,
                 RpbKeysClock#rpbkeysvalue.key,
-                riak_object:decode_vclock(RpbKeysClock#rpbkeysvalue.value)};
+                riak_object:decode_vclock(RpbKeysClock#rpbkeysvalue.value),
+                to_fetch};
         T ->
             {{T,
                 RpbKeysClock#rpbkeysvalue.bucket},
                 RpbKeysClock#rpbkeysvalue.key,
-                riak_object:decode_vclock(RpbKeysClock#rpbkeysvalue.value)}
+                riak_object:decode_vclock(RpbKeysClock#rpbkeysvalue.value),
+                to_fetch}
     end.
 
 -spec make_binarykey(riak_object:bucket(), riak_object:key()) -> binary().
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -439,7 +439,8 @@ make_binarykey(RObj) ->
     make_binarykey(riak_object:bucket(RObj), riak_object:key(RObj)).
 
 -spec unpack_keyclock_fun(#rpbkeysvalue{}) ->
-        {riak_object:bucket(RObj), riak_object:key(RObj), vclock:vclock()}.
+        {riak_object:bucket(RObj), riak_object:key(RObj), vclock:vclock(),
+            to_fetch}.
 unpack_keyclock_fun(RpbKeysClock) ->
     case RpbKeysClock#rpbkeysvalue.type of
         undefined ->
--- a/src/riak_kv_pb_object.erl
+++ b/src/riak_kv_pb_object.erl
@@ -29,6 +29,8 @@
 %%  9 - RpbGetReq
 %% 11 - RpbPutReq
 %% 13 - RpbDelReq
+%% 202 - RpbFetchReq
+%% 204 - RpbPushReq
 %% </pre>
 %%
 %% <p>This service produces the following responses:</p>
@@ -39,6 +41,8 @@
 %% 10 - RpbGetResp
 %% 12 - RpbPutResp - 0 length
 %% 14 - RpbDelResp
+%% 203 - RpbFetchResp
+%% 205 - RpbPushResp
 %% </pre>
 %%
 %% <p>The semantics are unchanged from their original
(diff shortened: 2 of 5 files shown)


rebar.config
```diff
@@ -56,9 +56,9 @@
     {sext, {git, "git://github.com/uwiger/sext.git", {tag, "1.4.1"}}},
     {riak_pipe, {git, "git://github.com/basho/riak_pipe.git", {tag, "riak_kv-3.0.5"}}},
     {riak_dt, {git, "git://github.com/basho/riak_dt.git", {tag, "riak_kv-3.0.0"}}},
-    {riak_api, {git, "git://github.com/basho/riak_api.git", {tag, "riak_kv-3.0.5"}}},
+    {riak_api, {git, "git://github.com/basho/riak_api.git", {branch, "mas-i1788-reconcileapi"}}},
```
ThomasArts: A little to-do after update of sub projects to point to new tag

src/riak_kv_pb_object.erl
```diff
@@ -379,6 +434,38 @@ process_stream(_,_,State) ->
...
 
+-spec make_binarykey(riak_object:riak_object()) -> binary().
+make_binarykey(RObj) ->
+    make_binarykey(riak_object:bucket(RObj), riak_object:key(RObj)).
+
+-spec unpack_keyclock_fun(#rpbkeysvalue{}) ->
+        {riak_object:bucket(RObj), riak_object:key(RObj), vclock:vclock(),
+            to_fetch}.
+unpack_keyclock_fun(RpbKeysClock) ->
+    case RpbKeysClock#rpbkeysvalue.type of
```
ThomasArts: This is code duplication. I wonder if it would not be easier to write 
```suggestion
    Bucket = case RpbKeysClock#rpbkeysvalue.type of
                             undefined -> RpbKeysClock#rpbkeysvalue.bucket;
                             T -> {T, RpbKeysClock#rpbkeysvalue.bucket}
                      end,
       {Bucket, RpbKeysClock#rpbkeysvalue.key,
                 riak_object:decode_vclock(RpbKeysClock#rpbkeysvalue.value),
                 to_fetch}
```

src/riak_kv_replrtq_snk.erl
```diff
@@ -530,18 +529,7 @@ map_peer_to_wi_fun({QueueName, Iteration, PeerInfo}) ->
                                 {certfile, CertfiicateFilename},
```
ThomasArts: This variable name has a typo, but nothing you did add.

src/riak_kv_wm_queue.erl
```diff
@@ -25,10 +25,16 @@
 %% ```
 %% GET /queuename/QueueName
 %%
-%% Will return an object id an object present on queue
+%% Will return an object if an object is present on the queue
```
ThomasArts:
```suggestion
%% Will return an object if an object is present in the queue
```

rebar.config
martinsumner: Yes, will pick this up as part of the release cycle

src/riak_kv_pb_object.erl
martinsumner: Agreed, will change

src/riak_kv_replrtq_snk.erl
martinsumner: Will still change though

src/riak_kv_wm_queue.erl
```diff
@@ -63,12 +79,12 @@ init(Props) ->
...
     {boolean(), #wm_reqdata{}, context()}.
 %% @doc Determine whether or not a connection to Riak
-%%      can be established. Also, extract query params.
+%%      can be established. 
 service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
     ClientID = riak_kv_wm_utils:get_client_id(RD),
     case riak_kv_wm_utils:get_riak_client(RiakProps, ClientID) of
         {ok, C} ->
-            {true, RD, Ctx#ctx{client = C}};
+            {true, RD, Ctx#ctx{client = C, method=wrq:method(RD)}};
```
ThomasArts, line 87: This surprises me... should you really update the context with the method type?

Is that because this is used in malformed_request? If so, I would rather get the method from RD in malformed_request instead of counting on this effect to take place.

src/riak_kv_wm_queue.erl
```diff
@@ -77,25 +93,100 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
...
-malformed_request(RD, Ctx) ->
-    QueueNameRaw = wrq:path_info(queuename, RD),
-    QueueName =
-        list_to_atom(riak_kv_wm_utils:maybe_decode_uri(RD, QueueNameRaw)),
+malformed_request(RD, Ctx) when Ctx#ctx.method =:= 'GET' ->
+    QueueName = malformed_queuename(RD),
     ObjectFormatRaw = wrq:get_qs_value(?Q_OBJECT_FORMAT, "internal", RD),
     ObjectFormat = list_to_atom(ObjectFormatRaw),
-    {false, RD, Ctx#ctx{queuename = QueueName, object_format = ObjectFormat}}.
+    {false, RD, Ctx#ctx{queuename = QueueName, object_format = ObjectFormat}};
```
ThomasArts: This is actually not "checking" for malformed ness. If one of the commands crashes, you may get a malformed message... not even sure, but basically you always get a `false` returned, meaning is is well formed.

I would expect that if the QueuName decoding fails or `get_qs_value` does not find the query, that you then return `true`... . Not that `wrq:get_qs_value` may return undefined, in which case the list_to_atom crashes. If you really want an error message  would just crash if the queue is not there. This would cause a 500 internal error instead of a 400 bad request.

src/riak_kv_wm_queue.erl
```diff
@@ -77,25 +93,100 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
...
+malformed_request(RD, Ctx) when Ctx#ctx.method =:= 'POST' ->
+    QueueName = malformed_queuename(RD),
+    KeyClockList =
```
ThomasArts: Same here... this kind of crashes or always succeeds

src/riak_kv_wm_queue.erl
```diff
@@ -77,25 +93,100 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
...
+-spec malformed_keyclocklist(iolist()) -> list(riak_kv_replrtq_src:repl_entry()).
+malformed_keyclocklist(ReqBody) ->
+    {struct, [{<<"keys-clocks">>, KCL}]} = mochijson2:decode(ReqBody),
+    lists:reverse(lists:foldl(fun decode_bucketkeyclock/2, [], KCL)).
+
+decode_bucketkeyclock({struct, BKC}, Acc) ->
+    B = 
+        case lists:keyfind(<<"bucket">>, 1, BKC) of
+            {<<"bucket">>, Bucket} when is_binary(Bucket) ->
+                case lists:keyfind(<<"bucket-type">>, 1, BKC) of
```
ThomasArts: Where is this encoded? How do I know for sure that the encoder uses `bucket-type` as key and not `bucketType` or so?

What I try to say.... it is nice if encoder and decoder are in the same module and if the use of decode here uses the module in which the encoder is defined

src/riak_kv_wm_queue.erl
```diff
@@ -77,25 +93,100 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
...
+                    {<<"bucket-type">>, BucketType} ->
+                        {BucketType, Bucket};
+                    false ->
+                        Bucket
+                end;
+            false ->
```
ThomasArts: So you allow JSON elements in this JSON Array that do not have "bucket" as key. That's a slippery slope I guess. This is where malformed would check that they are all well formed.

src/riak_kv_wm_queue.erl
```diff
@@ -77,25 +93,100 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
...
+    ok = riak_kv_replrtq_src:replrtq_ttaaefs(QueueName, KeyClockList),
+    R = 
+        case riak_kv_replrtq_src:length_rtq(QueueName) of
+            {QueueName, {FL, FSL, RTL}} ->
+                io_lib:format("Queue ~w: ~w ~w ~w", [QueueName, FL, FSL, RTL]);
+            _ ->
+                io_lib:format("No queue ~w", [QueueName])
+        end,
+    {true,
+        wrq:set_resp_body(
```
ThomasArts, line 204: This is not very REST. One would expect to get the response as `application/json` instead of text. That means, instead of creating the io:format above, build a JSON structure.

src/riak_kv_wm_queue.erl
```diff
@@ -107,9 +198,36 @@ produce_queue_fetch(RD, Ctx) ->
...
+format_response(internal_aaehash, {ok, {deleted, TombClock, RObj}}, RD, Ctx) ->
+    BK = make_binarykey(riak_object:bucket(RObj), riak_object:key(RObj)),
+    {SegmentID, SegmentHash} =
+        leveled_tictac:tictac_hash(BK, lists:sort(TombClock)),
+    SuccessMark = <<1:8/integer>>,
+    IsTombstone = <<1:8/integer>>,
+    ObjBin = encode_riakobject(RObj),
+    TombClockBin = term_to_binary(TombClock),
+    TCL = byte_size(TombClockBin),
+    {<<SuccessMark/binary, IsTombstone/binary,
```
ThomasArts, line 289: Should this really be part of web machine queue handler? So much detail in building a binary that I can hardly believe this is the right place. Rather have it somewhere in are or so...

But also, would you want this as a binary or rather as a JSON object if you're anyway doing it for having 3rd party apps handle the output.

ThomasArts (commented):
Looks ok on first sight. Wonder though about the encoding/decoding towards 3rd party applications. I guess the idea is to outward produce JSON instead of binaries in which encoding is hidden... but for efficiency that is an option, of course.

src/riak_kv_wm_queue.erl
```diff
@@ -77,25 +93,100 @@ service_available(RD, Ctx=#ctx{riak=RiakProps}) ->
...
+malformed_queuename(RD) ->
+    QueueNameRaw = wrq:path_info(queuename, RD),
+    list_to_atom(riak_kv_wm_utils:maybe_decode_uri(RD, QueueNameRaw)).
+
+-spec malformed_keyclocklist(iolist()) -> list(riak_kv_replrtq_src:repl_entry()).
+malformed_keyclocklist(ReqBody) ->
+    {struct, [{<<"keys-clocks">>, KCL}]} = mochijson2:decode(ReqBody),
+    lists:reverse(lists:foldl(fun decode_bucketkeyclock/2, [], KCL)).
+
+decode_bucketkeyclock({struct, BKC}, Acc) ->
```
ThomasArts: Suggestions, just map over the list and raise an error if <<"key">>, <<"clock">> or other expected fields are missing.

src/riak_kv_wm_queue.erl
martinsumner: This may have come about as a copy & paste from https://github.com/basho/riak_kv/blob/develop-3.0/src/riak_kv_wm_object.erl#L222.

I think there is a general problem that I've not been clear about the correct way of working with webmachine, and so this is rather cobbled from other examples from within the riak code.

The same is true with the misuse of malformed request, this is a pattern common elsewhere in riak.

I'm going to look through this agin, and see if it can be better organised.

src/riak_kv_wm_queue.erl
martinsumner: To  certain extent the use of a binary is locked in here, as this is how it is done for nextgenrepl.  I think the right thing to do for the future is to properly support an alternative to the `internal` and `internal_aaehash` encoding options which is to have a `external_json` alternative.

src/riak_kv_wm_queue.erl
martinsumner: Some general work done now to try and return errors in expected error cases

src/riak_kv_wm_queue.erl
martinsumner: This is a won't fix for now.  It is a common problem with the http api vs the pb api in riak - with the pb api there is the riak_pb library shared between clients and server, but no such thing for http.

A generic problem, rather than one we can fix entirely here.

src/riak_kv_wm_queue.erl
martinsumner: Acknowledged, but not fixed.

src/riak_kv_wm_queue.erl
martinsumner: Now errors

src/riak_kv_wm_queue.erl
martinsumner: Fixed

src/riak_kv_wm_queue.erl
martinsumner: Now any incorrect element in the list will lead to the request being considered malformed

src/riak_kv_wm_queue.erl
martinsumner: This has now changed so that we check for nay elements being filtered out, due to incorrect encoding, an the whole request will be malformed unless all elements are correct

PR #1794, opened 2021-09-14
Mas i350 selectivesync

martinsumner:
Add selective sync to Riak KV.  Prior to this change either all vnodes sync'd a PUT (i.e. flushed to disk as part of accepting the PUT) or no vnodes.  With no sync the flush decision is controlled by either the database (bitcask, eleveldb) or the operating system (leveled).

This changes allows for three options:

backend - rely on backend configuration as now
all - force a sync on all vnodes regardless of backend configuration
one - force a sync only on the coordinating vnode

The sync option can be set on a per-bucket basis, and over-written on a per-write basis.  This provides much more flexibility.

Previously customer who wanted to make sure data was flushed to disk in a least one place for some buckets, had to set backend sync settings to sync, and sync to all vnodes on every write on every bucket.  Throughput between sync and non-sync configurations have been discovered to be large even with hardware acceleration (e.g. flash-backed write caches) - see https://github.com/martinsumner/leveled/issues/350

### DIFF
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -298,6 +298,13 @@ validate([{QProp, MaybeQ}=Prop | T], ValidProps, Errors) when QProp =:= dw
         false ->
             validate(T, ValidProps, [{QProp, not_valid_quorum} | Errors])
     end;
+validate([{SyncProp, MaybeSync}=Prop | T], ValidProps, Errors) when SyncProp =:= sync_on_write ->
+    case is_valid_sync_param(MaybeSync) of
+        true ->
+            validate(T, [Prop | ValidProps], Errors);
+        false ->
+            validate(T, ValidProps, [{SyncProp, not_valid_sync_param} | Errors])
+    end;
 validate([Prop|T], ValidProps, Errors) ->
     validate(T, [Prop|ValidProps], Errors).
 
@@ -314,6 +321,17 @@ is_quorum(Q)  when Q =:= quorum
     true;
 is_quorum(_) ->
     false.
+
+-spec is_valid_sync_param(term()) -> boolean().
+is_valid_sync_param(SP) when SP =:= one
+                        orelse SP =:= all
+                        orelse SP =:= backend
+                        orelse SP =:= <<"one">>
+                        orelse SP =:= <<"all">>
+                        orelse SP =:= <<"backend">> ->
+   true;
+is_valid_sync_param(_) ->
+   false.
 
 %% @private some quorum options can be zero
 -spec is_opt_quorum(term()) -> boolean().
--- a/src/riak_kv_bucket.erl
+++ b/src/riak_kv_bucket.erl
@@ -298,12 +298,12 @@ validate([{QProp, MaybeQ}=Prop | T], ValidProps, Errors) when QProp =:= dw
         false ->
             validate(T, ValidProps, [{QProp, not_valid_quorum} | Errors])
     end;
-validate([{SyncProp, MaybeSync}=Prop | T], ValidProps, Errors) when SyncProp =:= sync_on_write ->
+validate([{sync_on_write, MaybeSync}=Prop | T], ValidProps, Errors) ->
     case is_valid_sync_param(MaybeSync) of
         true ->
             validate(T, [Prop | ValidProps], Errors);
         false ->
-            validate(T, ValidProps, [{SyncProp, not_valid_sync_param} | Errors])
+            validate(T, ValidProps, [{sync_on_write, not_valid_sync_param} | Errors])
     end;
 validate([Prop|T], ValidProps, Errors) ->
     validate(T, [Prop|ValidProps], Errors).
@@ -322,6 +322,10 @@ is_quorum(Q)  when Q =:= quorum
 is_quorum(_) ->
     false.
 
+%% validation of sync parameters
+%% one = sync coordinating node only
+%% all = sync all nodes
+%% backend = take sync value for all nodes from backend config (don't override)
 -spec is_valid_sync_param(term()) -> boolean().
 is_valid_sync_param(SP) when SP =:= one
                         orelse SP =:= all
--- a/src/riak_kv_eleveldb_backend.erl
+++ b/src/riak_kv_eleveldb_backend.erl
@@ -30,6 +30,7 @@
          start/2,
          stop/1,
          get/3,
+         flush_put/5,
          put/5,
          async_put/5,
          delete/4,
@@ -61,7 +62,7 @@
 
 -define(API_VERSION, 1).
 -define(CAPABILITIES, [async_fold, indexes, index_reformat, size,
-        iterator_refresh]).
+        iterator_refresh, flush_put]).
 -define(FIXED_INDEXES_KEY, fixed_indexes).
 
 -record(state, {ref :: eleveldb:db_ref(),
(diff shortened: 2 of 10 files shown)


martinsumner:
This PR adds selective sync support only with leveled and eleveldb backends (there is a backend_capability called flush_put and only backends with this capability receive non-standards PUT requests).  

Adding support for the bitcask backend is trivial, as the bitcask:sync/1 just needs to be called after a put when flush_put/4 is used.  This would have no change to any existing bitcask user, as the default bucket configuration is to revert to `backend` as now.

@martincox - would you be OK if bitcask support was added to this PR?  Normally I try and avoid bitcask changes, to avoid any risk for you guys.  The only change required will be in riak_kv_bitcask_backend (not in the actual bitcask repo).  Although you may not use the feature, it might be easier to implement across all backends rather than manage the ongoing caveat that it only works in some backends

martinsumner:
https://github.com/basho/riak/discussions/1080

martinsumner:
https://github.com/basho/riak_test/pull/1358

src/riak_kv_bucket.erl
```diff
@@ -299,6 +299,13 @@ validate([{QProp, MaybeQ}=Prop | T], ValidProps, Errors) when QProp =:= dw
         false ->
             validate(T, ValidProps, [{QProp, not_valid_quorum} | Errors])
     end;
+validate([{sync_on_write, MaybeSync}=Prop | T], ValidProps, Errors) ->
+    case is_valid_sync_param(MaybeSync) of
+        true ->
+            validate(T, [Prop | ValidProps], Errors);
+        false ->
+            validate(T, ValidProps, [{sync_on_write, not_valid_sync_param} | Errors])
```
ThomasArts, line 307: This is ok, but I see an inconsistency in error tuples in the valid function. The options `write_once` and `consistent` add a string to the error message, whereas the later ones add atoms.

This is type correct, just inconsistent.

src/riak_kv_eleveldb_backend.erl
```diff
@@ -173,19 +175,44 @@ get(Bucket, Key, #state{read_opts=ReadOpts,
...
                                                 legacy_indexes=WriteLegacy,
                                                 fixed_indexes=FixedIndexes}=State) ->
     %% Create the KV update...
     StorageKey = to_object_key(Bucket, PrimaryKey),
     Updates1 = [{put, StorageKey, Val} || Val /= undefined],
 
+    %% Setup write options...
+    WriteOpts2 = case Sync of
+        true ->
+            lists:keyreplace(sync,1,WriteOpts, {sync,Sync});
```
ThomasArts, line 209: Observe that keyreplace does not add sync in this parameter is missing in the WriteOpts. Is that really what you want?

This is not equal to `keydelete(sync, 1, WriteOpts) ++ [{sync, Sync}]`, which intuitively one probably wants.

src/riak_kv_eleveldb_backend.erl
```diff
@@ -173,19 +175,44 @@ get(Bucket, Key, #state{read_opts=ReadOpts,
...
+    Sync = false,
+    do_put(Bucket, PrimaryKey, IndexSpecs, Val, Sync, State).
+
+%% @doc put_flush capability - do a put with a flush to disk
+-spec flush_put(riak_object:bucket(), riak_object:key(), [index_spec()], binary(), state()) ->
+                 {ok, state()} |
+                 {error, term(), state()}.
+flush_put(Bucket, PrimaryKey, IndexSpecs, Val, State) ->
+    Sync = true,
+    do_put(Bucket, PrimaryKey, IndexSpecs, Val, Sync, State).
```
ThomasArts:
```suggestion
    do_put(Bucket, PrimaryKey, IndexSpecs, Val, true, State).
```

src/riak_kv_eleveldb_backend.erl
```diff
@@ -173,19 +175,44 @@ get(Bucket, Key, #state{read_opts=ReadOpts,
...
+put(Bucket, PrimaryKey, IndexSpecs, Val, State) ->
+    Sync = false,
+    do_put(Bucket, PrimaryKey, IndexSpecs, Val, Sync, State).
+
+%% @doc put_flush capability - do a put with a flush to disk
+-spec flush_put(riak_object:bucket(), riak_object:key(), [index_spec()], binary(), state()) ->
+                 {ok, state()} |
+                 {error, term(), state()}.
+flush_put(Bucket, PrimaryKey, IndexSpecs, Val, State) ->
+    Sync = true,
```
ThomasArts: check comment in `leveled_backend`, probably just write the boolean as argument

src/riak_kv_leveled_backend.erl
```diff
@@ -232,26 +234,27 @@ head(Bucket, Key, #state{bookie=Bookie}=State) ->
...
 
+-spec flush_put(riak_object:bucket(),
+                    riak_object:key(),
+                    [index_spec()],
+                    binary(),
+                    state()) ->
+                         {ok, state()} |
+                         {error, term(), state()}.
+flush_put(Bucket, Key, IndexSpecs, Val, State) ->
+    do_put(Bucket, Key, IndexSpecs, Val, true, State).
```
ThomasArts, line 245: compare this with code of `eleveldb_backend`, where you explicitly say `Sync = true` and then use Sync as argument.

I propose changing `eleveldb`

src/riak_kv_leveled_backend.erl
```diff
@@ -632,6 +635,32 @@ callback(Ref, UnexpectedCallback, State) ->
...
+                    binary(),
+                    boolean(),
+                    state()) ->
+                         {ok, state()} |
+                         {error, term(), state()}.
+do_put(Bucket, Key, IndexSpecs, Val, Sync, #state{bookie=Bookie}=State) ->
+    case leveled_bookie:book_put(Bookie,
+                                    Bucket, Key, Val, IndexSpecs,
+                                    ?RIAK_TAG,
+                                    infinity, Sync) of
```
ThomasArts, line 653: Verified `leveled` PR, this has effect in underlying library.

src/riak_kv_bucket.erl
```diff
@@ -316,6 +323,21 @@ is_quorum(Q)  when Q =:= quorum
 is_quorum(_) ->
     false.
 
+%% validation of sync parameters
+%% one = sync coordinating node only
+%% all = sync all nodes
+%% backend = take sync value for all nodes from backend config (don't override)
+-spec is_valid_sync_param(term()) -> boolean().
+is_valid_sync_param(SP) when SP =:= one
```
ThomasArts, line 331: Note that `default` does not validate. Is that really what you want?

compare to https://github.com/basho/riak_kv/blob/607275628745008814774898e198e18b47b23bc3/src/riak_kv_put_fsm.erl#L333

src/riak_kv_util.erl
```diff
@@ -160,6 +161,20 @@ expand_value(Type, default, BucketProps) ->
 expand_value(_Type, Value, _BucketProps) ->
     Value.
 
+expand_sync_on_write(default, BucketProps) ->
+    normalize_value(get_bucket_option(sync_on_write, BucketProps));
+expand_sync_on_write(Value, _BucketProps) ->
+    Value.
```
ThomasArts, line 167: should you not normalise here as well?

src/riak_kv_vnode.erl
```diff
@@ -2962,30 +2966,48 @@ perform_put({true, {_Obj, _OldObj}=Objects},
...
+                     sync_on_write=SyncOnWrite}) ->
     case ReadRepair of
       true ->
         MaxCheckFlag = no_max_check;
       false ->
         MaxCheckFlag = do_max_check
     end,
-    actual_put(BKey, Objects, IndexSpecs,
-                RB, ReqID, MaxCheckFlag, Coord, State).
+    case SyncOnWrite of
```
ThomasArts: This way of coding is not Erlang like. In OTP24 you may even get an error or warning.
You bind `Sync` in each branch. Instead you should bind it outside the case with the result of the case.

ThomasArts:
```suggestion
    Sync = case SyncOnWrite of
```

src/riak_kv_vnode.erl
```diff
@@ -2962,30 +2966,48 @@ perform_put({true, {_Obj, _OldObj}=Objects},
...
+        all ->
+           Sync = true;
```
ThomasArts:
```suggestion
           true;
```

src/riak_kv_vnode.erl
```diff
@@ -2962,30 +2966,48 @@ perform_put({true, {_Obj, _OldObj}=Objects},
...
+        %% 'one' does not override the backend value for the other nodes
+        %% therefore is only useful if the backend is configured to not sync-on-write
+        one ->
+           Sync = Coord;
```
ThomasArts:
```suggestion
           Coord;
```

src/riak_kv_vnode.erl
```diff
@@ -2962,30 +2966,48 @@ perform_put({true, {_Obj, _OldObj}=Objects},
...
+        _ ->
+           %% anything but all or one means do the default configured backend write
+           Sync = false
```
ThomasArts:
```suggestion
           false
```

ThomasArts (commented):
Some smaller comments, but did not find serious issues

src/riak_kv_bucket.erl
martinsumner: The strings are in the ensemble/write_once logic which is rarely/never used.  So being consistent with the use of atoms is the right path forward.

src/riak_kv_bucket.erl
martinsumner: `default` is not a valid property - it indicates to use the property on the bucket not on the request, and relies on the `riak_kv_util:expand_sync_on_write/2` being called.

https://github.com/basho/riak_kv/blob/mas-i350-selectivesync/src/riak_kv_util.erl#L164-L167

Perhaps there is a neater way of doing this. The riak_kv_bucket module is never directly referred to in the code - it gets registered as the validator of buckets within riak_core:

https://github.com/basho/riak_kv/blob/mas-i350-selectivesync/src/riak_kv_app.erl#L241-L246

So this only gets applied when updating the bucket properties.

Each API validates the per-request setting:

https://github.com/basho/riak_kv/blob/mas-i350-selectivesync/src/riak_kv_wm_object.erl#L423-L428

So `default` is a valid thing to say on a request, but not a valid thing to have as the bucket property.

src/riak_kv_eleveldb_backend.erl
martinsumner: This relies on the presence of sync in the eleveldb cuttlefish:

https://github.com/basho/eleveldb/blob/develop-3.0/priv/eleveldb.schema#L32-L39

which then becomes part of the write_opts at startup:

https://github.com/basho/riak_kv/blob/mas-i350-selectivesync/src/riak_kv_eleveldb_backend.erl#L633-L635

So it will always be there, base don the current logic.  However, this does appear to be quite a brittle chain.

src/riak_kv_leveled_backend.erl
martinsumner: Agree

src/riak_kv_util.erl
martinsumner: I think normalize_value may be unnecessary in both cases due to validation at the API/riak_kv_bucket

PR #1795, opened 2021-09-14
Mas i1793 repairkeysrange

martinsumner:
Add fold to accelerate read repair on ranges on keys.

The fold will simply prompt a GET on every key, which will in turn prompt a read_repair if that is necessary.

Some consideration was given to making a more targeted change - e.g. only GET those keys that exist on a specific node with a known repair requirement.  However, preference is to keep the change a simple as possible, with some flexibility in the API to divert from specifying `all` to repair in a future change.

### DIFF
--- a/src/riak_kv_vnode.erl
+++ b/src/riak_kv_vnode.erl
@@ -1845,6 +1845,39 @@ handle_aaefold({repl_keys_range,
                                 InitAcc, 
                                 [{clock, null}]),
     {select_queue(?AF4_QUEUE, State), {fold, Folder, ReturnFun}, Sender, State};
+handle_aaefold({repair_keys_range,
+                        Bucket, KeyRange,
+                        ModifiedRange,
+                        all},
+                    InitAcc, _Nval,
+                    IndexNs, Filtered, ReturnFun, Cntrl, Sender,
+                    State) ->
+    {ok, C} = riak:local_client(),
+    FetchFun = riak_kv_clusteraae_fsm:repair_fun(C),
+    FoldFun =
+        fun(BF, KF, _EFs, Acc) ->
+            {AccL, Count, all, BatchSize} = Acc,
+            case Count rem BatchSize of
+                0 ->
+                    lists:foreach(FetchFun, AccL), 
+                    {[{BF, KF}], Count + 1, all, BatchSize};
+                _ ->
+                    {[{BF, KF}|AccL], Count + 1, all, BatchSize}
+            end
+        end,
+    WrappedFoldFun = aaefold_withcoveragecheck(FoldFun, IndexNs, Filtered),
+    RangeLimiter = aaefold_setrangelimiter(Bucket, KeyRange),
+    ModifiedLimiter = aaefold_setmodifiedlimiter(ModifiedRange),
+    {async, Folder} = 
+        aae_controller:aae_fold(Cntrl, 
+                                RangeLimiter,
+                                all,
+                                ModifiedLimiter,
+                                false,
+                                WrappedFoldFun, 
+                                InitAcc, 
+                                []),
+    {select_queue(?AF4_QUEUE, State), {fold, Folder, ReturnFun}, Sender, State};
 handle_aaefold({find_keys, 
                         Bucket, KeyRange,
                         ModifiedRange,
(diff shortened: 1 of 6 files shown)


martinsumner:
Tes: https://github.com/basho/riak_test/pull/1357

martinsumner:
https://github.com/basho/riak/discussions/1081

src/riak_kv_vnode.erl
```diff
@@ -1845,6 +1845,39 @@ handle_aaefold({repl_keys_range,
...
+                    IndexNs, Filtered, ReturnFun, Cntrl, Sender,
+                    State) ->
+    {ok, C} = riak:local_client(),
+    FetchFun = riak_kv_clusteraae_fsm:repair_fun(C),
+    FoldFun =
+        fun(BF, KF, _EFs, Acc) ->
+            {AccL, Count, all, BatchSize} = Acc,
+            case Count rem BatchSize of
+                0 ->
+                    lists:foreach(FetchFun, AccL),
```
ThomasArts, line 1866: This may be worth a comment. Why fetching in the case of zero, but not when larger than zero

src/riak_kv_wm_aaefold.erl
```diff
@@ -81,12 +83,13 @@
 %%          the merkle tree. If absent then the default hash fun is used.
 %%   </li>
 %%   <li><tt>change_method=count|local|{job, Integer}</tt><br />
-%%          rangetrees trees query only
-%%          an integer that is an initialisation vector for the hash
-%%          method for the trees. Useful for avoiding hash collision,
-%%          this IV will be used to initialise the hash function that
-%%          will be used to hash version vectors that go into creating
-%%          the merkle tree. If absent then the default hash fun is used.
+%%          Used only on reape and erase queries.  If the change_method is set
```
ThomasArts:
```suggestion
%%          Used only on reap and erase queries.  If the change_method is set
```

src/riak_kv_wm_aaefold.erl
```diff
@@ -26,6 +26,8 @@
 %% GET /cachedtrees/nvals/NVal/keysclocks?filter
 %% GET /rangetrees/[types/Type/]buckets/Bucket/trees/Size?filter
 %% GET /rangetrees/[types/Type/]buckets/Bucket/keysclocks?filter
+%% GET /rangerepl/[types/Type/]bucket/Bucket?filter
```
ThomasArts: It is kind of convention in REST land to use plural here `buckets` instead of `bucket` as in the other endpoints.

src/riak_kv_wm_aaefold.erl
```diff
@@ -560,6 +564,48 @@ malformed_range_repl_request(Filter0, QN0, RD, Ctx) ->
...
     end.
 
+%% @private validate and parse a range repair request
+%% Needs a bucket as minimum - plus also maybe a key range and a
+%% date range
+-spec malformed_range_repair_request(#wm_reqdata{},
+                                    context()) ->
+                                        {boolean(), #wm_reqdata{}, context()}.
+malformed_range_repair_request(RD, Ctx) ->
+    case wrq:path_info(bucket, RD) of
```
ThomasArts, line 574: if you change to the more conventional `buckets` then change this key search

ThomasArts (commented):
Check whether you want to use the more standard REST notation
https://restfulapi.net/resource-naming/

src/riak_kv_wm_aaefold.erl
ThomasArts: In fact, if the Bucket is a bucket identifier, then the query `?filter` is a bit weird.
The `filter` here stands for a query in which several possible filters can be provided, such as `segments`, `all`, `start` and `end`. This is hard to express in http notation. One would list them all and have somewhere stated which combinations are possible, so never mind.

martinsumner: The bucket/buckets is a typo in the comments - buckets is actually what is being used here.  Will correct.

src/riak_kv_wm_aaefold.erl
martinsumner: This is set in riak_kv_web:raw_dispatch/1 - the path is buckets, but the atom for the extract from the path is `bucket`

PR #1798, opened 2021-10-07
Mas i1656 defaultobjecthashversion

martinsumner:
Default object_hash_version to 0.  Part of the capability check, if a node has registered and gossiped any capability, but not this one, it will fall back to RPC check to environment variable.  If this succeeds (node is up) but returns undefined - it will incorrectly default to legacy.

Should `legacy` be required, it can still be set in advanced.config.


martinsumner:
Reviewed with @ThomasArts.  Manual test by @martinsumner to confirm override.

PR #1802, opened 2021-10-27
Mas i1801 2iqueryperf

martinsumner:
Adds stats to secondary index queries.  Adds to riak stats for each vnode fold, and adds an optional log to the `riak_kv_index_fsm`.

The optional log can be enabled by setting `riak_kv.log_index_fsm` to true (or enabling this via riak.conf).

Dependencies are updated to included the revised `riak_core_coverage_plan:initiate_plan/5` function which offers significant performance advantages for larger ring_sizes.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -47,7 +47,7 @@
 ]}.
 
 {deps, [
-    {riak_core, {git, "https://github.com/basho/riak_core.git", {branch, "develop-3.0"}}},
+    {riak_core, {git, "https://github.com/basho/riak_core.git", {branch, "mas-i1801-monitorworkerq"}}},
     {sidejob, {git, "https://github.com/basho/sidejob.git", {tag, "2.1.0"}}},
     {bitcask, {git, "https://github.com/basho/bitcask.git", {tag, "2.1.0"}}},
     {redbug, {git, "https://github.com/massemanet/redbug", {tag, "1.2.2"}}},
--- a/rebar.config
+++ b/rebar.config
@@ -47,15 +47,15 @@
 ]}.
 
 {deps, [
-    {riak_core, {git, "https://github.com/basho/riak_core.git", {tag, "riak_kv-3.0.8"}}},
+    {riak_core, {git, "https://github.com/basho/riak_core.git", {branch, "mas-i1801-coverageplan"}}},
     {sidejob, {git, "https://github.com/basho/sidejob.git", {tag, "2.1.0"}}},
     {bitcask, {git, "https://github.com/basho/bitcask.git", {tag, "2.1.0"}}},
     {redbug, {git, "https://github.com/massemanet/redbug", {tag, "1.2.2"}}},
     {recon, {git, "https://github.com/ferd/recon", {tag, "2.5.2"}}},
     {sext, {git, "https://github.com/uwiger/sext.git", {tag, "1.4.1"}}},
-    {riak_pipe, {git, "https://github.com/basho/riak_pipe.git", {tag, "riak_kv-3.0.5"}}},
+    {riak_pipe, {git, "https://github.com/basho/riak_pipe.git", {branch, "mas-i1801-coverageplan"}}},
     {riak_dt, {git, "https://github.com/basho/riak_dt.git", {tag, "riak_kv-3.0.0"}}},
-    {riak_api, {git, "https://github.com/basho/riak_api.git", {tag, "3.0.8"}}},
+    {riak_api, {git, "https://github.com/basho/riak_api.git", {branch, "mas-i1801-coverageplan"}}},
     {hyper, {git, "https://github.com/basho/hyper", {tag, "1.1.0"}}},
     {kv_index_tictactree, {git, "https://github.com/martinsumner/kv_index_tictactree.git", {branch, "develop-3.0"}}},
     {riakhttpc, {git, "https://github.com/basho/riak-erlang-http-client", {tag, "3.0.8"}}}
--- a/rebar.config
+++ b/rebar.config
@@ -57,6 +57,6 @@
     {riak_dt, {git, "https://github.com/basho/riak_dt.git", {tag, "riak_kv-3.0.0"}}},
     {riak_api, {git, "https://github.com/basho/riak_api.git", {tag, "3.0.8"}}},
     {hyper, {git, "https://github.com/basho/hyper", {tag, "1.1.0"}}},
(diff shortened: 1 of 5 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1359

rebar.config
```diff
@@ -47,16 +47,16 @@
 ]}.
 
 {deps, [
-    {riak_core, {git, "https://github.com/basho/riak_core.git", {tag, "riak_kv-3.0.8"}}},
+    {riak_core, {git, "https://github.com/basho/riak_core.git", {branch, "mas-i1801-monitorworkerq"}}},
```
ThomasArts: work needed before merge (I mean, don't forget to reset to right branches)

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
     LastResults = sms:done(MergeSortBuffer),
-    DownTheWire = case (ResultsSent + length(LastResults)) > MaxResults of
-                      true ->
-                          lists:sublist(LastResults, MaxResults - ResultsSent);
-                      false ->
-                          LastResults
-                  end,
+    TotalResults = length(LastResults) + ResultsSent,
+    DownTheWire =
+        case TotalResults > MaxResults of
```
ThomasArts: We don't need TotalResults. The function lists:sublist takes care of this, taking only the MaxResult - ResultSent is sufficient, making the code substantially easier:

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
                     results_sent=ResultsSent,
                     max_results=MaxResults}) ->
     LastResults = sms:done(MergeSortBuffer),
-    DownTheWire = case (ResultsSent + length(LastResults)) > MaxResults of
-                      true ->
-                          lists:sublist(LastResults, MaxResults - ResultsSent);
-                      false ->
-                          LastResults
-                  end,
+    TotalResults = length(LastResults) + ResultsSent,
```
ThomasArts:
```suggestion
```

src/riak_kv_index_fsm.erl
ThomasArts:
```suggestion
```

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
+    DownTheWire =
+        case TotalResults > MaxResults of
+            true ->
```
ThomasArts:
```suggestion
```

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
+                lists:sublist(LastResults, MaxResults - ResultsSent);
```
ThomasArts:
```suggestion
      lists:sublist(LastResults, MaxResults - ResultsSent),
```

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
+            false ->
```
ThomasArts:
```suggestion
```

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
+                LastResults
```
ThomasArts:
```suggestion
```

src/riak_kv_index_fsm.erl
```diff
@@ -194,30 +221,38 @@ process_results({Bucket, Results},
...
+        end,
```
ThomasArts, line 252:
```suggestion
```

src/riak_kv_stat.erl
```diff
@@ -263,6 +263,11 @@ do_update({put_fsm_time, Bucket,  Microsecs, Stages, PerBucket, CRDTMod}) ->
     ok = create_or_update([P, ?APP, node, puts, Type, time], Microsecs, histogram),
     ok = do_stages([P, ?APP, node, puts, Type, time], Stages),
     do_put_bucket(PerBucket, {Bucket, Microsecs, Stages, Type});
+do_update({index_fsm_time, Microsecs, ResultCount}) ->
+    P = ?PFX,
```
ThomasArts, line 267: Correct, but I like 3 copies of ?PFX in arguments below better... Renaming a macro makes it not as easy to see that a macro is used.

In line 272 below the macro is also used, so more in line with that code

ThomasArts (approved):

src/riak_kv_index_fsm.erl
martinsumner: MaxResults, it turns out can be the atom all as well as an integer(), and this relies on the fact any integer() > atom() is false.

So if we just have lists:sublist we get badarith.

But it is a bit worse, as it is also possible to construct a query whereby MaxResults is undefined, but a page-sort query is still requested.  undefined isn't expected by dialyzer.

Going to make the case of all vs integer() more specific, and ensure that not integer() is always all

src/riak_kv_stat.erl
martinsumner: The convention in the module is that if ?PFX is required more than once in a  function, it is called then assigned rather than repeatedly called.  It might be a performance thing (this a very active piece of code, ignoring excess calls to get_env), or perhaps a consistency thing (so that the prefix can't change across a set of updates.

Without knowing why this is the convention, I don't wish to change it.

martinsumner:
Full list of stats added by this PR:

index_fsm_complete
 - this increments every time a 2i query complete (there is already a counter for every time one is started)

index_fsm_results_mean
index_fsm_results_median
index_fsm_results_95
index_fsm_results_99
index_fsm_results_100
- a histogram of the numbers of results seen in 2i queries on this node

index_fsm_time_mean
index_fsm_time_median,
index_fsm_time_95,
index_fsm_time_99
index_fsm_time_100
- a histogram of the time taken for 2i queries on this node

worker_vnode_pool_queuetime_mean"
worker_vnode_pool_queuetime_100
worker_vnode_pool_worktime_mean
worker_vnode_pool_worktime_100
- the max and mean time spent on each vnode running queries, both queueing to be run and actually running

worker_af1_pool_queuetime_mean"
worker_af1_pool_queuetime_100
worker_af1_pool_worktime_mean
worker_af1_pool_worktime_100
..... (i.e. repeated for af2_pool etc)
- as above, but for each worker pool which can run queries (i.e. those for aae_folds, rebuilds etc)

Issue #1805, opened 2021-11-12
Use Tictac AAE for hints for more repairs


martinsumner:
The tictac AAE process is deficient when compared with the "legacy" riak kv_index_hashtree approach to AAE, in that it repairs much slower.

The repairs are required to be slower as the cost of running the fetch_clocks comparison is relatively high in Tictac AAE, as they require a full scan of the vnode key store (albeit skipping any blocks that don't include relevant segments).  The cost of fetch_clocks is proportionate to the number of segments in the results as defined by `tictacaae_maxresults`:

```
%% @doc Max number of leaf IDs per exchange
%% To control the length of time for each exchange, only a subset of the
%% conflicting leaves will be compared on each exchange.  If there are issues
%% with query timeouts this may be halved.  Large backlogs may be reduced
%% faster by doubling.  There are 1M segments in a standard tree overall.
{mapping, "tictacaae_maxresults", "riak_kv.tictacaae_maxresults", [
  {datatype, integer},
  {default, 256},
  hidden
]}.
```

On very large vnodes, 256 segments being checked on a fold can be time consuming, and the AAE process will back off if too much time is consumed - but this limits the number of repairs per exchange to be approximately 256, and then the back-off throttles the repair speed even more.

The proposed enhancement is to add two new workers per node:

- riak_kv_tictacaae_monitor
- riak_kv_readrepair_pool

## Tictac AAE Monitor

Results from clock comparisons will now be sent to `riak_kv_tictacaae_monitor` rather than direct for read repair, along with the relevant n-val and partition information pertinent to the exchange.  The monitor will prompt the read repairs as now, but will also generate a per-bucket and per-nval list of LastModifiedDates on those keys where a repair was required.

If there is a bucket with > 25% of the repairs - the a new fetch clock comparison should be prompted between the vnodes.  This time the constrained should not be constrained by segment but by bucket and the last modified date range which covers at least 25% of the discovered repairs(i.e. using `fetch_clocks_range`).

Otherwise the nval list should be checked and for the nval with the most required repairs a last modified date range for a `fetch_clocks_nval` should be determined that covers at least 25% of the repairs.

Having discovered a potentially broken part of the store, the riak_kv_tictcaaae_monitor should now repeat a `fetch_clocks_range`/`fetch_clocks_nval` comparison for the discovered range, and prompt any discrepancies for read repair.

[...]

PR #1806, opened 2021-11-18
Mas i1805 hintedrepair

martinsumner:
In Tictac AAE, repairs are slow relative to the previous (and still default) kv_index_hashtree solution.  Comparisons to discover broken AAE segments are fast, but then a fetch_clocks query has to be run.

The fetch_clocks query is a full keystore scan - but each slot in the store has a segment index, so whole slots can be skipped when they contain no interesting segments.  The skipping improves speed, but there are issues:

- the higher `tictcaaae_maxresults` (the segment count) the less skipping can be done, and the slower the query.
- in order to address issues whereby a key is misrepresented in the cached AAE tree, extra work must be done to rebuild the segments in the AAE store as part of the query.
- to avoid issues with items in the key-store but not in the Journal (when running native leveled) the query has a journal presence check enabled.

What is implemented here is prompted repairs. There is as before, scheduled exchanges, however these by default only repair 128 segments not 256 (to reduce overheads).  These exchanges are otherwise unchanged.

If the exchanges highlight keys that need repairing, they are repaired as before.  However, the `riak_kv_tictcaaae_repairs:analyse_repairs/2` function examines those repaired keys - looking at the modified dates, buckets and (optionally) the key ranges.  The `analyse_repairs/2` will result potentially then in a new exchange being prompted, but this time with a filter that will accelerate the time taken to run the resulting fetch_clocks query (with either a bucket and modified time range, or just a modified time range).

These prompted exchanges should be able to find repairs (via fetch_clocks) an order of magnitude faster than standard exchanges, as:

- the additional constraints should greatly reduce the amount of the key space to be covered (one would normally expect a large delta to have a small range of last_modified times in particular);
- as normal exchanges are handling the outlier cases (no journal presence, broken cached tree), the extra activity (check presence and segment rebuild) is not required on these prompted exchanges.

There will be `tictcaaae_repairloops` (default 4) run following an exchange, until the loop count is exhausted or a loop finds insufficient keys to warrant continuing (less than 50% of requested max results).  These loops will have the `tictacaae_maxresults` boosted by a `tictacaae_rangeboost` integer (default 2 i.e. go to 256).


martinsumner:
https://github.com/basho/riak_test/pull/1360

martinsumner:
https://github.com/basho/riak_kv/issues/1805

As the pause system exists to space out any repairs, the proposal in the issue to have a new read-repair-manager process has been dropped.

martinsumner:
https://github.com/martinsumner/kv_index_tictactree/pull/101

martinsumner:
Test evidence of any performance/efficiency improvements still required before PR is merged

martinsumner:
For initial performance testing, this riak_test can be used for comparisons.

https://github.com/basho/riak_test/blob/171476caa2b2c2ba3a65906ca2c488089d4a0df6/tests/verify_tictac_aae_load.erl

With these settings (switching the MAX_RESULTS back to 256 when testing previous), approximately 60-70% reduction in repair time with this branch

martinsumner:
In a full environment volume test, the new code was able to repair deltas discovered via AAE at 3.5 x the rate of the previous code.  At this rate, it was using more CPU, but roughly only 30% - 50% more.  so the improvement is CPU efficient.

It could be de-tuned to gain the extra benefit at the equivalent speed (such as by reducing `tictacaae_maxresults` further, down to 64).

The volume test was done with a ring-size of 512, on an 8-node cluster with about 180M keys.  On one node a backup is taken, then more load is added, the node is stopped, and then more load is added - then the node is recovered from backup.  

When the node is recovered, there is a hinted handoff of recent writes (since the stop), but then the previous delta (between the backup and the stop) needs to be recovered via AAE.  In this case, this delta amounts to about 1.8M keys.

In Riak 3.0.9, recovering the delta completely takes just over 24 hours, although the majority is recovered within 16 hours.  The peak rate of recovery is just over 3K objects per minute.  During the recovery, about 10% of available CPU is consumed.  Chart shows decline in mismatched segments over 24 hours:

MismatchedSegments_Riak309

With this branch, recovering the delta completely takes about 10 hours, and the majority is recovered within 4 hours.  The peak rate of recovery is 11K objects per minute.  During the recovery, about 15% of available CPU is consumed.

MismatchedSegments_HintedRepair

If the `tictacaae_maxresults` is reduced further, halved to 64, with hinted repair - the CPU utilisation reverts to that found in Riak 3.0.9, but the recovery time is faster (though not as fats as with the higher `tictacaae_maxresults`:

MismatchedSegments_HintedRepair_MR64

With this setting, recovering the delta completely takes 17 hours, and the majority is recovered within 8 hours.  The peak rate of recovery is 7K objects per minute.  During the recovery less than 10% of available CPU is consumed.

martinsumner:
Note that there appear to be two "stripes" in the recovery.  There are two forms of delta created by the test:

- genuine deltas as a result of the node being started prior at a state prior to recent updates being received (that are not covered in handoffs)
- resurrected keys, as a result of running Riak in a non-keep delete mode, and as a side effect of the journal-compaction/backup/recovery process within leveled.  These resurrected keys are fewer in number, but concentrated in a smaller subset of vnodes (and hence take longer to clear).

priv/riak_kv.schema
```diff
@@ -121,8 +121,7 @@
 %% faster by doubling.  There are 1M segments in a standard tree overall.
 {mapping, "tictacaae_maxresults", "riak_kv.tictacaae_maxresults", [
   {datatype, integer},
-  {default, 128},
-  hidden
+  {default, 64}
```
ThomasArts, line 128: But need update in PR explanation as well

`What is implemented here is prompted repairs. There is as before, scheduled exchanges, however these by default only repair 128 segments not 256 (to reduce overheads). These exchanges are otherwise unchanged.`

This is 64 now, right? and if so, then do you need to re-explain :
`These loops will have the tictacaae_maxresults boosted by a tictacaae_rangeboost integer (default 2 i.e. go to 256).`
I would expect the default now to be 4???

priv/riak_kv.schema
```diff
@@ -121,7 +121,34 @@
...
+%% Problems related to invalid hash trees will need to be detected through the
+%% initial loop of the exchange, not repair loops
+{mapping, "tictacaae_repairloops", "riak_kv.tictacaae_repairloops", [
+  {datatype, integer},
+  {default, 4},
+  hidden
+]}.
+
+%% @doc Multiplier when following an exchange with a range-limited one
+%% After eache xchange where sufficient deltas are discovered there will be
```
ThomasArts:
```suggestion
%% After each exchange where sufficient deltas are discovered there will be
```

priv/riak_kv.schema
```diff
@@ -121,7 +121,34 @@
...
+%% `tictacaae_repairloops` number of range-limited queries (assuming
+%% sufficient results continue to be found).  Each of these may have the
+%% the number of max results boosted by this integer factor.
+{mapping, "tictacaae_rangeboost", "riak_kv.tictacaae_rangeboost", [
+  {datatype, integer},
+  {default, 2},
```
ThomasArts, line 163: Is this now 4 in line with `256 div 64` is 4? If that is the case, then I wonder if it should be a configurable constant or whether it is a derived constant. I did not really get this from the above explanation. Tried to correlated it what was written in the commit message:

`These loops will have the tictacaae_maxresults boosted by a tictacaae_rangeboost integer (default 2 i.e. go to 256).`

ThomasArts (commented):
not done yet with review... but want to save the state

priv/riak_kv.schema
martinsumner: The change in constants happened after the [volume test above](https://github.com/basho/riak_kv/pull/1806#issuecomment-975360319).

With maxresults=128 and rangeboost=2, the repair was faster, but it used more CPU than the current solution.

With maxresults=64 and rangeboost=2, the repair was still better than as-is (though not as goos as with 128), but the CPU use was the same as-is.

It was felt it was safer to make the introduction by default in such a way that it wouldn't use more resource, and leave it up to the user if they want to tune it higher and accept the trade off of  further improved performance for further resource utilisation.

src/riak_kv_tictacaae_repairs.erl
```diff
@@ -0,0 +1,376 @@
...
+            {transition_pause_ms, ExchangePause},
+            {purpose, kv_aae},
+            {max_results, MaxResults}],
+    
+    BlueList = 
+        [{riak_kv_vnode:aae_send(LocalVnode), [IndexN]}],
+    PinkList = 
+        [{riak_kv_vnode:aae_send(RemoteVnode), [IndexN]}],
+    PromptRehash =
+        case Filter of
```
ThomasArts: PromptRehash = (Filter == none)

src/riak_kv_tictacaae_repairs.erl
```diff
@@ -0,0 +1,376 @@
...
+%% one clock has an identifiably higher last modified time which can be
+%% considered a likely timestamp of an incident
+-spec analyse_repair({{riak_object:bucket(), riak_object:key()},
+                        {vclock:vclock(), vclock:vclock()}},
+                        repair_list()) -> repair_list().
+analyse_repair({{B, K}, {BlueClock, PinkClock}}, ByBucketAcc) ->
+    BlueTime = last_modified(BlueClock),
+    PinkTime = last_modified(PinkClock),
+    ModTime = 
+        case {BlueTime, PinkTime} of
```
ThomasArts: This probably would be clearer in an ìf`, the matching is a bit weird here.

ThomasArts (approved):

src/riak_kv_tictacaae_repairs.erl
martinsumner: Change made

src/riak_kv_tictacaae_repairs.erl
martinsumner: Change made

priv/riak_kv.schema
martinsumner: Clarified comments, included examples

PR #1809, opened 2021-12-09
Mas i1807 overflowqueues

martinsumner:
Alternative solution to the unbound queue problem (e.g. alternative to https://github.com/basho/riak_kv/pull/1808).

Rather than simply curtailing when size of queue hits an arbitrary limit - overflow to disk at a limit instead.

This solution also refactors the reaper/eraser to use a common behaviour to remove duplication of code.

### DIFF
--- a/README.org
+++ b/README.org
@@ -43,7 +43,7 @@ git push mine some-topic-branch
 # standard tests
 ./rebar3 do xref, dialyzer, eunit
 # property-based tests
-./rebar3 as test eqc
+./rebar3 as test eqc --testing_budget 600
 #+END_SRC
 
 For a more complete set of tests, update riak_kv in the full Riak application and run any appropriate [[https://github.com/basho/riak_test/tree/develop-3.0/groups][Riak riak_test groups]]
--- a/eqc/eraser_eqc.erl
+++ b/eqc/eraser_eqc.erl
@@ -18,8 +18,16 @@ start_args(_S) ->
   [gen_delete_mode()].
 
 start(DelMode) ->
-  {ok, Pid} = gen_server:start_link(riak_kv_eraser, [eqc_job, fun erase/2, DelMode], []),
-  Pid.
+    FilePath = riak_kv_test_util:get_test_dir("eraser_eqc"),
+    {ok, Pid} = riak_kv_eraser:start_link(FilePath),
+    case DelMode of
+        keep ->
+            riak_kv_eraser:override_redo(false);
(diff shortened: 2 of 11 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1361

hmmr:
The proposed change implements a disk-backed buffer for delete and reap request queues (as opposed to in-memory queues). It is a mitigation against said queues growing and consuming too much memory, and potentially causing OOM conditions on a node, observed when, for example, a node re-joins its cluster after substantial amount of data has been modified/deleted in the cluster, causing massive AAE activity with which the joining node is unable to cope.

I can see all pieces in the flow and I find eunit tests with good coverage, which I can confirm are passing.

One concern that I have is about stale disk_log files: I agree that ["wiping of wrong data"](https://github.com/basho/riak_kv/pull/1809/commits/fee56e5931d67090b8b4407982569b83ac91e3a8) is a valid argument to the contrary; but something still needs to be done to prevent stale files from appearing. What if `riak_kv.eraser_dataroot` and `riak_kv.reaper_dataroot` are placed in a temp (uuid-based) dir, specifically created at node startup, under some `riak_kv.disklog_queue_dir`? Users can then be able to distinguish between the dir currently in use from dirs left behind, and delete them without the need to stop the node.

Other than that, a couple of nits: no need to [++ strings](https://github.com/basho/riak_kv/pull/1809/files#diff-24efacea5e7b7d6be1d68972e4c37eb252e661437b2e6eb25c4cb193e58df103R126), and [lots](https://github.com/basho/riak_kv/pull/1809/files#diff-13190cb1d6b02cf95feecfe576f9c30f09e066b6e1ff102c409d463ba9040d02R96) of [trailing](https://github.com/basho/riak_kv/pull/1809/files#diff-0e69c47b07842327ba0cecd5c2f63c22852b8a0767fd99a3dc8c12b8b1f8d57fR83) [whitespace](https://github.com/basho/riak_kv/pull/1809/files#diff-24efacea5e7b7d6be1d68972e4c37eb252e661437b2e6eb25c4cb193e58df103R151). Bonus points for using a solution from stdlib (disk_log) instead of reinventing it!

martinsumner:
@hmmr With regards to the use of '++', is this a suggestion to use iolists instead, with the improved efficiency?

hmmr (approved):
I meant to say that `++` between strings is redundant. But that's not important.

Issue #1804, opened 2021-11-12
Auto-connect for nextgenrepl real-time


martinsumner:
As a general rule, a design decision was made for nextgenrepl to lean more heavily on setup being discovered by operator configuration rather than discovery.  This reduces complexity in the code.

However, this creates an overhead for operators, especially when considering cluster changes where nextgenrepl real-time replication is used.

Currently each node in the "sink" is configured with a static set of peers in the source to connect with:

```
%% @doc A list of peers is required to inform the sink node how to reach the
%% src.  All src nodes will need to have entries consumed - so it is
%% recommended that each src node is referred to in multiple sink node
%% configurations.
%% The list of peers is tokenised as host:port:protocol
%% In exceptional circumstances this definition can be extended to
%% queuename:host:port:protocol - but restricting the definitions of queuename
%% to the single queue specified in replrtq_sinkqueue is strongly recommended.
{mapping, "replrtq_sinkpeers", "riak_kv.replrtq_sinkpeers", [
    {datatype, string},
    {commented, "127.0.0.1:8098:http"}
]}.
```

However, if we join nodes into the source cluster, as soon as that node can start to coordinate PUTs a queue of real-time replication events will begin to form.  This now requires the operator to script real-time changes to the sink cluster to reflect this.

There exists too much scope for operator error in this case.

What is proposed instead is that there will be a new config item:

```
%% @doc Use peer list for discovery only
%% If enabled rplrtq_sinkpeers will be used for discovery only.
%% A call will be made to one of those peers (at random) to get a list of upnodes, and
%% their open port. Periodically (every 60s), the call to the peer will be repeated.  The
%% actual replrtq_sinkpeers used in the configuration will be those returned by this
%% request
{mapping, "replrtq_peer_discovery", "riak_kv.replrtq_peer_discovery", [
    {datatype, {flag, enabled, disabled}},
    {default, disabled},
    {commented, disabled}
]}.
```

Ideally, different sink peers should be configured with different source peers for discovery (or a load-balance address be used).

This will not work with NAT - in this case replrtq_peer_discovery must be disabled, and the NAT'd addresses used in the replrtq_sinkpeers configuration for static peer relationships.

martinsumner:
The implementation of this will use a separate process `riak_kv_replrtq_peer`.  This process will always be started, but will only actively poll if peer_discovery is enabled.

When triggered the peer discovery process will:

[...]

PR #1812, opened 2022-03-03
Mas i1804 peerdiscovery

martinsumner:
See https://github.com/basho/riak_kv/issues/1804

The heart of the problem is how to avoid needing configuration changes on sink clusters when source clusters are bing changed.  This allows for new nodes to be discovered automatically, from configured nodes.

Default behaviour is to always fallback to configured behaviour.

Worker Counts and Per Peer Limits need to be set based on an understanding of whether this will be enabled.  Although, if per peer limit is left to default, the consequence will be the worker count will be evenly distributed (independently by each node).  Note, if Worker Count mod (Src Node Count) =/= 0 - then there will be no balancing of the excess workers across the sink nodes.


src/riak_kv_replrtq_peer.erl
```diff
@@ -0,0 +1,240 @@
...
+
+-spec start_link() -> {ok, pid()}.
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+%% @doc
+%% Prompt for the discovery of peers
+-spec update_discovery(riak_kv_replrtq_snk:queue_name()) -> boolean().
+update_discovery(QueueName) ->
+    gen_server:call(?MODULE, {update_discovery, QueueName}, 60 * 1000).
```
ThomasArts: Do you really want 60 hardcoded here or rather use the macro `?DISCOVERY_TIMEOUT_SECONDS`.

src/riak_kv_replrtq_peer.erl
```diff
@@ -0,0 +1,240 @@
...
+            MinDelay = 
+                app_helper:get_env(riak_kv,
+                    replrtq_prompt_min_seconds,
+                    ?AUTO_DISCOVERY_MINIMUM_SECONDS),
+
+            lists:foreach(
+                fun({QueueName, _PeerInfo}) -> 
+                    erlang:send_after(MinDelay * 1000,
+                        self(),
+                        {prompt_discovery, QueueName})
```
ThomasArts: It takes a few moments to realize that there are 2 interfaces for `prompt discovery` where this one goes via `handle_info` and has only one argument that does a lookup for the PeerInfo at the time it executes, whereas the cast version gets the PeerInfo in the interface.


Probably one should comment this difference more clearly. In particular why one does not need to do a state lookup for the cast version

src/riak_kv_replrtq_peer.erl
```diff
@@ -0,0 +1,240 @@
...
+    end.
+
+handle_cast({prompt_discovery, {QueueName, PeerInfo}}, State) ->
+    _ = prompt_discovery(QueueName, PeerInfo, regular),
+    {noreply, State}.
+
+handle_info({prompt_discovery, QueueName}, State) ->
+    {QueueName, PeerInfo} =
+        lists:keyfind(QueueName, 1, State#state.discovery_peers),
+    ok = prompt_discovery({QueueName, PeerInfo}),
```
ThomasArts: I kind of dislike this... you leave the context of the genserver in the cast to later return to it... I wonder if it would not be better to actually call `prompt_discovery(QueueName, PeerInfo, regular)` at this point.

src/riak_kv_replrtq_peer.erl
```diff
@@ -0,0 +1,240 @@
...
+            _ ->
+                lists:usort(
+                    lists:map(
+                        fun({ID, _D, H, P, Prot}) ->
+                            {ID, StartDelayMS, H, P, Prot}
+                        end,
+                        riak_kv_replrtq_snk:current_peers(QueueName)))
+        end,
+    case discover_peers(PeerInfo, StartDelayMS) of
+        CurrentPeers ->
```
ThomasArts, line 197: I guess this is by design, but if you run with Type `count_change` then CurrentPeers is the empty list. So in case discover_peers returns an empty list, you don't know in which of the two cases you are.

Both cases return false, but the side effects are different.

src/riak_client.erl
```diff
@@ -137,6 +138,51 @@ maybe_update_consistent_stat(Node, Stat, Bucket, StartTS, Result) ->
...
+    end. 
+
+%% @doc Reset the worker count and per peer limit on each up node, returning
+%% a list of nodes to which the change was successfully applied
+-spec replrtq_reset_all_workercounts(
+    non_neg_integer(),
+    non_neg_integer()) -> list(node()).
+replrtq_reset_all_workercounts(WorkerC, PerPeerL) ->
+    UpNodes = riak_core_node_watcher:nodes(riak_kv),
+    lists:foldl(replrtq_resetcount_fun(WorkerC, PerPeerL), [], UpNodes).
```
ThomasArts: Personally I find the code clearer if you inline this function.

src/riak_kv_replrtq_snk.erl
```diff
@@ -409,13 +428,45 @@ handle_info({prompt_requeue, WorkItem}, State) ->
     requeue_work(WorkItem),
     {noreply, State}.
 
-terminate(_Reason, _State) ->
+terminate(_Reason, State) ->
```
ThomasArts, line 434: This terminate function may now take considerably more time.

If the supervisor terminates this server, how likely is it that this takes too long compared to the time you want it restarted?

src/riak_kv_replrtq_snk.erl
```diff
@@ -409,13 +428,45 @@ handle_info({prompt_requeue, WorkItem}, State) ->
...
+                            replrtq_sinkworkers,
+                            ?DEFAULT_WORKERCOUNT),
+    PerPeerLimit =
+        app_helper:get_env(riak_kv,
+                                replrtq_sinkpeerlimit,
+                                SnkWorkerCount),
+    {SnkWorkerCount, PerPeerLimit}.
+
+set_worker_counts(SnkWorkerCount, PerPeerLimit) ->
+    application:set_env(riak_kv, replrtq_sinkworkers, SnkWorkerCount),
```
ThomasArts, line 467: Why using app_helper for getting and application for setting? Is this an artefcat of OTP18 or so?

ThomasArts (approved):
Some smaller comments, nothing structural, just making sure you did things on purpose.

src/riak_kv_replrtq_snk.erl
martinsumner: Never thought about this much, just following inline with its use elsewhere.  Looking at the docs application:get_env/3 didn't exist until R16B ... and Riak was initially written prior to that ... so I guess it is just a legacy of this.

It might be one thing to add to the list for the OTP 24+ version of Riak, to refactor throughout and to take this out.

src/riak_kv_replrtq_peer.erl
martinsumner: Tried to avoid this confusion by not relying on the empty list mis-match - rather a specific mis-match between list and atom.  Commented as well.  Don't think it is super-clean still, but improved maybe.

Issue #1813, opened 2022-03-07
Replication, delete_mode not keep and key amnesia


martinsumner:
## Scenario

There are two clusters: ClusterA, ClusterB.  Both clusters are configured with the standard delete_mode (reaping on 3s timeout).

An object key (K1) belongs to a Preflist of {VnodeA1, VnodeA2, VnodeA3} on ClusterA, and {VnodeB1, VnodeB2, VnodeB3} on ClusterB.

An object is initiated with that key K1 on ClusterA (VnodeA1 say, acting as coordinator).  The creation is replicated.

The object is deleted on ClusterB (VnodeB1 say, acting as coordinator).  The deletion is replicated.

For some reason, on one cluster (ClusterA say) the reap does not occur on the timeout.

## Replication Goes Wrong

1. There is now an object on ClusterA, a tombstone with a vector clock like `[{VnodeA1, {1, TS1}}, {VnodeB1, {1, TS2}}]`, that does not exist on ClusterB.
2. nextgenrepl full-sync now runs with ClusterA as the src, and ClusterB as the snk.
3. The aae_exchange discovers that A > B for K1 as K1 not_found on B.  It puts a reference on the replication queue.
4. The fetch, which is run via riak_kv_get_fsm pulls across the tombstone object.  But as it prompts a read (via riak_kv_get_fsm), the read final action spots that this is a tombstone, and so triggers a maybe_delete, which causes the tombstone to be reaped from ClusterA.
5. The push of the tombstone to ClusterB, is performed as expected on VnodeB2 and VnodeB3.  However on VnodeB1 there is [key amnesia](https://github.com/basho/riak_kv/pull/1643) - this vnode previously coordinated the write, but has no backend memory because of the reap.
6. Due to key amnesia, the object is written with an updated vector clock.
7. The riak_client:push function which has replicated the object prompts a GET of the object after the `asis` PUT.  This GET is intended to prompt any delete actions (e.g. it is expected to reap).   However, in this case it prompts a read-repair instead (there can only be one final action), because the tombstone's vclock differs at VnodeB1 from VnodeB2 and VnodeB3 due to the new actor epoch generated by key amnesia.
8. This means that the object has been reaped from ClusterA, but exists on ClusterB with a vector clock following read repair like `[{VnodeA1, {1, TS1}}, {VnodeB1, {1, TS2}}, {VnodeB1.1, {1, TS3}}]`.
[...]

PR #1814, opened 2022-03-08
Allow for a dual final action - repair and delete

martinsumner:
If there is repair only (as a priority over delete - which really means trigger reap if required), then there are scenarios where tombstones can perpetually re-appear, especially were there is key amnesia.


martinsumner:
https://github.com/basho/riak_test/pull/1363

martinsumner:
@martincox - you may want to check this.  It is to fix a `nextgenrepl` issue, but will change functionality in other systems.  It is a very subtle change in the GET_FSM behaviour following GET.

Previously after completing a GET, only one action (other than `nop`) could be performed: `read_repair` or `delete` (i.e. trigger tombstone reaping).  If the GET revealed that both need to occur (i.e. there needs to be a repair, and also the winning object is a tombstone so tombstone trigger also required), only the repair would be done.

This indirectly led to a weird [issue](https://github.com/basho/riak_kv/issues/1813), which could cause some serious problems.

Now, as well as `read_repair` and `delete`, an additional `delete_repair` option can occur in this circumstance.  If `delete_repair` is triggered, the FSM will first repair then delete.  So a fetch will always trigger delete when required.

Issue #1817, opened 2022-03-09
Real-time replication - switch to `riak_kv_overflow_queue`


martinsumner:
The `riak_kv_overflow_queue` has been added as part of [this enhancement to the reaper/eraser queues](https://github.com/basho/riak_kv/pull/1809).

Currently the realtime replication queue is in memory, and of a limited fix size.  Change this to use the overflow queue, to allow for:

- larger backlogs;
- backlogs to persist across reboots.

[The reason](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) why originally this was simply an in-memory queue, was that full-sync will always fill the gaps.  Although full-sync is now cheaper than key-listing, especially when gaps are small, it is still a slow operation.

If there are o(100K) or greater gaps, full-sync is going to be more time consuming and expensive than draining from a queue, if such a queue exists.

In implementing this, the queue should persist across backlogs but start empty.  There should be an admin command to empty a previous real-time queue on the node, so that replaying such a persisted backlog is an operator choice.

PR #1818, opened 2022-03-14
Mas i1817 overflowrtq

martinsumner:
Expand on use of `riak_kv_overflow_queue` so that it is used by the riak_kv_replrtq_src, as well as riak_kv_reaper and riak_kv_eraser.

This means that larger queue sizes can be supported for `riak_kv_replrtq_src` without having to worry about compromising the memory of the node.  This should allow for `repl_keys_range` AAE folds to generate very large replication sets, without clogging the node worker pool by pausing so that real-time replication can keep up.

the overflow queues are deleted on shutdown (if there is a queue on disk).  The feature is to allow for larger queues without memory exhaustion, persistence is not used to persist queues across restarts.


martinsumner:
https://github.com/basho/riak_kv/issues/1817

martinsumner:
Overflow Queues extended to include a 'reader' queue which may be sused for read_repairs.  Currently this queue is only used for the repair_keys_range query and the read-repair triggered by key amnesia.

PR #1829, opened 2022-05-12
Mas i1817 overflowrtq

martinsumner:
Expand on use of riak_kv_overflow_queue so that it is used by the riak_kv_replrtq_src, as well as riak_kv_reaper and riak_kv_eraser.

This means that larger queue sizes can be supported for riak_kv_replrtq_src without having to worry about compromising the memory of the node. This should allow for repl_keys_range AAE folds to generate very large replication sets, without clogging the node worker pool by pausing so that real-time replication can keep up.

The overflow queues are deleted on shutdown (if there is a queue on disk). The feature is to allow for larger queues without memory exhaustion, persistence is not used to persist queues across restarts.

Overflow Queues extended to include a 'reader' queue which may be used for read_repairs. Currently this queue is only used for the repair_keys_range query and the read-repair trigger.


martinsumner:
https://github.com/basho/riak_kv/issues/1817

priv/riak_kv.schema
```diff
@@ -59,6 +59,46 @@
   {datatype, directory}
 ]}.
 
+%% @doc A path under which the reaper overload queue will be stored.
```
yorkshire-steve, line 62: Should this doc string be updated to indicate this location is used for all queues?

yorkshire-steve (approved):

priv/riak_kv.schema
martinsumner: There are separate locations for each queue, which you can't really see from the diff:

```
%% @doc A path under which the eraser overload queue will be stored.
{mapping, "eraser_dataroot", "riak_kv.eraser_dataroot", [
  {default, "$(platform_data_dir)/kv_eraser"},
  {datatype, directory}
]}.

%% @doc A path under which the reaper overload queue will be stored.
{mapping, "reaper_dataroot", "riak_kv.reaper_dataroot", [
  {default, "$(platform_data_dir)/kv_reaper"},
  {datatype, directory}
]}.

%% @doc A path under which the reader overload queue will be stored.
{mapping, "reader_dataroot", "riak_kv.reader_dataroot", [
  {default, "$(platform_data_dir)/kv_reader"},
  {datatype, directory}
]}.

%% @doc A path under which the repl real-time overload queue will be stored.
{mapping, "replrtq_dataroot", "riak_kv.replrtq_dataroot", [
  {default, "$(platform_data_dir)/kv_replrtqsrc"},
  {datatype, directory}
]}.
```

PR #1830, opened 2022-05-12
Mas i1813 keepamnesia

martinsumner:
Introduces a reader overflowq for doing read repair operations.  Initially this is used for:

- repair_keys_range aae_fold - avoids the pausing of the fold that would block the worker pool;
- repair on key_amnesia - triggers the required repair rather than causing an AAE delta;
- repair on finding a corrupted object when folding to rebuild aae_store - previously the fold would crash, and the AAE store would therefore never be rebuilt.  [This PR](https://github.com/martinsumner/kv_index_tictactree/pull/106) is required to make this consistent in both AAE solutions.


s2hc-johan:
Looks good, I think

martinsumner:
Related to https://github.com/basho/riak_kv/pull/1814

PR #1838, opened 2022-11-23
Mas i1121 reip3

martinsumner:
Add a `reip_manual/1` function to riak_kv_console that allows reip to be done with knowledge of ring_sir and cluster name - avoiding the need to load riak_core.

https://github.com/basho/riak_core/pull/992

https://github.com/basho/riak/issues/1121

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -42,7 +42,7 @@
 ]}.
 
 {deps, [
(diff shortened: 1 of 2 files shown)


hmmr:
+1

PR #1839, opened 2022-11-30
add missing function clause repair_keys_range in convert_fold

hmmr:
This is to unbreak `aae_fold` for `repair_keys_range`.

In collaboration with @pjaclark.

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -907,7 +907,7 @@ aae_fold(Query, {?MODULE, [Node, _ClientId]}) ->
         true ->
             riak_kv_clusteraae_fsm_sup:start_clusteraae_fsm(Node,
                                                             [{raw, ReqId, Me},
-                                                            [Query, TimeOut]]),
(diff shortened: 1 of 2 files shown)


martinsumner:
As I understand it convert_fold is only used normally via `riak_client:aae_fold/1`.  However this is broken too:

https://github.com/basho/riak_kv/blob/9ca718423ea2f37a653b933886ef1650a5e02de7/src/riak_client.erl#L905-L911

The `Q0` is used in the validity check, but not in the actual query (where `Query` is used).

Do you want to make this quick fix as part of this PR, and pass Q0 through to the query?

I have a riak_test ready.  This test extends `verify_aaefold_rangerepair` so that it works via riak_client when depending on the date conversion: https://github.com/basho/riak_test/blob/mas-1839-convertfold/tests/verify_aaefold_rangerepair.erl.

If you update the PR with the second fix, the test should pass

martinsumner (approved):

PR #1840, opened 2022-12-15
Update log

martinsumner:
Make situation clearer in log

Rather than change behaviour, just make it clear that the warning can be ignored when shutting down.  This should avoid unnecessary concern.


azharnisar:
+1

PR #1848, opened 2023-01-23
Mas i1847 putapi

martinsumner:
For both PUT and DELETE the behaviour of the HTTP API should be consistent with that of the PB API, in particular it should not do a fetch of the object, unless such a fetch is required to support conditional request headers (e.g. 'if-None-Match').

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -53,5 +53,5 @@
     {riak_api, {git, "https://github.com/basho/riak_api.git", {branch, "develop-3.0"}}},
     {hyper, {git, "https://github.com/basho/hyper", {tag, "1.1.0"}}},
     {kv_index_tictactree, {git, "https://github.com/martinsumner/kv_index_tictactree.git", {branch, "develop-3.0"}}},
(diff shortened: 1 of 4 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1370

stephenwood24-nhs (approved):

martinsumner:
Failure in test `verify_api_timeout`.  This test used to work by accident because of the riak_client:get within the webmachine workflow.  Not this bypassed, and riak_client:delete is prompted - but this does not look in options for a passed-in timeout, using the default instead.

martinsumner:
The original intention was to standardise the response to delete - such that HTTP and PB would both respond OK as a result of an attempt to delete an object which is not present.

There is a need to be correct with respect to the HTTP standard.  However, the HTTP standard is unclear.  Some argue, that due to the idempotent requirement - `204` is the correct response to a not_found delete, but equally others have argued for `404` as the correct response.  The `204` response is valid, but the `404` is also valid, and is more explicitly describing what has happened.

As there is no certainty as to the correctness of the response, and there are other cases where the response (especially error responses) vary between HTTP and PB - it seems better to be consistent with the previous version, rather than be consistent between the APIs.

PR #1852, opened 2023-02-21
Log Fragmentation

martinsumner:
With the leveled backend, memory fragmentation is an ongoing concern.  So, here the regular compaction callback is used to log information about carrier sizes and carrier block sizes for key allocators.

### DIFF
--- a/src/riak_kv_leveled_backend.erl
+++ b/src/riak_kv_leveled_backend.erl
@@ -659,7 +659,7 @@ log_fragmentation(Allocator) ->
     lager:info(
         "Memory for allocator=~p "
         "mbcs_block_size=~w mbcs_carrier_size=~w "
-        "sbcs_block_size=~w sbcs_carrier_size=~W",
+        "sbcs_block_size=~w sbcs_carrier_size=~w",
         [Allocator, MB_BS, MB_CS, SB_BS, SB_CS]).
 
 %% @private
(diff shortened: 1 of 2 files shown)


src/riak_kv_leveled_backend.erl
```diff
@@ -635,6 +637,31 @@ callback(Ref, UnexpectedCallback, State) ->
 %% Internal functions
 %% ===================================================================
 
+
+-spec log_fragmentation(eheap_alloc|binary_alloc) -> ok.
+log_fragmentation(Allocator) ->
+    FragStats =
+        lists:filter(
+            fun(MemStuff) ->
+                element(1, element(1, MemStuff)) == Allocator
```
ThomasArts: This is a bit risky.. assume data type changes in the future, then this may actually crash as a side effect of logging.
Rather make 3 match cases:
lists:filter(fun({{A, _}, _Stats}) -> A == Allocator;
                       (_) -> false
                       end,

ThomasArts: alternatively, build the new list straigth away:
FragStats = [ Stats || {{A, _}, Stats} <- econ_alloc:fragmentation(current), A == Allocator ]

src/riak_kv_leveled_backend.erl
```diff
@@ -635,6 +637,31 @@ callback(Ref, UnexpectedCallback, State) ->
...
+            end,
+            recon_alloc:fragmentation(current)),
+    {MB_BS, MB_CS, SB_BS, SB_CS} =
+        lists:foldl(
+            fun({{_A, _I}, Stats}, {MBAcc, MCAcc, SBAcc, SCAcc}) ->
```
ThomasArts: or here 
```suggestion
            fun({{A, _I}, Stats}, {MBAcc, MCAcc, SBAcc, SCAcc}) when A == Allocator->
```
and a separte clause for when it's not.

ThomasArts (approved):

PR #1854, opened 2023-03-08
Web machine small fixes for riak_kv_wm_object

ThomasArts:


martinsumner:
@ThomasArts Is the cost of the `maybe_decode_uri` in `service_available/2` significant?  am I right in thinking if all buckets and keys our plain ascii with no spaces, this could be disabled using the riak_kv, http_url_encoding environment variable - and the cost of this decoding could be avoided (and the header `X-Riak-URL-Encoding` used for exceptional cases).  Some keys can be quite long (60-100 characters).

martinsumner (approved):

PR #1855, opened 2023-03-14
Type fixes required by OTP26 dialyzer

martinsumner:
mochijson2 types are not exported by mochijson, and so for convenience they are simply redefined.

### DIFF
--- a/src/riak_kv_overflow_queue.erl
+++ b/src/riak_kv_overflow_queue.erl
@@ -397,7 +397,7 @@ generate_ordered_guid() ->
         [Year, Month, Day, H, M, C band 16#0fff, D band 16#3fff bor 16#8000, E]).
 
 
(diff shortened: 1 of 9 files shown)


src/riak_kv_overflow_queue.erl
```diff
@@ -397,7 +397,7 @@ generate_ordered_guid() ->
         [Year, Month, Day, H, M, C band 16#0fff, D band 16#3fff bor 16#8000, E]).
 
 
--spec disklog_filename(string(), string()) -> filename:filename().
+-spec disklog_filename(string(), string()) -> filename().
```
ThomasArts: In kv_index_tictactree you use:
`-spec disklog_filename(string(), string()) -> file:filename_all().`
The function is exactly the same (cf. aee_keystore.erl line 1245).
According to spec in filename.erl, you should use filename_all as type.

Issue #1857, opened 2023-04-19
Replication reference with node_confirms


martinsumner:
When replicating objects using nextgenrepl, the sink cluster will issue `fetch` requests from the source cluster.  These fetch requests will read from the real-time queue any item ready for replication.  the item will either be:

- An actual object which has recently been PUT;
- A reference to an object which exists in the store;
- A delete reference.

The second case will commonly occur when a `repl_keys_range` aae_fold has been made (but also when the real-time queue has grown during a busy period).

In the case of an object reference being read from the queue, a standard GET request will be used to return the actual object to the sink.

That GET request will have some specific options to improve performance:

https://github.com/basho/riak_kv/blob/9c138dc485eef7cb7cdeb0480422600ddf16523e/src/riak_client.erl#L205

These objects allow the object to be returned to the client as soon as an object matching the expected vector clock has been returned.  As this might happen after only 1 read, the R/PR settings over-ride any bucket property requiring a higher R value.

However, if the bucket has node_confirms set to a value more than 1 - the response will fail validation at this stage.

martinsumner:
It is possible to create unhandled pressure, especially when using repl range folds.  The standard mechanism for tuning this is manipulating the number of snk_workers fetching and pushing.

An alternative would to allow the r value on fetch, and the w value on push to be over-ridden.  This will slow replication, but by allowing for more than one vnode to confirm before completing the operation there will be a natural break on snk_workers when vnodes start developing backlogs.

PR #1858, opened 2023-04-19
Don't trigger node confirms failure if clock expected

martinsumner:
When fetching an object for replication, enough nodes are found if the expected clock is true - but that might then fail on node_confirms.  Don't generate the node_confirms response error if there has been a match on the expected clock.

### DIFF
--- a/src/riak_kv_get_core.erl
+++ b/src/riak_kv_get_core.erl
@@ -245,12 +245,15 @@ enough(_) ->
 %% Get success/fail response once enough results received
 -spec response(getcore()) -> {reply(), getcore()}.
 %% Met quorum for a standard get request/response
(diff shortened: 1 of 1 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1374

martinsumner:
https://github.com/basho/riak_kv/issues/1857

martinsumner:
@yorkshire-steve - this has been resolved a different way to our discussion, as I considered that simply over-riding node_confirms in the request options for the specific fetch case to be a bit a fragile.  Now the response failure for node_confirms is only generated when there has been no "Expected Clock" match

stephenwood24-nhs (approved):

yorkshire-steve (approved):

PR #1859, opened 2023-05-08
get_data_dir may return error

martinsumner:
Don't crash vnode in this case

https://github.com/martinsumner/leveled/issues/405


martinsumner:
@stephenwood24-nhs

stephenwood24-nhs (approved):

Issue #1864, opened 2023-07-13
Pruned clocks and failure of full-sync replication


martinsumner:
The vector clock on an object will increase in size for every new vnode_id which attempts to co-ordinate a PUT for an object.  The increase is limited by the pruning limits - `young_vclock`, `old_vclock`, `small_vclock` and `big_vclock`.  The primary limit is the `small_vclock` which will cause an object hitting this limit (50 unique vnodes by default) to be pruned.

The pruning will normally cut back by 1 entry only - i.e. a clock that has reached a size of 51 will be pruned back to 50.  There are some controls which may prevent this pruning - `young_vclock`, `old_vclock` - but by the defualt the `big_vclock` also being set to 50 means that clocks will remain at this limit.

In one key situation, pruning is disabled.  When an update is a consequence of read_repair - the clock will not be pruned, so objects can be persisted with clocks longer than 50.

In a multi-cluster scenario, after read repairs there can be a circumstance where some objects have clocks of length > `small_vclock` on one cluster, but of length == `small_vclock` on another.

If there is now an attempt to full-sync between the clusters, the delta can be detected - but when the object is dispatched from the cluster with the dominant (longer) clock, the receiving cluster will prune the inbound clock ... and the delta will persist.  In this case, the objects are the same, but the full-sync mechanism cannot determine they are the same - so continuously discovers differences.

Clocks being at the pruning length is more likely to occur on very long-lived clusters which have been subject to lots of leaves/joins - but as many users have been running Riak for > 10 years this is increasingly probable.

martinsumner:
There is a workaround to "touch" (i.e. GET, then PUT back the object) all objects which have exceeded the prune size.

i.e.

```
{ok, CH} = riak:local_client().
TouchFun = fun({B, K, _C}) -> {ok, Obj} = riak_client:get(B, K, CH), ok = riak_client:put(Obj, CH) end.
QueriesFun = fun(Bucket) -> lists:map(fun(I) -> {fetch_clocks_range, Bucket, all, {segments, lists:seq((128 * I) + 1, 128 *  (I + 1)), small}, all} end, lists:seq(0, 511)) end.
TouchGreaterThanFun = fun(B, M) -> lists:foreach(fun(Q) -> lists:foreach(TouchFun,lists:filter(fun({_B0, _K0, C0}) -> length(C0) > M end, element(2, riak_client:aae_fold(Q, CH)))) end, QueriesFun(B)) end.
[...]

PR #1866, opened 2023-07-14
Use 'RR' when a prunable vclock is replicated

martinsumner:
There may be some situations whereby a vector clock grows beyond the prescribed limits on the source cluster - in particular following read repair.

In these cases the new object needs to be replicated but with the same resulting vector clock (assuming no siblings).  If the same vector clock does not result on the sink - any full-sync operation may continuously detect the  delta, but not be able to resolve it (as the sink vnodes prune each time).

The 'rr' option will, on riak_kv_vnode, ensure pruning is bypassed so that we avoid pruning on a sink, if we have not pruned on a source.  The 'rr' option is only used when the clock is prunable (as otherwise the delta could occur in the reverse direction).

The 'rr' option also blocks some sibling constraint checks (e.g. maximum number of siblings.  However, as the most likely cause of it being applied is 'rr' on the src side - this is still generally a win for replication consistency).

https://github.com/basho/riak_kv/issues/1864

### DIFF
--- a/src/riak_client.erl
+++ b/src/riak_client.erl
@@ -241,8 +241,19 @@ push(RObjMaybeBin, IsDeleted, _Opts, {?MODULE, [Node, _ClientId]}) ->
     Me = self(),
     ReqId = mk_reqid(),
     W = application:get_env(riak_kv, replrtq_vnodecheck, 1),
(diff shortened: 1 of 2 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1379

stephenwood24-nhs (approved):

PR #1880, opened 2023-11-14
Configurable $key query on leveled (#8)

martinsumner:
https://github.com/nhs-riak/riak_kv/issues/7

This PR as added to make transition of delete_modes easier with leveled.  When moving to delete_mode of keep, there may exist application queries for list of keys that do not expect tombstones.  This allows for tombstones to excluded from $key queries, where the backend is leveled.


martinsumner:
Related test - https://github.com/basho/riak_test/pull/1384

